PR #9, opened 2018-03-22
Fix make vet fmt

### LINKED ISSUE #6: make vet fmt is broken
```
$ make vet fmt
cd /Users/dims/go/src/git.openstack.org/openstack/openstack-cloud-controller-manager && go vet ./...
# git.openstack.org/openstack/openstack-cloud-controller-manager/pkg/volume/cinder/provisioner
pkg/volume/cinder/provisioner/iscsi_test.go:61:4: err declared but not used
pkg/volume/cinder/provisioner/iscsi_test.go:105:4: err declared but not used
pkg/volume/cinder/provisioner/iscsi_test.go:153:4: err declared but not used
pkg/volume/cinder/provisioner/rbd_test.go:78:4: err declared but not used
vet: typecheck failures
make: *** [vet] Error 2
```

hogepodge:

Several tests had unused err variables. Removed them to fix `make vet fmt`

**Which issue this PR fixes**
fixes #6 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/volume/cinder/provisioner/iscsi_test.go
+++ b/pkg/volume/cinder/provisioner/iscsi_test.go
@@ -82,6 +82,7 @@ var _ = Describe("Iscsi Mapper", func() {
 			Expect(source.ISCSI.IQN).To(Equal("iqn"))
 			Expect(source.ISCSI.Lun).To(Equal(int32(3)))
 			Expect(source.ISCSI.SessionCHAPAuth).To(BeFalse())
+			Expect(err).NotTo(HaveOccurred())
 		})
 
 		Context("when CHAP authentication is enabled", func() {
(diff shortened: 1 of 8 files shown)


flaper87:
lgtm

dims:
one more thing to fix @hogepodge

pkg/volume/cinder/provisioner/iscsi_test.go
```diff
@@ -74,7 +73,7 @@ var _ = Describe("Iscsi Mapper", func() {
 
 		JustBeforeEach(func() {
 			secretName = getChapSecretName(conn, options)
-			source, err = mapper.BuildPVSource(conn, options)
```
anguslees: The idea of the lint is that ignoring the error is a bug - not that you should just ignore the error in a different way :-P

Gingko makes this easy with an assertion like:
`Expect(err).NotTo(HaveOccurred())`

hogepodge:
@dims done

pkg/volume/cinder/provisioner/iscsi_test.go
hogepodge: Gotcha. Had a bad feeling about just dropping all those errors, just wasn't sure how to go about it. Thanks.

hogepodge:
Ok, so golint is not available in the test environment. I've updated the Makefile to install it if it isn't available.

PR #11, opened 2018-03-22
Add Daniel Gonzalez (gonzolino) as reviewer

gonzolino:


OWNERS
```diff
@@ -3,8 +3,10 @@ approvers:
 - NickrenREN
 - dims
 - FengyunPan2
+- gonzolino
```
dims: Let's start as a reviewer and then we can move to approver after some time. is that ok?

gonzolino: Yes of course

PR #28, opened 2018-03-22
Add code cleanup utils - gofmt and golint

dims:
* Run `make fmt` or `make lint` to run these
* Use hack/update-gofmt.sh to fix problems with fmt


dims:
/approve

(self approval for utilities picked up from k/k)

PR #38, opened 2018-03-23
Remove useless error handler

Fedosin:
It is a duplicate of another handler, located several lines above. 
Therefore, this code will never be executed.


hogepodge:
Can you explain why removing logging of auth failure is useless?

Fedosin:
@hogepodge Yep, I updated the description...

In short, it's a duplicate and this code is never executed. Just take a look: https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/identity/keystone/keystone.go#L89-L96

hogepodge:
Oh yes, of course! If I had unfolded the code I would have seen it. Thanks!

/approve

PR #58, opened 2018-03-27
Add Keystone authorizer tests

Fedosin:


pkg/identity/keystone/authorizer.go
```diff
@@ -35,11 +35,6 @@ type Authorizer struct {
 }
 
 func resourceMatches(p policy, a authorizer.Attributes) bool {
-	if p.NonResourceSpec != nil && p.ResourceSpec != nil {
```
Fedosin: The check is a duplicate of https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/identity/keystone/authorizer.go#L154

dims: Ack

hogepodge:
/approve

Going to wait for another reviewer.

In the meantime @Fedosin, can you create an issue to expand the test coverage as you mentioned in the TODO? Issues are more actionable than an inline TODO.

Fedosin:
@hogepodge All lines are covered except these https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/identity/keystone/authorizer.go#L107-L113
I decided to use project name instead of id...

What tests are required:
1. Policies with * in 'verb', 'match', etc.;
2. APIGroups and Namespaces;
3. Missing things in policies. 'verb' or 'match' for instance.

PR #73, opened 2018-03-30
Change default az to empty when create cinder volume

### LINKED ISSUE #72: Create cinder volume failed with default availability zone 'nova'

>
/kind bug
> /kind feature

**What happened**:
create cinder volume failed against a specific cloud provider due to invalid parameters caused by incorrect availability zone, because the default value of 'nova' [1] in this repository doesn't match the real case in cloud provider.

[1] https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/volume/cinder/volumeservice/actions.go#L75

create cinder volume succeed by change the default value of availability zone to empty.

1. install openstack cinder standalone service
2. create a k8s cluster via local-up-cluster.sh
3. run the cinder-provisioner binary built from this repository
4. run test from this repository, kubectl apply -f examples/persistent-volume-provisioning/cinder/cinder-full.yaml

**Environment**:
- openstack-cloud-controller-manager version: latest
- OS (e.g. from /etc/os-release): Ubuntu 16.04.4 LTS
- Kernel (e.g. `uname -a`): Linux 4.4.0-116-generic
- Install tools: git, make
- Others:

animationzl:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #72 

**Release note**:

```release-note
```

### DIFF
--- a/pkg/volume/cinder/volumeservice/actions.go
+++ b/pkg/volume/cinder/volumeservice/actions.go
@@ -72,7 +72,7 @@ func CreateCinderVolume(vs *gophercloud.ServiceClient, options controller.Volume
 	// Cinder works with gigabytes, convert to GiB with rounding up
 	sizeGB := int((sizeBytes + 1024*1024*1024 - 1) / (1024 * 1024 * 1024))
 	volType := ""
-	availability := "nova"
+	availability := ""
 	// Apply ProvisionerParameters (case-insensitive). We leave validation of
 	// the values to the cloud provider.
 	for k, v := range options.Parameters {


animationzl:
signed

dims:
@zetaab Note that this is the new external volume service, not the regular cinder one from the main k/k. So no one is using this yet.

zetaab:
@dims ok but still I would like to get answer how it behaves if az is empty.

dims:
@animationzl please respond to @zetaab 's question

animationzl:
@zetaab yes, the cloud provider I use has a different az name than nova, the default az in cinder create api is None, when creating volume without a specific az(such as `cinder create 1`), a random az of specific volume host which the volume is scheduled to will be chose as result.

zetaab:
ok then I can approve that change :)

zetaab:
/approve

@dims add LGTM if you think this is ok to merge

PR #82, opened 2018-04-03
Fix reliability of TestRoutes and TestVolumes

dims:
In TestRoutes, let's make sure we get an active instance and use
that. If the instance is not active, then it will not have an ip address
and hence fail.

In TestVolumes, we should actually wait for a total of 30 seconds. We
are bailing out in one second right now. Make sure we sleep for the
initial delay before we start as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,8 @@ BASE_DIR := $(shell basename $(PWD))
 # Keep an existing GOPATH, make a private one if it is undefined
 GOPATH_DEFAULT := $(PWD)/.go
 export GOPATH ?= $(GOPATH_DEFAULT)
+TESTARGS_DEFAULT := "-v"
+export TESTARGS ?= $(TESTARGS_DEFAULT)
 PKG := $(shell awk  '/^package: / { print $$2 }' glide.yaml)
 DEST := $(GOPATH)/src/$(GIT_HOST)/$(BASE_DIR)
 DEST := $(GOPATH)/src/$(PKG)
@@ -74,7 +76,7 @@ k8s-keystone-auth: depend $(SOURCES)
 
 test: unit functional
 
-check: fmt vet lint
+check: depend fmt vet lint
 
 unit: depend
 	cd $(DEST) && go test -tags=unit $(shell glide novendor) $(TESTARGS)
--- a/pkg/cloudprovider/providers/openstack/openstack.go
+++ b/pkg/cloudprovider/providers/openstack/openstack.go
@@ -275,7 +275,7 @@ func readInstanceID(searchOrder string) (string, error) {
 		instanceID := string(idBytes)
 		instanceID = strings.TrimSpace(instanceID)
 		glog.V(3).Infof("Got instance id from %s: %s", instanceIDFile, instanceID)
-		if instanceID != "" {
+		if instanceID != "" && instanceID != "iid-datasource-none" {
 			return instanceID, nil
 		}
 		// Fall through to metadata server lookup
--- a/pkg/cloudprovider/providers/openstack/openstack_routes_test.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_routes_test.go
@@ -87,7 +87,10 @@ func TestRoutes(t *testing.T) {
 
 func getServers(os *OpenStack) []servers.Server {
 	c, err := os.NewComputeV2()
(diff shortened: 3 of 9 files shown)


pkg/cloudprovider/providers/openstack/openstack_test.go
```diff
@@ -52,10 +52,11 @@ const (
 
 func WaitForVolumeStatus(t *testing.T, os *OpenStack, volumeName string, status string) {
 	backoff := wait.Backoff{
-		Duration: volumeStatusInitDelay,
+		Duration: volumeStatusTimeoutSeconds,
```
NickrenREN: why change this here ? `Duration` in Backoff is the base duration

dims:
recheck

dims:
recheck

dims:
recheck

dims:
recheck

dims:
recheck

dims:
recheck

dims:
recheck

dims:
recheck

Makefile
```diff
@@ -11,6 +11,8 @@ BASE_DIR := $(shell basename $(PWD))
 # Keep an existing GOPATH, make a private one if it is undefined
 GOPATH_DEFAULT := $(PWD)/.go
 export GOPATH ?= $(GOPATH_DEFAULT)
+TESTARGS_DEFAULT := "-v"
+export TESTARGS ?= $(TESTARGS_DEFAULT)
```
dims: Here i am ensuring that we use verbose output when we run `make test` locally

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -275,7 +275,7 @@ func readInstanceID(searchOrder string) (string, error) {
 		instanceID := string(idBytes)
 		instanceID = strings.TrimSpace(instanceID)
 		glog.V(3).Infof("Got instance id from %s: %s", instanceIDFile, instanceID)
-		if instanceID != "" {
+		if instanceID != "" && instanceID != "iid-datasource-none" {
```
dims: some providers seem to return a fake id at least some of the time, so account for that - http://cloudinit.readthedocs.io/en/latest/topics/datasources/fallback.html

pkg/cloudprovider/providers/openstack/openstack_routes_test.go
```diff
@@ -87,7 +87,10 @@ func TestRoutes(t *testing.T) {
 
 func getServers(os *OpenStack) []servers.Server {
 	c, err := os.NewComputeV2()
-	allPages, err := servers.List(c, servers.ListOpts{}).AllPages()
+	opts := servers.ListOpts{
```
dims: Look for just "ACTIVE" servers

pkg/cloudprovider/providers/openstack/openstack_test.go
```diff
@@ -65,7 +65,7 @@ func WaitForVolumeStatus(t *testing.T, os *OpenStack, volumeName string, status
 			t.Logf("Volume (%s) status changed to %s after %v seconds\n",
 				volumeName,
 				status,
-				volumeStatusTimeoutSeconds)
+				time.Now().Second()-start)
```
dims: print the time taken since start. gives the wrong impression that that it took 30 seconds when it did not really

pkg/cloudprovider/providers/openstack/openstack_volumes.go
```diff
@@ -547,18 +554,32 @@ func (os *OpenStack) getDevicePathFromInstanceMetadata(volumeID string) string {
...
-	if devicePath == "" {
+func (os *OpenStack) GetDevicePath(volumeID string) (string, error) {
+	backoff := wait.Backoff{
+		Duration: operationFinishInitDelay,
+		Factor:   operationFinishFactor,
+		Steps:    operationFinishSteps,
+	}
+
+	var devicePath string
+	err := wait.ExponentialBackoff(backoff, func() (bool, error) {
```
dims: Try the operation a few times, sometimes when the status turns to `in-use` we still don't have the info we need

dims:
recheck

dims:
recheck

pkg/cloudprovider/providers/openstack/openstack_test.go
dklyle: nice catch

PR #84, opened 2018-04-04
Better environment vars support for standalone cinder service

dims:
during some refactoring, lost support for OS_CINDER_ENDPOINT, so adding
that back and also pick up other env vars.


dims:
cc @jianglingxia

PR #86, opened 2018-04-04
Migrate from Glide to dep for dependency management

### LINKED ISSUE #70: Migrate from glide to dep
Now we use [Glide](https://github.com/Masterminds/glide) for dependency management, but the project is not actively developing and the authors recommend to migrate to [dep](https://github.com/Masterminds/glide#golang-dep) until it's too late.

gonzolino:
This PR migrates from Glide to dep as dependency management tool as suggested by the Glide authors: https://github.com/Masterminds/glide#golang-dep

There is still work to be done for the migration to finish, e.g. remove glide.yaml and glide.lock, remove usage of glide in all scripts, etc.

I replaced all usages of glide in the Makefile, but I think this still needs thorough testing. 

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #70 

**Release note**:

```release-note
Migrate from Glide to dep as dependency management tool.
```


gonzolino:
@flaper87 You're right, I have marked this PR as WIP.
Still, it would be good if people would start testing, to ensure that the migration works as expected

Makefile
```diff
@@ -13,12 +13,12 @@ GOPATH_DEFAULT := $(PWD)/.go
 export GOPATH ?= $(GOPATH_DEFAULT)
 TESTARGS_DEFAULT := "-v"
 export TESTARGS ?= $(TESTARGS_DEFAULT)
-PKG := $(shell awk  '/^package: / { print $$2 }' glide.yaml)
+PKG := $(shell awk  -F "\"" '/^ignored = / { print $$2 }' Gopkg.toml)
```
j-griffith: I think this is going to return the entire list from the toml file:
    `ignored = ["k8s.io/cloud-provider-openstack"]`

When we need just:
    `k8s.io/cloud-provider-openstack`

Unless my awk is super rusty, which it probably is :)

Makefile
```diff
@@ -13,12 +13,12 @@ GOPATH_DEFAULT := $(PWD)/.go
...
 DEST := $(GOPATH)/src/$(GIT_HOST)/$(BASE_DIR)
 DEST := $(GOPATH)/src/$(PKG)
 SOURCES := $(shell find $(DEST) -name '*.go')
 HAS_MERCURIAL := $(shell command -v hg;)
-HAS_GLIDE := $(shell command -v glide;)
+HAS_DEP := $(shell command -v dep;)
```
j-griffith: `dep -v` isn't actually valid, although it will call dep and tell you if it's installed or not, so it works for the purpose here.

Makefile
```diff
@@ -32,15 +32,13 @@ depend: work
 ifndef HAS_MERCURIAL
 	pip install Mercurial
 endif
-ifndef HAS_GLIDE
-	go get -u github.com/Masterminds/glide
-endif
-ifeq ($(wildcard $(DEST)/vendor/.*),)
-		cd $(DEST) && glide install --strip-vendor
+ifndef HAS_DEP
+	go get -u github.com/golang/dep/cmd/dep
```
j-griffith: I'm not sure we should use the dev branch of dep here as opposed to pulling/installing an official release.  I'll let others comment on that, but the dep page does state pretty clearly that you shouldn't use the master branch for production etc.

Makefile
```diff
@@ -32,15 +32,13 @@ depend: work
...
 endif
+	dep ensure
 
 depend-update: work
-	cd $(DEST) && glide update --strip-vendor
+	dep ensure -update
```
j-griffith: Ahh, so for this do we in fact want to update the lock file and everything?  Kind of dangerous if so, as opposed to an update to a single package.

I thought this was just a "make sure everything is up to date" check in which case wouldn't just `dep ensure` be the right call?

Makefile
```diff
@@ -79,7 +77,7 @@ test: unit functional
 check: depend fmt vet lint
 
 unit: depend
-	cd $(DEST) && go test -tags=unit $(shell glide novendor) $(TESTARGS)
+	cd $(DEST) && go test -tags=unit $(shell go list ./... | grep -v vendor) $(TESTARGS)
```
j-griffith: go list ./... Seems to have some trouble here and outputs some errors; not sure it's harmful for the unit test.  But I don't think you need to add the ignore here anyway, tested in my version without any special considerations and it matched the number of unit tests executed as the original glide version of the code.

Makefile
```diff
@@ -98,7 +96,7 @@ vet: work
 	cd $(DEST) && go vet ./...
 
 cover: depend
-	cd $(DEST) && go test -tags=unit $(shell glide novendor) -cover
+	cd $(DEST) && go test -tags=unit $(shell go list ./... | grep -v vendor) -cover
```
j-griffith: same thoughts as above

j-griffith:
Looks good, sorry I didn't know you'd picked this up.  I noted just a couple minor comments/questions but it does look like we've done the same work :)

Makefile
gonzolino: The idea here is not to use `dep -v` but `command -v dep` (similar to how it was before with glide).
If dep is installed `command -v dep` will return the absolute path to dep. If it isn't installed it will return nothing.

Makefile
gonzolino: That's true... I think I will go your way and curl the install.sh script. That should be platform independent and use a stable release.

Makefile
gonzolino: OK, it was my understanding that `make depend` installs dependencies and `make depend-update` updates those dependencies (which might also include updates to the lock file).
But I might be wrong here... Lets wait for comments from others.

Makefile
j-griffith: Yep, my bad; thanks for pointing it out

gonzolino:
@j-griffith Thanks for your comments! I'll update the PR tomorrow and I'll try to address all your comments :)

dims:
ok-to-test

gonzolino:
Ok, it seems like the dep [install.sh](https://github.com/golang/dep/blob/master/install.sh#L76) Script expects the `$GOBIN` directory to exist. I tried to adapt the Makefile to fix that.

Lets see if this works.

Gopkg.toml
```diff
@@ -0,0 +1,83 @@
+# Gopkg.toml example
```
dims: Let's remove this commented out portion and we should be good to go!

gonzolino: Done :)

dims:
+1 to squash and then merge

PR #87, opened 2018-04-04
Migrate from Glide to Dep

### LINKED ISSUE #70: Migrate from glide to dep

j-griffith:
This change moves off of Glide and on to Dep and makes the
necessary changes to the Makefile as well.

This PR moves off of Glide (as per Glide teams recommendations) and on to dep

**Which issue this PR fixes** 
fixes #70 

This should be a "drop in" replacement / upgrade.  This change includes the necessary updates to the Makefile.  We shouldn't need any special args for `go test` or `ensure` other than what's been added.  I've verified we run the same tests in the case of unit and coverage runs but I could certainly have an error in there for a keen reviewer to catch.

```release-note
```


j-griffith:
Looks like https://github.com/kubernetes/cloud-provider-openstack/pull/86 beat me to it and passed the dep path parts. I'll go ahead and close this.

Issue #88, opened 2018-04-05
external network identified as internal


yanndegat:
/kind feature

**What happened**:
public ipv4 node address identified as Internal 

**What you expected to happen**:
public ipv4 node address identified as External 

**How to reproduce it (as minimally and precisely as possible)**:
Don't name your openstack public network "public"
Create an instance with an IP on this network

dims:
+1 @yanndegat want to help with a PR?

yanndegat:
ok

FengyunPan2:
+1

PR #94, opened 2018-04-06
Add support for networking opts

### LINKED ISSUE #89: Filter instances ipv4/6 addresses
/kind feature

**What happened**:
my nodes are identified either by their ipv4 addr or their ipv6, according 
to the numerical order of the adresses.
Once added to the the nodeAddresses list, the type of the addr is lost.

something more consistent & at least either ipv4 addrs or ipv6 addrs

setup an openstack network with ipv4 & ipv6 support such as your instances
get 2 addresses.

yanndegat:
- allows to customize public network name
  according to the cloud provider
- allows disabling ipv6 at config level

Fixes #89 #88

/assign @dims

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack.go
+++ b/pkg/cloudprovider/providers/openstack/openstack.go
@@ -66,6 +66,10 @@ var ErrMultipleResults = errors.New("multiple results where only one expected")
 // ErrNoAddressFound is used when we cannot find an ip address for the host
 var ErrNoAddressFound = errors.New("no address found for host")
 
+// ErrIPv6SupportDisabled is used when one tries to use IPv6 Addresses when
+// IPv6 support is disabled by config
+var ErrIPv6SupportDisabled = errors.New("IPv6 support is disabled")
+
 // MyDuration is the encoding.TextUnmarshaler interface for time.Duration
 type MyDuration struct {
 	time.Duration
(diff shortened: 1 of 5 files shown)


pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -253,6 +265,8 @@ func readConfig(config io.Reader) (Config, error) {
 	cfg.BlockStorage.IgnoreVolumeAZ = false
 	cfg.Metadata.SearchOrder = fmt.Sprintf("%s,%s", configDriveID, metadataID)
 	cfg.Metadata.DHCPDomain = "novalocal"
+	cfg.Networking.IPv6SupportDisabled = false
```
dims: Can you please do the same in the `configFromEnv` method above too?

yanndegat:
hi @dims 

thanks a lot for being so prompt

2 questions though: is there anywhere i should add some doc about these new params?
what is this "tide" / "lgtm" label thing in the PR's check?

thanks a lot

regards 

yann

dims:
@yanndegat that's the usual workflow in all kubernetes repositories. right now i have approved the changes, waiting for some of the other reviewers to apply a `lgtm` (looks good to me) label. once that is done one of the bots merges the change

dims:
@yanndegat once https://github.com/kubernetes/cloud-provider-openstack/pull/99 merges, we can update that doc with the new options

zioproto:
@yanndegat I am sorry I arrived late to review. This patch is already merged but introduces a change in Behavior. 

Now you really need in the config
```

**Comment by zioproto** [comment]
Hello @yanndegat @dims 
I apologize I arrived late to review this patch, and it is now already merge.
Please not that there is a change in the default behaviour after merging this patch.
The patch removes the default hardcoded referce to the `public` network. So it is mandatory now to have at least this bit of configuration:
```
[Networking]
public-network-name=public
```
The code does not use a default option when `public-network-name` is not specified in the config.
@yanndegat please fix this with a new patch, or write a Release note of users upgrading.
thanks !

dims:
@zioproto it does default to `public`. please look carefully :)

dims:
search for `cfg.Networking.PublicNetworkName = "public"` @zioproto

zioproto:
@yanndegat @dims you are right ! Better ! It is all OK then. Sorry for the noise, thanks for your contribution.

PR #98, opened 2018-04-06
Check for existing names on csi CreateVolume

### LINKED ISSUE #91: CSI Plugin needs to check for existing volumes with same name on create
kind bug

**What happened**:
The CSI spec states that the CreateVolume call is idempotent and should not create duplicate volumes if/when called multiple times with the same Volume Name.  The current implementation does not check for existing Volumes with the specified Name and thus allows creation of duplicate named Volumes.

The CSI plugin should respond with the existing Volume and NOT create a new Volume.

Run the CSI Plugin and issue a create call multiple times with the same name, then check out the Cinder List and you'll see multiple volumes with the same name.

PR is in progress by jgriffith

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

j-griffith:
**What this PR does / why we need it:

The CSI spec states that it's up to the Plugin to enforce idempotency
on Create and that we do NOT create duplicate named volumes if the same
name is specified in a subsequent Create request.

This patch adds some things to check that:
1. A GetVolumesByName function to openstack_volumes
2. A check in CreateVolume prior to issuing Create to
   see if any volumes with the given Name exist prior
   to Create

If we find a single volume with the given Name we return it.

If we find multiple volumes with the given Name we don't really
have enough info currently to *know* what the right thing is, so
we return an Error.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` 
fixes #91 


pkg/csi/cinder/controllerserver.go
```diff
@@ -58,14 +60,32 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return nil, err
 	}
 
-	// Volume Create
-	resID, resAvailability, err := cloud.CreateVolume(volName, volSizeGB, volType, volAvailability, nil)
+	// Verify a volume with the provided name doesn't already exist for this tenant
+	volumes, err := cloud.GetVolumesByName(volName)
```
edisonxiang: It is better to check if  ```volSizeGB```,```volType```,```volAvailability``` of the ```volumes``` are equal with the same fields in the```CreateVolumeRequest```.

hogepodge: Based on @j-griffith's description of the spec, I think it's helpful for debugging but not compliant with the requirements. If identity is equivalent to name, then name should be the only criteria we use for identity.

PR #99, opened 2018-04-06
copy doc from main repo

dims:
from
https://github.com/kubernetes/website/blob/master/docs/concepts/cluster-administration/cloud-providers.md


hogepodge:
This is a good start. This document will likely be broken up and refactored as we work through standards with wg-cloud-provider, but it's good to have this migrated into our repository.

/ok-to-test
/approve

PR #104, opened 2018-04-09
Standardize docker image repo location

dims:
Let's use a single repo for all the docker images from cloud-provider-openstack:
https://hub.docker.com/u/k8scloudprovider/dashboard/


dims:
Also see https://github.com/kubernetes/cloud-provider-openstack/issues/105

dims:
we don't have releases yet ... we have to talk about that first. until then my idea was to have CI jobs build and upload container images daily and we can use `latest`

dims:
cc @animationzl @kiwik @liu-sheng

dims:
@zetaab true, naming is hard, please suggest alternatives, we can switch over quite easily

hogepodge:
@zetaab cloud-provider-openstack is the name that was chosen by wg-cloud-provider to be consistent and discoverable for all providers.

zioproto:
@hogepodge I am not sure I understand the discussion. Should `k8scloudprovider` be changed in `cloud-provider-openstack` ? Is this what you mean ? I see the containers images are already uploaded to https://hub.docker.com/u/k8scloudprovider/ . Is that the definitive name ? Can we advertise those images to users ? thanks

dims:
@zioproto we haven't made a release yet. so i'd not advertise these artifacts as they reflect `latest` from master/HEAD. at the time when they were cut. 

- we need to think if `k8scloudprovider` is the final name. `cloud-provider-openstack` won't work as it has dashes
- we need to setup CI jobs to publish these images periodically. possibly multiple times a day when the get a good signal from the test suite.

So yes, please kick the tires and ask folks who are developers to try it out.

PR #109, opened 2018-04-10
Add pull policy for driver-registrar image to support local image built from latest source code

### LINKED ISSUE #110: Add pull policy for driver-registrar image to support local image built from latest source code

>
> /kind bug
/kind feature

**What happened**:
Currently, during the OpenLab CI job the driver-registrar image is always pulled from remote(docker hub) which may be obsolete against the source code where it is built from, so this issue wants to add a pull policy to support local image built from latest source code.

We can keep using the latest image in OpenLab CI job.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

animationzl:
Fix #110 
Add pull policy for driver-registrar image to support local image built from latest source code


dims:
cc @edisonxiang 

/lgtm

PR #120, opened 2018-04-13
Enter DEST dir before running dep ensure

flaper87:
If we don't enter the `$(DEST)` dir, `dep` will fail saying that the `cloud-provider-openstack` dir is not in one of the known `GOPATH/src` paths.

This commit also forces the creation of the symlink, otherwise it'll fail saying that a symlink exists already.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,10 +37,10 @@ endif
 ifndef HAS_DEP
 	curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
 endif
-	dep ensure
+	cd $(DEST) && dep ensure
 
 depend-update: work
-	dep ensure -update
+	cd $(DEST) && dep ensure -update
 
 build: openstack-cloud-controller-manager cinder-provisioner cinder-flex-volume-driver cinder-csi-plugin k8s-keystone-auth
 
@@ -140,7 +140,7 @@ $(GOBIN):
 
 $(DEST): $(GOPATH)
 	mkdir -p $(shell dirname $(DEST))
-	ln -s $(PWD) $(DEST)
+	ln -sf $(PWD) $(DEST)
 
 .bindep:
 	virtualenv .bindep


Makefile
```diff
@@ -37,7 +37,7 @@ endif
 ifndef HAS_DEP
 	curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
 endif
-	dep ensure
+	cd $(DEST) && dep ensure
 
 depend-update: work
 	dep ensure -update
```
gonzolino: I think you also need to`cd $(DEST)` in this line.

gonzolino:
This seems to be the reason for the failing tests:
```
2018-04-16 13:28:44.247170 | ubuntu-xenial-vexxhost | cd /home/zuul/src/k8s.io/cloud-provider-openstack && dep ensure
2018-04-16 13:28:44.253005 | ubuntu-xenial-vexxhost | both /home/zuul/src/k8s.io/cloud-provider-openstack and /home/zuul/src/github.com/kubernetes/cloud-provider-openstack are in the same GOPATH /home/zuul
2018-04-16 13:28:44.254329 | ubuntu-xenial-vexxhost | Makefile:35: recipe for target 'depend' failed
```

I think this is caused by the `ln -sf $(PWD) $(DEST)`. It creates the `/home/zuul/src/github.com/kubernetes/cloud-provider-openstack` symlink. But not sure why this happens only now and not before...

Makefile
```diff
@@ -37,10 +37,10 @@ endif
 ifndef HAS_DEP
 	curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
 endif
-	dep ensure
+	cd $(DEST) && dep ensure
```
j-griffith: I'm not sure I follow, is your intent to be able to run make from outside of the repos working directory?  The eval of PWD, BASE_DIR an finally DEST should tie all this together; but even so we should expect to be in the correct directory when make is called no?

dims:
Do we still need this @j-griffith @flaper87 ?

j-griffith:
> So, without this patch, calling make seems to fail when GOPATH is set. The failure I'm seeing is: dir is > not in one of the known GOPATH/src paths.

GOPATH can be so annoying if you want to do anything custom.  What is your GOPATH on your system, are you by chance not setting one and just picking up the default '$HOME/go/'?

Honestly I'm not sure that we shouldn't just expect a proper GO Env and skip some of the GOBIN/GOPATH stuff, but there's definitely arguments on both sides of that.

> The way to avoid this issue is by entering $GOPATH/src/cloud-provider-openstack and running these > commands from there.

That's weird to me, I would have assumed you were already in that directory so I don't follow the "entering" into the same directory (you're already there) changing this.  Unless the GOPATH wasn't set, the Makefile modifies/sets it, the CD picks up the changes?

I would try making sure you're code is checked out under a valid $GOPATH/src directory; I guess if folks want to just merge this to fix this case that's reasonable, but I'm trying to understand the root of the issue; I'd also personally like to simplify the Makefile a bit, but maybe that's not feasible.

j-griffith:
Ohhh!  I think I see what's happening here; I think this is a result of the trickery of creating the k8s.io directory and adding the symlink to your cwd.  I wonder if we actually need to do that?  @dims  thoughts?

One thing I've been working on related to this is converting the build to run Make in a go container (eliminate most of the path and setup issues that come up), I was playing around with that this morning and realized that's what @flaper87 is most likely running into since he's not using the standard '../src/k8s.io' directory in his process.

So I guess that explains the mystery; haven't dug into the impact of this particular change and why it's failing the openlab tests but for now I'd probably just recommend setting up your go env so that your repo is under $GOPATH/src, that way the symlink trick will work and you won't have to mess with the indirection things here.

dims:
I am happy for us to tell people to git clone into `$GOPATH/src/k8s.io/cloud-provider-openstack` and then run `make` let's just make that clear when that's not the the case (and fail with a better error message)

flaper87:
Here's a snippet to illustrate better what's exactly happening to me (sorry, I should have started with this, my bad):

```
x1hobo ~ $ cd /tmp
x1hobo /tmp $ git clone https://github.com/kubernetes/cloud-provider-openstack/
Cloning into 'cloud-provider-openstack'...
remote: Counting objects: 6236, done.
remote: Compressing objects: 100% (7/7), done.
remote: Total 6236 (delta 0), reused 3 (delta 0), pack-reused 6229
Receiving objects: 100% (6236/6236), 1.47 MiB | 1.87 MiB/s, done.
Resolving deltas: 100% (2502/2502), done.
x1hobo /tmp $ cd cloud-provider-openstack
x1hobo (master) /tmp/cloud-provider-openstack $ which dep
/usr/bin/dep
x1hobo (master) /tmp/cloud-provider-openstack $ export GOPATH=`pwd`/.go        
x1hobo (master) /tmp/cloud-provider-openstack $ make
find: ‘/tmp/cloud-provider-openstack/.go/src/k8s.io/cloud-provider-openstack’: No such file or directory
mkdir -p /tmp/cloud-provider-openstack/.go
echo "create gobin"
create gobin
mkdir -p /tmp/cloud-provider-openstack/.go/bin
mkdir -p /tmp/cloud-provider-openstack/.go/src/k8s.io
ln -s /tmp/cloud-provider-openstack /tmp/cloud-provider-openstack/.go/src/k8s.io/cloud-provider-openstack
dep ensure
/tmp/cloud-provider-openstack is not within a known GOPATH/src
make: *** [Makefile:39: depend] Error 1
```

Am I wrong to believe that this should be a valid use case?

j-griffith:
@flaper87 I think neither case is really "correct" here personally; I mean creating the gopath dir and the symbolic link isn't great, leading to an invalid gopath and some weirdness trying to figure out where you're at.  I do think it's best to just setup a go env and actually clone to a k8s.io directory (at least for now) and avoid some of this altogether.

If you're setup in a valid gopath we can make in any subdir under "go/src/", but really it just seems like requiring proper go setup (ie go/src/k8s.io) isn't terrible.

flaper87:
Closing this in favor of #135

PR #124, opened 2018-04-16
Create a function to receive keystone tokens on the client side

### LINKED ISSUE #54: Kubectl exec-based client auth provider for Keystone
We pulled out code from main repository related to the kubectl client side plugin in this repository:
https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/client-go/plugin/pkg/client/auth/openstack

However the code needs to be enabled as a exec based auth provider to actually work, see:
https://github.com/kubernetes/kubernetes/pull/59495

Once that is done, we need to remove the code from the main k/k repository.

cc @Fedosin @flaper87 @knikolla

Fedosin:
This code adds new function GetToken inside "keystone" module that should be used
in the executable file for the client's credential exec plugin [1].

It allows to work in both interactive and non-interactive modes.
In the first case, the user will be prompted to enter data from the console,
in the second the data will be read from the related environment variables.

[1] https://kubernetes.io/docs/admin/authentication/#client-go-credential-plugins

### DIFF
--- a/pkg/identity/keystone/token_getter.go
+++ b/pkg/identity/keystone/token_getter.go
@@ -1,0 +1,55 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package keystone
+
+import (
+	"fmt"
+
+	"github.com/gophercloud/gophercloud"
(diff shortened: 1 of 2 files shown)


pkg/identity/keystone/token_getter.go
```diff
@@ -0,0 +1,160 @@
...
+		} else {
+			return "", fmt.Errorf("error reading input for %s", field)
+		}
+	}
+	return result, err
+}
+
+// prompt pulls keystone auth url, domain, username and password from stdin.
+// url and domain are not asked if they are defined initially.
+func promt(url string, domain string) (gophercloud.AuthOptions, error) {
```
dims: promt -> prompt

dims:
Can you please add a quick/simple doc in the docs/ directory on how someone would enable and use this plugin?

FengyunPan2:
> Can you please add a quick/simple doc in the docs/ directory on how someone would enable and use this plugin?

+1

Fedosin:
@FengyunPan2 @dims Required documentation was added to the related pull request #126 https://github.com/Fedosin/cloud-provider-openstack/blob/60408d78c7f689d31de9cb72851f5e14c983aedf/docs/using-client-keystone-auth.md

Take a look when you have time :)

PR #126, opened 2018-04-17
Authenticator executable

### LINKED ISSUE #54: Kubectl exec-based client auth provider for Keystone

Fedosin:
This code adds new executable that should be used as a command for the client's
credential exec plugin [1].

It allows to work in both interactive and non-interactive modes.
In the first case, the user will be prompted to enter data from the console,
in the second the data will be read from the related environment variables.

[1] https://kubernetes.io/docs/admin/authentication/#client-go-credential-plugins

fixes #54


cmd/client-keystone-auth/main.go
```diff
@@ -0,0 +1,64 @@
...
+
+func main() {
+	keiData := os.Getenv("KUBERNETES_EXEC_INFO")
+	if keiData == "" {
+		fmt.Fprintln(os.Stderr, "KUBERNETES_EXEC_INFO env variable must be set.")
+		os.Exit(1)
+	}
+
+	kei := KuberneteExecInfo{}
+	json.Unmarshal([]byte(keiData), &kei)
```
dixudx: We should exit the err here when unmarshalling.

cmd/client-keystone-auth/main.go
```diff
@@ -0,0 +1,64 @@
...
+
+	if kei.APIVersion != "client.authentication.k8s.io/v1alpha1" {
```
dixudx: I'd prefer to have a separate validation method.

Also the errors can be aggregated.

dixudx: This kind hard coded version is not k8s style. I'd prefer to import the GVK from `k8s.io/client-go/pkg/apis/clientauthentication` instead.

pkg/identity/keystone/token_getter.go
```diff
@@ -0,0 +1,138 @@
...
+	} else {
+		url, username, password, err := prompt(os.Stdin)
+		if err != nil {
+			return "", err
+		}
+		options = gophercloud.AuthOptions{
+			IdentityEndpoint: url,
+			Username:         username,
+			Password:         password,
+			DomainName:       "default",
```
dixudx: We should not hard coded domain name here.

Also `DomainName` is only applicable for v3. For v2, we use `TenantName`.

Do you mean that only v3 is supported. Or we would support v2 as well.

pkg/identity/keystone/token_getter.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+	}
+
+	client, err := openstack.NewClient(options.IdentityEndpoint)
+	if err != nil {
+		msg := fmt.Errorf("Failed: Initializing openstack authentication client: %v", err)
+		return "", msg
+	}
+	v3Client, err := openstack.NewIdentityV3(client, gophercloud.EndpointOpts{})
```
dixudx: I'd prefer to use `openstack.Authenticate`, which will incorporate v2 and v3 together.

Makefile
```diff
@@ -17,7 +17,6 @@ TESTARGS_DEFAULT := "-v"
 export TESTARGS ?= $(TESTARGS_DEFAULT)
 PKG := $(shell awk  -F "\"" '/^ignored = / { print $$2 }' Gopkg.toml)
 DEST := $(GOPATH)/src/$(GIT_HOST)/$(BASE_DIR)
-DEST := $(GOPATH)/src/$(PKG)
```
FengyunPan2: Please do a rebase.

dims: +1 to rebase. then LGTM

Fedosin:
For some reason we get "cmd/client-keystone-auth/main.go:180:16: undefined: keystone.GetToken". In other words, we cannot use GetToken from k8s.io/cloud-provider-openstack/pkg/identity/keystone in cmd/client-keystone-auth/main.go until it's merged in the master... For this reason I separated my pull requests in two parts: #124 that contains the required function, and #126 with all other code.

If there is a way to get the "cmd" code to see the code from "pkg", then I'll be glad to hear about it.

lingxiankong:
Hi, @Fedosin, I'm not familiar with client-go credential plugin, but I am wondering what's the difference between the functionality of this patch and the openstack auth provider support in kubectl?

PR #128, opened 2018-04-19
Small fixes in controller-manager manifests

gonzolino:
Fix some small issues in controller-manager manifests

DaemonSet manifest:
- Remove double hostNetwork attribute
- Set nodeSelector to only schedule pods on master nodes

Pod manifest:
- Set correct component label


gonzolino:
Looks like the last test failed because of quota: `{"NeutronError": {"message": "Quota exceeded for resources: ['loadbalancer'].", "type": "OverQuota", "detail": ""}}`

I'll try to run it again

PR #135, opened 2018-04-24
Enable building without ln into k8s.io directory

j-griffith:
This PR modifies the Makefile a bit to give us some flexibility and lay some groundwork for future improvements.  The current Makefile does some work to set a GOPATH and then link the current directory in to a k8s.io directory.  This mostly works, but there can be some problems if you for example aren't in a valid gopath, or if you have multiple development repos being worked on simultaneously.

What this change does is instead takes out the GOPATH setup and symbolic linking, and rather requires that you have the code checked out in a valid GOPATH.  We check for this condition in the Makefile and inform the user with an error if this is not the case.

One other thing we do is move all of the built binaries into a single _out directory, just to help with organization a bit.  This update will also allow us to move the build process into a Container more easily if we decide we'd like to go that way.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,18 +6,14 @@
 
 GIT_HOST = k8s.io
 
-PWD := $(shell pwd)
-BASE_DIR := $(shell basename $(PWD))
-# Keep an existing GOPATH, make a private one if it is undefined
-GOPATH_DEFAULT := $(PWD)/.go
-export GOPATH ?= $(GOPATH_DEFAULT)
+PROJECT_DIR := ${CURDIR}
+BASE_DIR := $(shell basename $(PROJECT_DIR))
 GOBIN_DEFAULT := $(GOPATH)/bin
 export GOBIN ?= $(GOBIN_DEFAULT)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -161,7 +155,7 @@ realclean: clean
 	fi
 
 shell: work
-	cd $(DEST) && $(SHELL) -i
+	$(SHELL) -i
 
 images: image-controller-manager image-flex-volume-driver image-provisioner image-csi-plugin image-k8s-keystone-auth
```
gonzolino: The image builds below should also use $(BIN_DIR)

j-griffith: Oh GREAT catch!  I'm trying an experiment right now to see if that BIN_DIR change is screwing up the build, but new to the OpenLab stuff and haven't quite untangled yet.

dims:
@j-griffith wanna pull the plug on this one?

flaper87:
@j-griffith the PR you referenced in your last comment doesn't seem to address this. Were you referring to my PR or another one?

flaper87:
@j-griffith nvm, found it! :)

Issue #136, opened 2018-04-24
Add Snapshot support to Cinder/CSI


j-griffith:
/kind feature

We're getting close to a consensus for the Snapshots spec in CSI:
    https://github.com/container-storage-interface/spec/pull/207

Of course we'll want to implement this in the CSI driver here when it lands, but also in the meantime we'll need to implement the provider layer as well.

So, might as well start implementing an openstack_snapshots pkg and at least roughing in the CSI implementation.

FengyunPan2:
/featrue
Nice featrue!

PR #141, opened 2018-04-25
Add Snapshots to CSI Provider

j-griffith:
This PR adds the implementation for Snapshots support to the CSI/OpenStack provider.  CSI will hopefully be merging Snapshot support soon, and we'll want to update the CSI driver accordingly.

First though, we need to add Snapshot support to the OpenStack Provider for the CSI plugin.  We simply add Create, Delete and List and update our Interface for the OpenStack package.  We also add the new functions to the openstack_mock so that when we add the CSI Controller implementation we can add our tests.

partial #136 

This just exposes the Snapshot functions we need in Gophercloud

Rather than create another abstraction layer over the response types, I'm just leveraging the Snapshot type from Gophercloud directly.


pkg/csi/cinder/openstack/openstack_snapshots.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// ListSnapshots retrieves a list of active snapshots from Cinder for the cooresponding Tenant.  We also
+// provide the ability to provide limit and offset, however for accurate pagination suggest this is handled
+// instead by the caller.  In addition the filters argument provides a mechanism for passing in valid
+// filter strings to the list operation.  Valid filter keys are:  Name, Status, VolumeID (TenantID has no
+// effect)
+func (os *OpenStack) ListSnapshots(limit, offset int, filters map[string]string) ([]snapshots.Snapshot, error) {
+	opts := snapshots.ListOpts{}
+	pages, _ := snapshots.List(os.blockstorage, opts).AllPages()
```
zetaab: I think we should not skip error handling here

pkg/csi/cinder/openstack/openstack_snapshots.go
```diff
@@ -0,0 +1,63 @@
...
+	snaps, _ := snapshots.ExtractSnapshots(pages)
```
zetaab: I think we should not skip error handling here

pkg/csi/cinder/openstack/openstack_snapshots.go
j-griffith: Yep, thanks!

pkg/csi/cinder/openstack/openstack_snapshots.go
j-griffith: Ditto.. done, thanks!

pkg/csi/cinder/openstack/openstack_snapshots.go
```diff
@@ -0,0 +1,74 @@
...
+	if tags != nil {
+		opts.Metadata = *tags
+	}
+
+	snap, err := snapshots.Create(os.blockstorage, opts).Extract()
+	if err != nil {
+		return &snapshots.Snapshot{}, err
+	}
+	// There's little value in rewrapping these gophercloud types into yet another abstraction/type, instead just
+	// return the gohpercloud item
```
zioproto: nit: there is a typo in the word _gophercloud_

pkg/csi/cinder/openstack/openstack_snapshots.go
```diff
@@ -0,0 +1,74 @@
...
+	return snap, nil
+}
+
+// ListSnapshots retrieves a list of active snapshots from Cinder for the cooresponding Tenant.  We also
```
zioproto: nit: check spelling for _ cooresponding_

pkg/csi/cinder/openstack/openstack_snapshots.go
```diff
@@ -0,0 +1,74 @@
...
+		glog.V(3).Infof("Failed call to retrieve snapshots from Cinder: %v", err)
+		return nil, err
+	}
+	snaps, err := snapshots.ExtractSnapshots(pages)
+	if err != nil {
+		glog.V(3).Infof("Failed to extract snapshot pages from Cinder: %v", err)
+		return nil, err
+	}
+	// There's little value in rewrapping these gophercloud types into yet another abstraction/type, instead just
+	// return the gohpercloud item
```
zioproto: nit: _ gohpercloud_ spelling

pkg/csi/cinder/openstack/openstack_snapshots.go
```diff
@@ -0,0 +1,74 @@
...
+// Package openstack.snapshots provides an implementation of Snapshot related
+// cinder functions using Gohpercloud.
+package openstack
+
+import (
+	"github.com/golang/glog"
+	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/snapshots"
+)
+
+// CreateSnapshot issues a request to take a Snapshot of the specified Volume with the cooresponding ID and
```
zioproto: nit: check spelling for _ cooresponding_

pkg/csi/cinder/openstack/openstack_snapshots.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		return &snapshots.Snapshot{}, err
+	}
+	// There's little value in rewrapping these gophercloud types into yet another abstraction/type, instead just
+	// return the gohpercloud item
+	return snap, nil
+}
+
+// ListSnapshots retrieves a list of active snapshots from Cinder for the cooresponding Tenant.  We also
+// provide the ability to provide limit and offset, however for accurate pagination suggest this is handled
```
zioproto: I would write:
We give the possibility to provide limit and offset, so that the caller can handle accurate pagination.

pkg/csi/cinder/openstack/openstack_snapshots.go
```diff
@@ -0,0 +1,74 @@
...
+// instead by the caller.  In addition the filters argument provides a mechanism for passing in valid
+// filter strings to the list operation.  Valid filter keys are:  Name, Status, VolumeID (TenantID has no
+// effect)
+func (os *OpenStack) ListSnapshots(limit, offset int, filters map[string]string) ([]snapshots.Snapshot, error) {
+	opts := snapshots.ListOpts{}
+	pages, err := snapshots.List(os.blockstorage, opts).AllPages()
+	if err != nil {
+		glog.V(3).Infof("Failed call to retrieve snapshots from Cinder: %v", err)
```
zioproto: I would write:
"Failed to retrieve snapshots from Cinder"

the _call_ word does not add any information.

pkg/csi/cinder/openstack/openstack_snapshots.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	// There's little value in rewrapping these gophercloud types into yet another abstraction/type, instead just
+	// return the gohpercloud item
+	return snaps, nil
+
+}
+
+// DeleteSnapshot issues a request to delete the Snapshot with the specified ID from the Cinder backend
+func (os *OpenStack) DeleteSnapshot(snapID string) error {
+	err := snapshots.Delete(os.blockstorage, snapID).ExtractErr()
```
zioproto: You dont print any glog message in case of error ! This will silently fail

pkg/csi/cinder/openstack/openstack_snapshots.go
j-griffith: Huh.. yeah, that's not right :)

pkg/csi/cinder/openstack/openstack_snapshots.go
j-griffith: Gahhh

pkg/csi/cinder/openstack/openstack_snapshots.go
j-griffith: Yep

pkg/csi/cinder/openstack/openstack_snapshots.go
j-griffith: done

pkg/csi/cinder/openstack/openstack_snapshots.go
j-griffith: sure

pkg/csi/cinder/openstack/openstack_snapshots.go
j-griffith: fair enough

pkg/csi/cinder/openstack/openstack_snapshots.go
j-griffith: Well, it wouldn't be a silent failure really, the caller would receive the error and I would expect the error to be dealt with.  I get your point though and I'll add a log message.  Thanks!

pkg/csi/cinder/openstack/openstack_snapshots.go
```diff
@@ -0,0 +1,73 @@
...
+		Name:        name,
+		Description: description,
+	}
+	if tags != nil {
+		opts.Metadata = *tags
+	}
+
+	snap, err := snapshots.Create(os.blockstorage, opts).Extract()
+	if err != nil {
+		return &snapshots.Snapshot{}, err
```
zetaab: we could add log print here as well

pkg/csi/cinder/openstack/openstack_snapshots.go
zetaab: Like I commented to another error: called can take also some responsibility of this? We could log that also in caller function

pkg/csi/cinder/openstack/openstack_snapshots.go
j-griffith: yeah, my opinion was that we probably don't need so much logging in the package, but rather the caller should be grown up and do what needs to be done when it gets an empty Snapshot and an error.  I'm fine either way though and don't mind adding a log message if people prefer it.

PR #142, opened 2018-04-25
Enforce running Makefile from $DEST

dims:
We have had several problems:
- folks checking out to a random directory outside their GOPATH
- CI jobs checking out directory to github.com/kubernetes/xyz

both resulting in problems with build, problems updating dependencies, sometimes things work on the dev environment and fail in CI causing folks to scratch their heads.

So let's standardize on the best practice "this git repo should be checked out under $GOPATH/src/k8s.io" and enforce it. Makefile will bail out when the directory is not correct (either because we checked out to the wrong repo, bad symlinks or wrong PWD)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,17 @@ VERSION ?= $(shell git describe --exact-match 2> /dev/null || \
                  git describe --match=$(git rev-parse --short=8 HEAD) --always --dirty --abbrev=8)
 REGISTRY ?= k8scloudprovider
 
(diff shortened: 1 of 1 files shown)


dims:
@flaper87 @j-griffith thanks for the support for this approach. Please lgtm

kiwik:
OpenLab side PR https://github.com/theopenlab/openlab-zuul-jobs/pull/178

PR #153, opened 2018-05-02
Allow to read username and password from env variables

Fedosin:
Now, in case of interactive session, users have to enter their
logins and passwords every time in the interactive sessions.
This is not very convenient, so it's better to allow to store
this data in env variables and ask to enter the values only if
they are unset.

This commit adds two new flags for the client-keystone-auth executable:
--user-name and --password, which are defaulted to OS_USERNAME and
OS_PASSWORD respectively.

Addtionally, this commit fixes several small issues:
1. Since Stdout is redirected the executable cannot write data there,
   when the user is asked for his login or password. Therefore, Stderr
   should be used for the output in the interactive sessions.
2. The Exec plugin expects to get expiresAt data in RFC3339Nano format,
   otherwise it is ignored. So the executable converts it before printing.
3. Domain name was defaulted to the string "default". But it's better to
   get this value from OS_DOMAIN_NAME env variable.

### DIFF
--- a/cmd/client-keystone-auth/main.go
+++ b/cmd/client-keystone-auth/main.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"time"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
@@ -54,7 +55,9 @@ const respTemplate string = `{
 }`
 
 func promptForString(field string, r io.Reader, show bool) (result string, err error) {
-	fmt.Printf("Please enter %s: ", field)
+	// We have to print output to Stderr, because Stdout is redirected and not shown to the user.
+	fmt.Fprintf(os.Stderr, "Please enter %s: ", field)
+
 	if show {
 		_, err = fmt.Fscan(r, &result)
 	} else {
@@ -62,7 +65,7 @@ func promptForString(field string, r io.Reader, show bool) (result string, err e
 		if terminal.IsTerminal(int(os.Stdin.Fd())) {
 			data, err = terminal.ReadPassword(int(os.Stdin.Fd()))
 			result = string(data)
-			fmt.Println()
+			fmt.Fprintf(os.Stderr, "\n")
 		} else {
 			return "", fmt.Errorf("error reading input for %s", field)
 		}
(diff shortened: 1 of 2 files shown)


cmd/client-keystone-auth/main.go
```diff
@@ -54,53 +55,64 @@ const respTemplate string = `{
 }`
 
 func promptForString(field string, r io.Reader, show bool) (result string, err error) {
-	fmt.Printf("Please enter %s: ", field)
+	// We have to print output to Stderr, because Stdout is redirected and not shown to the user.
+	fmt.Fprintf(os.Stderr, "Please enter %s: ", field)
```
FengyunPan2: Using 'os.Stderr'? Is there other better way to do it?

Fedosin: Good question... We can't use stdout anyway, so we have to invent some sort of hack to show this output to the user.
I'm not proud that we have to do this trick, but I don't see any better solution right now... Leaving the user without any prompt strings is very annoying.

ericchiang: Forgoing the "spirit of stderr and stdout" plenty of other auth tools do this

```
$ gcloud auth login > /dev/null
Your browser has been opened to visit:

    https://accounts.google.com/o/oauth2/auth?redirect_uri=(stuff)


        WARNING: `gcloud auth login` no longer writes application default credentials.
If you need to use ADC, see:
  gcloud auth application-default --help

You are now logged in as [eric.chiang.m@gmail.com].
Your current project is [None].  You can change this setting by running:
  $ gcloud config set project PROJECT_ID
```

I also find it annoying when tools prompt me over stdout. It makes it impossible to do something like:

```
tool-that-requires-passwd | xclip -selection c
```

Your prompt ends up piped into the next tool, which doesn't seem like the intention.

My vote is: leave stdout for things you'd actually want to pipe into the next command. Use stderr from prompting.

dims:
/hold

looks like this needs a bit more work

Fedosin:
@dims what kind of work do you mean?

cmd/client-keystone-auth/main.go
Fedosin: I agree with @ericchiang ! This is just what this code does - uses stderr instead of stdout to print user-facing messages.

I hope @dims and @FengyunPan2 don't have any objections.

Fedosin:
@dims done :)

PR #155, opened 2018-05-03
Makefile targets for cross build and distributions

dims:
example:
`VERSION=0.1.0 make build-cross dist`

The command line above should build a bunch of tars and zips that can
be uploaded when we cut a release.

Code was borrowed from https://github.com/kubernetes/helm/blob/master/Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,10 +21,17 @@ SOURCES := $(shell find $(DEST) -name '*.go')
 HAS_MERCURIAL := $(shell command -v hg;)
 HAS_DEP := $(shell command -v dep;)
 HAS_LINT := $(shell command -v golint;)
+HAS_GOX := $(shell command -v gox;)
+GOX_PARALLEL ?= 3
+TARGETS ?= darwin/amd64 linux/amd64 linux/386 linux/arm linux/arm64 linux/ppc64le
+DIST_DIRS         = find * -type d -exec
 
 GOOS ?= $(shell go env GOOS)
 VERSION ?= $(shell git describe --exact-match 2> /dev/null || \
                  git describe --match=$(git rev-parse --short=8 HEAD) --always --dirty --abbrev=8)
+GOFLAGS   :=
+TAGS      :=
+LDFLAGS   := "-w -s -X 'main.version=${VERSION}'"
 REGISTRY ?= k8scloudprovider
 
 ifneq ("$(DEST)", "$(PWD)")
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -231,5 +237,29 @@ upload-images: images
 version:
 	@echo ${VERSION}
 
+.PHONY: build-cross
+build-cross: LDFLAGS += -extldflags "-static"
+build-cross:
+ifndef HAS_GOX
+	go get -u github.com/mitchellh/gox
+endif
+	CGO_ENABLED=0 gox -parallel=3 -output="_dist/{{.OS}}-{{.Arch}}/{{.Dir}}" -osarch='$(TARGETS)' $(GOFLAGS) $(if $(TAGS),-tags '$(TAGS)',) -ldflags '$(LDFLAGS)' $(GIT_HOST)/$(BASE_DIR)/cmd/openstack-cloud-controller-manager/
```
j-griffith: Any reason to not make the number of parallel compiles configurable?

Makefile
```diff
@@ -231,5 +237,29 @@ upload-images: images
 version:
 	@echo ${VERSION}
 
+.PHONY: build-cross
+build-cross: LDFLAGS += -extldflags "-static"
+build-cross:
```
j-griffith: Doesn't need to be a part of this change, but I'm wondering if we should consider having "make dep" be required task for some of these, or at least a check if vendor exists and just do it if not.  Again, not an issue for this change but this is a good opportunity to capture the idea.

Makefile
j-griffith: output to _dist makes me SOOO HAPPY!

Makefile
```diff
@@ -231,5 +237,29 @@ upload-images: images
...
+.PHONY: dist
+dist:
+	( \
+		cd _dist && \
+		$(DIST_DIRS) cp ../LICENSE {} \; && \
+		$(DIST_DIRS) cp ../README.md {} \; && \
+		$(DIST_DIRS) tar -zcf cloud-provider-openstack-$(VERSION)-{}.tar.gz {} \; && \
+		$(DIST_DIRS) zip -r cloud-provider-openstack-$(VERSION)-{}.zip {} \; \
+	)
+
```
j-griffith: This is great, but we should consider cleaning up all the artifacts.  The way it stands now we end up with 3 items for each dist:
```
<dist>.tar.gz
<dist>.zip
<dist>/
```

I assume by default most that run this are going to be interested in just the tgz, or the zip.  The other option providing the most flexibility I guess is just provide the binaries (no tar, no zip) and then add a "make bundle" option to bundle them up

Makefile
dims: k added dependencies to the build-cross and dist target

Makefile
dims: Done

Makefile
dims: ack, please help tweak it. i just wanted us to get going

flaper87: Great suggestion. I'll push this through so we can iterate over what's here.

flaper87:
The error is not related to this change so I'm sending it through.

/lgtm

PR #165, opened 2018-05-07
Revert "Disable test-lb-octavia openlab job"

dims:
Thanks to Mohammed Naser <mnaser@vexxhost.com>

This reverts commit c4027dc07560fa5f460566d4a154bcf1beb5dc60.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -2,10 +2,9 @@
     cloud-provider-openstack-unittest:
       jobs:
         - cloud-provider-openstack-unittest
-# TODO: re-enable this job after https://github.com/theopenlab/openlab/issues/54 gets fixed
-#    cloud-provider-openstack-acceptance-test-lb-octavia:
-#      jobs:
-#        - cloud-provider-openstack-acceptance-test-lb-octavia
+    cloud-provider-openstack-acceptance-test-lb-octavia:
+      jobs:
+        - cloud-provider-openstack-acceptance-test-lb-octavia
     cloud-provider-openstack-acceptance-test-keystone-authentication-authorization:
       jobs:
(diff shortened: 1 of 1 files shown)


liu-sheng:
not sure why VMs in vexxhost cannot request metadata from 169.254.169.254, which problem cause many jobs failed.

mnaser:
Hey,

Where do you see traces of the metadata service not responding?  The VMs would never go up if the metadata service wasn’t responding, I believe. 

I don’t think these tests depend on the metadata service, or do they 

Thanks,
Mohammed

mnaser:
I just checked this morning and I don't see any issues in a VM that I created

Please let me know if you have any more info

mnaser:
OK, could be something tenant related, checking.

mnaser:
Looks like the specific network had an issue, reloaded metadata and it's responding now.

mrhillsman:
@liu-sheng @animationzl @kiwik lets move to using configdrive instead; we tested this quite a few times and it is intermittent, i cannot see a quicker way to resolve right now.

we can create another label/image with configdrive enabled and use it for these jobs only within VEXXHOST only - https://git.openstack.org/cgit/openstack-infra/project-config/tree/nodepool/nodepool.yaml#n41

liu-sheng:
@mnaser @mrhillsman thanks for help, yes, now I have changed the job to use config driver at first, and will also try connecting with the metadata service as an alternative, see # theopenlab/openlab-zuul-jobs/pull/211

Thanks
Liu sheng

dims:
/assign @flaper87 
/assign @dklyle 

Can one of you please lgtm?

PR #171, opened 2018-05-14
Improve the doc of k8s-keystone-auth

openstacker:

**Release note**:

```release-note
```

### DIFF
--- a/docs/using-keystone-webhook-authenticator-and-authorizer.md
+++ b/docs/using-keystone-webhook-authenticator-and-authorizer.md
@@ -64,8 +64,48 @@ This command creates the following entry in your ~/.kube/config
     auth-provider:
       name: openstack
 ```
-- Run `kubectl config set-context --cluster=kubernetes --user=openstackuser openstackuser@kubernetes`
+- Run `kubectl config set-context --cluster=mycluster --user=openstackuser openstackuser@kubernetes`
 - Run `kubectl config use-context openstackuser@kubernetes` to activate the context
+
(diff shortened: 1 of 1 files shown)


docs/using-keystone-webhook-authenticator-and-authorizer.md
```diff
@@ -64,9 +64,49 @@ This command creates the following entry in your ~/.kube/config
     auth-provider:
       name: openstack
 ```
-- Run `kubectl config set-context --cluster=kubernetes --user=openstackuser openstackuser@kubernetes`
+- Run `kubectl config set-context --cluster=<your cluster name> --user=openstackuser openstackuser@kubernetes`
```
gonzolino: In this command you use "<your cluster name>" as cluster name, in the kubeconfig below you use "mycluster".
It's just a personal opinion, but I would prefer to use the same name in both examples.

openstacker: @gonzolino Thanks for the comments, do you mean using --cluster=mycluster at here?

gonzolino: Yes that would be my suggestion. Or, alternatively, you could use "<your cluster name>" in the kubeconfig example.
Either way is ok for me, but I think it should be consistent.

openstacker: Ok, cool. I will do that. The current '--cluster=kuberntes' is very confusing, hence why I'd like to update it. Thanks. Will propose a new patch set.

PR #172, opened 2018-05-14
Sync latest master to cpo may 14

dims:
Sync a few commits from master that did not make it here yet


dims:
@gonzolino @zetaab @wenlxie @databus23 Can you please verify if the changes (that you made in kubernetes/kubernetes master) that i am copying over to this repository are good?

wenlxie:
@dims  My fix is ok. Thanks

PR #176, opened 2018-05-16
Restore pre 1.10 openstack instance naming behavior

dims:
Sync with the change from @liggitt in k/k master - https://github.com/kubernetes/kubernetes/pull/63903/commits


dims:
/assign @hogepodge 
/assign @dklyle 

cc @NickrenREN @dixudx @FengyunPan2

PR #177, opened 2018-05-16
Track coverage using codecov.io

dims:
We don't really have a good idea about the code coverage using our unit tests. The first step is to update the Makefile, need to figure out how to add support to our Zuul v3 jobs to talk to codecov.io

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,6 +127,18 @@ vet:
 
 cover: depend
 	go test -tags=unit $(shell go list ./...) -cover
+	@rm -rf coverage.txt
+	@for d in `go list ./... | grep -v vendor`; do \
+		t=$$(date +%s); \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -127,6 +127,18 @@ vet:
...
+		t=$$(date +%s); \
+		go test -coverprofile=cover.out -covermode=atomic $$d || exit 1; \
+		echo "Coverage test $$d took $$(($$(date +%s)-t)) seconds"; \
+		if [ -f cover.out ]; then \
+			cat cover.out >> coverage.txt; \
+			rm cover.out; \
+		fi; \
+	done
+	@echo "Uploading coverage results..."
+	@curl -s https://codecov.io/bash | bash
```
gonzolino: Are we sure we want to do this? Piping untrusted code to bash seems dangerous to me :P

PR #179, opened 2018-05-17
Cinder provisioner fixes

aglitke:
**What this PR does / why we need it**: Apply a few fixes that have been queued while the provisioner code was moved to this repo from the kubernetes-incubator.  I would like to focus my development effort now on this repository instead of the old location.

**Which issue this PR fixes**: This PR fixes the provisioning flow when cinder.create takes longer than 5 seconds.  It also makes it easier to use certain iscsi storage where initiatorName is verified during volume attachment.  The rest of the commits make small fixes and refactors to prepare for the addition of cloning support.

**Release note**:

```release-note
Usability and reliability improvements to the cinder-provisioner.
```

### DIFF
--- a/pkg/volume/cinder/provisioner/iscsi.go
+++ b/pkg/volume/cinder/provisioner/iscsi.go
@@ -25,6 +25,7 @@ import (
 )
 
 const iscsiType = "iscsi"
+const initiatorName = "iqn.2018-01.io.k8s:a13fc3d1cc22"
 
 type iscsiMapper struct {
 	volumeMapper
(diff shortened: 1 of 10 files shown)


aglitke:
@hogepodge This my first attempt at merging anything substantive into this repository since the cinder provisioner code moved here.  Is there anything else I can do to help this one along?

pkg/volume/cinder/provisioner/iscsi.go
```diff
@@ -25,6 +25,7 @@ import (
 )
 
 const iscsiType = "iscsi"
+const initiatorName = "iqn.2018-01.io.k8s:a13fc3d1cc22"
```
hogepodge: Can you help me understand why this name is hard coded?

j-griffith: @hogepodge You mean rather than say something like generating something unique per cluster or making this a variable?  I guess a variable with a default to ensure they match might be a good option here.

aglitke: This stems from the way we handle volume attachment in the standalone cinder provisioner.  After we create a volume, the provisioner calls the os-initialize_connection API in cinder to retrieve the connection information for the volume.  Then, we translate that information into a native kubernetes PV if it uses a backend we support (currently iscsi and rbd).  When the volume needs to be attached to a node kubernetes handles this using the information we populated in the PV.  Since at provision time we can't know which node will be attached we hard-code an initiatorName and ask kubernetes to always use this initiatorName (instead of the host-specific one).  This has the effect of bypassing initiatorName-based controls that might be used by the storage (eg, xtremio).  The advantage is that the administrator does not need to manually edit /etc/iscsi/initiatorname.iscsi on each node in the cluster in order to use this provisioner.

PR #182, opened 2018-05-22
Manila external provisioner

gman0:
Adds Manila external provisioner

Originated from [kubernetes-incubator/external-storage](https://github.com/kubernetes-incubator/external-storage). Some of the advantages are: ability to handle different share types, OpenStack credentials are passed per-StorageClass instead of per-provisioner.
I'll add unit tests in the following commits, I'd just like to have some feedback if the design is ok.

Tested with a Ceph cluster + [csi-cephfs](https://github.com/ceph/ceph-csi) and Kubernetes' CephFS mounter.

**Release note**:

```release-note
NONE
```

/cc @dims


dims:
cc @tomassedovic @flaper87

flaper87:
@gman0 hey, thanks a bunch for this new plugin.

As a general knowledge question: Why did you pick this approach over building a CSI based driver for Manila (like the Cinder one)? Just curious to know the motivation.

gman0:
@flaper87 Mainly for separation of concerns and less code duplication. The `Provisioner` interface seems a bit better suited for this use case - only create the share, and let something else take care of the resulting `PersistentVolume`.

Now, if CSI could accommodate drivers that only do`CreateVolume/DeleteVolume` (i.e. the `Provisioner` interface) and then forward the `CreateVolumeResponse` to another CSI driver, we would get this exact same pattern. I'll ask at the CSI spec repo if that's something worth pursuing and if it's not then yes - in the long run it would be better to do it as csi-cinder does, since all these storage-handling drivers will move to CSI eventually.

dims:
/approve
/lgtm

Let's get it in and iterate!

gman0:
Thanks a lot guys!

PR #183, opened 2018-05-22
Add cloning via PVC annotation

aglitke:
This PR enables provisioning new volumes as a clone of an existing PVC.  To
request a clone, the user should add a clone request annotation to the PVC:

```
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: clone-claim
  annotations:
    k8s.io/CloneRequest: default/source-pvc
```

The provisioner will locate the indicated PVC and determine the underlying
associated cinder volume ID.  The provisioner will use this ID as the
source-vol-id parameter when provisioning the new volume resulting in a clone
at the storage level.

Cloning will only work if the cinder backend supports "smart-cloning".  This
means that the clone will complete without the need for cinder to perform a
bit-for-bit copy using the host.  Because of this, the cinder provisioner will
not attempt to clone unless the associated storage class contains the
parameter: "smartclone" to indicate that volumes provisioned from this storage
class can be cloned correctly.

If the provisioner was able to clone the volume it will apply the
'k8s.io/CloneOf' annotation to the PVC.

### DIFF
--- a/pkg/volume/cinder/provisioner/clusterbroker.go
+++ b/pkg/volume/cinder/provisioner/clusterbroker.go
@@ -17,14 +17,20 @@ limitations under the License.
 package provisioner
 
 import (
+	"fmt"
+
 	"github.com/golang/glog"
 	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
 )
 
 // clusterBroker provides a mechanism for tests to override calls kubernetes with mocks.
 type clusterBroker interface {
 	createSecret(p *cinderProvisioner, ns string, secret *v1.Secret) error
 	deleteSecret(p *cinderProvisioner, ns string, secretName string) error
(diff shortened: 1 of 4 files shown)


aglitke:
/cc @hogepodge 
Can you mark this one ok-to-test?

zetaab:
if the cinder backend supports "smart-cloning" - how I can see does our cinder backend support this? Is this using basically volume clone command?

aglitke:
@zetaab The short answer is "depends on the configuration".  I define "smart-cloning" as the ability to provision a new volume clone utilizing exclusively the storage backend.  When a source volume id is passed into the create verb cinder will request a clone from the backend storage if supported.  Otherwise, cinder will perform a host-assisted clone.  Host assisted clone is also needed when cloning across volume types.  Cinder host-assisted clone will obviously not work well in a kubernetes context so we want to avoid this.  This is why this patch requires the smartclone: true parameter to be added to the storage class being serviced by the provisioner.

dims:
/approve
/lgtm

@aglitke please add some notes from here to the docs/

PR #188, opened 2018-06-01
[WIP]Pip quiet install in Makefile

kiwik:
Add "--quiet" option for pip install Mercurial to avoid
stdio to be interrupted, that cause incompleteness of OpenLab log.

@dims


kiwik:
Fix the issue in OpenLab side, close the PR.

PR #190, opened 2018-06-01
Add configmap support for k8s-keystone-auth

### LINKED ISSUE #189: ConfigMap support for k8s-keystone-auth
**Is this a BUG REPORT or FEATURE REQUEST?**: /kind feature

**What happened**: N/A

**What you expected to happen**: Allow the admin user to define the authorization policy by using configmap.

**How to reproduce it (as minimally and precisely as possible)**: N/A

**Anything else we need to know?**: N/A

lingxiankong:
**What this PR does / why we need it**: Add configmap support for k8s-keystone-auth

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #189

With this change, the k8s-keystone-auth can accept either policy config
file or a configmap name in 'kube-system' namespace, because in real
deployment, it's admin user's responsibility to update the policy file
for authorization.

In order to connect to k8s api, either a kubeconfig file or service
account should be provided.

The next step is to support on-the-fly change for the policy without
restarting k8s-keystone-auth service.

**Release note**:

```release-note
```


Fedosin:
Thanks for the patch! I'm looking.

pkg/identity/webhook/webhook.go
```diff
@@ -171,11 +171,17 @@ func (h *WebhookHandler) authorizeToken(w http.ResponseWriter, r *http.Request,
...
 	}
 
-	allowed, reason, err := h.Authorizer.Authorize(attrs)
-	glog.V(4).Infof("<<<< authorizeToken: %v, %v, %v\n", allowed, reason, err)
-	if err != nil {
-		http.Error(w, reason, http.StatusInternalServerError)
-		return
+	var allowed authorizer.Decision
+	if h.Authorizer != nil {
+		allowed, reason, err := h.Authorizer.Authorize(attrs)
```
Fedosin: you should use = instead of := here. Otherwise "allowed" value won't be seen outside of this section and always be equal 0 (i.e. DecisionDeny)
The tests fail because of this.

lingxiankong: Yeah, you are right, thanks!

Issue #192, opened 2018-06-05
Allow to synchronize Keystone projects with k8s namespaces


Fedosin:
To improve integration between keystone and k8s it's recommended to add a possibility to synchronize projects/namespaces.

In other words, if the user belongs to the project in Keystone, then when attempting to authenticate in Kubernetes using [k8s-keystone-auth](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/using-keystone-webhook-authenticator-and-authorizer.md), a k8s namespace with the same name as the project id will automatically be created for him.

/kind feature

lingxiankong:
Thanks for tracking this, @Fedosin

PR #197, opened 2018-06-10
Support to change the policy configmap for k8s-keystone-auth on the fly

lingxiankong:
**What this PR does / why we need it**:
Support to change the policy configmap for k8s-keystone-auth on the fly

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #195

**Special notes for your reviewer**:
- Watch the configmap change for authorization policy and apply to the
  service dynamically.
- Improve the shape of k8s-keystone-auth implementation

**Release note**:

```release-note
The authorization functionality of k8s-keystone-auth service can be running either by specifying a policy file or a configmap in kube-system namespace. If the configmap is provided, the authorization policy could be changed during the service running without service restart.
```

PR #198, opened 2018-06-11
Allow synchronization between Keystone projects and K8s namespaces

Fedosin:
To improve integration between keystone and k8s it's recommended
to add a possibility to synchronize projects/namespaces.

In other words, if the user belongs to the project in Keystone, then, when
he attepts to authenticate in Kubernetes using k8s-keystone-auth, a k8s
namespace with the same name as the project id will automatically be
created for him.

To enable the feature two new arguments were added to the k8s-keystone-auth binary:
"--sync-config-file" - points to a local file with sync configuration.
"--sync-configmap-name" defines a configmap containing the configuration.
If any of them is set, new namespace will be automatically created when the user is
authenticated in Keystone.

fixes: #192

### DIFF
--- a/pkg/identity/keystone/keystone.go
+++ b/pkg/identity/keystone/keystone.go
@@ -28,6 +28,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack"
 	"github.com/gophercloud/gophercloud/openstack/utils"
 	"github.com/gorilla/mux"
+	"gopkg.in/yaml.v2"
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -113,7 +114,7 @@ func (k *KeystoneAuth) enqueueConfigMap(obj interface{}) {
 		return
 	}
 
-	if namespace == cmNamespace && name == k.config.PolicyConfigMapName {
+	if namespace == cmNamespace && (name == k.config.PolicyConfigMapName || name == k.config.SyncConfigMapName) {
 		k.queue.Add(key)
 	}
 }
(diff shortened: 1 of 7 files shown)


cmd/k8s-keystone-auth/main.go
```diff
@@ -58,7 +59,8 @@ func main() {
 	pflag.StringVar(&keystoneCaFile, "keystone-ca-file", os.Getenv("KEYSTONE_CA_FILE"), "File containing the certificate authority for Keystone Service.")
 	pflag.StringVar(&policyFile, "keystone-policy-file", os.Getenv("KEYSTONE_POLICY_FILE"), "File containing the policy, if provided, it takes precedence over the policy configmap.")
 	pflag.StringVar(&configMap, "policies-configmap-name", "", "ConfigMap in kube-system namespace containing the policy configuration, the ConfigMap must contain the key 'policies'")
-	pflag.StringVar(&kubeConfig, "kubeconfig", "", "Kubeconfig file to connect to Kubernetes API to get policy configmap.")
+	pflag.StringVar(&syncConf, "syncconfig", "", "File containing config values for data synchroniztion beetween Keystone and Kubernetes.")
```
zioproto: spelling: synchronization

lingxiankong:
Hi, @Fedosin do you mind rebasing this change on #197? Since the `--kubeconfig` option is already supported in that patch and the code structure has been polished a lot.

Fedosin:
@lingxiankong done :) thank you for your work, it helps a lot!

pkg/identity/keystone/sync.go
```diff
@@ -0,0 +1,28 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package keystone
+
+// SyncConfig contains a config for data synchronization between Keystone and Kubernetes
+type SyncConfig struct {
+	// TODO(mfedosin) Add more config options.
```
Fedosin: There are several ideas what we can configure here:
1. What data we want to sync (this will be relevant when we also start syncing user roles).
2. Various black lists: by id, by name, by projects (for the roles), by domain.
3. namespace naming format. now it's just a project id, but I think this could be configurable to be more mnemonic.

dims:
is this ready to merge @Fedosin ? looks like some more stuff left over above

Fedosin:
Yes, it's still WIP and requires some updates. I just uploaded the code so that @zioproto could test it. But now, when #197 is merged, I am able to add the required changes.

If you have ideas what else we can add to the config, please share :)

hogepodge:
/hold
Until ready to merge. Thanks for you work on this @Fedosin

zioproto:
I tested this PR and it works for me. The namespace is created automatically.

pkg/identity/keystone/keystone.go
```diff
@@ -149,6 +150,41 @@ func (k *KeystoneAuth) processNextItem() bool {
...
+	glog.Info("ConfigMap created or updated, will update the sync configuration.")
+
+	var sc *syncConfig
+	if err := json.Unmarshal([]byte(cm.Data["syncConfig"]), sc); err != nil {
+		runtimeutil.HandleError(fmt.Errorf("failed to parse sync config defined in the configmap %s: %v", key, err))
+	}
+
+	k.authz.mu.Lock()
+	k.authn.syncConfig = sc
+	k.authz.mu.Unlock()
```
lingxiankong: Not sure if it's the best practice to use mux from a struct when updating fields of another struct.

Fedosin: Lol :) You're right! for sure we should lock authn, not authz

PR #213, opened 2018-06-20
Support Cinder API version configuration

### LINKED ISSUE #212: Support Cinder API version configuration in CSI plugin

/kind feature

**What happened**: 

There are lots of OpenStack users who are using Cinder v1 and Cinder v2 API. 
But now the Cinder CSI Plugin is only supporting Cinder V3.  For the end users, it has no way to set other Cinder API version like v1 and v2.

Cinder version should be supported in Cinder CSI Plugin.
It can be configured like [kubernetes in tree cloud provider](https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/).

For example:
```
[BlockStorage]
bs-version=v2
```

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

edisonxiang:

Support Cinder API version configuration in Cinder CSI Plugin.
It can be configured in the cloud.conf.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #212

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/etc/cloud.conf
+++ b/pkg/csi/cinder/etc/cloud.conf
@@ -4,3 +4,6 @@ password=pass
 auth-url=https://<keystone_ip>/identity/v3
 tenant-id=c869168a828847f39f7f06edd7305637
 domain-id=2a73b8f597c04551a0fdc8e95544be8a
+
+[BlockStorage]
+bs-version=v3
--- a/pkg/csi/cinder/openstack/openstack.go
+++ b/pkg/csi/cinder/openstack/openstack.go
@@ -17,12 +17,13 @@ limitations under the License.
 package openstack
 
 import (
+	"errors"
+	"fmt"
 	"os"
 
 	"github.com/golang/glog"
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
-	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/snapshots"
 	"gopkg.in/gcfg.v1"
 )
 
@@ -35,14 +36,14 @@ type IOpenStack interface {
 	WaitDiskDetached(instanceID string, volumeID string) error
 	GetAttachmentDiskPath(instanceID, volumeID string) (string, error)
 	GetVolumesByName(name string) ([]Volume, error)
-	CreateSnapshot(name, volID, description string, tags *map[string]string) (*snapshots.Snapshot, error)
-	ListSnapshots(limit, offset int, filters map[string]string) ([]snapshots.Snapshot, error)
+	CreateSnapshot(name, volID, description string, tags *map[string]string) (*Snapshot, error)
+	ListSnapshots(limit, offset int, filters map[string]string) ([]Snapshot, error)
 	DeleteSnapshot(snapID string) error
 }
 
 type OpenStack struct {
-	compute      *gophercloud.ServiceClient
-	blockstorage *gophercloud.ServiceClient
+	compute *gophercloud.ServiceClient
+	volumes volumeService
 }
 
 type Config struct {
@@ -57,6 +58,12 @@ type Config struct {
 		DomainName string `gcfg:"domain-name"`
 		Region     string
 	}
+	BlockStorage BlockStorageOpts
+}
+
+// BlockStorageOpts is used to talk to Cinder service
+type BlockStorageOpts struct {
+	BSVersion string `gcfg:"bs-version"`
 }
 
 func (cfg Config) toAuthOptions() gophercloud.AuthOptions {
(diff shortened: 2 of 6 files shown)


flaper87:
I'm actually a bit torn on this one. Do we want for old versions to be used? (especially v1)

What value would this bring (besides supporting environments with older versions) compared to the maintenance cost of those APIs?

I can see the value of this if there was a new Cinder API but I really hope there won't be :smile: 

@j-griffith thoughts?

edisonxiang:
@flaper87 thanks your reply. The value I see is that some users are using cinder v2.  We could not give up the users who is using the old cinder API. Actually I also totally agree your view point:)
> I can see the value of this if there was a new Cinder API

flaper87:
@edisonxiang mmh, could we then limit this to cinder version >=2? I'm still a bit unsure on whether it's worth adding this code now or just wait until everything else matures a bit before we do it. We don't really have many users of this plugin yet (AFAIK, at least).

j-griffith:
I really don't want to mix in support for V1 or V2 API's here.  The thing is; V1 has been out of support for "years" now.  I'd strongly object to adding the support burden of that.

As far as V2, it's sort of pointless in terms of functionality (V3 is a superset), but it is possible the deployment is still stuck back on Pre Newton I guess?  IMO thought that's still a >2 year old deployment and I'm not sure about supporting it with CSI.  We have enough challenges getting V3 and microversions all sorted IMO.

:-1:

flaper87:
@j-griffith yeah, exactly the feedback I was looking for and I fully agree with you.

@edisonxiang Do you agree with our comments here?

embik:
Hey @dims @j-griffith @edisonxiang, we don't have cinder v3 running yet and I would like to see this being discussed further. The CSI plugin is basically my last hope for dynamic provisioning and I don't have any influence on cinder v3 being activated.

j-griffith:
@edisonxiang 
> We may not support cinder v1. But cinder v2 support is very necessary for the end users.
So I can remove codes about cinder v1 and remain cinder v2.
Do you agree?

Seems like a reasonable compromise, the one detail I was trying to point out though is that in fact **_V3.0 is V2_**; so rather than mess with that nuance just implement V3, we can "make it work" with a V2 deployment.  Looking forward that means that we *only* have to deal with micro versions going forward.  

@embik points out
> we don't have cinder v3 running yet and I would like to see this being discussed further. The CSI plugin is basically my last hope for dynamic provisioning and I don't have any influence on cinder v3 being activated.

I think we could solve that internally but that might be "too much magic" so I'm good with just dropping V1 and implementing V2 and V3

dims:
So, @edisonxiang what do we want to do here?

dims:
/close

xlucas:
Is this PR still being considered? If so, can I help ?

adisky:
@dims @flaper87 @edisonxiang @j-griffith ^

nabheet:
Is there any way we could get this re-opened and merged? I am not sure my employer will ever upgrade to v3. It would be wonderfully great to have this support Cinder V2. I would help but it might a bit beyond my skills at this point.

adisky:
@dims @flaper87 @j-griffith ^ It seems many people wants v2 support!!

dims:
cc @hogepodge

hogepodge:
Reopening for discussion.

jianglingxia:
i hope commity can supprot v2,many people now use the v2 version and maybe in short time to upgrade  to v3,thanks

nabheet:
Sorry to repeatedly harass you guys - can someone pretty please with sugar on top merge this PR? Or help fix the tests? I am not sure how much time @edisonxiang has to devote to this. But y’all would be a life saver if V2 support gets included in CSI Cinder. I woul offer a bribe but I am not sure how ethical it would be. 😀

nabheet:
Yeah I saw that ... I was making a general request for someone to help fix the issues, without using those words directly. 😀

I was politely pleading primarily because it will take me a while to upgrade my skills sets to fix the tests.

nabheet:
Or if it’s just merge conflicts maybe I could fork and do it in a separate PR ... I should be able to look at it next week.

adisky:
@nabheet give me sometime, I will try to fix it

jianglingxia:
excuse me! i see the code there has computev2 package to call , the issue mean that csi plugin only support v3 blockstorage but support nova v2 version? is that right?thanks all!

nabheet:
@adisky were you able to spend any time on this issue?

adisky:
@nabheet  sorry I couldn't get time to work on this, if you still want this, Open an issue in this repo, and tag it as help wanted!!

adisky:
as https://github.com/kubernetes/cloud-provider-openstack/issues/212, this issue is stale now, and it is for V1 and V2, after discussions here, we agreed to support V2, so it is better to open a new issue

adisky:
/close 
no update from long time

PR #219, opened 2018-06-25
Sync role assignments

### LINKED ISSUE #45: Kubernetes RBAC <-> OpenStack Keystone - sync
Can we write a controller (CRD?) to make it easier to refresh RBAC info from Keystone information?

cc @Fedosin

Fedosin:
When new namespace is automatically created from a Keystone project, new role
bindings are also should be created to allow the user having the same rights
as in the keystone project.

fixes #45


zioproto:
@Fedosin I tested and it works for me !

PR #221, opened 2018-07-02
Format tables in manila provisioner doc

adisky:
Tables used to describe configuration options in Manila provisoner 
document are not easily readable, This patch corrects the 
table formatting

This PR fixes table formatting in manila provisioner doc

**Release note**:

```release-note
```


hogepodge:
One possibility for fixing the content is to add a new 'required' column that has a check if the parameter is required. Then the default can be reserved for the default value and the description can include acceptable values.

Another possibility would be to follow the formatting in this PR:
https://github.com/kubernetes/cloud-provider-openstack/pull/224

PR #223, opened 2018-07-03
Add cluster name to LB description

### LINKED ISSUE #222: Support to differentiate cloud load balancers created for 'LoadBalancer' type services
/kind feature

**What happened**:
In current k8s deployment on openstack, the only way to get all the load balancers created for the services of `LoadBalancer` type for a specific k8s cluster is:

- Get all the services in that k8s cluster
- 'Calculate' the load balancer name according to the code [here](https://github.com/kubernetes/kubernetes/blob/a3d30dc939c102be2eec3d7d5c7f33f63a1e9893/pkg/cloudprovider/cloud.go#L67)

which is not an easy task for the end users or operators.

As an end users or operators, I can easily get all the load balancers created for a specific k8s cluster. 

Suggestion: As the cloud provider, we could set the param `--cluster-name` of controller manager for different k8s clusters and inject the cluster name in the load balancer description.

N/A

N/A

**Environment**:
N/A

gonzolino:
Add cluster name to the description of Loadbalancers to help operators with the differentiation of clusters.
The cluster name is passed to the cloud-controller-manager using the `--cluster-name` parameter. If it is not set 'kubernetes' is used as default cluster name.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #222

**Release note**:

```
Add cluster name to LB description
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
@@ -440,10 +440,10 @@ func createNodeSecurityGroup(client *gophercloud.ServiceClient, nodeSecurityGrou
 	return nil
 }
 
-func (lbaas *LbaasV2) createLoadBalancer(service *v1.Service, name string, internalAnnotation bool) (*loadbalancers.LoadBalancer, error) {
+func (lbaas *LbaasV2) createLoadBalancer(service *v1.Service, name, clusterName string, internalAnnotation bool) (*loadbalancers.LoadBalancer, error) {
 	createOpts := loadbalancers.CreateOpts{
 		Name:        name,
-		Description: fmt.Sprintf("Kubernetes external service %s", name),
+		Description: fmt.Sprintf("Kubernetes external service %s from cluster %s", name, clusterName),
 		VipSubnetID: lbaas.opts.SubnetID,
 		Provider:    lbaas.opts.LBProvider,
 	}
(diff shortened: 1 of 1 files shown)


lingxiankong:
The change looks good to me, thanks @gonzolino

pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
```diff
@@ -440,10 +440,10 @@ func createNodeSecurityGroup(client *gophercloud.ServiceClient, nodeSecurityGrou
 	return nil
 }
 
-func (lbaas *LbaasV2) createLoadBalancer(service *v1.Service, name string, internalAnnotation bool) (*loadbalancers.LoadBalancer, error) {
+func (lbaas *LbaasV2) createLoadBalancer(service *v1.Service, name, clusterName string, internalAnnotation bool) (*loadbalancers.LoadBalancer, error) {
 	createOpts := loadbalancers.CreateOpts{
 		Name:        name,
-		Description: fmt.Sprintf("Kubernetes external service %s", name),
+		Description: fmt.Sprintf("%s external service %s", clusterName, name),
```
FengyunPan2: Can we use that:
Description: fmt.Sprintf("Kubernetes external service %s from cluster %s", name, clusterName),

FengyunPan2: BTW, I want to add 'clusterName' into the description of cloudprovider's resources, such as LoadBalancer, cinder volume, Securty Group, vip, Floatingip ...

FengyunPan2: Just like L1058.

PR #226, opened 2018-07-09
Manila: provisioning existing shares

gman0:
Makes it possible to use existing Manila shares in the manila external-provisioner: the user supplies the share name/ID and the share access ID in StorageClass params.
This has implications on dealing with `reclaimPolicy`: if the share has been created by the external-provisioner, it will be deleted if `reclaimPolicy: Delete`. If the user has supplied share name/ID, it won't delete the underlying share regardless of the reclaim policy.

Added 3 new PV annotations:
- Provision type: can be either dynamic or static (newly created share or an existing one)
- ShareSecretName, ShareSecretNamespace: secret ref for the share secrets

A big chunk of code deals with `shareoptions`, which now supports `value` tag with these options: `optional`, `default=`, `coalesce=`, `requires=`

**Release note**:

```release-note
NONE
```

/cc @dims @flaper87


gman0:
It seems like the version of `gophercloud` library is too old, the CI is reporting undefined `shares.IDFromName` - this got in 2 days ago.

pkg/share/manila/shareoptions/common.go
```diff
@@ -18,11 +18,16 @@ package shareoptions
 
 // CommonOptions contains options common for any backend/protocol
 type CommonOptions struct {
-	Zones    string `name:"zones"`
-	Type     string `name:"type"`
+	Zones    string `name:"zones" value:"default=default"`
+	Type     string `name:"type" value:"default=nova"`
```
gman0: oops...defaults for `zones` and `type` are swapped

gman0:
Any news on this?

gman0:
Thanks for the merge!

PR #234, opened 2018-07-19
Sync with Kubernetes master repo

dims:

- We had commits in k/k that were not applied here. 10 commits in total.
- Switched from 10.x versions to 11.x versions for kubernetes dependencies
- Removed some unnecessary files (vendored them instead of having a copy that we need to keep in sync)
- Removed a temporary `getVolumeSource` and use the `getVolumeInfo` as in the main k/k master

**Release note**:

```release-note
NONE
```


lingxiankong:
thanks for the synchronization, those commits look reasonable to me.
/lgmt

PR #236, opened 2018-07-20
Add keystone-webhook-auth sample for ubuntu 16.04

oomichi:

Sometimes we had questions how to configure keystone-webhook-auth
on ubuntu 16.04, because that is a little far from current doc
explanation. So this adds the sample configuration for ubuntu 16.04
users.

**Release note**: `NONE`


zioproto:
Hello @oomichi thanks for your contribution.
Why this documentation is specific to Ubuntu ? It looks generic to me. I think we can remove the reference to Ubuntu.

oomichi:
@zioproto
because I am testing the feature on ubuntu only and I am not sure the file paths of configurations under /etc are the same on the other distributions.
If you are sure that, I'd like to remove ubuntu from the doc :-)

dims:
Let's leave this as-is for now. 

/lgtm
/approve

PR #239, opened 2018-07-23
Add documentation for k8s-keystone-auth

lingxiankong:
Add the missing documentation for k8s-keystone-auth service. It only covers how to config and run the k8s-keystone-auth as a k8s service rather than the static pod.

N/A

N/A

**Release note**:

```release-note
```


lingxiankong:
Seems we have some issue in CPO CI.

PR #241, opened 2018-07-24
Fix authentication error for client-auth-plugin

### LINKED ISSUE #240: client-keystone-auth: Project name is missing for authentication
/kind bug

**What happened**:
client-keystone-auth doesn't work

client-keystone-auth should work after kubeconfig configuration.

1. `kubectl config set-credentials openstackuser-clientauth`
2. Edit ~/.kube/config file, add the information client-keystone-auth needs

```yaml
users:
- name: openstackuser-clientauth
  user:
    exec:
      command: "/home/ubuntu/client-keystone-auth"
      apiVersion: "client.authentication.k8s.io/v1alpha1"
      env:
        - name: "OS_USERNAME"
          value: "demo"
        - name: "OS_PASSWORD"
          value: "password"
        - name: "OS_PROJECT_NAME"
          value: "demo"
        - name: "OS_TENANT_NAME"
          value: "demo"
        - name: "OS_REGION_NAME"
          value: "RegionOne"
        - name: "OS_IDENTITY_API_VERSION"
          value: "3"
        - name: "OS_DOMAIN_NAME"
          value: "default"
      args:
        - "--keystone-url=http://10.140.81.86/identity/v3"
        - "--domain-name=default"
```

3. In webhook service policy definition, allow user `demo` with `kube-viewer` role to get pods
4. Run `kubectl get pods`, the authorization failed.

```shell
$ kubectl get po
Keystone auth url is "http://10.140.81.86/identity/v3"
Domain name is "default"
User name is "demo"
Password is set
Error from server (Forbidden): pods is forbidden: User "demo" cannot list pods in the namespace "default"
```

The reason is the token client-keystone-auth gets from keystone is an unscoped token because the project name is not specified when fetching token.

lingxiankong:
Fix authentication error for client-auth-plugin

fixes #240

With this patch, the client-keystone-auth could work the same with the traditional way. You can config the kubeconfig file as the following, you can use environment variables and don't need to add any params:

```yaml
users:
- name: openstackuser-clientauth
  user:
    exec:
      command: "/home/ubuntu/client-keystone-auth"
      apiVersion: "client.authentication.k8s.io/v1alpha1"
```

Then run `kubectl get pods`.

**Release note**:

```release-note
```

### DIFF
--- a/cmd/client-keystone-auth/main.go
+++ b/cmd/client-keystone-auth/main.go
@@ -75,7 +75,7 @@ func promptForString(field string, r io.Reader, show bool) (result string, err e
 
 // prompt pulls keystone auth url, domain, username and password from stdin,
 // if they are not specified initially (i.e. equal "").
-func prompt(url string, domain string, user string, password string) (gophercloud.AuthOptions, error) {
+func prompt(url string, domain string, user string, project string, password string) (gophercloud.AuthOptions, error) {
 	var err error
 	var options gophercloud.AuthOptions
 
@@ -84,8 +84,6 @@ func prompt(url string, domain string, user string, password string) (gopherclou
 		if err != nil {
 			return options, err
 		}
-	} else {
-		fmt.Fprintf(os.Stderr, "Keystone auth url is \"%s\"\n", url)
 	}
 
 	if domain == "" {
@@ -93,8 +91,6 @@ func prompt(url string, domain string, user string, password string) (gopherclou
 		if err != nil {
 			return options, err
 		}
-	} else {
-		fmt.Fprintf(os.Stderr, "Domain name is \"%s\"\n", domain)
 	}
 
 	if user == "" {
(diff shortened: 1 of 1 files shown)


cmd/client-keystone-auth/main.go
```diff
@@ -150,13 +150,15 @@ func main() {
...
 	var user string
+	var project string
 	var password string
 	var options gophercloud.AuthOptions
 	var err error
 
 	pflag.StringVar(&url, "keystone-url", os.Getenv("OS_AUTH_URL"), "URL for the OpenStack Keystone API")
 	pflag.StringVar(&domain, "domain-name", os.Getenv("OS_DOMAIN_NAME"), "Keystone domain name")
 	pflag.StringVar(&user, "user-name", os.Getenv("OS_USERNAME"), "User name")
+	pflag.StringVar(&project, "project-name", os.Getenv("OS_PROJECT_NAME"), "Keystone project name")
```
oomichi: The environment variable OS_PROJECT_NAME is valid for the project name in OpenStack as http://git.openstack.org/cgit/openstack/python-openstackclient/tree/README.rst#n85 so LGTM

cmd/client-keystone-auth/main.go
```diff
@@ -84,40 +84,40 @@ func prompt(url string, domain string, user string, password string) (gopherclou
 		if err != nil {
 			return options, err
 		}
-	} else {
-		fmt.Fprintf(os.Stderr, "Keystone auth url is \"%s\"\n", url)
```
oomichi: According to the reported issue, the above message in the console seems a little odd, so it is nice to remove the line in this PR,

Fedosin:
Thanks for the patch :) I wanted to do it myself could not find the time for that

PR #243, opened 2018-07-27
Allow to provide user, domain and project ids to the client authenticator

Fedosin:
In many cases it's necessary to use ids instead of names to provide information
for authentication.

This patch adds new options that allow to get the related id parameters from
OpenStack env variable, from command arguments and from the terminal.

Also this patch allows to leave project(id/name) empty, because these parameters
are not mandatory for authentication - in this case domain-scoped token will be
received.

### DIFF
--- a/cmd/client-keystone-auth/main.go
+++ b/cmd/client-keystone-auth/main.go
@@ -59,7 +59,7 @@ func promptForString(field string, r io.Reader, show bool) (result string, err e
 	fmt.Fprintf(os.Stderr, "Please enter %s: ", field)
 
 	if show {
-		_, err = fmt.Fscan(r, &result)
+		_, err = fmt.Fscanln(r, &result)
 	} else {
 		var data []byte
 		if terminal.IsTerminal(int(os.Stdin.Fd())) {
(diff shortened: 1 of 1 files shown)


cmd/client-keystone-auth/main.go
```diff
@@ -70,59 +70,92 @@ func promptForString(field string, r io.Reader, show bool) (result string, err e
...
+	}
+
 	return result, err
 }
 
-// prompt pulls keystone auth url, domain, username and password from stdin,
-// if they are not specified initially (i.e. equal "").
-func prompt(url string, domain string, user string, project string, password string) (gophercloud.AuthOptions, error) {
+// prompt pulls missing auth parameters from stdin, if they were not specified initially (i.e. equal "")
+func prompt(options *gophercloud.AuthOptions) error {
```
Fedosin: Instead of using multiple string parameters I decided to provide here a previously created AuthOption structure and use this function just to fill missing auth options in the structure

cmd/client-keystone-auth/main.go
```diff
@@ -70,59 +70,92 @@ func promptForString(field string, r io.Reader, show bool) (result string, err e
 			return "", fmt.Errorf("error reading input for %s", field)
 		}
 	}
+
+	// intentionally ignore "unexpected newline" errors to allow read empty lines from the console
+	if err != nil && err.Error() == "unexpected newline" {
```
Fedosin: This is sort of hack, but I didn't find a better solution to allow reading empty lines from the console

cmd/client-keystone-auth/main.go
```diff
@@ -70,59 +70,92 @@ func promptForString(field string, r io.Reader, show bool) (result string, err e
...
+
+		if options.TenantName == "" {
+			options.UserID, err = promptForString("project ID", os.Stdin, true)
+			if err != nil {
+				return err
+			}
+		}
+
+		if options.TenantID == "" && options.TenantName == "" {
+			fmt.Fprintf(os.Stderr, "Neither project ID nor project name have been specified. Proceed with domain-scoped token.")
```
hogepodge: Is there any reason we might want to allow domain name do be set to 'default' if there is no input?

cmd/client-keystone-auth/main.go
```diff
@@ -70,59 +70,92 @@ func promptForString(field string, r io.Reader, show bool) (result string, err e
...
 	}
 
-	options = gophercloud.AuthOptions{
-		IdentityEndpoint: url,
-		Username:         user,
-		TenantName:       project,
-		Password:         password,
-		DomainName:       domain,
+	if options.Password == "" {
+		options.Password, err = promptForString("password", nil, false)
```
hogepodge: Is there a secure option for specifying the password?

dims:
@Fedosin can you please rebase?

dims:
@Fedosin can you please rebase this?

dims:
/close

PR #253, opened 2018-08-03
Add readable and configurable loadbalancer naming

imdigitaljim:
Currently LB's are named in Neutron based on what Google and AWS needs by using an in-tree cloudprovider function. Since this driver is now out of tree, we can make this a feature improve the experience for users on Horizon and API's and correlate the loadbalancer name and description in a more meaningful way.


imdigitaljim:
will try this again soon

PR #259, opened 2018-08-10
Improve docs about auth data syncronization

zioproto:

Improves documentation about Auth data synchronization between Keystone and Kubernetes

### DIFF
--- a/docs/using-auth-data-synchronization.md
+++ b/docs/using-auth-data-synchronization.md
@@ -69,3 +69,57 @@ projects_black_list: ["id1", "id2"]
 # List of data types to synchronize
 "data_types_to_sync": ["projects", "role_assignments"]
 ```
+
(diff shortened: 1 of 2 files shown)


docs/using-auth-data-synchronization.md
```diff
@@ -69,3 +69,57 @@ projects_black_list: ["id1", "id2"]
...
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: _member_
+rules:
+- apiGroups: [""] # "" indicates the core API group
+  resources: ["pods"]
+  verbs: ["get", "watch", "list"]
+```
+
+Where the important bit of information is `_member_`: the string that represents the role of a user in keystone. If your role is called differently in your keystone setup, like `Member` or `member` you have to change this.
```
Fedosin: I believe I'll add a mapper in this case keystone_role->k8s_role, but for now it's true.

PR #262, opened 2018-08-13
Skip auth data synchronization in the case of unscoped tokens

Fedosin:
When a user sends an unscoped token to the server, Keystone still can authenticate
him, but doesn't include information about user's project and roles in the response.

In this case we have to skip the synchronization part, to prevent possible errors.


pkg/identity/keystone/keystone.go
```diff
@@ -246,7 +246,8 @@ func (k *KeystoneAuth) Handler(w http.ResponseWriter, r *http.Request) {
 		userInfo := k.authenticateToken(w, r, token, data)
 
 		// Do synchronization
-		if k.syncer.syncConfig != nil && userInfo != nil {
+		// In the case of unscoped tokens, when project id is not defined, we have to skip this part
+		if k.syncer.syncConfig != nil && userInfo != nil && userInfo.Extra["alpha.kubernetes.io/identity/project/id"][0] != "" {
```
FengyunPan2: How about the following codes?
if k.syncer.syncConfig != nil && userInfo != nil && len(userInfo.Extra["alpha.kubernetes.io/identity/project/id"]) != 0 {

Fedosin: Done. Thanks!

PR #263, opened 2018-08-13
Start using client.authentication.k8s.io/v1beta1 api for the client auth plugin

Fedosin:
Recently client authentication api was updated from v1alpha1 to v1beta1.
https://github.com/kubernetes/client-go/commit/f58a8f4fd079a6e2d027afeba4eb8c2ef3a233f5

This commit switches keystone-client-auth on the new api and removes the outdated
one.

The difference between two versions is that in alpha interactive mode was checked
in client-go and provided to the plugin with the environment variable
KUBERNETES_EXEC_INFO. In beta the plugin uses a TTY check to determine if it's
appropriate to prompt a user interactively. That's why KUBERNETES_EXEC_INFO is
not used anymore and can be removed.

Another change is that ExecCredentialSpec struct doesn't contain any fields now,
which means the plugin shouldn't write anything in this section during printing
response output.

https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins

### DIFF
--- a/cmd/client-keystone-auth/main.go
+++ b/cmd/client-keystone-auth/main.go
@@ -17,7 +17,6 @@ limitations under the License.
 package main
 
 import (
-	"encoding/json"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


dixudx:
defer to others for lgtm.

This PR has some changes on using TTY check to determine if it's appropriate to prompt a user interactively instead of reading from env vars.

PR #272, opened 2018-08-22
Pulling docker for the build

hyperbolic2346:
Fixing the snap build process. The launchpad builders didn't like using docker, so I switched it out to just build directly.


dims:
merging by hand since the test suite does not really use the snap

PR #286, opened 2018-09-14
[WIP] Barbican as Key Manager Service for Kubernetes

### LINKED ISSUE #44: Barbican as a Kubernetes KMS provider
Similar to https://github.com/Azure/kubernetes-kms

background research:
- https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/
- https://github.com/kubernetes/kubernetes/pulls?utf8=%E2%9C%93&q=is%3Apr+++EncryptionConfig+
- https://github.com/kubernetes/kubernetes/pulls?utf8=%E2%9C%93&q=is%3Apr+++experimental-encryption-provider-config

adisky:
This PR adds Feature to use Barbican as Key manager service for kubernetes

fixes #44 

This Plugin cannot be used in production until[1] in barbican gets implemented.
As per Kubernetes documents[2], we need to return encrypted dek to kubernetes, It stores encrypted dek in etcd, only kek is stored in Key Manager Service. OpenStack Barbican currently does not support to return encrypted secret, It only returns secret id. This Plugin returns the secret id to Kubernetes rather than encrypted secret. It results in problem of storing encrypted data at two different locations and since there is no notification from kuberenetes about secret deletion, user has to manually delete secrets stored in barbican. Once feature[1] gets implemented barbican needs to store only single kek and encrypted secret will be stored in kubernetes etcd only.
[1] https://review.openstack.org/#/c/598389/
[2] https://kubernetes.io/docs/tasks/administer-cluster/kms-provider/

**Release note**:

`NONE`.

### DIFF
--- a/cmd/barbican-kms/main.go
+++ b/cmd/barbican-kms/main.go
@@ -1,0 +1,66 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+	"k8s.io/cloud-provider-openstack/pkg/kms/barbican/server"
+)
+
+var (
+	socketpath  string
+	cloudconfig string
+)
+
+func init() {
+	flag.Set("logtostderr", "true")
+}
+
+func main() {
+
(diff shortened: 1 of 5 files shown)


dims:
discussion from slack

dims   [9:11 AM] @adisky around? wanted to chat about the barbican kms (edited)
adisky [9:28 AM] @dims yes, please tell
dims   [9:36 AM] may be just try here… so if you take the aescbc provider, the key needs to be specified in the config right? both encrypt and decrypt are done in the go code itself without talking to a service. So what if we follow the same pattern, except we specify a uuid of a key in the config. Then get the key from barbican and do the rest in the go code. is that possible?
adisky [9:37 AM] @dims yes, it is possible if we do encryption decryption in our code
dims   [9:38 AM]So let’s start there, then we can make things more complex slowly as needed. sounds good? (sorry took longer than expected to get back to you)
adisky [9:40 AM] @dims sure, thanks for your inputs :) , in this way it is non blocking and does not depend on openstack specs
adisky [9:44 AM] @dims But there is one concern here, the key should never leave the KMS and in this case for doing enc dec we need to get the key from barbican
dims   [9:45 AM] @adisky we do this in cinder now. get the key from KMS and use it to encrypt volumes

dims:
/close

PR #289, opened 2018-09-19
Add debug output for cfg options

jichenjc:
spend a couple of time in finding why the error:
Cloud provider could not be initialized: could not init
cloud provider "openstack": Authentication failed

turn out if we can log those with -v flag it will be helpful

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack.go
+++ b/pkg/cloudprovider/providers/openstack/openstack.go
@@ -171,11 +171,25 @@ type Config struct {
 	Networking   NetworkingOpts
 }
 
+func logcfg(cfg Config) {
+	klog.V(5).Infof("AuthURL: %s", cfg.Global.AuthURL)
+	klog.V(5).Infof("Username: %s", cfg.Global.Username)
+	klog.V(5).Infof("UserID: %s", cfg.Global.UserID)
+	klog.V(5).Infof("TenantID: %s", cfg.Global.TenantID)
+	klog.V(5).Infof("TenantName: %s", cfg.Global.TenantName)
+	klog.V(5).Infof("DomainName: %s", cfg.Global.DomainName)
+	klog.V(5).Infof("DomainID: %s", cfg.Global.DomainID)
+	klog.V(5).Infof("TrustID: %s", cfg.Global.TrustID)
+	klog.V(5).Infof("Region: %s", cfg.Global.Region)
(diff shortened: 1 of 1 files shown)


jichenjc:
I signed it

jichenjc:
"I signed it"

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -172,11 +172,26 @@ type Config struct {
 	Networking   NetworkingOpts
 }
 
+func logcfg(cfg Config) {
+	glog.V(3).Infof("AuthURL: %s", cfg.Global.AuthURL)
+	glog.V(3).Infof("Username: %s", cfg.Global.Username)
+	glog.V(3).Infof("UserID: %s", cfg.Global.UserID)
+	glog.V(3).Infof("Password: %s", cfg.Global.Password)
```
zetaab: printing password to log is not good thing

zetaab:
with this code in this PR, this will be added to be executed always. As I see it this is not needed always. And should we really print passwords to log even in debug cases? (its not that difficult to check cloud-config file)

jichenjc:
you are right, I removed the password stuffs

but the log output is helpful to debug as I actually spent some time in sync open rc and cloud conf and figure out the issue, if I can log it then it should be much easier

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -172,11 +172,25 @@ type Config struct {
 	Networking   NetworkingOpts
 }
 
+func logcfg(cfg Config) {
+	glog.V(3).Infof("AuthURL: %s", cfg.Global.AuthURL)
```
dims: @jichenjc can we move it to `V(5)`? since `V(3)` will log always?

dims:
@jichenjc please update with suggestion inline.

pkg/cloudprovider/providers/openstack/openstack.go
jichenjc: ok, will do ,thanks for the reminder

hogepodge:
@adisky the code changed upstream and created a merge conflict. I manually fixed the conflict with the inline github editor, which explains the additional commit (and why I'm the author of it).

jichenjc:
should this ready to merge? seems not find something block this PR?

PR #298, opened 2018-09-27
Fix csi test

### LINKED ISSUE #297: Cinder CSI CI Tests are failing
**What happened**:
Cinder CSI CI Tests are failing with the error "attacher.MountDevice failed: driver name csi-cinderplugin not found in the list of registered CSI drivers".
It is happening after PR https://github.com/kubernetes/kubernetes/pull/68200/, which defualts enable
KubeletPluginsWatcher, we need to update yaml files for driver-registrar sidecar container as described here https://kubernetes-csi.github.io/docs/print.html#kubelet-plugin-watcher

Also After the PR https://github.com/kubernetes/kubernetes/pull/67110, Adding NodeGetInfo call in node server is required

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

adisky:
This PR Fixes CSI CI Tests, It updates driver registrar configurations and add NodeGetInfo, NodeGetCapabilties rpc calls

**Which issue this PR fixes** : fixes #297 

**Release note**:
 `NONE`.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1386,7 +1386,6 @@
     "k8s.io/client-go/kubernetes/fake",
     "k8s.io/client-go/listers/core/v1",
     "k8s.io/client-go/listers/extensions/v1beta1",
-    "k8s.io/client-go/pkg/apis/clientauthentication/v1alpha1",
     "k8s.io/client-go/rest",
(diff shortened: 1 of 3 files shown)


dims:
This fixes the long standing issue with the CSI CI tests. So let's get this in.

PR #303, opened 2018-09-29
Add detail info about controller public/unpublish

jichenjc:
Add detail info about controller publish/unpublish

This doc does not have enough info.
Adding more description for someone who is familiar with
openstack but not with cloud provider.


ranasheel2000:
NIT:
If you find it ok, please use following as commit message:

--------------------------------------------------------

Add detail info about controller publish/unpublish

This doc does not have enough info.
Adding more description for someone who is familiar with 
openstack but not with cloud provider.

--------------------------------------------------------

PR #308, opened 2018-10-02
Batch update members in octavia-ingress-controller

lingxiankong:
Use batch operation API to reduce API calls to Octavia

<https://developer.openstack.org/api-ref/load-balancer/v2/index.html#batch-update-members>

**Release note**:

```release-note
```


lingxiankong:
@dims is there any known issue happened to CI?

PR #336, opened 2018-10-24
Add capabilities

jichenjc:
This PR will add more doc about the controller capabilities ,user should know the 
capability of the csi controller ,e.g what's it can do 

N/A
N/A
**Release note**:
no release notes needed because it's doc change

### DIFF
--- a/docs/using-cinder-csi-plugin.md
+++ b/docs/using-cinder-csi-plugin.md
@@ -59,6 +59,15 @@ $ csc identity plugin-capabilities --endpoint tcp://127.0.0.1:10000
 CONTROLLER_SERVICE
 ```
 
+#### Get controller implemented capabilities
+```
+$ csc controller get-capabilities  --endpoint tcp://127.0.0.1:10000
+&{type:CREATE_DELETE_VOLUME }
+&{type:PUBLISH_UNPUBLISH_VOLUME }
+&{type:LIST_VOLUMES }
+
+```
+
 #### Create a volume
 Following sample creates a volume named ``CSIVolumeName`` and the
 volume id returned is ``8a55f98f-e987-43ab-a9f5-973352bee19c`` with size ``1073741824`` bytes (1Gb)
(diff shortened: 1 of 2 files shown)


dims:
/ok-to-test

cc @adisky

docs/using-cinder-csi-plugin.md
```diff
@@ -59,6 +59,14 @@ $ csc identity plugin-capabilities --endpoint tcp://127.0.0.1:10000
 CONTROLLER_SERVICE
 ```
 
+#### Get controller implemented capabilities
+```
+$ csc controller get-capabilities  --endpoint tcp://127.0.0.1:10000
```
adisky: Thanks for improving CSI docs, Are you able to get the output?? as i cannot see ControllerGetCapabilities implemented here 
https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/controllerserver.go

jichenjc: actually, the output is listed here: 

&{type:CREATE_DELETE_VOLUME }
&{type:PUBLISH_UNPUBLISH_VOLUME }

so I believe it's the parent class handled this request and seems no need for us to override this method?

jichenjc: I mean , the command does give the output as listed above....

FengyunPan2: IMO, 'output a readable example' should be more helpful, like Line 75.

adisky: @jichenjc ok, so it think we are getting the default output
@FengyunPan2 IMHO, we should show the actual ouput in doc also, and the actual output has this weird format, so i think it is fine as of now.

FengyunPan2: +1 : show the actual output

dims:
/override openlab/cloud-provider-openstack-acceptance-test-flexvolume-cinder

PR #344, opened 2018-10-31
[csi] add snapshot support

jichenjc:
Add snapshot function for csi including:
add snapshot
delete snapshot
list snapshot

add snapshot functions for csi cinder volume
not a bug fix

**Release note**:

```
add snapshot csi functions, create-snapshot, delete-snapshot and list-snapshot are added 
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -18,8 +18,10 @@ package cinder
 
 import (
 	"errors"
+	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi/v0"
+	ossnapshots "github.com/gophercloud/gophercloud/openstack/blockstorage/v3/snapshots"
 	csicommon "github.com/kubernetes-csi/drivers/pkg/csi-common"
 	"github.com/pborman/uuid"
 	"golang.org/x/net/context"
@@ -197,7 +199,6 @@ func (cs *controllerServer) ControllerUnpublishVolume(ctx context.Context, req *
 }
 
 func (cs *controllerServer) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (*csi.ListVolumesResponse, error) {
-
 	// Get OpenStack Provider
 	cloud, err := openstack.GetOpenStackProvider()
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/csi/cinder/controllerserver.go
```diff
@@ -221,5 +221,88 @@ func (cs *controllerServer) ListVolumes(ctx context.Context, req *csi.ListVolume
...
+
+	name := req.Name
+	volumeId := req.SourceVolumeId
+	// No description from csi.CreateSnapshotRequest now
+	description := ""
+
+	// Delegate the check to openstack itself
+	snap, err := cloud.CreateSnapshot(name, volumeId, description, &req.Parameters)
+	if err != nil {
+		glog.V(3).Infof("Faled to Create snapshot: %v", err)
```
adisky: s/faled/failed

pkg/csi/cinder/controllerserver.go
```diff
@@ -221,5 +221,88 @@ func (cs *controllerServer) ListVolumes(ctx context.Context, req *csi.ListVolume
...
+		return nil, err
+	}
+
+	name := req.Name
+	volumeId := req.SourceVolumeId
+	// No description from csi.CreateSnapshotRequest now
+	description := ""
+
+	// Delegate the check to openstack itself
+	snap, err := cloud.CreateSnapshot(name, volumeId, description, &req.Parameters)
```
adisky: 1. various checks missing here, like if CO request to create a snapshot that already exists, It should not respond any error and return the details of existing snapshot.
2. Snapshot with same name on two different volume-id not allowed.

If you prefer to implement this checks later, add in todo as
TODO: Idempotency Checks

adisky: @jichenjc for q2 i know it is allowed on openstack side, but after reading csi specs, i think it is not allowed, for q1 you should not respond with any error [It is clearly mentioned in csi specs]

jichenjc: ok, I will add more to do and some checks here..

adisky:
@jichenjc can we add snapshot tests in CI also??

dims:
@adisky @flaper87 this looks ready?

PR #350, opened 2018-11-07
Add support for reading clouds.yaml

### LINKED ISSUE #323: Use standard clouds.yaml file as config file
/kind feature

We're currently migrating the `cluster-api-provider-openstack` config format from a custom implementation to using `clouds.yaml`, which is a known format in the OpenStack community.

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/16

Just like in the `cluster-api-provider-openstack` case, it'd be great if we could port the `cloud-provider-openstack` config format to use `clouds.yaml`. This would not only make both projects consistent but it'd also bring in a known format in the openstack community making the deployment of this provider easier to operators.

We could port the implementation that we're working on into `cloud-provider-openstack` and then ventor it in `cluster-api-provider-openstack` making the former a dependency for the latter.

iamemilio:
**What this PR does / why we need it**:  enables use of clouds.yaml in place of environment variables

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: Fixes: https://github.com/kubernetes/cloud-provider-openstack/issues/323

**Special notes for your reviewer**: First pr in this repo :1st_place_medal: 

@flaper87

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack.go
+++ b/pkg/cloudprovider/providers/openstack/openstack.go
@@ -39,8 +39,9 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/identity/v3/extensions/trusts"
 	tokens3 "github.com/gophercloud/gophercloud/openstack/identity/v3/tokens"
 	"github.com/gophercloud/gophercloud/pagination"
+	"github.com/gophercloud/utils/openstack/clientconfig"
 	"github.com/mitchellh/mapstructure"
-	"gopkg.in/gcfg.v1"
+	gcfg "gopkg.in/gcfg.v1"
 
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -169,6 +170,8 @@ type Config struct {
 		DomainName string `gcfg:"domain-name"`
 		Region     string
 		CAFile     string `gcfg:"ca-file"`
+		UseClouds  bool   `gcfg:"use-clouds"`
+		CloudsFile string `gcfg:"clouds-file,omitempty"`
 	}
 	LoadBalancer LoadBalancerOpts
 	BlockStorage BlockStorageOpts
@@ -219,6 +222,7 @@ func (cfg Config) toAuth3Options() tokens3.AuthOptions {
 
 // configFromEnv allows setting up credentials etc using the
 // standard OS_* OpenStack client environment variables.
+// TODO: Replace this with gophercloud upstream once community moves away from cloud.conf
 func configFromEnv() (cfg Config, ok bool) {
 	cfg.Global.AuthURL = os.Getenv("OS_AUTH_URL")
 	cfg.Global.Username = os.Getenv("OS_USERNAME")
@@ -261,6 +265,7 @@ func configFromEnv() (cfg Config, ok bool) {
 	return
 }
 
+// ReadConfig reads values from environment variables and the cloud.conf, prioritizing cloud-config
 func ReadConfig(config io.Reader) (Config, error) {
 	if config == nil {
 		return Config{}, fmt.Errorf("no OpenStack cloud provider config file given")
@@ -277,7 +282,47 @@ func ReadConfig(config io.Reader) (Config, error) {
 	cfg.Networking.PublicNetworkName = "public"
 
 	err := gcfg.FatalOnly(gcfg.ReadInto(&cfg, config))
+	if cfg.Global.UseClouds {
+		if cfg.Global.CloudsFile != "" {
+			os.Setenv("OS_CLIENT_CONFIG_FILE", cfg.Global.CloudsFile)
+		}
+		err = ReadClouds(&cfg)
+		if err != nil {
+			return Config{}, err
+		}
+	}
 	return cfg, err
+}
+
+// replaceEmpty is a helper function to replace empty fields with another field
+func replaceEmpty(a string, b string) string {
+	if a == "" {
+		return b
+	}
+	return a
+}
+
+// ReadClouds reads Reads clouds.yaml to generate a Config
+// Allows the cloud-config to have priority
+func ReadClouds(cfg *Config) error {
+	co := new(clientconfig.ClientOpts)
+	cloud, err := clientconfig.GetCloudFromYAML(co)
+	if err != nil && err.Error() != "unable to load clouds.yaml: no clouds.yaml file found" {
+		return err
+	}
+
+	cfg.Global.AuthURL = replaceEmpty(cfg.Global.AuthURL, cloud.AuthInfo.AuthURL)
+	cfg.Global.Username = replaceEmpty(cfg.Global.Username, cloud.AuthInfo.Username)
+	cfg.Global.UserID = replaceEmpty(cfg.Global.UserID, cloud.AuthInfo.UserID)
+	cfg.Global.Password = replaceEmpty(cfg.Global.Password, cloud.AuthInfo.Password)
+	cfg.Global.TenantID = replaceEmpty(cfg.Global.TenantID, cloud.AuthInfo.ProjectID)
+	cfg.Global.TenantName = replaceEmpty(cfg.Global.TenantName, cloud.AuthInfo.ProjectName)
+	cfg.Global.DomainID = replaceEmpty(cfg.Global.DomainID, cloud.AuthInfo.UserDomainID)
+	cfg.Global.DomainName = replaceEmpty(cfg.Global.DomainName, cloud.AuthInfo.UserDomainName)
+	cfg.Global.Region = replaceEmpty(cfg.Global.Region, cloud.RegionName)
+	cfg.Global.CAFile = replaceEmpty(cfg.Global.CAFile, cloud.CACertFile)
+
+	return nil
 }
 
 // caller is a tiny helper for conditional unwind logic
@@ -336,6 +381,7 @@ func checkOpenStackOpts(openstackOpts *OpenStack) error {
 	return checkMetadataSearchOrder(openstackOpts.metadataOpts.SearchOrder)
 }
 
+// NewOpenStack creates a new new instance of the openstack struct from a config struct
 func NewOpenStack(cfg Config) (*OpenStack, error) {
 	provider, err := openstack.NewClient(cfg.Global.AuthURL)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -255,12 +256,65 @@ func configFromEnv() (cfg Config, ok bool) {
...
+	clientOpts := new(clientconfig.ClientOpts)
+	cloud, err := clientconfig.GetCloudFromYAML(clientOpts)
+
+	// Exit as a failure if unable to read clouds.yaml
+	if err != nil {
+		fmt.Println(err)
+		return cfg, err
+	}
+
+	cfg.Global.AuthURL = cloud.AuthInfo.AuthURL
```
flaper87: I think this is a good time to refactor the [`Config`](https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/cloudprovider/providers/openstack/openstack.go#L154)  struct. We could replace the current Global definition with a field that holds a `clientconfig.Cloud` object.

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -255,12 +256,65 @@ func configFromEnv() (cfg Config, ok bool) {
...
+	cfg.Global.DomainID = cloud.AuthInfo.UserDomainID
+	cfg.Global.DomainName = cloud.AuthInfo.UserDomainName
+	cfg.Global.CAFile = cloud.CACertFile
+
+	// Retaining support for trusts via environment variables
+	// due to lack of support by GetCloudFromYAML(gohpercloud)
+	cfg.Global.TrustID = os.Getenv("OS_TRUST_ID")
+
+	// Supporting injecting passwords via env variables
+	if cfg.Global.Password == "" {
```
flaper87: We shouldn't need this now that we're using `clientconfig`.

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -255,12 +256,65 @@ func configFromEnv() (cfg Config, ok bool) {
...
 	if config == nil {
 		return Config{}, fmt.Errorf("no OpenStack cloud provider config file given")
 	}
+	var cfg Config
 
-	cfg, _ := configFromEnv()
+	// Prioritizes reading clouds.yaml over envorinment variables
+	cfg, err := configFromYAML()
+	if err != nil && err.Error() == "unable to load clouds.yaml: no clouds.yaml file found" {
+		cfg, _ = configFromEnv()
```
flaper87: To properly keep backwards compatibility, I think we should first attempt to load configs using the existing method, warn the user that this behavior will change (only if the configs were set using the old method), then attempt to load them using `clouds.yaml`.

iamemilio: I'm not sure this is an issue. According to the unit tests, the app has full backwards compatibility. I personally think that clouds.yaml should take priority, since it is generated directly by openstack, is a more reliable source of the data we need, and is ultimately the configuration method we want to support moving forward.

pkg/cloudprovider/providers/openstack/openstack.go
flaper87: A new commit sounds good. I'd have it in this PR, tho.

pkg/cloudprovider/providers/openstack/openstack.go
flaper87: Agreed that `clouds.yaml` is a better source for this config. However, there may be deployments currently using the "old method". If one of these deployments updates the provider and has a valid `clouds.yaml` in a valid path, the deployment may end up using that instead of the currently known method. We don't want to change the current order without first warning OPs that this change is coming.

pkg/cloudprovider/providers/openstack/openstack.go
flaper87: mmh, if that's the case, I'd move the Trusts field to the main Config struct. We can still use it from there.

dims:
/override openlab/cloud-provider-openstack-acceptance-test-flexvolume-cinder

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -201,66 +251,119 @@ func (cfg Config) toAuthOptions() gophercloud.AuthOptions {
...
+		Password:         cfg.Global.AuthInfo.Password,
+		DomainID:         cfg.Global.AuthInfo.UserDomainID,
+		DomainName:       cfg.Global.AuthInfo.UserDomainName,
 		AllowReauth:      true,
 	}
 }
 
 // configFromEnv allows setting up credentials etc using the
 // standard OS_* OpenStack client environment variables.
 func configFromEnv() (cfg Config, ok bool) {
```
flaper87: We may be able to simplify this function further by relying completely on clientconfig to read most of the env variables.

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -201,66 +251,119 @@ func (cfg Config) toAuthOptions() gophercloud.AuthOptions {
 
 func (cfg Config) toAuth3Options() tokens3.AuthOptions {
```
flaper87: We might not need this function now that we're loading configs using `clientconfig`

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -149,8 +150,8 @@ type OpenStack struct {
 	localInstanceID string
 }
 
-// Config is used to read and store information from the cloud configuration file
-type Config struct {
+// Workaround is used to read and store information from the cloud configuration file
+type Workaround struct {
```
flaper87: I'd rather try to get rid of this struct entirely. Think of this workflow:

1) Attempt to load this configuration (warn the user that it'll be deprecated if a config like this is found)
2) If found, load the data directly into an `AuthOpts` object
3) If not found, attempt to load the configs using `clientconfig`

this way we preserve backwards compatibility, we reduce the code and we move the internal structures to using `clientconfig` and the upstream standard

pkg/volume/cinder/volumeservice/connection.go
```diff
@@ -39,7 +39,24 @@ import (
 )
 
 type cinderConfig struct {
-	openstack_provider.Config
```
flaper87: I had completely forgotten that part of this code was duplicated in the drivers. :(

pkg/cloudprovider/providers/openstack/openstack.go
iamemilio: Yeah, the issue is that in order to read from the custom config file, we need a struct with the gcgf annotations, so we will have to have something similar to the workaround struct until we fully phase out config files. Other than that though, this makes sense.

pkg/volume/cinder/volumeservice/connection.go
iamemilio: should this also be clientconfigified?

chaosaffe:
xref: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/31

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -274,7 +259,45 @@ func configFromLegacy(cfg *Config, legacyconfig io.Reader) error {
...
+
+	if v := os.Getenv(envPrefix + "CLOUD"); v != "" {
+		cloudName = v
+	}
+	return cloudName
+}
+
+
+func getUnsupportedFieldsFromYAML(cfg *Config, opts *clientconfig.ClientOpts) error{
+	if oscloud := osCloud(opts); oscloud != "" {
```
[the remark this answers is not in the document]
iamemilio: I only added it because it was in the AuthOptions function that way. No other reason really. Also, what is a configuration lead flow?

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -255,7 +258,10 @@ func configFromEnv() (cfg Config, ok bool) {
 	return
 }
 
+// ReadConfig reads values from environment variables and the cloud.conf, prioritizing cloud-config
+// WARNING: depricated
 func ReadConfig(config io.Reader) (Config, error) {
```
iamemilio: I think i am going to move this to an else statement at line 286 instead. The current implementation makes no sense

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -255,7 +258,10 @@ func configFromEnv() (cfg Config, ok bool) {
...
+	klog.Warning("Use of cloud-configs are depricated, and may no longer be supported in the future")
```
iamemilio: I think i am going to move this to an else statement at line 286 instead. The current implementation makes no sense

dims:
@iamemilio @flaper87 is this ready to merge?

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -255,6 +259,8 @@ func configFromEnv() (cfg Config, ok bool) {
 	return
 }
 
+// ReadConfig reads values from environment variables and the cloud.conf, prioritizing cloud-config
+// WARNING: depricated
```
chaosaffe:
```suggestion
// WARNING: deprecated
```

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -271,9 +277,51 @@ func ReadConfig(config io.Reader) (Config, error) {
...
+	if cfg.Global.UseClouds {
+		if cfg.Global.CloudsFile != "" {
+			os.Setenv("OS_CLIENT_CONFIG_FILE", cfg.Global.CloudsFile)
+		}
+		err = ReadClouds(&cfg)
+		if err != nil {
+			return Config{}, err
+		}
+	} else {
+		klog.Warning("Use of cloud-configs are depricated, and we are planning to move away from them in the future")
```
chaosaffe:
```suggestion
		klog.Warning("Use of cloud-configs are deprecated, and we are planning to move away from them in the future")
```

pkg/cloudprovider/providers/openstack/openstack.go
iamemilio: i think we should consider excluding or changing this message btw

iamemilio: ok, i just took both of these out entirely

Issue #361, opened 2018-11-15
report size when create volume again with same name


jichenjc:
/kind bug

**What happened**:
root@kvm-017407:~# ~/go/src/github.com/rexray/gocsi/csc/csc controller create-volume --endpoint tcp://127.0.0.1:10000 CSIVolumeName1
"27aea4f0-9a93-4b99-9396-6a26a5bbf6e7"  1073741824      "availability"="nova"

root@kvm-017407:~# ~/go/src/github.com/rexray/gocsi/csc/csc controller create-volume --endpoint tcp://127.0.0.1:10000 CSIVolumeName1
"27aea4f0-9a93-4b99-9396-6a26a5bbf6e7"  0       "availability"="nova"

**What you expected to happen**:
0 ==>　1073741824      
**How to reproduce it (as minimally and precisely as possible)**
every time

**Anything else we need to know?**:

**Environment**:
master branch

jichenjc:
PR already merged

PR #362, opened 2018-11-15
Add size into output

jichenjc:
when create a volume with name already there,
we don't give the size of the volume in output and lack of log.

Fix bug #361 

**Release note**:

```release-note
create volume with same name will get the actual size other than 0 now
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -73,6 +73,9 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	if len(volumes) == 1 {
 		resID = volumes[0].ID
 		resAvailability = volumes[0].AZ
+		resSize = volumes[0].Size
+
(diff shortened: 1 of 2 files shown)


pkg/csi/cinder/controllerserver.go
```diff
@@ -73,6 +73,9 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	if len(volumes) == 1 {
 		resID = volumes[0].ID
 		resAvailability = volumes[0].AZ
+		resSize = volumes[0].Size
+
+		glog.V(4).Infof("Return created volume %s in Availability Zone: %s of size %d GiB", resID, resAvailability, resSize)
```
adisky: It Would be better if we also log, that volume already exist.
glog.V(4).Infof("Volume %s  exists with same name in Availability Zone: %s  of  size %d GiB, resID, resAvailability, resSize)

adisky:
@jichenjc Thanks for the PR, Please check one comment inline

jichenjc:
Updated the log per request..

dims:
/test  cloud-provider-openstack-format

PR #370, opened 2018-11-20
Manila: Add StorageClass parameter to specify NFS share client

tsmetana:
The NFS share created by Manila provisioner is by default read/write by the world. This is not always welcome behaviour. I have added a new StorageClass option that allows to specify the NFS clients that would have access to the share.

Moreover the default value ('0.0.0.0/0') doesn't work with Ganesha NFS server as it doesn't seem to like the '/0' part so I changed the default to '0.0.0.0'

```release-note
Manila provisioner recognizes a new StorageClass option to specify allowed NFS share client.
```

### DIFF
--- a/pkg/share/manila/sharebackends/nfs.go
+++ b/pkg/share/manila/sharebackends/nfs.go
@@ -45,11 +45,11 @@ func (NFS) BuildSource(args *BuildSourceArgs) (*v1.PersistentVolumeSource, error
 	}, nil
 }
 
-// GrantAccess to NFS share. Allows read-write access to everyone!
+// GrantAccess to NFS share. Allows read-write access to everyone by default!
 func (NFS) GrantAccess(args *GrantAccessArgs) (*shares.AccessRight, error) {
 	return shares.GrantAccess(args.Client, args.Share.ID, shares.GrantAccessOpts{
 		AccessType:  "ip",
-		AccessTo:    "0.0.0.0/0",
+		AccessTo:    args.Options.BackendOptions.NFSShareClient,
 		AccessLevel: "rw",
 	}).Extract()
 }
--- a/pkg/share/manila/shareoptions/backend.go
+++ b/pkg/share/manila/shareoptions/backend.go
@@ -19,6 +19,6 @@ package shareoptions
 // BackendOptions contains backend-specific options
 type BackendOptions struct {
 	CSICEPHFSdriver string `name:"csi-driver" backend:"csi-cephfs" protocol:"CEPHFS"`
-
+	NFSShareClient  string `name:"nfs-share-client" backend:"nfs" protocol:"NFS" value:"default=0.0.0.0"`
 	// Add more backend options here
 }
--- a/docs/using-manila-provisioner.md
+++ b/docs/using-manila-provisioner.md
@@ -39,6 +39,7 @@ None.
 Key | For backend | For protocol  | Required | Default Value | Description
 --- | ----------- | ------------- | ------------- | ----------- |---------
 `csi-driver` | `csi-cephfs` | `CEPHFS` | Yes | None | Name of the CSI driver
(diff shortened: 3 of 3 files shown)


pkg/share/manila/sharebackends/args.go
```diff
@@ -39,6 +39,7 @@ type GrantAccessArgs struct {
 	Options   *shareoptions.ShareOptions
 	Clientset clientset.Interface
 	Client    *gophercloud.ServiceClient
+	AccessTo  string
```
gman0: Is the `AccessTo` attribute needed here?

pkg/share/manila/shareoptions/backend.go
```diff
@@ -19,6 +19,6 @@ package shareoptions
 // BackendOptions contains backend-specific options
 type BackendOptions struct {
 	CSICEPHFSdriver string `name:"csi-driver" backend:"csi-cephfs" protocol:"CEPHFS"`
-
+	NFSShareClient  string `name:"nfs-share-client" backend:"nfs" protocol:"nfs"`
```
gman0: Two points to consider here:
* those attribute tags are case-sensitive and since the documentation says `NFS`, let's make the protocol tag for `NFSShareClient` upper-case as well. You make a good point here though, it should be case-insensitive (otherwise the parser will get confused), I'll post a PR.
* you could add an `optional` tag here ([see this for example](https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/share/manila/shareoptions/common.go#L21)), making the whole tag field look like this `name:"nfs-share-client" backend:"nfs" protocol:"NFS" value:"default=0.0.0.0"`. This will simplify the code in the `NFS.GrantAccess` method.

pkg/share/manila/sharebackends/nfs.go
```diff
@@ -45,11 +45,18 @@ func (NFS) BuildSource(args *BuildSourceArgs) (*v1.PersistentVolumeSource, error
...
+
+	if len(args.Options.BackendOptions.NFSShareClient) == 0 {
+		accessTo = "0.0.0.0"
+	} else {
+		accessTo = args.Options.BackendOptions.NFSShareClient
+	}
 	return shares.GrantAccess(args.Client, args.Share.ID, shares.GrantAccessOpts{
 		AccessType:  "ip",
-		AccessTo:    "0.0.0.0/0",
+		AccessTo:    accessTo,
```
gman0: See the comment [here](https://github.com/kubernetes/cloud-provider-openstack/pull/370/files#r236597414), with `value:"default=..."` this can be simplified into

```go
func (NFS) GrantAccess(args *GrantAccessArgs) (*shares.AccessRight, error) {
	return shares.GrantAccess(args.Client, args.Share.ID, shares.GrantAccessOpts{
		AccessType:  "ip",
		AccessTo:    args.Options.BackendOptions.NFSShareClient,
		AccessLevel: "rw",
	}).Extract()
}
```

pkg/share/manila/shareoptions/backend.go
tsmetana: Good idea with the default setting. I will have to re-test the case for the protocol -- I think I originally had it uppercase (basically I copied the CSICEPHFSdriver string) but it didn't work.

pkg/share/manila/sharebackends/nfs.go
tsmetana: Yup. Thanks.

pkg/share/manila/shareoptions/backend.go
tsmetana: Most likely. All the other attribute values are lower case so it's pretty possible I just made a mistake. I tested with the requested changes and everything seems to be working fine now.

gman0: Perfect! Yes the current behavior is really confusing and needs to be changed.

gman0:
cc @dims

PR #376, opened 2018-11-22
Move loadbalancer annotations to the service

### LINKED ISSUE #375: Loadbalancer examples have the annotations on the wrong resource
/kind bug

**What happened**:

The annotations in examples/loadbalancers are set on the `Deployment` not on the `Service`. This causes every loadbalancer to be an external one (at least in my setup). To have any effect, the annotations have to be made on the `Service`.

These annotations hint to the provider if the loadbalancer should grab a floating IP for the loadbalancer or if its meant to remain internal.

Just apply the `internal-http-nginx.yaml` and you'll find that it will make an external one.

ainmosni:
For the loadbalancer annotations to have any effect, they need to be in
the service definition not in the deployment.

It moves the loadbalancer annotations in the loadbalancers examples to the right resource. These annotations hint to the provider if the loadbalancer should grab a floating IP for the loadbalancer or if its meant to remain internal.

 fixes #375

### DIFF
--- a/examples/loadbalancers/external-http-nginx.yaml
+++ b/examples/loadbalancers/external-http-nginx.yaml
@@ -3,9 +3,6 @@ apiVersion: apps/v1beta1
 kind: Deployment
 metadata:
   name: external-http-nginx-deployment
-  annotations:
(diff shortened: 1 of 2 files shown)


dims:

/lgtm

PR #377, opened 2018-11-23
Implement Barbican as KMS provider for kubernetes

### LINKED ISSUE #44: Barbican as a Kubernetes KMS provider

adisky:
This PR adds Feature to use Barbican as Key manager service for kubernetes

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #44

**Release note**:

```release-note
```

### DIFF
--- a/manifests/barbican-kms/pod.yaml
+++ b/manifests/barbican-kms/pod.yaml
@@ -1,0 +1,24 @@
+apiVersion: v1 
+kind: Pod 
+metadata: 
+ name: barbican-kms
+spec: 
+  containers: 
+    - name: barbican-kms 
+      image: docker.io/k8scloudprovider/barbican-kms-plugin:latest
+      args:
+        - "--socketpath=/kms/kms.sock"
+        - "--cloud-config=/etc/kubernetes/cloud-config"
+      volumeMounts:
+        - name: cloud-config
+          mountPath: /etc/kubernetes/
+        - name: socket-dir
+          mountPath: /kms/
+  volumes:
+  - name: config
+    hostPath:
+      path: /etc/kubernetes    
+  - name: socket-dir
+    hostPath:
+      path: /var/lib/kms/
+      type: DirectoryOrCreate
--- a/pkg/kms/barbican/barbican.go
+++ b/pkg/kms/barbican/barbican.go
@@ -1,0 +1,95 @@
+package barbican
+
+import (
+	"github.com/golang/glog"
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack"
+	"github.com/gophercloud/gophercloud/openstack/keymanager/v1/secrets"
+)
+
+type BarbicanService interface {
+	GetSecret(cfg *Config) ([]byte, error)
+}
+
+type KMSOpts struct {
+	KeyID string `gcfg:"key-id"`
+}
+
+//Config to read config options
+type Config struct {
+	Global struct {
+		AuthURL    string `gcfg:"auth-url"`
+		Username   string
+		UserID     string `gcfg:"user-id"`
+		Password   string
+		TenantID   string `gcfg:"tenant-id"`
+		TenantName string `gcfg:"tenant-name"`
+		DomainID   string `gcfg:"domain-id"`
+		DomainName string `gcfg:"domain-name"`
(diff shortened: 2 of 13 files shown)


lingxiankong:
hi @adisky, as we talked last time, i'd appreciate if you could also provide some doc for this feature so that i can follow for testing? It's also very useful for other potential users for this addon, thanks!

adisky:
@lingxiankong I have created the doc PR 
https://github.com/kubernetes/cloud-provider-openstack/pull/393.
It should work with kubernetes v1.12 for 1.13 some flags are changed.

pkg/kms/barbican/barbican.go
```diff
@@ -0,0 +1,95 @@
...
+type Config struct {
+	Global struct {
+		AuthURL    string `gcfg:"auth-url"`
+		Username   string
+		UserID     string `gcfg:"user-id"`
+		Password   string
+		TenantID   string `gcfg:"tenant-id"`
+		TenantName string `gcfg:"tenant-name"`
+		DomainID   string `gcfg:"domain-id"`
+		DomainName string `gcfg:"domain-name"`
```
strigazi: Can you add support of trustee and application credentials?
Using a trust-id is supported by gophercloud.

strigazi: https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/share/manila/shareoptions/openstack.go#L27

adisky: @strigazi will add Trust support.

manifests/barbican-kms/pod.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v1 
+kind: Pod
```
strigazi: We could make this a DS to run on all master nodes. Similar to [0] and [1].
[0] https://github.com/kubernetes/cloud-provider-openstack/blob/master/manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml#L26
[1] https://github.com/kubernetes/website/blob/1d1ccd974d4a6910cd24877ea96b24391ce3a7d1/content/en/examples/admin/cloud/ccm-example.yaml#L57

adisky: @strigazi we cannot run this as kubernetes construct, It has to be run as docker conatiner or systemd services, I will explore to run it as static pod later

strigazi: When you add trust support I can start testing in our setup.

Static pods works perfectly in a single master cluster. We can see for multimaster.

dims:
/hold

@adisky is this ready?

adisky:
@dims yes it is ready, but i have not added trust support yet as pointed out by @strigazi

manifests/barbican-kms/pod.yaml
lingxiankong: > @strigazi we cannot run this as kubernetes construct, It has to be run as docker conatiner or systemd services, I will explore to run it as static pod later

I think you are talking about different deployment models. I guess @adisky 's original plan is to deploy the kms plugin together with k8s master before the master provision, k8s cannot function well without the kms plugin because it needs to store/retrieve things from etcd. But @strigazi 's meaning is we can deploy kms plugin after k8s installation is completed with no etc encryption and config the kms plugin afterwards.

However, i think static pod can do things pretty well, the only concern is that static pod may be deprecated in the future according to k8s official doc.

adisky:
@lingxiankong This is just the basic implementation for key rotation and production env, we need to keep improving this

pkg/kms/barbican/barbican.go
strigazi: Is support for trust-id added in another PR? Did you find any difficulties?

PR #379, opened 2018-11-26
Add a member number check

### LINKED ISSUE #378: create LB without node (master and taint only) leads to Internal Server Error from openstack
/kind bug

**What happened**:
Just a minor update, to provide more info when create LB member failed

**Environment**:
openstack devstack (created from commit e3e9ea299601665a295e31a98e90dd9587165850)
Nov 5

jichenjc:
No members to be updated and calling openstack API will leads to
update error, instead of provide internal error to end user
we can provide more helpful info here.

Closes: #378

enhancing log output when no node existing and creating a LB
fixes #378 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/ingress/controller/openstack/octavia.go
+++ b/pkg/ingress/controller/openstack/octavia.go
@@ -492,6 +492,11 @@ func (os *OpenStack) EnsurePoolMembers(deleted bool, poolName string, lbID strin
 		}
 		members = append(members, member)
 	}
+	// only allow >= 1 members or it will lead to openstack octavia issue
+	if len(members) == 0 {
+		return nil, fmt.Errorf("error because no members in pool: %s", pool.ID)
+	}
+
 	if err := pools.BatchUpdateMembers(os.octavia, pool.ID, members).ExtractErr(); err != nil {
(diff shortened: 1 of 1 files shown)


lingxiankong:
I noticed it's related to octavia-ingress-controller, thanks for submitting this PR, i'll take a look tomorrow.

pkg/ingress/controller/openstack/octavia.go
```diff
@@ -492,6 +492,11 @@ func (os *OpenStack) EnsurePoolMembers(deleted bool, poolName string, lbID strin
 		}
 		members = append(members, member)
 	}
+	// only allow >= 1 members or it will lead to openstack octavia issue
+	if len(members) == 0 {
```
lingxiankong: Why not fail immediately after querying all the nodes?

jichenjc: because there are logic check above to avoid node failure case, 
so this check is to avoid either no nodes or has nodes but all nodes are failed

to be honest it seems openstack side change might be better (to avoid 500 error, return 400 bad request), but seems add this kind of 'no members' is a plus to have in provider side

dims:
/assign @lingxiankong

PR #384, opened 2018-11-29
manila-provisioner: added mounter option for csi-cephfs backend

gman0:
Adds a new StorageClass param for the csi-cephfs backend, users can now choose whether to use the Ceph kernel client or Ceph FUSE driver for mounting.

```release-note
Manila provisioner recognizes a new StorageClass option to specify mount tool for csi-cephfs backend
```

### DIFF
--- a/pkg/share/manila/sharebackends/csicephfs.go
+++ b/pkg/share/manila/sharebackends/csicephfs.go
@@ -17,6 +17,8 @@ limitations under the License.
 package sharebackends
 
 import (
+	"fmt"
+
 	"github.com/gophercloud/gophercloud/openstack/sharedfilesystems/v2/shares"
 	"k8s.io/api/core/v1"
 )
@@ -36,6 +38,10 @@ func (CSICephFS) BuildSource(args *BuildSourceArgs) (*v1.PersistentVolumeSource,
 		return nil, err
 	}
 
+	if args.Options.CSICEPHFSmounter != "fuse" && args.Options.CSICEPHFSmounter != "kernel" {
+		return nil, fmt.Errorf("unrecognized mounter option '%s'", args.Options.CSICEPHFSmounter)
+	}
+
 	return &v1.PersistentVolumeSource{
 		CSI: &v1.CSIPersistentVolumeSource{
 			Driver:       args.Options.CSICEPHFSdriver,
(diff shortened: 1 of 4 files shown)


gman0:
cc @dims @adisky

pkg/share/manila/sharebackends/sharebackend_test.go
```diff
@@ -72,6 +72,7 @@ func TestCSICephFSBuildSource(t *testing.T) {
 	}
 
 	args.Options.CSICEPHFSdriver = "csi-cephfs"
+	args.Options.CSICEPHFSmounter = "fuse"
```
dixudx: This can also be `kernel`, right?
Please add a test to keep us from falling back.

pkg/share/manila/sharebackends/csicephfs.go
```diff
@@ -36,6 +38,10 @@ func (CSICephFS) BuildSource(args *BuildSourceArgs) (*v1.PersistentVolumeSource,
 		return nil, err
 	}
 
+	if args.Options.CSICEPHFSmounter != "fuse" && args.Options.CSICEPHFSmounter != "kernel" {
```
adisky: I think default is 'fuse' as you mentioned in document, so in case of empty it should not return error.

gman0: This is a no - `CSICEPHFSmounter` may never be empty. It must be either:
- populated by the user using the storage class parameters, or
- populated by the options parser with the default value, if the attribute has not been previously populated by the user

It's an error to have this value empty. In this case it would mean there's a bug in the parser (because it allowed an empty value here, which it shouldn't), so at least it won't go unnoticed.

adisky: ok got that, so the options parser will always populate it with default value.

pkg/share/manila/sharebackends/sharebackend_test.go
gman0: @dixudx addressed your comment

gman0:
Could someone have a look at the [openlab/cloud-provider-openstack-acceptance-test-e2e-conformance](http://logs.openlabtesting.org/logs/84/384/c797141cca5bd71331e4c4698f30d7b6e102bdeb/cloud-provider-openstack-acceptance-test-e2e-conformance/cloud-provider-openstack-acceptance-test-e2e-conformance/f87996e/) test please? I don't think the failures reported in the logs are cause by the code from this PR.

gman0:
Can we get this merged? Or is there something I can do about that failed test?

gman0:
Thanks!

Issue #391, opened 2018-12-05
Delete a node then readd it in LB ingress has issue


jichenjc:
/kind bug

**What happened**:
root@kvm-017407:~/go/src/k8s.io/cloud-provider-openstack# kubectl get ing
NAME            HOSTS         ADDRESS       PORTS   AGE
test-ingress1   foo.bar.com   172.24.4.22   80      44h

root@kvm-017407:~/go/src/k8s.io/cloud-provider-openstack# kubectl get nodes
NAME         STATUS   ROLES    AGE   VERSION
kvm-017407   Ready    master   49d   v1.12.1
kvm-017734   Ready    <none>   21m   v1.12.2

kubectl delete node
delete node (17734) but LB doesn't have any update, I think we should consider to update openstack member 

Also
has following log
time="2018-12-05T02:53:36Z" level=info msg="Detected change in list of current cluster nodes. New node set: []"
time="2018-12-05T02:53:37Z" level=error msg="Failed to handle ingress" ingress=test-ingress1

means we should avoid doing up in this case (node = [])

jichenjc:
I will submit some patches..

PR #392, opened 2018-12-05
Fix part of bug 391

jichenjc:
avoid update loadbalancer member if new node is []

Fix part of bug 391

```release-note
NONE
```


adisky:
@jichenjc Please add some description in PR headline, and fixes #issueNo. in comment

pkg/ingress/controller/controller.go
```diff
@@ -350,6 +350,13 @@ func (c *Controller) nodeSyncLoop() {
 
 	log.Infof("Detected change in list of current cluster nodes. New node set: %v", nodeNames(newNodes))
 
+	// if no new nodes, then avoid update member
+	if len(newNodes) == 0 {
+		c.knownNodes = newNodes
```
lingxiankong: if no nodes now, the lb pool needs to be updated, no?

dims:
/hold

waiting for @jichenjc to respond to @lingxiankong comments

pkg/ingress/controller/controller.go
jichenjc: let me dig more here... originally I think nodes == [] we should avoid update lb
but another thought here might defer the check to later...

chaosaffe:
After taking a deeper look into this patch, if we have no active nodes we will never update the LB, meaning that the LB will have nodes in the pool that are not in a `Ready` state, and may have even been deleted.

Unfortunately, the existing variable name `newNodes` is misleading, because we retrieve all ready nodes that are not labeled as masters, not only 'new' nodes that we haven't seen before. A more appropriate name for `newNodes` would be `readyWorkerNodes`  to clarify what is being returned from `ListWithPredicate()`

It looks like https://github.com/kubernetes/cloud-provider-openstack/issues/391 is caused by the empty node set hitting [this code block](https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/ingress/controller/openstack/octavia.go#L495-L498) you added in https://github.com/kubernetes/cloud-provider-openstack/pull/379, which triggers a return without updating or deleting pool members in Octavia when there are no nodes in the set. 

Instead of short-circuiting when there is an empty set of nodes, we should call `EnsurePoolMembers()` with `deleted=true` to remove the pool from Octavia as we cannot provide an empty pool, or even better IMO we can fix `EnsurePoolMembers()` to handle an empty node set as an implicit delete of the pool, but I would like @dims or @lingxiankong to chime in on that path first

@jichenjc thank you for reporting #391 and your work so far on patching it up!

lingxiankong:
@chaosaffe @jichenjc how about we do nothing if the readyWorkerNodes=[], because if there is no node available, there is actually some problem in the cluster that needs to fix, the ingress won't behave properly. After the cluster is healthy, the ingress-controller will reconcile to the correct state.

@jichenjc are you still working on this?

lingxiankong:
thanks!, This is what I actually suggested.

PR #396, opened 2018-12-11
Trivial Fix: update incorrect log

jichenjc:
the log indicate wrong info, fix them

fix log

**Release note**:

```release-note
None
```


chaosaffe:
@jichenjc Thank you for your suggestion. 
In this case the original error message is more correct than your proposed change. We waited for the load balancer to be created, but the creation timed out or encountered an error. We append the actual error message to the returned error, correctly advising the user that we could not create the loadbalancer.

PR #398, opened 2018-12-11
Fix part of bug 372, try to use metadata to retrieve data first

### LINKED ISSUE #372: Use OpenStack metadata service to acquire instance ID
/kind feature

**What happened**:
Currently, when setting up the provider on a machine with a hostname that differs from the name in OpenStack the provider can't find the instance ID. The provider should use the [metadata service](https://docs.openstack.org/nova/rocky/user/metadata-service.html)  to acquire this ID to make this process more reliable.

Note that `/var/lib/cloud/data` is less reliable, as operating systems like container Linux have moved away from cloud-init.

jichenjc:
first, let's try use metadata to retrieve data then use cloud-init data
after all, the data in metadata/config drive won't be changed but
it's likely the data in cloud-init generated might be changed.

use metadata/config drive first then cloud-init generated data

fixes #372 

**Release note**:
```release-note
None
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack.go
+++ b/pkg/cloudprovider/providers/openstack/openstack.go
@@ -287,6 +287,13 @@ func (c *caller) call(f func()) {
 }
 
 func readInstanceID(searchOrder string) (string, error) {
+	// First, try to get data from metadata service because local
+	// data might be changed by accident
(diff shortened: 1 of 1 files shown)


adisky:
@jichenjc Please add some description in PR headline, and fixes #issueNo. in comment

Issue #301, opened 2018-09-28
Add Features in Cinder CSI Driver 


adisky:
/kind feature

**What happened**:
With CSI v0.3.0 New Plugin Capabilties are added including Snapshot and Topology API, This features needs to be included in Cinder CSI driver.

Following CSI Interface are not implemented in cinder csi driver
 **Identity Service** 
 **Controller Service** 
    - ListVolumes, GetCapacity, ValidateVolumeCapabilities , ControllerGetCapabilities, CreateSnapshot, 
       DeleteSnapshot, ListSnapshots 
 **Node Service** 
    - NodeStageVolume , NodeUnstageVolume , NodeGetVolumeStats , NodeGetCapabilities

We need to check which interface we can implement, if not possible to implement we can atleast add them with default implementation.

https://github.com/container-storage-interface/spec/blob/master/spec.md

adisky:
I am working on this.

jichenjc:
ok, I don't know you are working on this :) as I am trying to try CSI functions and find some missing parts
so I will keep an eye on this and find something that you are not working to wok

adisky:
@jichenjc  No problem, we can work on this together :), just mention here (I will also) when you start working on a particular feature, so that it does not conflict

jichenjc:
I made some changes to create snapshot and seems it works, need add UT ...
so record here to avoid conflict :)

ramineni:
I'm working on topology support

jichenjc:
ControllerExpandVolume need to be added as well..

PR #413, opened 2018-12-27
Remove dependency from csi-common

### LINKED ISSUE #359: [Cinder CSI] Remove Dependency from csi-common package

/kind feature

**What happened**:
Cinder CSI Driver heavily depends on https://github.com/kubernetes-csi/drivers/tree/master/pkg/csi-common, we should remove its dependency from csi-common package as it is not suggested for production deployments and not maintained actively.

To Remove its Dependency we need to develop -
- Non Blocking Grpc server, To run controller, node, Identity services
- Identity Service

adisky:
This PR removes dependency of cinder csi driver from
https://github.com/kubernetes-csi/drivers/tree/master/pkg/csi-common.
This is necessary to have own identity rpc plugin.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #359
partialfixes: #301 

**Release note**:
`NONE`

```release-note
```

### DIFF
--- a/pkg/csi/cinder/identityserver.go
+++ b/pkg/csi/cinder/identityserver.go
@@ -1,0 +1,65 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cinder
+
+import (
+	"github.com/container-storage-interface/spec/lib/go/csi/v0"
+	"golang.org/x/net/context"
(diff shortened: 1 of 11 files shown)


dims:
/approve

leaving the lgtm for @flaper87

pkg/csi/cinder/identityserver.go
```diff
@@ -0,0 +1,65 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cinder
+
+import (
+	"github.com/container-storage-interface/spec/lib/go/csi/v0"
+	"github.com/golang/glog"
```
ramineni: using glog package here, using klog will be taken care in seperate commit?

pkg/csi/cinder/server.go
```diff
@@ -0,0 +1,112 @@
...
+*/
+
+package cinder
+
+import (
+	"net"
+	"os"
+	"sync"
+
+	"github.com/golang/glog"
```
ramineni: ditto

pkg/csi/cinder/identityserver.go
adisky: Thanks for pointing, will update here only.

PR #433, opened 2019-01-14
Add description for the floating ip created for external service

### LINKED ISSUE #432: Add description for the LoadBalancer type service floating IP
/kind feature

**What happened**:
In Magnum k8s cluster, after creating floating ip for LoadBalancer type service inside the cluster, it's impossible to delete the floating ip when the cluster is deleted.

Add description for the fip similar to the lb description.

- create magnum k8s cluster
- create external service inside the cluster, a load balancer is created in octavia and a floating ip is created in neutron and associated with the load balancer vip.
- delete the cluster
- magnum can delete the load balancer automatically, but the floating ip is left.

**Environment**:
- OS (e.g. from /etc/os-release): Ubuntu 16.04
- Kernel (e.g. `uname -a`): 4.4.0-139-generic
- Install tools: magnum
- Others:

lingxiankong:
fixes #432

When creating floating ip for external service, add the cluster and service information in the floating ip description field, so it's easy to query and delete outside of the cluster.

**Release note**:

```release-note
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -279,7 +279,7 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:84f4a50111aeb906b336f15cee8837bbabe7c39c378af9bf529c95b15c361cff"
+  digest = "1:e2161cf62eb56389630bbcdab989c2656e8808edca3dc7cadf25cffda88d6a16"
   name = "github.com/gophercloud/gophercloud"
   packages = [
     ".",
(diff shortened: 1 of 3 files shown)


jichenjc:
> 
> 
> @lingxiankong just one question, after deleting the cluster shouldnt the floating ip automatically released to pool marking it as free , is it required to delete it everytime cluster/service is deleted? sorry if I misunderstood something here.

Have similar thought, k8s cluster should touch LB (ingress) but not the floating ip ?

lingxiankong:
> @lingxiankong just one question, after deleting the cluster shouldnt the floating ip automatically released to pool marking it as free , is it required to delete it everytime cluster/service is deleted? sorry if I misunderstood something here.

This improvement is actually for magnum. In magnum, imagine that a user creates a k8s cluster and created some services of LoadBalancer type inside the cluster, there will be some octavia load balancers and new floating ips created accordingly. 

The lbs and floating ips will be deleted automatically if the user deletes the service explicitly inside the k8s cluster. However, In Magnum, if the user deletes the cluster without deleting the service first, even with [this](https://review.openstack.org/#/c/625767/) patch, the floating ip is not going to be deleted. In the public cloud, 'released to the pool' means the customer still need to be charged for the allocated resources.

This PR is aiming to make it easy to find and delete those floating ips.

Hope that explanation helps :-)

PR #438, opened 2019-01-17
Improve the cfg options for octavia ingress controller

lingxiankong:
Add/change some options for octavia ingress controller configuration, make most of the options consistent with cloud-config, so it's easy to config octavia-ingress-controller together with openstack-cloud-controller-manager.

**Release note**:

```release-note
```

### DIFF
--- a/pkg/ingress/config/config.go
+++ b/pkg/ingress/config/config.go
@@ -18,11 +18,12 @@ package config
 
 import (
 	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/tokens"
 )
 
 // Config struct contains ingress controller configuration
 type Config struct {
-	ClusterName string
+	ClusterName string        `mapstructure:"cluster_name"`
 	Kubernetes  kubeConfig    `mapstructure:"kubernetes"`
 	OpenStack   osConfig      `mapstructure:"openstack"`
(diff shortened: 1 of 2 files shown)


pkg/ingress/config/config.go
```diff
@@ -39,31 +40,50 @@ type kubeConfig struct {
...
 		Password:         cfg.OpenStack.Password,
 		TenantID:         cfg.OpenStack.ProjectID,
-		DomainName:       "default",
+		DomainID:         cfg.OpenStack.DomainID,
+		DomainName:       cfg.OpenStack.DomainName,
+		AllowReauth:      true,
+	}
+}
+
+func (cfg Config) ToV3AuthOptions() tokens.AuthOptions {
```
ramineni: Openstack now supports only v3 right, so maintaining 2 configs is necessary? ToAuthOptions & ToV3AuthOptions? could you let me know if there is any guideline that cloud-provider-openstack maintains compatibility upto this release of openstack?

ramineni: @lingxiankong yes, its been removed in queens release. https://docs.openstack.org/releasenotes/keystone/queens.html#other-notes

lingxiankong: Thanks for the link, so yeah, we don't need to maintain that. But I'd rather do that in a separate PR, thoughts? BTW, just curious, are you using octavia-ingress-controller?

PR #440, opened 2019-01-17
add a log when ccm start failed

jichenjc:
has following error when start ccm after a unexpected update in
configuration of openstack config file

error building controller context: cloud provider could not be initialized:
could not init cloud provider "openstack": Resource not found

'Resource not found' is not really helpful here..

helpful for wrong conf debug

**Release note**:
```release-note
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack.go
+++ b/pkg/cloudprovider/providers/openstack/openstack.go
@@ -189,7 +189,11 @@ func init() {
 		if err != nil {
 			return nil, err
 		}
-		return NewOpenStack(cfg)
+		cloud, err := NewOpenStack(cfg)
+		if err != nil {
+			klog.V(1).Infof("New openstack client created failed with config")
+		}
+		return cloud, err
 	})
 }
 


pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -189,7 +189,11 @@ func init() {
 		if err != nil {
 			return nil, err
 		}
-		return NewOpenStack(cfg)
+		cloud, err := NewOpenStack(cfg)
+		if err != nil {
+			klog.V(1).Infof("New openstack client created failed with config")
```
zetaab: maybe we should print the error as well?

jichenjc: this actually comes from following error:

error building controller context: cloud provider could not be initialized:
could not init cloud provider "openstack": Resource not found

The `resource not found` will anyway be printed in the log but it's not helpful to know what
is the error come from... so add the info in this PR helps to find where's error 
because resource not found (the real error) anyway will be printed, we don't need to add here?

Issue #420, opened 2019-01-07
manila-provisioner missing share-network configuration option


UristMcMiner:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**What happened**:

Trying to provision a share via manila-provisioner in kubernetes with the option driver_handles_share_servers = True.
The provisioning failed with 400: BAD REQUEST due to the missing share_network parameter

**What you expected to happen**:

share_network to be configurable

**How to reproduce it (as minimally and precisely as possible)**:

Use manila-provisioner to provision a share with DHSS enabled

**Anything else we need to know?**:

**Environment**:
- openstack-cloud-controller-manager version:
  N/A
- OS (e.g. from /etc/os-release):
  Ubuntu 16.04
- Kernel (e.g. `uname -a`):
  4.4.0-140-generic
- Install tools:
  N/A
- Others:
Image of manila-provisioner used:
k8scloudprovider/manila-provisioner:1.13.1

PR #443, opened 2019-01-18
New options validator for manila-provisioner

gman0:
Revamp of the previous version of the options validator: cleaner code, easier testing, added a few internal features - this will make https://github.com/kubernetes/cloud-provider-openstack/issues/420 a lot easier to implement with proper input validation

Makes following code no longer necessary as we'll get valid input every time:
https://github.com/kubernetes/cloud-provider-openstack/blob/a63c14dd484381995562dc8c02861b5c557e60db/pkg/share/manila/sharebackends/csicephfs.go#L41-L43

Fixes a bug described here https://github.com/kubernetes/cloud-provider-openstack/pull/370#discussion_r236616224

https://github.com/kubernetes/cloud-provider-openstack/commit/9c066a03ec19d64db7c29729cfc7dc0aa3bc90f8 : unrelated, but the newer version of `external-storage` lib in this repo breaks the deployment because leader election requires access to `endpoints` - this commit fixes it. Hopefully it can be squeezed into this PR.

**Release note**:

```release-note
manila-provisioner: shareSecretNamespace option now defaults to "default" (as opposed to coalescing with osSecretNamespace)
```


dims:
/hold

needs rebase

gman0:
@dims finally passing CI, PTAL

I've added one more commit here https://github.com/kubernetes/cloud-provider-openstack/pull/443/commits/ef56215b90cac0cf92d1f750f2ab2e88e2ec01d5 - turns out I was a bit too hasty with accepting case-insensitive inputs. The `protocol` field still can accept case-insensitive values though, since Manila doesn't care either:

https://github.com/openstack/manila/blob/3017c3239e5056a1cfc992e91bd822baaa589940/manila/share/api.py#L129-L138

PR #451, opened 2019-01-25
cinder-csi-plugin: Add cluster argument to the driver

### LINKED ISSUE #450: cinder-csi-plugin: Volume does not contain any information of the cluster and pv
/kind feature

**What happened**:
The volumes created by cinder-csi-plugin doesn't have any information e.g. the cluster name, the reclaim policy, etc. which are important for the resource management of cloud provider.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

lingxiankong:
Add `cluster` parameter for the driver, admin user could specify the CO's identifier, e.g.  in Magnum, the `cluster` could be the magnum cluster id.

When using cinder-csi-plugin together with kubernetes in the cloud environment, we need some information to recognize the volumes created for the PVs from different kubernetes clusters, which is important for the cloud resource management.

This PR is only adding cluster name to the driver, so that the driver could pass the cluster name to the cinder volume metadata.

**Which issue this PR fixes** : fixes #450

This PR is only adding cluster name to the driver, so that the driver could pass the cluster name to the cinder volume metadata.

There are several ways to inject the cluster name to the cinder volume property, e.g. kubernetes admin can provide parameters when creating the StorageClass, the cloud admin could config `volume-name-prefix` to include the cluster as part of cinder volume name, etc. However, in order to support a more flexible and extensible way, I prefer to add all those information to the cinder volume medata.

**Release note**:

```release-note
cinder-csi-driver supports `--cluster` argument to get the the CO's identifier.
```

### DIFF
--- a/manifests/cinder-csi-plugin/csi-provisioner-cinderplugin.yaml
+++ b/manifests/cinder-csi-plugin/csi-provisioner-cinderplugin.yaml
@@ -47,6 +47,7 @@ spec:
             - /bin/cinder-csi-plugin
             - "--nodeid=$(NODE_ID)"
             - "--endpoint=$(CSI_ENDPOINT)"
+            - "--cluster=$(CLUSTER_NAME)"
             - "--cloud-config=$(CLOUD_CONFIG)"
           env:
             - name: NODE_ID
@@ -57,6 +58,8 @@ spec:
               value: unix://csi/csi.sock
             - name: CLOUD_CONFIG
               value: /etc/config/cloud.conf
+            - name: CLUSTER_NAME
+              value: kubernetes
           imagePullPolicy: "IfNotPresent"
           volumeMounts:
             - name: socket-dir
--- a/cmd/cinder-csi-plugin/main.go
+++ b/cmd/cinder-csi-plugin/main.go
@@ -31,6 +31,7 @@ var (
 	endpoint    string
 	nodeID      string
 	cloudconfig string
+	cluster     string
 )
 
 func init() {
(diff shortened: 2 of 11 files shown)


lingxiankong:
@adisky, as per discussion, the cluster name should be a configuration like what I proposed here, and PV's information should be passed by the external-provisioner. So I think I can carry on with this PR, right?

adisky:
@lingxiankong +1

manifests/cinder-csi-plugin/csi-provisioner-cinderplugin.yaml
```diff
@@ -47,6 +47,7 @@ spec:
             - /bin/cinder-csi-plugin
             - "--nodeid=$(NODE_ID)"
             - "--endpoint=$(CSI_ENDPOINT)"
+            - "--cluster=kubernetes"
```
adisky: you want to point it to CO cluster name not on the CO name??
--cluster=<CO-cluster-name>

lingxiankong: I am using `kubernetes` as the default value just like `--cluster-name string     Default: "kubernetes"` for kube-controller-manager service.

adisky: ok, can you add it via env?? as other arguments. and also update somewhere in doc, about this param and how to get this

lingxiankong: Done

jichenjc:
looks good~

PR #457, opened 2019-01-29
Add support for topology aware dynamic volume provisioning

### LINKED ISSUE #301: Add Features in Cinder CSI Driver

ramineni:
This PR implements following:

1. update metadata struct to return Availability zone information.
2. update NodeGetInfo function to return topology information eg. accessible_topology =
{"topology.cinder.csi.openstack.org/zone": "nova"}
2. consumes topology requirement if present in CreateVolume request and creates
volume using the given zone while favoring preferred topologies

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: 
fixes #301 

**Release note**:

```release-note
```


jichenjc:
@ramineni @adisky not direct relate to this patch but I think we need give a thought (maybe a follow up patch ?) is there any history reason do you know why csi/cinder/openstack/openstack_metadata.go only get metadata from metadata service and not include config drive? 
I know some cases metadata is not mandatory stuff but here csi use metadata service anyway

ramineni:
>>  is there any history reason do you know why csi/cinder/openstack/openstack_metadata.go only get metadata from metadata service and not include config drive?
@jichenjc not sure about why it is not done. But I can see, cloud provider supports the same , we could utilize the same in csi also I suppose. I'll look into it and do it seperate patch.

ramineni:
> @ramineni any question, I seems remember cinder/nova have their az respectively (not 100% sure), and the az info you got is from nova? but csi target to volume creation? will they align?

@jichenjc I have updated NodeGetInfo to retrieve the availability zone of the node only , which would be used by external provisoner . volumes topological requirement  could be specified in AllowedTopologies field of storage class, both would be considered while scheduling of pod.

adisky:
@ramineni as discussed you want to add topology supoort after 1.0.0 update??

adisky:
/hold 
till #470

ramineni:
@adisky right, mainly because I need external-provisoner 1.01 release which has some bugs related to topology addressed which is inturn dependent on CSI 1.0 .

ramineni:
@adisky updated with csi spec 1.0

adisky:
/lgtm 
we can have a test in CI (provided kubernetes multi node cluster, and Openstack with atleast two az)

PR #470, opened 2019-02-04
Update Cinder CSI Driver to CSI Spec 1.0.0

### LINKED ISSUE #365: Update Cinder CSI driver to CSI v1.0.0

>
> /kind bug
> /kind feature

**What happened**:
All CSI sidecar containers and kubernetes  are updated to CSI v1.0.0, we should soon upgrade cinder CSI drive to 1.0.0
https://github.com/container-storage-interface/spec/releases/tag/v1.0.0

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

adisky:
Update cinder csi driver to csi specs 1.0.0

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #365

CSI v0.3.0 support is deprecated with kubernetes 1.13 and will be removed in v1.15

**Release note**:

```release-note
```
`NONE`

### DIFF
--- a/manifests/cinder-csi-plugin/csi-nodeplugin-cinderplugin.yaml
+++ b/manifests/cinder-csi-plugin/csi-nodeplugin-cinderplugin.yaml
@@ -17,17 +17,21 @@ spec:
       serviceAccount: csi-nodeplugin
       hostNetwork: true
       containers:
-        - name: driver-registrar
-          image: quay.io/k8scsi/driver-registrar:v0.4.1
+        - name: node-driver-registrar
+          image: quay.io/k8scsi/csi-node-driver-registrar:v1.0.1
           args:
             - "--v=5"
             - "--csi-address=$(ADDRESS)"
             - "--kubelet-registration-path=$(DRIVER_REG_SOCK_PATH)"
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sh", "-c", "rm -rf /registration/cinder.csi.openstack.org /registration/cinder.csi.openstack.org-reg.sock"]
           env:
             - name: ADDRESS
               value: /csi/csi.sock
             - name: DRIVER_REG_SOCK_PATH
-              value: /var/lib/kubelet/plugins/csi-cinderplugin/csi.sock
+              value: /var/lib/kubelet/plugins/cinder.csi.openstack.org/csi.sock
             - name: KUBE_NODE_NAME
               valueFrom:
                 fieldRef:
@@ -78,11 +82,11 @@ spec:
       volumes:
         - name: socket-dir
           hostPath:
-            path: /var/lib/kubelet/plugins/csi-cinderplugin
+            path: /var/lib/kubelet/plugins/cinder.csi.openstack.org
             type: DirectoryOrCreate
         - name: registration-dir
           hostPath:
-            path: /var/lib/kubelet/plugins/
+            path: /var/lib/kubelet/plugins_registry/
             type: Directory
         - name: pods-mount-dir
           hostPath:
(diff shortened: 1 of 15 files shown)


manifests/cinder-csi-plugin/csi-nodeplugin-cinderplugin.yaml
```diff
@@ -17,17 +17,21 @@ spec:
       serviceAccount: csi-nodeplugin
       hostNetwork: true
       containers:
-        - name: driver-registrar
-          image: quay.io/k8scsi/driver-registrar:v0.4.1
+        - name: node-driver-registrar
+          image: quay.io/k8scsi/csi-node-driver-registrar:v1.0.1
           args:
```
ramineni: why we are not adding cluster-driver-registrar here, it creates a cluster-level CSIDriver object , and says its recommended to easily discover the CSI drivers installed on their clusters.

pkg/csi/cinder/driver.go
```diff
@@ -31,7 +31,7 @@ const (
 )
 
 var (
-	version = "0.3.0"
+	version = "1.0.0"
```
ramineni: Do we maintain always the same versioning according to CSI spec ? or could we add alpha, beta etc to the same to indicate the stability etc?

manifests/cinder-csi-plugin/csi-nodeplugin-cinderplugin.yaml
adisky: @ramineni yes we can deploy cluster-driver-registrar as well, but it necessary now as we are not using any of its features like pod info and skip attach, and one problem with current deployment yamls is we do not have single yaml for controller-plugin, I would prefer to add to cluster-driver-registrar while rearranging yamls for cinder csi plugin.

ramineni: @adisky ok, thanks. sounds good.

pkg/csi/cinder/controllerserver.go
```diff
@@ -210,14 +212,13 @@ func (cs *controllerServer) ListVolumes(ctx context.Context, req *csi.ListVolume
 	vlist, err := cloud.ListVolumes()
 	if err != nil {
 		klog.V(3).Infof("Failed to ListVolumes: %v", err)
-		return nil, err
```
jichenjc: why remove this line? on purpose?

pkg/csi/cinder/controllerserver.go
```diff
@@ -266,13 +267,18 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		klog.V(3).Infof("CreateSnapshot %s on %s", name, volumeId)
 	}
 
+	ctime, err := ptypes.TimestampProto(snap.CreatedAt)
+	if err != nil {
+		klog.Errorf("Error to convert time to timestamp: %v", err)
```
jichenjc: this will return ctime as 0 ? we will use this value even error occur here..

pkg/csi/cinder/driver.go
jichenjc: guess keep with CSI spec will make it easier to sync and understand......

pkg/csi/cinder/controllerserver.go
adisky: yes, it is mandatory to return as type Timestamp as per CSI compliance

pkg/csi/cinder/controllerserver.go
adisky: will fix this

pkg/csi/cinder/driver.go
adisky: keeping it as it is

Issue #476, opened 2019-02-08
client-keystone-auth: pipe and arguments doesn't work


johscheuer:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
>
/kind bug

**What happened**:

I use the `client-keystone-auth` with only arguments passed (see the following snippet):

```yaml
- name: redacted
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - --domain-name=redacted
      - --keystone-url=redacted
      - --project-name=redacted
      - --user-name=redacted
      - --password=redacted
      command: ./bin/client-keystone-auth
      env: null
```

everything works as expected when using `kubectl` without a pipe otherwise I get the following error message:

```bash
Failed to read openstack env vars: Missing environment variable [OS_AUTH_URL]
```

**What you expected to happen**:

I expected that kubectl authenticate against Keystone and passes the output to the next command.

**How to reproduce it (as minimally and precisely as possible)**:

Set only arguments in the kubeconfig and run a command like:  `kubectl get po --all-namespaces | grep 'a'`

**Anything else we need to know?**:

I already found the error: This command checks if STDIN is a terminal https://github.com/kubernetes/cloud-provider-openstack/blob/master/cmd/client-keystone-auth/main.go#L166 which returns in the case of `kubectl ...` `false` but in the case of `kubectl ... | do stuff` `true`. In the next step we jump in the [openstack.AuthOptionsFromEnv()](https://github.com/kubernetes/cloud-provider-openstack/blob/master/cmd/client-keystone-auth/main.go#L167) method which ignores the arguments.

There are two possible options to tackle this problem:

1.) Implement a fallback (if env variables are not set) with the specified arguments
2.) Set the required Env Variables for the auth method and unset them afterwards

I also would expect that an argument overwrites the value of an env variable (like it's coded).

**Environment**:
- openstack-cloud-controller-manager version: 1.13.1
- OS (e.g. from /etc/os-release): darwin / linux
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #477, opened 2019-02-08
Implement arguments fallback for client-keystone-auth

johscheuer:

Without this PR it's not possible to only use arguments in the `client-keystone-auth` and a pipe.

Fixes: https://github.com/kubernetes/cloud-provider-openstack/issues/476

**Release note**:

```release-note
client-keystone-auth: change the default behavior to use cli arguments if set otherwise use env variables for authentication
```

### DIFF
--- a/cmd/client-keystone-auth/main.go
+++ b/cmd/client-keystone-auth/main.go
@@ -119,6 +119,10 @@ func prompt(url string, domain string, user string, project string, password str
 	return options, nil
 }
 
+// consolidate input and env
+// set env (and a list to rember)
+// after run clean up
+
 func main() {
 	// Glog requires this otherwise it complains.
 	flag.CommandLine.Parse(nil)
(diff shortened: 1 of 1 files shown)


johscheuer:
A side effect of this approach is the line `Failed to read openstack env vars: Missing environment variable [OS_AUTH_URL]`:

```bash
kubectl get po --all-namespaces | grep 'example'
Failed to read openstack env vars: Missing environment variable [OS_AUTH_URL]
example         details-v1-68c7c8666d-gr9h2                           2/2     Running     0          8h
example         productpage-v1-54d799c966-f4lnt                       2/2     Running     0          8h
example         ratings-v1-8558d4458d-p87hs                           2/2     Running     4          14h
example         reviews-v1-cb8655c75-tmc25                            2/2     Running     0          8h
example         reviews-v2-7fc9bb6dcf-ng67k                           2/2     Running     0          8h
example         reviews-v3-c995979bc-zxcgj                            2/2     Running     0          8h
```

We could omit this error if the token fetch was successfully and otherwise print it.

hogepodge:
This patch needs a release note and a documentation update since it's making a major change to the auth interface.

johscheuer:
@hogepodge I added a release note actually nothing (at least from the user point of view) has changed here: https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/using-client-keystone-auth.md#input-and-output-formats expect that now "only arguments" are working.

PR #485, opened 2019-02-18
Add cluster info into GetPluginInfo

### LINKED ISSUE #484: Add cluster info into GetPluginInfoResponse manifest output
/kind feature
>
> /kind bug
> /kind feature

**What happened**:
we return Name and VendorVersion into GetPluginInfo,
add cluster info would be better 

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

jichenjc:
GetPluginInfo can container cluster name for reference
so this patch added the cluster.

add cluster info into GetPluginInfo CSI call 
fixes #484 

**Release note**:
```release-note
Added the cluster info into CSI call GetPluginInfo 
```

### DIFF
--- a/pkg/csi/cinder/identityserver.go
+++ b/pkg/csi/cinder/identityserver.go
@@ -39,9 +39,12 @@ func (ids *identityServer) GetPluginInfo(ctx context.Context, req *csi.GetPlugin
 		return nil, status.Error(codes.Unavailable, "Driver is missing version")
 	}
 
+	manifest := map[string]string{"cinder.csi.openstack.org/cluster": ids.Driver.cluster}
+
 	return &csi.GetPluginInfoResponse{
 		Name:          ids.Driver.name,
 		VendorVersion: ids.Driver.version,
+		Manifest:      manifest,
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/csi/cinder/identityserver.go
```diff
@@ -39,9 +39,12 @@ func (ids *identityServer) GetPluginInfo(ctx context.Context, req *csi.GetPlugin
 		return nil, status.Error(codes.Unavailable, "Driver is missing version")
 	}
 
+	manifest := map[string]string{"cinder.csi.openstack.org/cluster": ids.Driver.cluster}
```
adisky: This method meant to be called form, CSI sidecars(registrar, provisioner etc). IMO, it is not needed to pass cluster info to kubernetes sidecars

lingxiankong: Agree, for now the cluster name is only used in the cinder-provisioner and specified by the k8s cluster admin or cloud operator, I don't see there is a reason for COs to get that information.

jichenjc: ok, I am still new to CSI core stuffs... I will read more about side cars, thanks for the comments~

PR #502, opened 2019-02-22
[octavia-ingress-controller] Support security group management

lingxiankong:
Support security group management in octavia-ingress-controller. In most of the production environments, the security group of the cluster nodes is more strict and the default port range of service ports may not be exposed. As a result, it's neccessary for octavia-ingress-controller to manage the security groups on kubernetes cluster admin's behalf.

This PR introduces a new bool type option `manage_security_groups` in octavia-ingress-controller configuration with the default value `false`, so the ingress controller doesn't manage security groups. The kubernetes cluster admin could change it to `true` so that the security group could be automatically managed.

Requirements to the OpenStack environment:
- Neutron needs to support [Standard Attributes Tag Extension](https://developer.openstack.org/api-ref/network/v2/index.html#standard-attributes-tag-extension)

**Release note**:

```release-note
```


adisky:
/lgtm
@lingxiankong any reason you are using klog and logrus both?or octavia ingress controller not switched to klog?

lingxiankong:
> @lingxiankong any reason you are using klog and logrus both?or octavia ingress controller not switched to klog?

Actually, the logrus is mainly used in octavia-ingress-controller, the klog is only used for eventBroadcaster, I am not sure if I could replace that with logrus, so just leave it as is.

hogepodge:
I'm a bit late on reviewing this, but I'd like to remind everyone that when we add new features they really need to be documented both with a release note (to make it easier to spot all of the changes that happened in the last release) and usage documentation so that other users can take advantage of them.

lingxiankong:
> we add new features they really need to be documented both with a release note (to make it easier to spot all of the changes that happened in the last release) and usage documentation

Hi Chris, thanks for the reminder. Maybe I should also mention in the PR description that I will submit a separate PR for the documentation, because currently I'm still working on some improvements for octavia-ingress-controller, and I am planning to change the doc after all the coding work is done.

Issue #483, opened 2019-02-18
Cinder CSI should add error probe


jichenjc:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind feature
> Uncomment only one, leave it on its own line: 
>
> /kind bug
> /kind feature

**What happened**:
Current CSI plugin always assume cinder is not healthy, we need check it
**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- openstack-cloud-controller-manager version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #508, opened 2019-02-27
Add probe for CSI

jichenjc:
Add CSI identity CSI so that we can return CSI status back

fixes: #483 

some discussion might be considered , whether we need create a client to auth or use cinder to do some query etc...

**Release note**:

```release-note
```

### DIFF
--- a/pkg/csi/cinder/identityserver.go
+++ b/pkg/csi/cinder/identityserver.go
@@ -21,6 +21,7 @@ import (
 	"golang.org/x/net/context"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"k8s.io/cloud-provider-openstack/pkg/csi/cinder/openstack"
 	"k8s.io/klog"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/csi/cinder/identityserver.go
```diff
@@ -46,6 +47,11 @@ func (ids *identityServer) GetPluginInfo(ctx context.Context, req *csi.GetPlugin
 }
 
 func (ids *identityServer) Probe(ctx context.Context, req *csi.ProbeRequest) (*csi.ProbeResponse, error) {
+	_, err := openstack.CreateOpenStackProvider()
+	if err != nil {
+		klog.V(3).Infof("Failed to CreateOpenStackProvider: %v", err)
```
lingxiankong: Shouldn't the log level be ERROR?

jichenjc: right, but actually I copied the code :)
we may need another PR to fix those..

lingxiankong: ok, good to know, thanks!

lingxiankong:
Could you please change the PR description? Put a `#` before the issue ID, so the issue could be correctly closed after this PR merged

jichenjc:
Thanks for the reminder, done

PR #509, opened 2019-02-27
use CSI error code instead of own defined error

jichenjc:
CSI spec has error code defined

NOTE: this should have more included, but create a PR for discussion and see whether something need to be corrected or use additional stuff that I am not aware...

**Release note**:

```release-note
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -18,6 +18,7 @@ package cinder
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/golang/protobuf/ptypes"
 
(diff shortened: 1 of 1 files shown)


pkg/csi/cinder/controllerserver.go
```diff
@@ -283,14 +284,16 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+		return nil, status.Error(codes.FailedPrecondition, err)
 	} else {
 		// TODO: Delegate the check to openstack itself and ignore the conflict
 		snap, err = cloud.CreateSnapshot(name, volumeId, description, &req.Parameters)
 		if err != nil {
-			klog.V(3).Infof("Failed to Create snapshot: %v", err)
-			return nil, err
+			errStr := fmt.Sprintf("Failed to Create snapshot: %v", err)
+			klog.V(3).Infof(errStr)
+			return nil, status.Error(codes.Aborted, errStr)
```
ramineni: It should be codes.Internal instead? codes.Aborted Indicates that there is already an operation pending for the specified snapshot which might not be correct in this case?

directly status.Errorf can be used instead. 
`status.Errorf(codes.Internal, "Failed to create snapshot %v", err)`

pkg/csi/cinder/controllerserver.go
```diff
@@ -303,8 +306,9 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 
 	err = cloud.WaitSnapshotReady(snap.ID)
 	if err != nil {
-		klog.V(3).Infof("Failed to WaitSnapshotReady: %v", err)
-		return nil, err
+		errStr := fmt.Sprintf("Failed to WaitSnapshotReady: %v", err)
+		klog.V(3).Infof(errStr)
+		return nil, status.Error(codes.Aborted, errStr)
```
ramineni: codes.Internal ?

pkg/csi/cinder/controllerserver.go
jichenjc: ok, thanks for the info~

pkg/csi/cinder/controllerserver.go
jichenjc: got it

adisky:
/hold
lot of failures
@jichenjc can we add this after csi-sanity merge?? there are lot more error code needs to be updated.
but it can take a while, if you want to this to add, please fix the failures

jichenjc:
sure, let me wait for sanity merge and work on this after that...thanks

jichenjc:
@adisky because you are doing update for CSI sanity check ,should we close this PR ? I believe you already updated everything?

adisky:
@jichenjc mostly error codes fixed in https://github.com/kubernetes/cloud-provider-openstack/pull/629

adisky:
Thanks for your efforts @jichenjc 
/close

PR #515, opened 2019-03-01
Trivial: Remove meaningless charactor in doc

oomichi:

TBH, I am not sure why there is.

**Release note**:

```release-note
NONE
```


oomichi:
@lingxiankong Thanks for your comment, you are right.
It is fixed already with 4c9ab513910abcff005f0753a362e785b2e74134

PR #518, opened 2019-03-04
Support new annotation for Service

lingxiankong:
Support to specify port ID to create Service of LoadBalancer type.

Use case:
Sometimes, we need to know the service IP address before a Service is actually created, so other applications who replies on the Service external IP could be deployed with no dependencies, e.g. the kube-apiserver certificate could be created before the service is actually created.

In order to specify port ID to create Service, a new annotation for Service is introduced, e.g.:
```yaml
annotations:
  loadbalancer.openstack.org/port-id: "849ee2ed-72a6-44a4-911e-6a8be46dbe8e"
```

Please be aware that after the Service is created, port ID cannot be changed.

**Release note**:

```release-note
Introduce a new annotation(`loadbalancer.openstack.org/port-id`) for Service of LoadBalancer type to specify a particular Neutron port for creating the Octavia load balancer.
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -277,7 +277,7 @@
     "testhelper/client",
   ]
   pruneopts = "UT"
-  revision = "b18d22ae2c8b9ef046c980e62422369b5aea953f"
+  revision = "f83aee3da90f3f337d024070e2e68b04fb035149"
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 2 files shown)


jichenjc:
>>>service IP address before a Service is actually created
can you help to share why this is needed? just curious the use case.. thanks

so the port id which we got from a port need to be queried ? seems this expose too much detail on the cloud itself..

lingxiankong:
> can you help to share why this is needed? just curious the use case.. thanks
> so the port id which we got from a port need to be queried ? seems this expose too much detail on the cloud itself..

Think about creating k8s in k8s, before the api service is up and running, we need to create kube-apiserver certificate which needs the all the dnsNames such as 127.0.0.1, kubernetes.default, private ip and potential floating ip, etc, which means we need to know the ip addresses information before the kube-apiserver Service is created.

That is a real use case from us in order to do k8s service management in the public cloud.

Make sense?

PR #533, opened 2019-03-13
Allow to specify default value in cloud-config for openstack-internal…

JrCs:
…-load-balancer

Signed-off-by: JrCs <90z7oey02@sneakemail.com>

Add on option (`internal-lb`) in the cloud-config configuration file so we can set the default value for openstack-internal-load-balancer

### DIFF
--- a/docs/provider-configuration.md
+++ b/docs/provider-configuration.md
@@ -202,6 +202,8 @@ file.
   Where `true` is specified and an Octavia LBaaS V2 entry can not be found, the
   provider will fall back and attempt to find a Neutron LBaaS V2 endpoint
   instead. The default value is `false`.
+* `internal-lb`: Determines whether or not to create an internal load balancer
+  (no floating IP) by default. The default value is `false`.
 
 #### Block Storage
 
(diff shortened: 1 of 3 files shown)


lingxiankong:
Why we need this? We already have the `service.beta.kubernetes.io/openstack-internal-load-balancer` annotation which could do the same thing, and it's flexible. I didn't see this PR brings much value here.

lingxiankong:
/hold

Wait for author's response.

JrCs:
@lingxiankong Yes we can use the `service.beta.kubernetes.io/openstack-internal-load-balancer` annotation to change the behavior for a particular service. But by default a floating IP is always created if you don't specify the annotation.
In my company we do not use an internal openstack network for the VM but 2 "external" networks: 1 for the VM and 1 for the VIP of loadbalancer. So in our case we don't want to create a floationg IP for every service.
So the solution to not create an annotation for EVERY service to disable the creation of floating IP is to have an option to change the **default** value of the  `openstack-internal-load-balancer` option.
I any case the **default** value can be overwritten by an annotation.

JrCs:
Is it ok for you ?

lingxiankong:
i'm ok with this change, please rebase your commit and make sure all tests are passed.

lingxiankong:
/lgtm
/approve

Thanks!

PR #542, opened 2019-03-19
Add missing /var/lib/cloud mount to pod volumes

### LINKED ISSUE #541: missing  pods-cloud-data in openstack-cloud-controller-manager-ds.yaml
> /kind bug

**What happened**:
I was trying to run the cloud-controller-manager and encountered a lot of the following messages:
> Error determining if node master-1 shutdown in cloud: ProviderID "" didn't match expected format "openstack:///InstanceID" 

See also the discussion in #540 

It turns out, that this happens because  [readInstanceID](https://github.com/kubernetes/cloud-provider-openstack/blob/664540459d96aaf10b809352e33a2a3c129241ef/pkg/cloudprovider/providers/openstack/openstack.go#L364) tries to read the file `/var/lib/cloud/data/instance-id`
which does not exist in the container.

This is already fixed in [csi-nodeplugin-cinderplugin](https://github.com/kubernetes/cloud-provider-openstack/blob/888e3c5b16ef992fb51b675693362c4c03aee3b0/manifests/cinder-csi-plugin/csi-nodeplugin-cinderplugin.yaml#L75)

The function [readInstanceID](https://github.com/kubernetes/cloud-provider-openstack/blob/664540459d96aaf10b809352e33a2a3c129241ef/pkg/cloudprovider/providers/openstack/openstack.go#L364) should return the instance properly.

**Environment**:
- openstack-cloud-controller-manager version: 1.13.1
- OS : Ubuntu 16.04

oz123:
This fixes #541.

See also #540.

**Which issue this PR fixes** :
fixes #541 

**Release note**:

```release-note
```

### DIFF
--- a/manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml
+++ b/manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml
@@ -56,6 +56,9 @@ spec:
               readOnly: true
             - mountPath: /usr/libexec/kubernetes/kubelet-plugins/volume/exec
               name: flexvolume-dir
+            - name: pods-cloud-data
+              mountPath: /var/lib/cloud/data
+              readOnly: true
           resources:
             requests:
               cpu: 200m
@@ -79,3 +82,7 @@ spec:
       - name: cloud-config-volume
         secret:
           secretName: cloud-config
+      - name: pods-cloud-data
+        hostPath:
+          path: /var/lib/cloud/data
+          type: Directory


oz123:
@adisky, this line
 384,

	idBytes, err := ioutil.ReadFile(instanceIDFile)

Already returned error, how should this be handled? Should this stop the execution? 
The function return this error to the next call, should this be handled there? 
Can you please suggest a fix and post a snippet here?

adisky:
@oz123 Are you sure you get some error in case of not found?? if it is returning error this should get printed in logs
https://github.com/kubernetes/cloud-provider/blob/master/cloud.go#L97
you can fix by checking file is mounted or not, not relying on error from, ioutil.ReadFile(instanceIDFile)

oz123:
@adisky check the logs posted by my colleague  @joroec  in #540 . No where we see an error about the file not found.

adisky:
@oz123 It needs to be checked more, i think the check can be added in a seperate PR, revert your PR to previous commit only, will add the check later

oz123:
> @oz123 It needs to be checked more, i think the check can be added in a seperate PR, revert your PR to previous commit only, will add the check later

@adisky I reverted the changes.

lingxiankong:
Please take a look at my reply in the issue.

/hold

oz123:
@lingxiankong,
- why this is added then in other manifests? see csi-nodeplugin-cinderplugin
- also, how do you explain that adding this causes the messages 
`Error determining if node master-1 shutdown in cloud: ProviderID "" didn't match expected format "openstack:///InstanceID"` disappeared? see #540

lingxiankong:
> @lingxiankong,
> 
> * why this is added then in other manifests? see csi-nodeplugin-cinderplugin
> * also, how do you explain that adding this causes the messages
>   `Error determining if node master-1 shutdown in cloud: ProviderID "" didn't match expected format "openstack:///InstanceID"` disappeared? see #540

Let's discuss in the issue first before we actually review this PR.

adisky:
The relevant issue is closed, closing the PR
Feel Free to Reopen if necessary!!
/close

oz123:
@adisky  the issue was closed without any discussion or no real explanation why other components have the config file and the controller does not. I'd really love to understand this.
As for re-opening: obviously I can't, and I don't want to create spam by creating a new PR.
I'd would be more than happy to know why the cloud.config file isn't needed in the controller controller.
Can you shed some light on this here?

lingxiankong:
@oz123, first, the PR is totally not needed, that manifest could work well without any volume mapping. Second, as I said in issue #540, you need to fix your cluster problem first(Response Status: 409 for Node update)

oz123:
@lingxiankong  I would be happy to fix with, but I am totally helpless and the issue is not solved with your ansible playbooks. 
Can you explain where is the controller getting it's configuration from if not from the config file?

lingxiankong:
@oz123 My ansible playbook works in our openstack cloud, not sure how your cloud was deployed but I suggest you understand how the playbook works before actually running. BTW, I moved the original ansible role to a separate repo [here](https://github.com/lingxiankong/ansible-kubernetes-on-openstack).

The controller is getting its configuration from the secret [here](https://github.com/kubernetes/cloud-provider-openstack/blob/7f1127e54587615d4102a90f6afe7454e1eb84e6/manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml#L81)

PR #543, opened 2019-03-20
initialize explictly to enable verbose option

### LINKED ISSUE #539: enable log
> /kind feature

**What happened**:
I have to debug communication between the pod and my openstack. I cannot pass -v4 or -v5 to the kubernetes manifest

```
# kubectl exec -it csi-provisioner-cinderplugin-0 -n kube-system -c cinder -- /bin/sh

sh-4.2# /bin/cinder-csi-plugin
Error: required flag(s) "cloud-config", "endpoint", "nodeid" not set
Usage:
  Cinder [flags]

Flags:
      --cloud-config string     CSI driver cloud config
      --cluster string          The identifier of the cluster that the plugin is running in.
      --endpoint string         CSI endpoint
  -h, --help                    help for Cinder
      --httptest.serve string   if non-empty, httptest.NewServer serves on this address and blocks
      --nodeid string           node id
```

I expected that -v would be available in the flags

Just launch the manifest and check /bin/cinder-csi-plugin inside one of the pods

ramineni:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #539 

**Release note**:

```release-note
```

### DIFF
--- a/cmd/cinder-csi-plugin/main.go
+++ b/cmd/cinder-csi-plugin/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"k8s.io/cloud-provider-openstack/pkg/csi/cinder"
+	"k8s.io/component-base/logs"
 	"k8s.io/klog"
 )
 
@@ -80,6 +81,9 @@ func main() {
 	cmd.MarkPersistentFlagRequired("cloud-config")
 
 	cmd.PersistentFlags().StringVar(&cluster, "cluster", "", "The identifier of the cluster that the plugin is running in.")
+
+	logs.InitLogs()
+	defer logs.FlushLogs()
 
 	if err := cmd.Execute(); err != nil {
 		fmt.Fprintf(os.Stderr, "%s", err.Error())


dricoco:
Can we change the label of this issue for the pending test?

jichenjc:
is this ongoing?

dricoco:
What do you mean by ongoing ?

jichenjc:
The title of this PR is wip: so I think it's work in progress?

dricoco:
I don't get why there is wip in the title as this is only one line enabling debug in klog.
I don't see what extra work would be added in this merge request
@adisky have you an idea?

dims:
/hold

need to look deeper, apparently openstack-cloud-controller-manager works but this binary fails, so we need to check what's different between the two. We are calling `klog.InitFlags` a few lines below that

ramineni:
@dims updated the patch as discussed , PTAL

dricoco:
Sorry This isn't working

I had to remove the init function and move   flag.Set("logtostderr", "true")

as first line in main() to have it work

PR #549, opened 2019-03-20
Improve load balancer name

lingxiankong:
- Using a more meaningful name for load balancer created by Service.
- Support backward compatibility.
- Adding doc for LoadBalancer type Service guide.

**Release note**:

```release-note
The load balancer name created by Service has been changed, now the name includes cluster name, Service namespace and Service name. The services created before this change are not affected.
```


dims:
@lingxiankong  is there any implication during upgrades? (move from older to newer version of the external cloud provider)

PR #560, opened 2019-03-22
Update CI conformance jobs - 1.14/master

mrhillsman:

Conformance jobs should only be running relevant to the branch they appear on. The periodic jobs for 1.10 and 1.11 are being removed due to being EOL and when run pushing the wrong binaries with the tag latest to dockerhub.

We need to merge this as soon as possible due to it pushing the wrong binaries to dockerhub using the latest tag. I recommend ignoring OpenLab CI errors and forcing it.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -71,22 +71,6 @@
               - ^.*\.md$
               - ^OWNERS$
               - ^SECURITY_CONTACTS$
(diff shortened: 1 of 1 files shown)


ramineni:
@mrhillsman looks like this patch not needed now?

mrhillsman:
@ramineni it is not, thx

PR #575, opened 2019-04-01
Adding support for CA specfication

hyperbolic2346:
This PR adds the ability to specify a CA that isn't installed on the system to use for the client-kubernetes-auth plugin. Handled the same way that client cert and key are handled and keys off the OS_CACERT environment variable.

**Release note**:
```release-note
Added support for specifying a certificate authority that isn't installed on the client machine via OS_CACERT environment variable.
```

### DIFF
--- a/cmd/client-keystone-auth/main.go
+++ b/cmd/client-keystone-auth/main.go
@@ -173,10 +173,10 @@ func main() {
 	pflag.StringVar(&password, "password", os.Getenv("OS_PASSWORD"), "Password")
 	pflag.StringVar(&clientCertPath, "cert", os.Getenv("OS_CERT"), "Client certificate bundle file")
 	pflag.StringVar(&clientKeyPath, "key", os.Getenv("OS_KEY"), "Client certificate key file")
+	pflag.StringVar(&clientCAPath, "cacert", os.Getenv("OS_CACERT"), "Certificate authority file")
 	pflag.StringVar(&applicationCredentialID, "application-credential-id", os.Getenv("OS_APPLICATION_CREDENTIAL_ID"), "Application Credential ID")
 	pflag.StringVar(&applicationCredentialName, "application-credential-name", os.Getenv("OS_APPLICATION_CREDENTIAL_NAME"), "Application Credential Name")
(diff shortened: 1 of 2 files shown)


cmd/client-keystone-auth/main.go
```diff
@@ -175,6 +176,7 @@ func main() {
 	pflag.StringVar(&applicationCredentialID, "application-credential-id", os.Getenv("OS_APPLICATION_CREDENTIAL_ID"), "Application Credential ID")
 	pflag.StringVar(&applicationCredentialName, "application-credential-name", os.Getenv("OS_APPLICATION_CREDENTIAL_NAME"), "Application Credential Name")
 	pflag.StringVar(&applicationCredentialSecret, "application-credential-secret", os.Getenv("OS_APPLICATION_CREDENTIAL_SECRET"), "Application Credential Secret")
+	pflag.StringVar(&clientCAPath, "cacert", os.Getenv("OS_CACERT"), "Certificate authority file")
```
chaosaffe: nit: Group with other `client*Path`

pkg/identity/keystone/token_getter.go
```diff
@@ -64,7 +68,28 @@ func GetToken(options Options) (*tokens3.Token, error) {
...
+	}
+
+	if options.ClientCAPath != "" {
+		certBytes, err := ioutil.ReadFile(options.ClientCAPath)
+		if err != nil {
+			msg := fmt.Errorf("failed: Cannot read CA file: %v", err)
+			return token, msg
+		}
+
+		certpool, err := x509.SystemCertPool()
```
ramineni: could we use kk lib instead here "k8s.io/client-go/util/cert" which makes consistent with other modules ? and directly use cert.NewPool(options.ClientCAPath) which handles the reading file etc?

hyperbolic2346: @ramineni Thanks for the pointer. I like that much better and consistency is best!

ramineni:
@hyperbolic2346 Thanks for the update. Looks like format job is failing because of imports order. You could run hack/update-gofmt.sh , hack/verify-gofmt.sh to check the same locally. Otherwise patch looks good to me.

hyperbolic2346:
@ramineni Thanks for the pointer. I didn't realize that the ordering would be crucial and had tried to follow the pattern. Updated.

PR #578, opened 2019-04-04
Do not get load balancers in deleted or deleting status

### LINKED ISSUE #576: Service: Load balancer not created for the Service
/kind bug

**What happened**:
Create a LoadBalancer type Service first then delete, and quickly create another one, Octavia load balancer is not created. The OCCM log said the load balancer already exists. Although after for a while, the load balancer is created eventually but that sounds like a bug to the end user.

Load balancer should be created for the second Service.

- Create a LoadBalancer type Service1
- Delete Service1
- Quickly create a second one with the same name

**Environment**:
- openstack-cloud-controller-manager version: master branch as of 2 Apr
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

lingxiankong:
Ignore the load balancers in deleting or deleted status when doing query from load balancing service.

fixes #576

**Release note**:

```release-note
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
@@ -171,7 +171,10 @@ func getLoadBalancers(client *gophercloud.ServiceClient, opts loadbalancers.List
 	return allLoadbalancers, nil
 }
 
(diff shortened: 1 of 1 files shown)


lingxiankong:
@mrhillsman @dims Could you please take a look at the CI failure, we can't merge any PR. Seems like it's related to the k8s master installation, but I'm not familiar with the CI job.

PR #580, opened 2019-04-05
Allow to specify a Tenant Domain Name/ID

Fedosin:
This commit adds a possibility to specify a domain name/id for a tenant other than the default one.

/kind feature

```release-note
Allow to specify a domain name/id for a tenant other than the default one for the OpenStack cloud provider.
```

### DIFF
--- a/docs/provider-configuration.md
+++ b/docs/provider-configuration.md
@@ -103,11 +103,10 @@ file.
   OpenStack control panels, this can be found at Access and Security > API
   Access > Credentials.
 * `password`: Refers to the password of a valid user set in keystone.
-* `tenant-id`: Used to specify the ID of the project where you want
-  to create your resources. When using Keystone V3 - which changed the 
-  identifier `tenant` to `project` - the `tenant-id` value is automatically
-  mapped to the project construct in the API.
 * `username`: Refers to the username of a valid user set in keystone.
+
+It is also required to provide either `tenant-id` only or `tenant-name` together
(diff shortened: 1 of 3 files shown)


ramineni:
Thanks. Code looks good. May be we could document the support for the same in docs as well. https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/provider-configuration.md

Fedosin:
@adisky hi! done

docs/provider-configuration.md
```diff
@@ -103,10 +103,6 @@ file.
   OpenStack control panels, this can be found at Access and Security > API
   Access > Credentials.
 * `password`: Refers to the password of a valid user set in keystone.
-* `tenant-id`: Used to specify the ID of the project where you want
```
ramineni: tenant-id /tenant-name one of it should be part of required parameter list rather than optional?

Fedosin: tenant-id is an optional parameter, you may not specify it in a case you have tenant-name along with either domain-name or domain-id.
I think the best option would be to add information about this behavior in the "required" section.

ramineni: Thanks. sounds good

ramineni:
/lgtm
Thanks

adisky:
> /test cloud-provider-openstack-acceptance-test-e2e-conformance

@Fedosin This is gate failure, openlab team is working on fix

Fedosin:
@adisky yep, I see :)

PR #582, opened 2019-04-05
Adding TrustID to authentification

### LINKED ISSUE #581: trust-id not used in auth for CSI
/kind bug

**What happened**:
I'm working on a template with openstack magnum using CSI. Magnum provides a trustee to the template and this is used to have the cloud-config config file done.
The problem is that trust-id parameter is not used in the authentification process between the CSI pod and this leads to the following error : 

> I0405 11:07:46.749363       1 server.go:108] Listening for connections on address: &net.UnixAddr{Name:"/csi/csi.sock", Net:"unix"}
I0405 11:07:46.810438       1 utils.go:73] GRPC call: /csi.v1.Identity/Probe
I0405 11:07:46.810457       1 utils.go:74] GRPC request:
I0405 11:07:47.517388       1 identityserver.go:52] Failed to CreateOpenStackProvider: No suitable endpoint could be found in the service catalog.

in the pods log.

in the JSON sent from the pod to openstack we can see: 

> {"auth":{"identity":{"methods":["password"],"password":{"user":{"domain":{"id":"c5503d1bd8b14141b782cad252c6b7cb"},"name":"b67ae13d-5c93-4374-8902-712f291de94f_968b910e7f0b46b586ad967a1082a693","password":"CCC"}}}}}

we don't have any trust id.

If I reproduce those parameter using CLI in openstack I get 

> user@stack:~# openstack port list
The service catalog is empty.

If I add the 
OS_TRUST_ID=xxx
in my CLI file, then it works fine

CSI should work with a trustee user as it is stated in the docs
create a trustee user, configure the cloud-config, launch a pod

dricoco:
We need to have Trust-id working with openstack magnum
fixes #581

**Release note**:

```release-note
```


dricoco:
/assign @dixudx

can you help see why some unrelated test for this issue are failing ?

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -61,13 +63,27 @@ type Config struct {
...
 		DomainId   string `gcfg:"domain-id"`
 		DomainName string `gcfg:"domain-name"`
 		Region     string
 		CAFile     string `gcfg:"ca-file"`
 	}
 }
 
+func logcfg(cfg Config) {
+	klog.V(3).Infof("AuthURL: %s", cfg.Global.AuthUrl)
+	klog.V(3).Infof("Username: %s", cfg.Global.Username)
```
adisky: v(5) is better, v(3) will log always

dricoco: Hi,
Just changed to 5.

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -61,13 +63,27 @@ type Config struct {
...
+		TrustID    string `gcfg:"trust-id"`
 		DomainId   string `gcfg:"domain-id"`
 		DomainName string `gcfg:"domain-name"`
 		Region     string
 		CAFile     string `gcfg:"ca-file"`
 	}
 }
 
+func logcfg(cfg Config) {
+	klog.V(5).Infof("AuthURL: %s", cfg.Global.AuthUrl)
```
dixudx: Why bothering adding these creds into logs?

dricoco: because it's easier to debug then. this code part is present in the controller image also.
If I would have had them, I would'nt had to use ngrep between the kubernetes and my openstack controller to figure out that the trust-id wasn't transmitted I would have seen it in the logs.

this also helps to see that your cloud-config file is correct. There is no password, so probably no security issues.

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -84,6 +100,18 @@ func (cfg Config) toAuthOptions() gophercloud.AuthOptions {
 	}
 }
 
+func (cfg Config) toAuth3Options() tokens3.AuthOptions {
```
ramineni: only v3 is supported in csi, so no need to create new cfg here , we can reuse existing one

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -167,7 +196,17 @@ func CreateOpenStackProvider() (IOpenStack, error) {
 		provider.HTTPClient.Transport = netutil.SetOldTransportDefaults(&http.Transport{TLSClientConfig: config})
 	}
 
-	err = openstack.Authenticate(provider, authOpts)
+	if cfg.Global.TrustID != "" {
+		opts := cfg.toAuth3Options()
+		authOptsExt := trusts.AuthOptsExt{
+			TrustID:            cfg.Global.TrustID,
+			AuthOptionsBuilder: &opts,
```
ramineni: we can reuse cfg.toAuthOptions() here 
opts := cfg.toAuthOptions()

pkg/csi/cinder/openstack/openstack.go
ramineni: we could move to common place as used in multiple places . may be pkg/util/util.go?

pkg/csi/cinder/openstack/openstack.go
dricoco: are you meaning cinder v3 ?

pkg/csi/cinder/openstack/openstack.go
dricoco: just to be sure it identity v3 only ?

ramineni: then you can change it to use only v3 , cfg.toAuth3Options() ?  I dont see the need for maintaining both . wdyt?

dricoco: I'll check if this is working without the tokensv3 import, I'll let you know

dricoco: I think some people may not have the rights endpoints enabled in their openstack, or using old openstack versions where identity v3 is not yet present.
Thats 3 lines, so not a lot to maintain

dricoco:
if I can be of some help, just let me know

dricoco:
@adisky : are the changes ok for you, could we merge it ?

pkg/csi/cinder/openstack/openstack.go
lingxiankong: @ramineni hi, any related doc mentions this(only v3 is supported in csi)?

adisky: @dricoco _gophercloud.AuthOptions_ is intended to support for v2 and v3 both https://github.com/gophercloud/gophercloud/blob/master/auth_options.go#L30
IMO no need to add this  _tokens3.AuthOptions_ seperately.

adisky:
@dricoco It mostly looks good, just left comment

adisky:
@dricoco please update the PR

dricoco:
Sorry will do this week !

pkg/csi/cinder/openstack/openstack.go
jim-bach: outputting the configuration/flags is pretty standard behavior for V(4) in Kubernetes.. fwiw
we talked about this in the OS Summit in Denver. The loadbalancers, csi, keystone, and kms we've all made basically use the same bit of code downstream already, might as well make it a importable utility/library instead.

pkg/csi/cinder/openstack/openstack.go
jim-bach: i think using the same bit of code found https://ghosthub.corp.blizzard.net/kubernetes/cloud-provider-openstack/blob/import/master/pkg/cloudprovider/providers/openstack/openstack.go#L413
until we can converge on a library might be the way to go. which it looks like it was OP's intent

dricoco: Sorry your link cannot be seen outside blizzard apparently, could you paste the lines ?

jim-bach: oh whoops, meant to link upstream 

> i think using the same bit of code found https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/cloudprovider/providers/openstack/openstack.go#L411
until we can converge on a library might be the way to go. which it looks like it was OP's intent

jim-bach:
I think the PR is great for short term and should get merged, LGTM.
longterm auth utility/library to use!

adisky:
@dricoco please address the inline comment and also resolve conflict

> gophercloud.AuthOptions is intended to support for v2 and v3 both https://github.com/gophercloud/gophercloud/blob/master/auth_options.go#L30
> IMO no need to add this tokens3.AuthOptions seperately.

dricoco:
Sorry got not enough time this week, but I'll address this comment quickly. This needs quite a lot of testing on my side prior to commit this modification. (And also it works very well as it is).

dricoco:
> @dricoco please address the inline comment and also resolve conflict
> 
> > gophercloud.AuthOptions is intended to support for v2 and v3 both https://github.com/gophercloud/gophercloud/blob/master/auth_options.go#L30
> > IMO no need to add this tokens3.AuthOptions seperately.

Hi @adisky,
I've been checking this issue a little bit more and I think this explicit V3 is needed because you need to make an explicit V3 authentification call to have it working. Even if authenticate support v2 and v3 we still need to make a v3 only call for that.

I've been running test on code relying only on authenticate and it's not working.

adisky:
@dricoco run hack/update-gofmt.sh to pass gofmt check

dricoco:
@adisky I just did, came back without any error.

dricoco:
looks there is some problem with the test :

                                "stdout": "+ go get -u github.com/tebeka/go2xunit\n+ go2xunit -input /home/zuul/workspace/test_results.txt -output /home/zuul/workspace/test_results.xml\nerror: can't open /home/zuul/workspace/test_results.txt for reading: open /home/zuul/workspace/test_results.txt: no such file or directory\n+ true",

dricoco:
there is also an error here
https://logs.openlabtesting.org/logs/82/582/520af892bf899749dbcb771df68b03afc7dfc494/cloud-provider-openstack-acceptance-test-e2e-conformance/cloud-provider-openstack-acceptance-test-e2e-conformance/2085358/logs/kubernetes/junit_runner.xml
that is not related to the code here

ramineni:
@dricoco failure is because of import order here https://github.com/kubernetes/cloud-provider-openstack/pull/582/files#diff-252011069a949a55c80b81de97af21f9R27 , you could run hack/update-gofmt.sh and also hack/verify-gofmt.sh , it will give solution

dricoco:
thans @ramineni I got it fixed

dricoco:
@adisky can we merge ?

ramineni:
looks ok to merge for now. Later can be refactored as necessary as suggested
/lgtm

PR #587, opened 2019-04-11
k8s-keystone-auth authorization policy definition version 2

### LINKED ISSUE #550: k8s-keystone-auth: Policy setting improvement
/kind feature

**What happened**:
The authorization policy setting is too complex and not flexible, especially for the fine-grained controls for the resource.

I'd propose the following setting format:
```JSON
[
  {
    "match": {
      "role": ["member"],
      "project": ["demo"]
    },
    "resource_permissions": [
      {"*.service": ["*"]},
      {"*.pod": ["get", "list", "watch"]},
      {"!kube-system.*": ["get", "list", "watch"]},
      {"*.*": ["get", "list", "watch"]},
      {"*.!secret": ["get", "list", "watch"]}
    ],
    "nonresource_permissions": [
      {"/healthz": ["get"]}
    ]
  },
  {...}
]
```

- The key in `resource_permissions` is in the format of `<namespace>.<resource>`
- `*` means all the namespaces or resources are allowed.
- `!` means all the namespaces or resources except the following one are allowed.
- The policy is whitelist and the config orders in `resource_permissions` matters.
- resource_permissions and nonresource_permissions can't be both used at the same time(same with the current behavior).

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

lingxiankong:
Improve k8s-keystone-auth authorization policy definition.

The current authorization policy setting is too complex and not flexible, especially for the fine-grained resource access control. With this PR, the cluster admin could define authorization policy as described in docs/using-keystone-webhook-authenticator-and-authorizer.md

The new version policy definition is still based on whitelist, which means the operation is allowed if any rule in the permissions is satisfied.

Version 1 definition is still supported and could be used together with version 2.

**Which issue this PR fixes**: fixes #550

You can use the image: docker.io/lingxiankong/k8s-keystone-auth:policy-version2 for testing.

**Release note**:
```release-note
k8s-keystone-auth service now supports two versions of authorization policy definition. Version 2 is recommended because of its flexibility and ongoing support, although both versions can be used together, version 1 may be considered deprecated in the future. Refer to [k8s-keystone-auth guide](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/using-keystone-webhook-authenticator-and-authorizer.md) for more information.
```

### DIFF
--- a/docs/using-keystone-webhook-authenticator-and-authorizer.md
+++ b/docs/using-keystone-webhook-authenticator-and-authorizer.md
@@ -1,13 +1,17 @@
 # k8s-keystone-auth
 
-Kubernetes webhook authentication and authorization for OpenStack Keystone.
+[Kubernetes webhook authentication and authorization](https://kubernetes.io/docs/reference/access-authn-authz/webhook/)
+for OpenStack Keystone. With k8s-keystone-auth, the Kubernetes cluster
+administrator only need to know the OpenStack project names or roles,
+it's up to the OpenStack project admin for user management, as a result,
+the OpenStack users could have access to the Kubernetes cluster.
 
-The k8s-keystone-auth can be running either as a static pod(controlled by
-kubelet) or a normal kubernetes service.
+The k8s-keystone-auth can be running either as a static pod(controlled
+by kubelet) or a normal kubernetes service.
 
 ## Prerequisites
 
-- You already have an available kubernetes cluster(version >= 1.9.3) and you
+- You already have a Kubernetes cluster(version >= 1.9.3) up and running and you
   have the admin permission for the cluster.
 - You have an OpenStack environment and admin credentials.
 
@@ -16,28 +20,35 @@ kubelet) or a normal kubernetes service.
 
 ## Running k8s-keystone-auth as a Kubernetes service
 
-### Prepare the authorization policy
-
-The authorization policy can be specified using an existing configmap name in
-the cluster, by doing this, the policy could be changed dynamically without the
-k8s-keystone-auth service restart. We need to create the configmap before
-running the k8s-keystone-auth service.
-
-Currently, k8s-keystone-auth service supports four types of policies:
-
-- user. The Keystone user ID or name.
-- project. The Keystone project ID or name.
-- role. The user role defined in Keystone.
-- group. The group is not a Keystone concept actually, it's supported for
-  backward compatibility, you can use group as project ID.
-
-For testing purpose, in the following configmap, we only allow the users in
-project `demo` with `k8s-viewer` role in OpenStack to query the pod information
-from all the namespaces. We create the configmap in `kube-system` namespace
-because we will also run k8s-keystone-auth service there.
+First, create a folder in which we will put all the manifest files and
+config files.
 
 ```shell
-cat <<EOF | kubectl apply -f -
+$ mkdir -p /etc/kubernetes/keystone-auth
+```
+
+### Prepare the authorization policy
+
+The authorization policy can be specified using an existing ConfigMap name in
+the cluster, by doing this, the policy could be changed dynamically without the
+k8s-keystone-auth service restart. The ConfigMap needs to be created before
+running the k8s-keystone-auth service.
+
+k8s-keystone-auth service supports two versions of policy definition.
+Version 2 is recommended because of its better flexibility. However,
+both versions are described in this guide. You can see more information
+of version 2 in `Authorization policy definition(version 2)` section
+below.
+
+For testing purpose, in the following ConfigMap, we only allow the users in
+project `demo` with `member` role in OpenStack to query the Pods information
+from all the namespaces. We create the ConfigMap in `kube-system` namespace
+because we will also run k8s-keystone-auth service there.
+
+Version 1:
+
+```shell
+$ cat <<EOF > /etc/kubernetes/keystone-auth/policy-config.yaml
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -56,7 +67,7 @@ data:
         "match": [
           {
             "type": "role",
-            "values": ["k8s-viewer"]
+            "values": ["memberr"]
           },
           {
             "type": "project",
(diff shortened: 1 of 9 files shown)


docs/using-keystone-webhook-authenticator-and-authorizer.md
```diff
@@ -16,28 +20,35 @@ kubelet) or a normal kubernetes service.
 
 ## Running k8s-keystone-auth as a Kubernetes service
 
+First, create a folder in which we will put all the manifest files and
```
dixudx: First, create a folder to store all the manifest files and config files.

docs/using-keystone-webhook-authenticator-and-authorizer.md
```diff
@@ -334,53 +410,111 @@ and both as well, depending on your requirement.
...
+    ```
+
+- Wait for the API server to restart successfully until you can see all the
+  pods are running in `kube-system` namespace.
+
+## Authorization policy definition(version 2)
+
+The version 2 definition could be used together with version 1 but will
+take precedence over version 1 if both are defined. The version 1
+definition is still supported but is considered deprecated.
```
dixudx: Version 2 is not mature yet. Better not deprecate version 1 in such a hurry.

docs/using-keystone-webhook-authenticator-and-authorizer.md
```diff
@@ -334,53 +410,111 @@ and both as well, depending on your requirement.
...
+        name: webhook
+    users:
+      - name: webhook
+    contexts:
+      - context:
+          cluster: webhook
+          user: webhook
+        name: webhook
+    current-context: webhook
+    ```
```
dixudx: Why remove `EOF` here? It is intended to be here.

docs/using-keystone-webhook-authenticator-and-authorizer.md
```diff
@@ -334,53 +410,111 @@ and both as well, depending on your requirement.
...
-
-- Wait for the API server to restart successfully until you can get all the
-  pods in `kube-system` namespace by running `kubectl get pod -n kube-system`
+- Create webhook config file. We reuse the folder `/etc/kubernetes/pki/`
+  because it's already mounted and accessible by API server pod.
+
+    ```shell
+    cat <<EOF > /etc/kubernetes/pki/webhookconfig.yaml
+    ---
+    apiVersion: v1
```
dixudx: Did anything get changed? Format?
Why is Git telling us these lines modified?
Same concern on above modified lines.

docs/using-keystone-webhook-authenticator-and-authorizer.md
```diff
@@ -334,53 +410,111 @@ and both as well, depending on your requirement.
...
+  non-resource endpoint such as `/healthz`, the value defines the
+  allowed operations.
+
+Some examples:
+
+- Any operation is allowed on any resource in any namespace.
+
+    ```json
+    "resource_permissions": {
+      "*.*": ["*"]
```
dixudx: So `*.*` here is mainly used to set rules on operating resources of a certain namespace?
We could use other separators, e.g. `/`, instead of  `.`, which is misleading and has special meanings in regex.

dixudx: Actually I'd prefer to have `resources` and `namespaces` set separately.

e.g.
```json
"resource_permissions": [
  {
    "namespaces": "foo*",
    "resources": "bar,baz",
    "verbs": "get,list,delete"
  }
]
```

docs/using-keystone-webhook-authenticator-and-authorizer.md
```diff
@@ -334,53 +410,111 @@ and both as well, depending on your requirement.
...
+    ```shell
+    sed -i '/image:/ i \ \ \ \ - --authentication-token-webhook-config-file=/etc/kubernetes/pki/webhookconfig.yaml' /etc/kubernetes/manifests/kube-apiserver.yaml
+    sed -i '/image:/ i \ \ \ \ - --authorization-webhook-config-file=/etc/kubernetes/pki/webhookconfig.yaml' /etc/kubernetes/manifests/kube-apiserver.yaml
+    sed -i "/authorization-mode/c \ \ \ \ - --authorization-mode=Node,Webhook,RBAC" /etc/kubernetes/manifests/kube-apiserver.yaml
+    ```
+
+- Wait for the API server to restart successfully until you can see all the
+  pods are running in `kube-system` namespace.
+
+## Authorization policy definition(version 2)
```
dixudx: We should have discussions on version2. Too many designs and details should be talked.

docs/using-keystone-webhook-authenticator-and-authorizer.md
lingxiankong: Why? I don't see there is much difference.

docs/using-keystone-webhook-authenticator-and-authorizer.md
lingxiankong: Will change to `The version 1 definition is still supported but may be considered deprecated in the future.`

docs/using-keystone-webhook-authenticator-and-authorizer.md
lingxiankong: Yeah, the EOF is missing, need to add back

docs/using-keystone-webhook-authenticator-and-authorizer.md
lingxiankong: I intentionally avoided using the keys such as namespaces, resources and verbs, because in this way, there is not much difference with version one which is tedious and redundant.

lingxiankong: > So `*.*` here is mainly used to set rules on operating resources of a certain namespace?
> We could use other separators, e.g. `/`, instead of `.`, which is misleading and has special meanings in regex.

Make sense, will consider.

chelios:
I like the idea behind this. I think this is a much cleaner, more legible approach to managing resource access than the version 1 approach.

Something that is easier to read is easier to understand, troubleshoot and share.

adisky:
@dixudx all your discussions resolved?? @lingxiankong This is big change, could you please add release notes??
cc @hogepodge @dims

dims:
+1 from me. please add release note as mentioned before.

/approve

lingxiankong:
@adisky @dims release note added.

PR #422, opened 2019-01-09
Add annotations for setting listener timeouts

emonty:
**What this PR does / why we need it**:

The octavia loadbalancer supports setting timeouts for client and member
data connections. For some operations, such as pushing git large git
repositories over an ssh connection, it is very important to be able to
increase the timeout.

This adds two annotations that allows a user to create a load balancer
and set its timeout_client_data and timeout_member_data fields.

**Special notes for your reviewer**:

Depends-On: https://github.com/gophercloud/gophercloud/pull/1394

PR #589, opened 2019-04-11
Support listener timeouts as service annotations

### LINKED ISSUE #588: Support listener timeouts as service annotations
**Is this a BUG REPORT or FEATURE REQUEST?**: Feature request

>
> /kind bug
/kind feature

**What happened**:

Currently, we are not able to specify any timeout values when creating Openstack load balancers despite the options being [available in the API.](https://developer.openstack.org/api-ref/load-balancer/v2/index.html?expanded=create-listener-detail#id34)

Kubernetes service annotations expressed as metadata determine listener timeout values when cloud-provider-openstack creates new LoadBalancers

N/A

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

itsc0lin:
**What this PR does / why we need it**: Extends support for listener [timeout values](https://developer.openstack.org/api-ref/load-balancer/v2/index.html?expanded=create-listener-detail#id34) to Kubernetes services via annotation

**Which issue this PR fixes**: fixes #588 by allowing the default listener timeout values of the loadbalancer Octavia creates for each Kubernetes LoadBalancer service to be individually overridden. For longer-living connections, unexpected disconnects/RST can be observed when a successfully negotiated connection reaches the [default timeout](https://github.com/kubernetes/cloud-provider-openstack/issues/588#issuecomment-482819223) values without activity. For example, if the `kube-apiserver` is placed behind an Octavia loadbalancer with the default listener timeouts and `kubectl logs -f <pod>` is used, after 50 seconds the client receives a connection reset and logs `error: unexpected EOF`. Tuning these values allows for LoadBalancer services to be configured to accommodate longer-lived connections.

**Special notes for your reviewer**: the additional createOpts for the listener are supported in gophercloud by virtue of this change https://github.com/gophercloud/gophercloud/pull/1544

**Release note**:

```Adds support for setting timeout_client_data, timeout_member_connect, timeout_member_data, and timeout_tcp_inspect values on load balancers```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -231,7 +231,7 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:ba2ade5da932202872cfbe51377e7b108f4e82f981a15c1f07045678a4818464"
+  digest = "1:c6ae6215aaacaa06d48c8cd231a6c28396b8f3a1efea1ebb0b3563062e4ef26a"
   name = "github.com/gophercloud/gophercloud"
   packages = [
     ".",
@@ -277,7 +277,7 @@
     "testhelper/client",
   ]
   pruneopts = "UT"
-  revision = "a127a1b294c369554364af633d29d213ba88ef16"
+  revision = "2c55d17f707cc8333ca4f49690cb2970d12a25f6"
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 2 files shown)


lingxiankong:
Hi, before reviewing this PR, i need more information of the original feature request, already left a comment in the issue. 

just FYI, there is already a PR doing the same thing, https://github.com/kubernetes/cloud-provider-openstack/pull/422.

itsc0lin:
I've elaborated on the linked issue in a way that I hope will explain the intent of the feature. I took a look at #422 and see that it does include `TimeoutClientData` and `TimeoutMemberData` in the updateOpts, but notice that it does not address the absence of the other two timeout values (`timeout_member_connect` and `timeout_tcp_inspect`) that this change includes.

hogepodge:
/lgtm

I'm satisfied with the reasons outlined in this comment. https://github.com/kubernetes/cloud-provider-openstack/issues/588#issuecomment-483434994

PR #594, opened 2019-04-18
Add Max Vol Limit for Cinder CSI

### LINKED ISSUE #572: Add MaxVolumesPerNode limit to node info in Cinder CSI
**What happened**:
`NodeGetInfo` in pkg/csi/cinder/nodeserver.go should return also `MaxVolumesPerNode`. As openstack has default limit of 26 Cinder volumes per VM. 

There should be a way to set `MaxVolumesPerNode` value and it should default to 26 if nothing is provided. Other safe value for limit would be 25 so it would take a count the situation when you are using Cinder volume as root volume.

Way to provider configurable value for `MaxVolumesPerNode` could be via openstack cloud config file. i.e
```
type Config struct {
 ...
  BlockStorage struct {
    MaxVolumesPerNode int64 `gcfg:"max-volumes-per-node"`
  }
}
```
Then we would need to introduce global variable (similar to `OsInstance` and `configFile`) for `maxVolumes` as it would be static after initialization. Also function is needed so that we can access it from nodeserver.go.
```
var maxVolumes int64 = 26

func GetMaxVolumes() int64 {
	return maxVolumes
}
```
We would just need to set the defined value to the global variable. If `cfg.BlockStorage.MaxVolumesPerNode` is defined. (int64 should default to 0 so we ignore it and also all non positive values)
```
func CreateOpenStackProvider() (IOpenStack, error) {
...
  cfg, epOpts, err := GetConfigFromFile(configFile)
  if err == nil {
    ...
    if cfg.BlockStorage.MaxVolumesPerNode > 0 {
       maxVolumes = cfg.BlockStorage.MaxVolumesPerNode
    }
...
```
Then last in nodeserver.go we could just call `GetMaxVolumes` 
```
func (ns *nodeServer) NodeGetInfo(ctx context.Context, ...
...
return &csi.NodeGetInfoResponse{
  NodeId:             nodeID,
  AccessibleTopology: topology,
  MaxVolumesPerNode:  openstack.GetMaxVolumes(),
}, nil
```

**Environment**:
- Cinder CSI

/kind feature

adisky:
This PR adds support for returning max volume limit in _CSI NodeGetInfo_ call, It returns MaxVolumesPerNode in CSINodeGetInfo response, which defaults to 256(same as intree driver), This can be configured via _node-volume-attach-limit_ from cloud config file.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #572

**Release note**:

```release-note
```

### DIFF
--- a/pkg/csi/cinder/openstack/openstack.go
+++ b/pkg/csi/cinder/openstack/openstack.go
@@ -53,6 +53,10 @@ type OpenStack struct {
 	blockstorage *gophercloud.ServiceClient
 }
 
+type BlockStorageOpts struct {
+	NodeVolumeAttachLimit int64 `gcfg:"node-volume-attach-limit"`
+}
+
 type Config struct {
 	Global struct {
 		AuthUrl    string `gcfg:"auth-url"`
@@ -66,6 +70,7 @@ type Config struct {
 		Region     string
 		CAFile     string `gcfg:"ca-file"`
 	}
+	BlockStorage BlockStorageOpts
 }
 
 func (cfg Config) toAuthOptions() gophercloud.AuthOptions {
@@ -125,11 +130,14 @@ func GetConfigFromEnv() (gophercloud.AuthOptions, gophercloud.EndpointOpts, erro
 	return authOpts, epOpts, nil
 }
 
+const maxVol int64 = 256
+
 var OsInstance IOpenStack = nil
 var configFile = "/etc/cloud.conf"
+var cfg Config
 
-func InitOpenStackProvider(cfg string) {
-	configFile = cfg
+func InitOpenStackProvider(cfgFile string) {
+	configFile = cfgFile
 	klog.V(2).Infof("InitOpenStackProvider configFile: %s", configFile)
 }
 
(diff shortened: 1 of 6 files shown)


pkg/csi/cinder/openstack/openstack.go
```diff
@@ -206,3 +214,13 @@ func GetOpenStackProvider() (IOpenStack, error) {
 
 	return OsInstance, nil
 }
+
+//GetMaxVolLimit returns max vol limit
+func GetMaxVolLimit() int64 {
+	if cfg.BlockStorage.NodeVolumeAttachLimit > 0 {
```
ramineni: do we need to check here if its greater than 26 ?
And also , is this configurable on cinder side also? if yes, then we need to get exact configured value for that node from cinder?

adisky: @ramineni yes, will update the PR for greater than 26,
It is not currently configurable, work in progress in openstack

lingxiankong:
could you please rebase?

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -206,3 +214,13 @@ func GetOpenStackProvider() (IOpenStack, error) {
 
 	return OsInstance, nil
 }
+
+//GetMaxVolLimit returns max vol limit
+func GetMaxVolLimit() int64 {
+	if cfg.BlockStorage.NodeVolumeAttachLimit > 0 && cfg.BlockStorage.NodeVolumeAttachLimit <= 256 {
```
ramineni: doc update for this new parameter seems to be missing?

jichenjc: yes, this need some doc to indicate the limit..

jichenjc: others look good

adisky:
@jichenjc @ramineni Doc is added, PTAL

ramineni:
Thanks. Looks good

PR #595, opened 2019-04-22
Get real device path from node

### LINKED ISSUE #531: Cinder CSI should not trust openstack device path
**Is this a BUG REPORT**:
/kind bug

**What happened**:
Cinder CSI do not validate path from the node, but trusts the path given by Openstack. This causes that non trivial environment will eventually fail to mount volumes as device path on node and one given by Openstack are out of sync. 

This causes either that volume can not be mounted or wrong volume is mounted to the pod.

Cinder CSI (nodeserver) should check the device path from node and not trust the path given by Openstack. This is implemented correctly in CCM openstack volume implementation. 

Attach and detach multiple volumes simultaneously and eventually Openstack and Nodes will get out of sync. 
Just creating multiple pods with volumes and then deleting them might not trigger this behavior as usually OS will allocated device path with same as given by OS. 

Simple solution is just copy implementation from CCM. There get GetDevicePath function that is the one what should be called when trying to determine the path at node.

Better solution would be to merge these two openstack API implementations to one. 
Basically as CCM do not use volume implementation at all it could be quite heavily modified during merge as it will not break anything. I.e we could just remove support for cinder V1 and V2 apis as Cinder CSI do not support them now and they are any case deprecated. This would help that there would not be need for wrapper functions that duplicate code, as every api version needs own functions. Also we would need to copy snapshot implementation from CSI to CCM. 

This merge we would eliminate the issue that there are multiple implementations of same functions.

**Effects**:
All Cinder CSI versions.

ramineni:
This PR is to get the real device path from the node and not
to trust the device path returned by cinder which often leads to failed mount

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #531 

**Release note**:

```release-note
```

### DIFF
--- a/pkg/csi/cinder/mount/mount.go
+++ b/pkg/csi/cinder/mount/mount.go
@@ -20,9 +20,11 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path"
 	"strings"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/kubernetes/pkg/util/mount"
 	utilexec "k8s.io/utils/exec"
 
@@ -30,13 +32,17 @@ import (
 )
 
 const (
-	probeVolumeDuration = 1 * time.Second
-	probeVolumeTimeout  = 60 * time.Second
-	instanceIDFile      = "/var/lib/cloud/data/instance-id"
+	probeVolumeDuration      = 1 * time.Second
+	probeVolumeTimeout       = 60 * time.Second
+	operationFinishInitDelay = 1 * time.Second
+	operationFinishFactor    = 1.1
+	operationFinishSteps     = 15
+	instanceIDFile           = "/var/lib/cloud/data/instance-id"
 )
 
 type IMount interface {
 	ScanForAttach(devicePath string) error
+	GetDevicePath(volumeID string) (string, error)
 	IsLikelyNotMountPointAttach(targetpath string) (bool, error)
 	FormatAndMount(source string, target string, fstype string, options []string) error
 	IsLikelyNotMountPointDetach(targetpath string) (bool, error)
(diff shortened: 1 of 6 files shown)


adisky:
@ramineni block volume is depnedent on it?? this has to be merged before block volume support?

ramineni:
> @ramineni block volume is dependent on it?? this has to be merged before block volume support?

no, its not related to block volume , it fails when cinder doesnt give exact mounted path

pkg/csi/cinder/mount/mount.go
```diff
@@ -81,6 +87,63 @@ func probeVolume() error {
...
+		for _, c := range candidateDeviceNodes {
+			if c == f.Name() {
+				klog.V(4).Infof("Found disk attached as %q; full devicepath: %s\n",
+					f.Name(), path.Join("/dev/disk/by-id/", f.Name()))
+				return path.Join("/dev/disk/by-id/", f.Name())
+			}
+		}
+	}
+
+	klog.V(4).Infof("Failed to find device for the volumeID: %q by serial ID", volumeID)
```
lingxiankong: `Failed to find device for the volumeID: %q` makes more sense.

pkg/util/metadata/metadata.go
```diff
@@ -165,6 +165,49 @@ func GetFromMetadataService(metadataVersion string) (*Metadata, error) {
...
+	// Nova Hyper-V hosts cannot override disk SCSI IDs. In order to locate
+	// volumes, we're querying the metadata service. Note that the Hyper-V
+	// driver will include device metadata for untagged volumes as well.
+	//
+	// We're avoiding using cached metadata (or the configdrive),
+	// relying on the metadata service.
+	instanceMetadata, err := GetFromMetadataService(defaultMetadataVersion)
+
+	if err != nil {
+		klog.V(4).Infof("Could not retrieve instance metadata. Error: %v", err)
```
lingxiankong: Please use WARNING or ERROR log level

pkg/util/metadata/metadata.go
```diff
@@ -165,6 +165,49 @@ func GetFromMetadataService(metadataVersion string) (*Metadata, error) {
...
+					"could not retrieve disk path for volumeID: %q. Error filepath.Glob(%q): %v",
+					volumeID, diskPattern, err)
+				return ""
+			}
+
+			if len(diskPaths) == 1 {
+				return diskPaths[0]
+			}
+
+			klog.V(4).Infof("expecting to find one disk path for volumeID %q, found %d: %v",
```
lingxiankong: Please use WARNING log level

pkg/util/metadata/metadata.go
```diff
@@ -165,6 +165,49 @@ func GetFromMetadataService(metadataVersion string) (*Metadata, error) {
...
+				volumeID, len(diskPaths), diskPaths)
+
+		}
+	}
+
+	klog.V(4).Infof("Could not retrieve device metadata for volumeID: %q", volumeID)
```
lingxiankong: Please use WARNING or ERROR log level

pkg/csi/cinder/mount/mount.go
```diff
@@ -81,6 +87,63 @@ func probeVolume() error {
 	return nil
 }
 
+// GetDevicePath returns the path of an attached block storage volume, specified by its id.
```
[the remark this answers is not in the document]
ramineni: @lingxiankong ok anyway getDevicePathBySerialID is internal function so doc string is not required. I will remove the same to avoid confusion.

pkg/util/metadata/metadata.go
lingxiankong: I don't think so, what if the metadata service is not accessible for some reason?

All my concerns about the log level come from the perspective of the cluster operator that if there is some error that will eventually lead to a user-facing failure, we should print the log in non-INFO level.

PR #596, opened 2019-04-22
Consolidate CSI Manifest

adisky:
Consolidate CSI yamls to node and controller plugin. Currently we have seperate yamls for each sidecar container and we need to deploy cinder plugin with each sidecar, so we have 3-4 instances of cinder plugin running, it is difficult to maintain and debug. For this reason consolidating csi manifests to node plugin and controller plugin  so that only two instances of cinder plugin run.

**Release note**:

```release-note
```
`NONE`


ramineni:
/lgtm
Thanks.

PR #617, opened 2019-05-17
csi: Support xfs filesystem for csi-node-provisioner

chrischdi:

This PR adds support for the csi-node-plugin to format disks using xfs.
When building the csi-node-plugin using `make image-csi-plugin` there will currently be no `xfsprogs` installed.

**Release note**:

```release-note
Added support for xfs to csi-node-provisioner
```

### DIFF
--- a/cluster/images/cinder-csi-plugin/Dockerfile
+++ b/cluster/images/cinder-csi-plugin/Dockerfile
@@ -3,7 +3,7 @@ LABEL maintainers="Kubernetes Authors"
 LABEL description="Cinder CSI Plugin"
 
 # Install e4fsprogs for format
-RUN apk add --no-cache ca-certificates e2fsprogs eudev
+RUN apk add --no-cache ca-certificates e2fsprogs eudev xfsprogs
 
 ADD cinder-csi-plugin /bin/
 


adisky:
/hold I am not sure to add this in default docker image, if there are specific requirements a separate image can be built.
cc @hogepodge @lingxiankong @ramineni

chrischdi:
I did a rebase because of a merge conflict.

@adisky just for context :-) :
In our kubernetes clusters a user could define a storageclass for ext4 (which works in the current image) but also define xfs (which won't work in the current image).

ramineni:
> /hold I am not sure to add this in default docker image, if there are specific requirements a separate image can be built.
> cc @hogepodge @lingxiankong @ramineni

@adisky As there is option to specify fstype in storageclass and k8s supports xfs , it would be ok to add the same in image to avoid failures ? wdyt?

adisky:
@ramineni @chrischdi  k8s supports other fsType also should we add all?https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#storageclass-v1-storage-k8s-io

ramineni:
> @ramineni @chrischdi k8s supports other fsType also should we add all?https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#storageclass-v1-storage-k8s-io

only ntfs is remaining right? we could add if requires diff binary and is used in production

adisky:
@ramineni @chrischdi IMO It is not like that we are not supporting xfs, k8s does support xfs its just we are not building it into default image, these images are for reference and CI, people can customize there images for there own requirment.

chrischdi:
I think xfs is very common to use. We did start to use the default image, thinking everything should work with that one :-)

If you like to I could add the `ntfs-3g` package too, to fullfill the examples of the documentation.

I would be also ok with the statement to build our own images. 
I think adding `xfs` (and maybe `ntfs-3g`) would fullfill the most use cases for the node-provisioner and most users will use the same image :-)

ramineni:
> I think xfs is very common to use. We did start to use the default image, thinking everything should work with that one :-)
> 
> If you like to I could add the `ntfs-3g` package too, to fullfill the examples of the documentation.
> 
> I would be also ok with the statement to build our own images.
> I think adding `xfs` (and maybe `ntfs-3g`) would fullfill the most use cases for the node-provisioner and most users will use the same image :-)

I agree , if its commonly used by all , we should add in default image.

adisky:
@chrischdi @ramineni ok, considering it will be commonly used by all :)
/lgtm

PR #282, opened 2018-09-06
Get a token using Keystone Application Credential

zioproto:
**What this PR does / why we need it**:

This PR makes possibile to login using Keystone Application Credentials.
fixes #232 

**Special notes for your reviewer**:

This PR requires this gophercloud PR to be merged first 
https://github.com/gophercloud/gophercloud/pull/1224

PR #618, opened 2019-05-20
Improve Keystone Application Credential docs

zioproto:

It adds documentation for the changes pushed in #282

### DIFF
--- a/docs/using-client-keystone-auth.md
+++ b/docs/using-client-keystone-auth.md
@@ -108,6 +108,11 @@ arguments
 (`--keystone-url`, `--domain-name`, `--user-name`, `--password` and `--project-name`), respectively.
 If they are not specified, the user will be prompted to enter them at the time of the interactive
 session.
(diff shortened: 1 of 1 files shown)


docs/using-client-keystone-auth.md
```diff
@@ -108,6 +108,11 @@ arguments
 (`--keystone-url`, `--domain-name`, `--user-name`, `--password` and `--project-name`), respectively.
 If they are not specified, the user will be prompted to enter them at the time of the interactive
 session.
+It is also possible to authenticate with [Keystone Application Credentials](https://docs.openstack.org/keystone/queens/user/application_credentials.html).
```
jichenjc: suggest to use https://docs.openstack.org/keystone/latest/user/application_credentials.html
so we always recommend latest

zioproto: You are right ! Fixed

PR #619, opened 2019-05-21
Added configurable Loadbalancer classes

afritzler:
If you have multiple `FloatingIPPools` and/or `FloatingIPSubnets` it might be desirable to offer the users logical meanings for LoadBalancers like `internetFacing` or `DMZ` instead of requiring the user to select dedicated network or subnet ids at the service object as an annotation.

This PR allows first to configure a set of `LoadBalancerClasses` on cloud provider level containing FloatingIPPoolId, FloatingSubnetID and/or provisioning subnet ID. At the service level a new annotation has been added to select a dedicated class by name.

CloudProvider Configuration Example:
```
    [Global]
    auth-url="https://someurl"
    domain-name="mydomain"
    tenant-name="mytenant"
    username="myuser"
    password="mypassword"
    [LoadBalancer]
    lb-version=v2
    lb-provider="f5networks"
    floating-network-id="a57af0a0-da92-49be-a98a-345ceca004b3"
    floating-subnet-id="a02eb6c3-fc69-46ae-a3fe-fb43c1563cbc"
    subnet-id="fa6a4e6c-6ae4-4dde-ae86-3e2f452c1f03"
    create-monitor=true
    monitor-delay=60s
    monitor-timeout=30s
    monitor-max-retries=5
    [LoadBalancerClass "internetFacing"]
    floating-network-id="c57af0a0-da92-49be-a98a-345ceca004b3"
    floating-subnet-id="f90d2440-d3c6-417a-a696-04e55eeb9279"
    [LoadBalancerClass "dmz"]
    floating-subnet-id="a374bed4-e920-4c40-b646-2d8927f7f67b"
    [LoadBalancerClass "office"]
    floating-subnet-id="b374bed4-e920-4c40-b646-2d8927f7f67b"
```

Service example
```
apiVersion: v1
kind: Service
metadata:
  annotations:
    loadbalancer.openstack.org/class: internetFacing
  name: nginx-internet
spec:
  ports:
  - port: 80
    targetPort: 80
  selector:
    app: nginx
  type: LoadBalancer
```

**Release note**:
```release-note
Added support for LoadBalancer Classes when using multiple FloatingPool subnets and/or FloatingPool networks
```


afritzler:
Looks like 2 tests failed for various other reasons here?!?

jim-bach:
i like this idea, +1 when its ready

afritzler:
Can somebody help me with the e2e conformance tests issues here? I don't see any load balancer related issues in the logs. Is the e2e conformance suite currently functional?

lingxiankong:
Hi @afritzler, our CI is encountering some issues at the moment, the openlab team is working on that.

afritzler:
Thanks @lingxiankong for the info!

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -43,7 +43,7 @@ import (
 	"github.com/mitchellh/mapstructure"
 	gcfg "gopkg.in/gcfg.v1"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
gonzolino: Whats the reason for this change?

pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
```diff
@@ -863,7 +868,29 @@ func (lbaas *LbaasV2) EnsureLoadBalancer(ctx context.Context, clusterName string
...
+	internalAnnotation := false
+	var lbClass *LBClass
+	var err error
+	class := getStringFromServiceAnnotation(apiService, ServiceAnnotationLoadBalancerClass, "")
+	if class != "" {
+		lbClass = lbaas.opts.LBClasses[class]
+		if lbClass == nil {
+			return nil, fmt.Errorf("invalid loadbalancer class %q", class)
+		}
+		internalAnnotation = lbClass.SubnetID != ""
```
gonzolino: If I understand it correctly, the global option `lbaas.opts.InternalLB` will be ignored when a lb class is set.
Instead the controller will create a internal lb when a class with a defined subnet ID is set.
Not sure if this is the behaviour that you desire here, but it sounds a bit unintuitive to me...

Also, when a class sets a subnet ID and a floating network ID, wouldn't that prevent the creation of the floating ip with your code?

pkg/cloudprovider/providers/openstack/openstack.go
afritzler: Good point!

pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
afritzler: When an LBclass is requested via service annotation but can't be found in the controller configuration, an error will be thrown.

pkg/cloudprovider/providers/openstack/openstack.go
afritzler: Apparently this a new convention. The linter is doing that automatically.

pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
afritzler: Ok, we addressed this issue. Throwing an error now if the LBclass is misconfigured.

pkg/cloudprovider/providers/openstack/openstack.go
gonzolino: IMHO it would make more sense to name it corev1 then.

afritzler: Alright, using now `corev1`.

afritzler:
Alright, added documentation on how to use this feature. I guess once this is rolled out, we also need to adjust official docs here https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/

PR #622, opened 2019-05-24
Add magnum-auto-healer for the Kubernetes cluster

lingxiankong:
Add k8s-auto-healer for the Kubernetes cluster running on top of cloud environment, OpenStack is the default cloud provider implementation.

k8s-auto-healer ensures that the running Kubernetes nodes are healthy by monitoring the nodes’ status periodically, searching for unhealthy instances and triggering replacements when needed, maximizing your cluster’s efficiency and performance.

The cluster provider and health check mechanism are both pluggable and easy to integrate with other clouds and customized health check approaches.

For the openstack cloud provider:
- Only the Magnum cluster is supported, doesn't support the Kubernetes cluster created in other ways(kops, cluster-api, kubeadm, etc.)
- Auto-healing is achieved via Heat stack update, which replaces the unhealthy resources with new ones.

**Which issue this PR fixes**: fixes #

**Release note**:
```release-note
```

### DIFF
--- a/docs/using-magnum-auto-healer.md
+++ b/docs/using-magnum-auto-healer.md
@@ -1,0 +1,227 @@
+# Using magnum-auto-healer
+
+## What is magnum-auto-healer
+
+Kubernetes is self-healing container orchestration platform that will detect failures from your pods and redeploy that workload, however, magnum-auto-healer is a self-healing cluster management service that will automatically recover a failed master or worker node within your Magnum cluster. Basically, magnum-auto-healer ensures the running Kubernetes nodes are healthy by monitoring the nodes' status periodically, searching for unhealthy instances and triggering replacements when needed, maximizing your cluster's high availability and reliability, and protecting your application from downtime when the node it's running on fails.
+
+The other side of concerns for Kubernetes cluster is scalability. Kubernetes [cluster-autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) can scale the worker pools in your cluster automatically to increase or decrease the number of worker nodes based on the sizing needs of the scheduled workloads. cluster-autoscaler periodically scans the cluster to adjust the number of worker nodes in response to your workload resource requests and any custom settings that you configure, such as scanning intervals. The main purpose of cluster-autoscaler is autoscaling, not autohealing. cluster-autoscaler can be deployed together with magnum-auto-healer.
+
+Like cluster-autoscaler, magnum-auto-healer is implemented to use together with cloud providers as well, [OpenStack Magnum](https://docs.openstack.org/magnum/latest/user/) is supported by default.
+
+## magnum-auto-healer Design
+
+There are some considerations when we were designing the magnum-auto-healer service:
+
+- We want to have a single component for the cluster autohealing purpose. There are already some other components out there in the community to deal with some specific tasks separately, combining them together with some customization may work, but will lead to much complexity and maintenance overhead.
+- Support both master nodes and worker nodes.
+- Cluster administrator is able to disable the autohealing feature on the fly, which is very important for the cluster operations like upgrade or scheduled maintenance.
+- The Kubernetes cluster is not necessary to be exposed to either the public or the OpenStack control plane. For example, In Magnum, the end user may create a private cluster which is not accessible even from Magnum control services.
+- The health check should be pluggable. Deployers should be able to write their own health check plugin with customized health check parameters.
+- Support different cloud providers.
+
+## Deploying and testing magnum-auto-healer
+
+### Prerequisites
+
+1. A multi-node cluster(3 masters and 3 workers) is created in Magnum.
+
+    ```
+    $ openstack coe cluster list
+    +--------------------------------------+-----------------------------+-----------------+------------+--------------+-----------------+
+    | uuid                                 | name                        | keypair         | node_count | master_count | status          |
+    +--------------------------------------+-----------------------------+-----------------+------------+--------------+-----------------+
+    | c418c335-0e52-42fc-bd68-baa8d264e072 | lingxian_por_test_1.12.7_ha | lingxian_laptop |          3 |            3 | CREATE_COMPLETE |
+    +--------------------------------------+-----------------------------+-----------------+------------+--------------+-----------------+
+    $ openstack server list --name lingxian-por-test-1-12-7-ha
+    +--------------------------------------+---------------------------------------------------+--------+-----------------------------------------+-------------------------+---------+
+    | ID                                   | Name                                              | Status | Networks                                | Image                   | Flavor  |
+    +--------------------------------------+---------------------------------------------------+--------+-----------------------------------------+-------------------------+---------+
+    | 908957c2-ac88-4b54-a1fc-91f9cc8f98f1 | lingxian-por-test-1-12-7-ha-bbgjts5g4xhb-minion-2 | ACTIVE | lingxian_net=10.0.10.33, 150.242.42.234 | fedora-atomic-27-x86_64 | c1.c4r8 |
+    | 8f0c3ad9-caf5-45b6-bf3a-97b3bb6de623 | lingxian-por-test-1-12-7-ha-bbgjts5g4xhb-minion-0 | ACTIVE | lingxian_net=10.0.10.32, 150.242.42.233 | fedora-atomic-27-x86_64 | c1.c4r8 |
+    | a6ae4cee-7cf2-4b25-89bc-a5c6cb2c364d | lingxian-por-test-1-12-7-ha-bbgjts5g4xhb-minion-1 | ACTIVE | lingxian_net=10.0.10.34, 150.242.42.245 | fedora-atomic-27-x86_64 | c1.c4r8 |
+    | 2af96203-cc6f-4b55-8fb2-062340207ebb | lingxian-por-test-1-12-7-ha-bbgjts5g4xhb-master-2 | ACTIVE | lingxian_net=10.0.10.31, 150.242.42.226 | fedora-atomic-27-x86_64 | c1.c2r4 |
+    | 10bef366-b5a8-4400-b2c3-82188ec06b13 | lingxian-por-test-1-12-7-ha-bbgjts5g4xhb-master-1 | ACTIVE | lingxian_net=10.0.10.30, 150.242.42.22  | fedora-atomic-27-x86_64 | c1.c2r4 |
+    | 9c17f034-6825-4e49-b3cb-0ecddd1a8dd8 | lingxian-por-test-1-12-7-ha-bbgjts5g4xhb-master-0 | ACTIVE | lingxian_net=10.0.10.29, 150.242.42.213 | fedora-atomic-27-x86_64 | c1.c2r4 |
+    +--------------------------------------+---------------------------------------------------+--------+-----------------------------------------+-------------------------+---------+
+    ```
+
+2. The kubeconfig file of the cluster is in place.
+
+### Deploy magnum-auto-healer
+
+It's recommended to run magnum-auto-healer service as a DaemonSet on the master nodes, the service is running in active-passive mode using leader election mechanism. There is a sample manifest file in `manifests/magnum-auto-healer/magnum-auto-healer.yaml`, you need to change some variables as needed before actually running `kubectl apply` command. The following commands are just examples:
+
+```shell
+magnum_cluster_uuid=c418c335-0e52-42fc-bd68-baa8d264e072
+keystone_auth_url=https://api.nz-por-1.catalystcloud.io:5000/v3
+user_id=ceb61464a3d341ebabdf97d1d4b97099
+user_project_id=b23a5e41d1af4c20974bf58b4dff8e5a
+password=password
+region=RegionOne
+image=lingxiankong/magnum-auto-healer:0.1.0
+
+cat <<EOF | kubectl apply -f -
+---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  name: magnum-auto-healer
+  namespace: kube-system
+
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: magnum-auto-healer
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
+subjects:
+  - kind: ServiceAccount
+    name: magnum-auto-healer
+    namespace: kube-system
+
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: magnum-auto-healer-config
+  namespace: kube-system
+data:
+  config.yaml: |
+    cluster-name: ${magnum_cluster_uuid}
+    dry-run: false
+    monitor-interval: 15s
+    check-delay-after-add: 20m
+    leader-elect: true
+    healthcheck:
+      master:
+        - type: Endpoint
+          params:
+            unhealthyDuration: 30s
+            protocol: HTTPS
+            port: 6443
+            endpoints: ["/healthz"]
+            okCodes: [200]
+        - type: NodeCondition
+          params:
+            unhealthyDuration: 1m
+            types: ["Ready"]
+            okValues: ["True"]
+      worker:
+        - type: NodeCondition
+          params:
+            unhealthyDuration: 1m
+            types: ["Ready"]
+            okValues: ["True"]
+    openstack:
+      auth-url: ${keystone_auth_url}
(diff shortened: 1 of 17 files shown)


lingxiankong:
cc @adisky @dims @hogepodge

hogepodge:
One thing needs to be added, and we also need to discuss one major issue.

- Major new features need documentation. The patch can't be considered complete until there is a corresponding document that describes how to use it.

- It appears that the patch depends upon Magnum, which makes it not widely usable. We need to think about how we organize provider code that presumes the existence of clusters managed by Magnum, to make it clear that the solution is not a general one. Ideally we would want to have a more general solution, but that may not be possible given the last of desired cluster state without a management layer like Magnum. If we can think of general structures that would allow us to abstract the ideal cluster state, it would be preferable.

Docs is a solid requirement we need meet before merging. The Magnum requirement just needs to be documented, and we might want to think about a different location in the repository for now.

/hold

/assign @chrigl

adisky:
> One thing needs to be added, and we also need to discuss one major issue.
> 
> * Major new features need documentation. The patch can't be considered complete until there is a corresponding document that describes how to use it.
> * It appears that the patch depends upon Magnum, which makes it not widely usable. We need to think about how we organize provider code that presumes the existence of clusters managed by Magnum, to make it clear that the solution is not a general one. Ideally we would want to have a more general solution, but that may not be possible given the last of desired cluster state without a management layer like Magnum. If we can think of general structures that would allow us to abstract the ideal cluster state, it would be preferable.
> 
> Docs is a solid requirement we need meet before merging. The Magnum requirement just needs to be documented, and we might want to think about a different location in the repository for now.
> 
> /hold
> 
> /assign @chrigl

+1

lingxiankong:
> One thing needs to be added, and we also need to discuss one major issue.
> 
> * Major new features need documentation. The patch can't be considered complete until there is a corresponding document that describes how to use it.
> * It appears that the patch depends upon Magnum, which makes it not widely usable. We need to think about how we organize provider code that presumes the existence of clusters managed by Magnum, to make it clear that the solution is not a general one. Ideally we would want to have a more general solution, but that may not be possible given the last of desired cluster state without a management layer like Magnum. If we can think of general structures that would allow us to abstract the ideal cluster state, it would be preferable.
> 
> Docs is a solid requirement we need meet before merging. The Magnum requirement just needs to be documented, and we might want to think about a different location in the repository for now.

Thanks for the suggestion, Chris. I'm working on the doc and a demo at the moment.

I agree we should have a general solution, however, at the moment, I can't think of an existing solution that is production ready in openstack for kubernetes-as-a-service, except Magnum(please correct me if i missed something). A similar discussion happened in cluster-autoscaler[0] as well. Additionally, the cloud provider layer provides the abstract interface which makes things pluggable.

[0]: https://github.com/kubernetes/autoscaler/pull/1690

jichenjc:
yes, please add some document and @lingxiankong when you say plugable, does this mean later on if we have other solution we may potentially write some glue code in order to make k8s-auto-healer to use it ?

and looks like there are some existing considerations on the healthy of node
https://github.com/openshift/machine-api-operator/#machine-healthcheck-controller , FYI

lingxiankong:
> yes, please add some document and @lingxiankong when you say pluggable, does this mean later on if we have other solution we may potentially write some glue code in order to make k8s-auto-healer to use it ?

Don't understand what you mean by the `glue code`. By pluggable, I mean different cloud providers could have their own implementation.

> 
> and looks like there are some existing considerations on the healthy of node
> https://github.com/openshift/machine-api-operator/#machine-healthcheck-controller , FYI

Thanks for the link, @jichenjc. We are living in OpenStack world, I'd prefer to use Machine API Operator if the k8s cluster is created using Cluster API rather than Magnum.

lingxiankong:
> Hi.
> 
> This has the same purpose as work already done in magnum: https://review.opendev.org/#/c/631378/

I'm very glad you mentioned this patch, I was involved in the design and discussion of that patch, mostly with the author because we are in the same team, and Feilong is also the co-author of this PR actually. The current solution(NPD+Draino+autoscaler) in Magnum has some problems and cannot satisfy our production requirement, e.g. it doesn't take master nodes into consideration, it cannot guarantee the original node count, etc. What's more, the whole solution involves too many components which brings complexity and maintenance overhead, and cluster-autoscaler is not designed to deal with autohealing scenario.

> 
> That's not to say that a solution on the kubernetes side can't also exist, but since this proposal is limited to magnum clusters only there isn't really much benefit.

It's not limited to Magnum, but with Magnum as the default implementation.

> 
> Something that would definitely have to be changed is that the cluster label `auto_healing_enabled` is already used in the magnum implementation, so can't be used [here](https://github.com/kubernetes/cloud-provider-openstack/pull/622/files#diff-58744bdfc314511a1a2f35f2c707a087R39).

Once we contributed the auto-healer in CPO, Magnum will be changed to use it when `auto_healing_enabled=True`

> 
> One final comment, in magnum this was achieved just by a combination of pre-existing components - draino, autoscaler, node problem detector - I would say that that makes it more configurable and easier to maintain than adding a new controller here.

IMHO, as I mentioned above, they are not doing good job in autohealing but brings complexity.

I'm not sure if you are already using that combo in production, we were, and we found some problems which motivated this PR.

tghartland:
@lingxiankong 
Thanks for explaining, this makes more sense to me now.

So you would still be using node problem detector and draino for their roles in detecting/cordoning bad nodes, but this controller instead of the autoscaler? We aren't using the new auto healing in magnum so I can't comment on how well it works.

> it doesn't take master nodes into consideration

This would maybe be possible after node groups are introduced in train. The master nodes will be their own node group and it would be possible to scale them.

I just tried the mechanism you're using (shut off VM, mark stack resource unhealthy, update stack), and I encounter one thing which is CERN specific but makes this mechanism unusable for us. All our VMs get a hostname on our network and our DNS only updates every 15 minutes, so it is not possible to delete and immediately remake a VM with the same name. When I send the update to the stack, it tries to replace the marked minion, and fails:
`Resource UPDATE failed: resources.kube_minions: Resource CREATE failed: BadRequest: resources[0].resources.kube-minion: Hostname already in use (HTTP 400)`

This can be avoided by putting the stack index of the unhealthy minion into the `removed_rsrc_list` of the kube_minions stack, so that the new minion takes a new name, but I don't know if that list should (or even can) be modified manually.

lingxiankong:
> So you would still be using node problem detector and draino for their roles in detecting/cordoning bad nodes, but this controller instead of the autoscaler?

No, auto-healer will replace the whole combo and designed as the autohealing purpose specifically. The NPD and Draino are both great projects and professional in their specific areas, I believe they could be integrated in other solutions that more suitable for them.  In the end, we only need autoscaler for autoscaling and autohealer for autohealing.

lingxiankong:
> All our VMs get a hostname on our network and our DNS only updates every 15 minutes, so it is not possible to delete and immediately remake a VM with the same name.

I think that'd be problematic whether or not you are using auto-healer. Suppose you are using cluster-autoscaler, what if the CA scales down a node but immediately scales up a new one within the DNS refresh period? I guess the new node name will be same with the old one, no?

Because that's a specific issue in CERN, one possible solution I can think of is to maintain your own cloud provider (maybe the only difference with the openstack one is one extra function to remove the DNS record after removing the VM somehow.)

tghartland:
@lingxiankong when a node is deleted via heat, it's stack index is put into `removed_rsrc_list` property on the stack, and not used again.

For example if you have nodes [0, 1, 2] and the autoscaler removes minion-1 then adds one more node, the new node will be minion-3.

lingxiankong:
> @lingxiankong when a node is deleted via heat, its stack index is put into `removed_rsrc_list` property on the stack, and not used again.
> 
> For example if you have nodes [0, 1, 2] and the autoscaler removes minion-1 then adds one more node, the new node will be minion-3.

Thanks for the answer, then it's ok for autoscaler.

lingxiankong:
@hogepodge @adisky @tghartland documentation is added.

strigazi:
> > All our VMs get a hostname on our network and our DNS only updates every 15 minutes, so it is not possible to delete and immediately remake a VM with the same name.
> 
> I think that'd be problematic whether or not you are using auto-healer. Suppose you are using cluster-autoscaler, what if the CA scales down a node but immediately scales up a new one within the DNS refresh period? I guess the new node name will be same with the old one, no?
> 
> Because that's a specific issue in CERN, one possible solution I can think of is to maintain your own cloud provider (maybe the only difference with the openstack one is one extra function to remove the DNS record after removing the VM somehow.)

Well, if we start maintaining our own, we can't give much feedback since it is unusable by us.

lingxiankong:
> Well, if we start maintaining our own, we can't give much feedback since it is unusable by us.
> Scaling down and then up fast works, since it does not create vms with the same names.
> Deleted vms are appended in the list of deleted resources by heat (this is the default behavior
> in heat ResourceGroup).

The current Magnum implementation for node repair relies heavily on the Heat ["Mark Unhealthy Resources"](https://specs.openstack.org/openstack/heat-specs/specs/mitaka/mark-unhealthy.html) feature, which replaces the unhealthy resource with a new one but using the same name. Although `DNS only updates every 15 minutes` sounds a bit weird to me, I will try to find if there is a better option. Your suggestions are also appreciated.

strigazi:
> > Well, if we start maintaining our own, we can't give much feedback since it is unusable by us.
> > Scaling down and then up fast works, since it does not create vms with the same names.
> > Deleted vms are appended in the list of deleted resources by heat (this is the default behavior
> > in heat ResourceGroup).
> 
> The current Magnum implementation for node repair relies heavily on the Heat ["Mark Unhealthy Resources"](https://specs.openstack.org/openstack/heat-specs/specs/mitaka/mark-unhealthy.html) feature, which replaces the unhealthy resource with a new one but using the same name. Although `DNS only updates every 15 minutes` sounds a bit weird to me, I will try to find if there is a better option. Your suggestions are also appreciated.

`DNS only updates every 15 minutes` The reasons for this are not relevant to this discussion.

You can delete and create a node using magnum's cluster resize api.

lingxiankong:
> My suggestion: delete and create a node using magnum's cluster resize api.

Resize API doesn't support master for now.

strigazi:
> > My suggestion: delete and create a node using magnum's cluster resize api.
> 
> Resize API doesn't support master for now.

So we can make it support it instead of implementing this new heat based thing. Which option seems more reasonable to your needs and why?

adisky:
This component may be useful and may have better design than counterparts, but i am skeptical about having it in cloud-provider-openstack repo, as here we generally have implementation of plugins that integrates kubernetes with openstack services.

lingxiankong:
> plugins that integrates kubernetes with openstack services.

The auto-healer is actually a service running on Kubernetes but need the capability of the clouds (with OpenStack services as the first citizen)

lingxiankong:
> So we can make it support it instead of implementing this new heat based thing. Which option seems more reasonable based on your needs and why?

I agree using resize API is a better option at the time when it's fully implemented in Magnum. Before that, I didn't see there is an easier way to do the autohealing than using the Heat API in OpenStack. We are not developing a service for the future, but for the Kubernetes cluster created (via Magnum) in the current OpenStack cloud.

> 
> In any case, this is up to the cloud-provider-openstack maintainers to decide. Sorry for the noise, from our point of view, we can not help on it.

Appreciate your suggestion.

openstacker:
As the original author of Magnum /resize API and the auto-healing feature in Magnum, I'm working with Lingxian very closely on this auto healer. And as Lingxian mentioned above, the main requirement to have a dedicated auto healer is because the combination of (NPD + Draino + Cluster Autoscaler) can't 100% meet the requirement of a small public cloud or user don't want a size-unpredicted cluster when doing auto-healing.

As for the /resize API, that's on our roadmap. As long as we get the /resize API working with master nodes in Magnum, we will switch current (heat mark unhealthy) way to Magnum /resize.  This is a very useful service for any cloud provider they want the auto-healing, and I'm pretty sure it will be adopted by more cloud providers. So it would be nice if the #sig-openstack can accept this until the community can figure out a better place. Thank you.

lingxiankong:
@hogepodge Can we move the PR forward? Is there anything else we could do?

hogepodge:
In thinking about how to support the community on these efforts, I propose that we accept the code if the name is changed. The name k8s-auto-healer has a large potential to confuse end users, and to make it clear that this requires magnum I propose that we change it to "magnum-auto-healer" or magnum-k8s-auto-healer"

lingxiankong:
@hogepodge the binary has been renamed to `magnum-auto-healer` (and all other places related)

pkg/autohealing/cloudprovider/cloudprovider.go
```diff
@@ -0,0 +1,66 @@
...
+	"k8s.io/cloud-provider-openstack/pkg/autohealing/config"
+	"k8s.io/cloud-provider-openstack/pkg/autohealing/healthcheck"
+)
+
+var (
+	providers = make(map[string]RegisterFunc)
+)
+
+// CloudProvider is an abstract, pluggable interface for cloud providers.
+type CloudProvider interface {
```
jim-bach: does this intended support other cloud providers than OpenStack? should this be just in the Kubernetes org?

pkg/autohealing/cloudprovider/openstack/provider.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	// In order to rebuild the nodes by Heat stack update, we need to know the parent stack ID of the resources and
+	// mark them unhealthy first.
+	allMapping, err := provider.getAllStackResourceMapping(clusterStackName, cluster.StackID)
+	if err != nil {
+		return fmt.Errorf("failed to get the resource stack mapping for cluster %s, error: %v", clusterName, err)
+	}
+
+	for _, n := range nodes {
```
jim-bach: might i suggest batch size and maybe false positives detection, in a network/partial failure scenario, this will be deleting all the nodes.

pkg/autohealing/cloudprovider/cloudprovider.go
lingxiankong: We could move to kubernetes.org after there are other providers implemented and maintainers from other clouds, but for now, there is no plan yet. I'm very happy that people could help to implement other cloud providers and we could find a better place for the service then.

pkg/autohealing/cloudprovider/openstack/provider.go
lingxiankong: Yes, batch size repair is on our roadmap. I'm waiting for this service to be merged and we could add the support in Magnum, so that we could do the integration test in our preprod.

jim-bach: nice, did i miss it? is the roadmap published somewhere?

lingxiankong: I will submit the feature request after the service is approved, otherwise, I don't think it makes sense to discuss the roadmap for a private service.

docs/using-magnum-auto-healer.md
```diff
@@ -0,0 +1,227 @@
...
+    | 10bef366-b5a8-4400-b2c3-82188ec06b13 | lingxian-por-test-1-12-7-ha-bbgjts5g4xhb-master-1 | ACTIVE | lingxian_net=10.0.10.30, 150.242.42.22  | fedora-atomic-27-x86_64 | c1.c2r4 |
+    | 9c17f034-6825-4e49-b3cb-0ecddd1a8dd8 | lingxian-por-test-1-12-7-ha-bbgjts5g4xhb-master-0 | ACTIVE | lingxian_net=10.0.10.29, 150.242.42.213 | fedora-atomic-27-x86_64 | c1.c2r4 |
+    +--------------------------------------+---------------------------------------------------+--------+-----------------------------------------+-------------------------+---------+
+    ```
+
+2. The kubeconfig file of the cluster is in place.
+
+### Deploy magnum-auto-healer
+
+It's recommended to run magnum-auto-healer service as a DaemonSet on the master nodes, the service is running in master-slave mode using leader election mechanism. There is a sample manifest file in `manifests/magnum-auto-healer/magnum-auto-healer.yaml`, you need to change some variables as need before actually running `kubectl apply` command. The following commands are just examples:
```
hogepodge: Referring back to this previously reported issue[1], the use of the terminology "master-slave" is discouraged within the Kubernetes community, and the preferred nomenclature is "control-plane"[2] and "node"[3].

[1] https://github.com/kubernetes/website/issues/6525
[2] https://kubernetes.io/docs/reference/glossary/?fundamental=true#term-control-plane
[3] https://kubernetes.io/docs/reference/glossary/?fundamental=true#node

lingxiankong: Changed to "active-passive"

PR #504, opened 2019-02-25
[WIP] cleanup mount package

ramineni:
This commit reuses kubernetes/pkg/util/mount in places
required. And also removes, existing copy of k8s mount
package in repository.

**What this PR does / why we need it**:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
```

PR #631, opened 2019-06-04
Fix csi cinder disk teardown on EIO errors

chrischdi:
Handle teardown of cinder disks which return an "Input/Output" during `os.Stat` system call.
It is a reproducible problem which exists and persists if the disk is already detached via `nova volume-detach` but still mounted inside the operating system.

Fixes the issue on the Cinder CSI side mentioned in https://github.com/kubernetes/kubernetes/issues/77969

As discussed in Mattermost :-)

The problem is reproducible as follows:
* create statefulset [gist](https://gist.github.com/chrischdi/39141aaf5fc3b2cb48200d216b5ce7b4)
* detach cinder disk manually (to reproduce the problem, may be a race-condition) `nova volume-detach`
* wait for disk to get I/O error on vm
  ```
  # ls /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-fc52b5a4-77bc-11e9-8554-fa163e0629eb/globalmount
  ls: cannot access '/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-fc52b5a4-77bc-11e9-8554-fa163e0629eb/globalmount': Input/output error
  ```
* call pod deletion `kubectl delete pod test-statefulset-0`
Now the pod is stuck in termination and the logs show `input/output` error

**Release note**:

```release-note
```

### DIFF
--- a/pkg/csi/cinder/mount/mount.go
+++ b/pkg/csi/cinder/mount/mount.go
@@ -238,3 +238,7 @@ func (m *Mount) GetInstanceID() (string, error) {
 	}
 	return "", err
 }
+
+func IsCorruptedMnt(err error) bool {
+	return mount.IsCorruptedMnt(err)
+}
--- a/pkg/csi/cinder/nodeserver.go
+++ b/pkg/csi/cinder/nodeserver.go
@@ -135,10 +135,10 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 	m := ns.Mount
 
 	notMnt, err := m.IsLikelyNotMountPointDetach(targetPath)
-	if err != nil {
+	if err != nil && !mount.IsCorruptedMnt(err) {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
-	if notMnt {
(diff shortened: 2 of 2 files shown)


adisky:
@chrischdi the fix looks good to me, assigning it to @ramineni, as she has worked on mount pkg.
/assign @ramineni

pkg/csi/cinder/mount/mount.go
```diff
@@ -238,3 +238,7 @@ func (m *Mount) GetInstanceID() (string, error) {
 	}
 	return "", err
 }
+
+func IsCorruptedMnt(err error) bool {
```
hogepodge: I guess I'm a bit confused as to why we wrap this rather than use it directly.

adisky: @ramineni is trying to clean it up
https://github.com/kubernetes/cloud-provider-openstack/pull/504

ramineni:
@chrischdi +1 , Thanks for the fix. Looks good to me as well .

adisky:
cc @lingxiankong for approval :)

PR #641, opened 2019-06-12
Add more information to gophercloud user-agent string

### LINKED ISSUE #606: Gophercloud user-agent should include component version and cluster UUID
/kind feature

The version of the cloud-provider-openstack components and the UUID of the cluster that is running them should be added to the gophercloud user-agent, to be able to group and filter request logs and to provide better information in general.

Setting the version is relatively simple, and can be done in the same way as [I am doing in cluster-autoscaler](https://github.com/kubernetes/autoscaler/pull/1986). Unfortunately the cluster UUID can not be passed through from the CLI parameters in `main.go` to where it is needed, as the function in which [the provider is created](https://github.com/kubernetes/cloud-provider-openstack/blob/20b7744ee505a1da05550a866051bf5323e4feba/pkg/cloudprovider/providers/openstack/openstack.go#L412) is only [registered with k8s.io/cloud-provider](https://github.com/kubernetes/cloud-provider-openstack/blob/20b7744ee505a1da05550a866051bf5323e4feba/pkg/cloudprovider/providers/openstack/openstack.go#L209), and not called directly.

Possible solutions are:

- Adding the cluster UUID to the config file and struct
- Setting a global variable like the version, but for the cluster UUID

The first option seems like it would probably be disruptive to a lot of places that use the same struct.

For the second option the UUID variable could either be stored in its own package like the version, which adds a lot just for a single variable, or the variable could be set right in the package where it needs to be used. For example `cmd/openstack-cloud-controller-manager/main.go` would set a variable in `pkg/cloudprovider/providers/openstack/openstack.go`.

Other `cmd/*/main.go` would have to set the cluster UUID in whichever packages they use, i.e `cmd/k8s-keystone-auth` -> `pkg/identity/keystone/keystone.go`.

Should I go ahead with this design?

tghartland:

Adds the cloud provider version string to gophercloud user-agent string, and adds a `user-agent` flag which can be used to add any other information.

I.e in Magnum, for the cloud controller manager configuration `--user-agent="cluster/${CLUSTER_UUID}"`, to be able to see which clusters are responsible for which requests.

If there are no objections to the code for the cloud controller manager then I will make the same changes for the keystone and barbican components (and any others that are suggested)

**Which issue this PR fixes** : fixes #606 

Since there is only a single flag, to add more than one thing to the user-agent would be like
`--user-agent="abc/123 cluster/${CLUSTER_UUID}"`

but in yaml it would look like
```yaml
command:
  - ......
  - --user-agent=abc/123 cluster/${CLUSTER_UUID}
```
as the quotes are not necessary since the whole line is already a string, but this might not be intuitive. If the quotes are left in then they are included in the user agent.

Instead it would be possible to make `user-agent` a flag that can be used multiple times with different arguments. Is that preferable?

**Release note**:

```release-note
Set gophercloud user-agent in all components, and add flags for custom user-agent data in openstack-cloud-controller-manager, k8s-keystone-auth, manila-csi-plugin, manila-provisioner, cinder-csi-plugin, cinder-provisioner
```

### DIFF
--- a/cmd/cinder-provisioner/main.go
+++ b/cmd/cinder-provisioner/main.go
@@ -23,6 +23,7 @@ import (
 
 	"k8s.io/cloud-provider-openstack/pkg/version"
 	"k8s.io/cloud-provider-openstack/pkg/volume/cinder/provisioner"
+	"k8s.io/cloud-provider-openstack/pkg/volume/cinder/volumeservice"
 	"sigs.k8s.io/sig-storage-lib-external-provisioner/controller"
 
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -45,6 +46,8 @@ func main() {
 	pflag.StringVar(&kubeconfig, "kubeconfig", "", "Absolute path to the kubeconfig")
 	pflag.StringVar(&id, "id", "", "Unique provisioner identity")
 	pflag.StringVar(&cloudconfig, "cloud-config", "", "Path to OpenStack config file")
+
+	volumeservice.AddExtraFlags(pflag.CommandLine)
 
 	// Glog requires this otherwise it complains.
 	flag.CommandLine.Parse(nil)
--- a/cmd/cinder-provisioner/main.go
+++ b/cmd/cinder-provisioner/main.go
@@ -21,6 +21,7 @@ import (
 	"github.com/spf13/pflag"
 	"k8s.io/klog"
 
+	"k8s.io/cloud-provider-openstack/pkg/version"
 	"k8s.io/cloud-provider-openstack/pkg/volume/cinder/provisioner"
 	"sigs.k8s.io/sig-storage-lib-external-provisioner/controller"
 
@@ -37,7 +38,6 @@ var (
 	kubeconfig  string
 	id          string
 	cloudconfig string
-	version     string
 )
 
 func main() {
@@ -66,7 +66,7 @@ func main() {
 	logs.InitLogs()
 	defer logs.FlushLogs()
 
-	klog.V(1).Infof("cinder-provisioner version: %s", version)
+	klog.V(1).Infof("cinder-provisioner version: %s", version.Version)
 
 	var config *rest.Config
 	var err error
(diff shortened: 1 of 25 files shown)


tghartland:
@jim-bach, @hogepodge, @strigazi

strigazi:
lgtm

jim-bach:
lgtm thanks for getting this in @tghartland

tghartland:
I added the user-agent flag to k8s-keystone-auth as well, and changed the flag type to StringArray.
That means the flag would be used multiple times in the container command like so:
```
--user-agent=cluster/e58868d8-79cb-4567-adf9-c18ea9a5b350
--user-agent=owner/tghartland
```
and the user-agent would then be set as
```
"owner/tghartland cluster/e58868d8-79cb-4567-adf9-c18ea9a5b350 openstack-cloud-controller-manager/1.14.0 gophercloud/2.0.0" 
```

tghartland:
Rebased to master.

tghartland:
/hold

Version strings in the other cmd/... components need to be changed to use the string from the new location.

tghartland:
Added flags for custom data to:
* openstack-cloud-controller-manager
* k8s-keystone-auth
* manila-csi-plugin
* manila-provisioner
* cinder-csi-plugin
* cinder-provisioner

Set the gophercloud user-agent to "component/version" in:
* client-keystone-auth
* magnum-auto-healer
* octavia-ingress-controller

cmd/cinder-provisioner/main.go
```diff
@@ -46,6 +47,8 @@ func main() {
 	pflag.StringVar(&id, "id", "", "Unique provisioner identity")
 	pflag.StringVar(&cloudconfig, "cloud-config", "", "Path to OpenStack config file")
 
+	volumeservice.AddExtraFlags(pflag.CommandLine)
```
hogepodge: Since we are adding a new flag to the API with associated functionality, can we please add unit tests to check the default and the custom behavior, including using multiple flags.

tghartland: I have added tests for the components which have the new flag; [diff](https://github.com/kubernetes/cloud-provider-openstack/compare/00ae5d268e23dd8aa5c80411a099a367d591af0c..e20a17f40ee9f19bd596f96b8a17be195e939611).

hogepodge:
@tghartland You'll need to remove the hold once you're ready to merge.

Issue #630, opened 2019-06-04
Add CI jobs for Manila CSI Driver


adisky:
/kind Feature

**What happened**:

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- openstack-cloud-controller-manager version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

adisky:
cc @gman0

gman0:
@adisky yes, CI tests are in the plan

gman0:
@adisky can you assign me please?

adisky:
@gman0 Please apply for k8s community membership
https://github.com/kubernetes/community/blob/master/community-membership.md

gman0:
xref for csi-sanity tests https://github.com/kubernetes/cloud-provider-openstack/issues/714

gman0:
xref for Zuul acceptance job https://github.com/theopenlab/openlab-zuul-jobs/pull/624

PR #652, opened 2019-06-21
csi-manila: added initial support for snapshots

gman0:
This PR adds initial support for snapshots and restoring volumes from snapshots. Docs and examples updated as well + some minor refactoring, clean up

Things that are still missing and will be added in the upcoming PRs:
* support for snapshot restoration of CephFS shares: Manila doesn't support _"create share from snapshot"_ for CephFS just yet (see [this support matrix](https://docs.openstack.org/manila/stein/admin/share_back_ends_feature_support_mapping.html)), meaning in this case the driver will have to do the storage-specific heavy lifting on its own
* too generic error messages from backend (not necessarily linked to snapshots only): if something goes wrong (e.g. storage quota exceeded when creating a new share), the relevant resource will go into an error state. The driver currently only reports that the operation has failed due to the resource being in the aforementioned error state, but not much more - not exactly helpful to the end user. There is an API to fetch the actual error messages from Manila, but we're lacking its client-side support in the gophercloud framwork. This needs to be added to gophercloud in order to improve error reporting.
* CI jobs: we're lacking CI jobs (https://github.com/kubernetes/cloud-provider-openstack/issues/630), the plan is to test the whole life-cycle of a share, that is create->snapshot->restore->mount->delete

**Release note**:

```release-note
CSI Manila: added initial support for snapshots
```


gman0:
cc @tsmetana @gouthampacha @tombarron @adisky @ramineni

pkg/csi/manila/snapshot.go
```diff
@@ -0,0 +1,98 @@
...
+			if snapshot, createErr = snapshots.Create(manilaClient, req).Extract(); createErr != nil {
+				return nil, createErr
+			}
+
+		} else {
+			// Something else is wrong
+			return nil, fmt.Errorf("failed to probe for snapshot: %v", err)
+		}
+	} else {
+		klog.V(4).Infof("snapshot %s (snapshot ID %s) already exists", snapID, snapshot.ID)
```
tsmetana: I think we should also check whether the sourceShareID is the same. Calling CreateSnapshot with the existing SnapshotID but different source is also an error.

pkg/csi/manila/util.go
```diff
@@ -24,18 +24,27 @@ import (
...
 	bytesInGiB = 1024 * 1024 * 1024
 )
 
+func newVolumeID(volumeName string) volumeID {
+	return volumeID(fmt.Sprintf("csi-manila-%s", volumeName))
+}
+
+func newSnapshotID(snapshotName string) snapshotID {
+	return snapshotID(fmt.Sprintf("csi-manila-%s", snapshotName))
+}
```
tsmetana: Shouldn't the IDs be more unique? Won't this end with too frequent ID collisions?

pkg/csi/manila/snapshot.go
gman0: Good point, thanks!

pkg/csi/manila/util.go
gman0: I don't think so - at least from the perspective of a single CO because the resource names it generates are implied to be unique for that particular instance of CO (otherwise it couldn't guarantee idempotency). We could use the actual resource ID from Manila as a volume/snapshot ID though - that would make it easier to find the resource on the backend...so this is definitely something to reconsider but it probably won't help with uniqueness.

If what you mean is the situation where 2 COs were to generate the same name, then yes, only one of them would create the resource and the other one would just fetch the existing storage asset - and this is indeed a collision because in this situation the intention was to have two distinct resources. So if we have a single Manila instance and multiple COs (each running an instance of csi-manila) claiming assets on that Manila instance, each instance of the driver could receive some unique string during startup (e.g. cmd args / env vars)...then we could have a hashing function where
_hash(input-unique-string, resource-name) = resource-id_

What do you think?

tsmetana: Snapshots are namespaced (just like PVCs): user in namespace A wants snapshot `snap`, user in namespace B wants also snapshot `snap` (I tend to do this myself -- reusing yamls for different things in one cluster) and this code would assign both the snapshots the same ID.

gman0: Sorry, I'm not following...sure, PVCs and snapshots can have matching names (in their object metadata) in separate namespaces but those are not exposed to CSI. CSI `Create{Volume/Snapshot}` receives UIDs of those objects (stored in `CreateVolumeRequest.Name` and `CreateSnapshotRequest.Name` respectively) and these must be unique for sure.

I tried to reproduce the scenario you've described:
I've created 2 PVCs named `new-nfs-share-pvc` in namespaces `default` and `kube-public`

```
I0622 11:38:09.066939       1 driver.go:265] [ID:14] GRPC call: /csi.v1.Controller/CreateVolume
I0622 11:38:09.067183       1 driver.go:266] [ID:14] GRPC request: {"capacity_range":{"required_bytes":1073741824},"name":"pvc-30f456e8-94e2-11e9-809c-52540013b951","parameters":{"type":"default"},"secrets":"***stripped***","volume_capabilities":[{"AccessType":{"Mount":{"fs_type":"ext4"}},"access_mode":{"mode":5}}]}
I0622 11:38:10.512865       1 volumesource.go:37] creating a share for volume csi-manila-pvc-30f456e8-94e2-11e9-809c-52540013b951

I0622 11:40:53.646209       1 driver.go:265] [ID:19] GRPC call: /csi.v1.Controller/CreateVolume
I0622 11:40:53.646422       1 driver.go:266] [ID:19] GRPC request: {"capacity_range":{"required_bytes":1073741824},"name":"pvc-91097c3e-94e2-11e9-809c-52540013b951","parameters":{"type":"default"},"secrets":"***stripped***","volume_capabilities":[{"AccessType":{"Mount":{"fs_type":"ext4"}},"access_mode":{"mode":5}}]}
I0622 11:40:53.755975       1 volumesource.go:37] creating a share for volume csi-manila-pvc-91097c3e-94e2-11e9-809c-52540013b951
```
Those are 2 `CreateVolume` requests for 2 PVCs with the same name but different namespace.
```
$ kubectl get pv
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM                           STORAGECLASS     REASON   AGE
pvc-30f456e8-94e2-11e9-809c-52540013b951   1Gi        RWX            Delete           Bound    default/new-nfs-share-pvc       csi-manila-nfs            57m
pvc-91097c3e-94e2-11e9-809c-52540013b951   1Gi        RWX            Delete           Bound    kube-public/new-nfs-share-pvc   csi-manila-nfs            55m

$ kubectl describe pv pvc-30f456e8-94e2-11e9-809c-52540013b951
Name:            pvc-30f456e8-94e2-11e9-809c-52540013b951
Labels:          <none>
Annotations:     pv.kubernetes.io/provisioned-by: manila.csi.openstack.org
Finalizers:      [kubernetes.io/pv-protection]
StorageClass:    csi-manila-nfs
Status:          Bound
Claim:           default/new-nfs-share-pvc
Reclaim Policy:  Delete
Access Modes:    RWX
VolumeMode:      Filesystem
Capacity:        1Gi
Node Affinity:   <none>
Message:         
Source:
    Type:              CSI (a Container Storage Interface (CSI) volume source)
    Driver:            manila.csi.openstack.org
    VolumeHandle:      csi-manila-pvc-30f456e8-94e2-11e9-809c-52540013b951
    ReadOnly:          false
    VolumeAttributes:      cephfs-mounter=fuse
                           shareAccessID=95a4ca19-ae10-4457-b5e0-c57e955249de
                           shareID=5e123251-fe73-44aa-8124-4e624b401c1d
                           storage.kubernetes.io/csiProvisionerIdentity=1561201976886-8081-manila.csi.openstack.org

$ kubectl describe pv pvc-91097c3e-94e2-11e9-809c-52540013b951
Name:            pvc-91097c3e-94e2-11e9-809c-52540013b951
Labels:          <none>
Annotations:     pv.kubernetes.io/provisioned-by: manila.csi.openstack.org
Finalizers:      [kubernetes.io/pv-protection]
StorageClass:    csi-manila-nfs
Status:          Bound
Claim:           kube-public/new-nfs-share-pvc
Reclaim Policy:  Delete
Access Modes:    RWX
VolumeMode:      Filesystem
Capacity:        1Gi
Node Affinity:   <none>
Message:         
Source:
    Type:              CSI (a Container Storage Interface (CSI) volume source)
    Driver:            manila.csi.openstack.org
    VolumeHandle:      csi-manila-pvc-91097c3e-94e2-11e9-809c-52540013b951
    ReadOnly:          false
    VolumeAttributes:      cephfs-mounter=fuse
                           shareAccessID=5cda7e52-350a-4fff-980d-8fc8790dcea1
                           shareID=ebf07997-83bf-436f-b305-45d96d32ed71
                           storage.kubernetes.io/csiProvisionerIdentity=1561201976886-8081-manila.csi.openstack.org
```
First snapshot of `default/new-nfs-share-pvc`:
```
I0622 11:42:02.044179       1 driver.go:265] [ID:24] GRPC call: /csi.v1.Controller/CreateSnapshot
I0622 11:42:02.044200       1 driver.go:266] [ID:24] GRPC request: {"name":"snapshot-bcd02708-94e2-11e9-809c-52540013b951","secrets":"***stripped***","source_volume_id":"csi-manila-pvc-30f456e8-94e2-11e9-809c-52540013b951"}
I0622 11:42:02.206078       1 controllerserver.go:252] creating snapshot csi-manila-snapshot-bcd02708-94e2-11e9-809c-52540013b951 of volume csi-manila-pvc-30f456e8-94e2-11e9-809c-52540013b951

$ kubectl describe volumesnapshot.snapshot.storage.k8s.io/new-nfs-share-snap                                                                                                                                     
Name:         new-nfs-share-snap
Namespace:    default
Labels:       <none>
Annotations:  <none>
API Version:  snapshot.storage.k8s.io/v1alpha1
Kind:         VolumeSnapshot
Metadata:
  Creation Timestamp:  2019-06-22T11:41:56Z
  Finalizers:
    snapshot.storage.kubernetes.io/volumesnapshot-protection
  Generation:        5
  Resource Version:  1849397
  Self Link:         /apis/snapshot.storage.k8s.io/v1alpha1/namespaces/default/volumesnapshots/new-nfs-share-snap
  UID:               bcd02708-94e2-11e9-809c-52540013b951
```
snapshot of `kube-system/new-nfs-share-pvc`:
```
I0622 11:55:17.826043       1 driver.go:265] [ID:38] GRPC call: /csi.v1.Controller/CreateSnapshot
I0622 11:55:17.826171       1 driver.go:266] [ID:38] GRPC request: {"name":"snapshot-0eee019e-94e4-11e9-809c-52540013b951","secrets":"***stripped***","source_volume_id":"csi-manila-pvc-91097c3e-94e2-11e9-809c-52540013b951"}
I0622 11:55:17.992888       1 controllerserver.go:252] creating snapshot csi-manila-snapshot-0eee019e-94e4-11e9-809c-52540013b951 of volume csi-manila-pvc-91097c3e-94e2-11e9-809c-52540013b951

$ kubectl -n kube-public describe volumesnapshot.snapshot.storage.k8s.io/new-nfs-share-snap
Name:         new-nfs-share-snap
Namespace:    kube-public
Labels:       <none>
Annotations:  <none>
API Version:  snapshot.storage.k8s.io/v1alpha1
Kind:         VolumeSnapshot
Metadata:
  Creation Timestamp:  2019-06-22T11:51:23Z
  Finalizers:
    snapshot.storage.kubernetes.io/volumesnapshot-protection
  Generation:        5
  Resource Version:  1850153
  Self Link:         /apis/snapshot.storage.k8s.io/v1alpha1/namespaces/kube-public/volumesnapshots/new-nfs-share-snap
  UID:               0eee019e-94e4-11e9-809c-52540013b951
Spec:
  Snapshot Class Name:    csi-manila-snapclass
  Snapshot Content Name:  snapcontent-0eee019e-94e4-11e9-809c-52540013b951
  Source:
    API Group:  <nil>
    Kind:       PersistentVolumeClaim
    Name:       new-nfs-share-pvc
```
None of those resources' IDs match, despite their matching object metadata names in k8s.

> And think about special cases where user creates a snapshot, deletes it and creates another one with the same name... This must not break either.

```
I0622 13:07:24.104486       1 driver.go:265] [ID:17] GRPC call: /csi.v1.Controller/CreateSnapshot
I0622 13:07:24.104630       1 driver.go:266] [ID:17] GRPC request: {"name":"snapshot-acfa5443-94ee-11e9-a2e2-52540013b951","secrets":"***stripped***","source_volume_id":"csi-manila-pvc-9798c34e-94ee-11e9-a2e2-52540013b951"}
I0622 13:07:24.338836       1 controllerserver.go:252] creating snapshot csi-manila-snapshot-acfa5443-94ee-11e9-a2e2-52540013b951 of volume csi-manila-pvc-9798c34e-94ee-11e9-a2e2-52540013b951

I0622 13:09:04.165481       1 driver.go:265] [ID:24] GRPC call: /csi.v1.Controller/CreateSnapshot
I0622 13:09:04.165500       1 driver.go:266] [ID:24] GRPC request: {"name":"snapshot-e8c30c85-94ee-11e9-a2e2-52540013b951","secrets":"***stripped***","source_volume_id":"csi-manila-pvc-9798c34e-94ee-11e9-a2e2-52540013b951"}
I0622 13:09:04.321829       1 controllerserver.go:252] creating snapshot csi-manila-snapshot-e8c30c85-94ee-11e9-a2e2-52540013b951 of volume csi-manila-pvc-9798c34e-94ee-11e9-a2e2-52540013b951
```
It can't break, their UIDs are different.

The CSI spec says that volume/snapshot names are there as a suggested ID ([CreateVolume, DeleteVolume, ListVolume RPC interactions](https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume-deletevolume-listvolumes)) and to maintain idempotency between requests with the same name in a CSI request. If those names were not unique, I don't think they would be able to maintain idempotency.

What do you think? Or can you elaborate more regarding where we might be in a risk of ID collision?

pkg/csi/manila/snapshot.go
adisky: +1 if sourceShareID is different, we need to return codes.AlreadyExists error

pkg/csi/manila/util.go
adisky: I have not read the whole conversation, but what is the benefit of generating id's here?? why we cannot use the id from openstack as it is??

adisky:
@gman0 Thanks for the PR!!
Few questions inline and also It is suggested to add sanity testing(in a separate PR ) 
https://github.com/kubernetes-csi/csi-test/tree/master/pkg/sanity, it is easier to track validation and return error codes.

pkg/csi/manila/util.go
tsmetana: My bad: I didn't realize where do these names come from. They should be unique enough for nothing I was afraid of to happen... I got triggered by the fact we call ID something that relies on external source for uniqueness.
@adisky Yes, this would be also an option, @gman0 mentiones this in the comment above too. I think the reason to use our own ID is simply because it's easier and good enough for the moment.

adisky:
@gman0 This PR needs a rebase

pkg/csi/manila/snapshot.go
gman0: ID validation added in https://github.com/kubernetes/cloud-provider-openstack/pull/652/commits/a8bdd2356094f3b93d85eff20cf1eefc40a6d488

pkg/csi/manila/util.go
gman0: Using OpenStack resource IDs as volume/snapshot IDs: https://github.com/kubernetes/cloud-provider-openstack/pull/652/commits/a8bdd2356094f3b93d85eff20cf1eefc40a6d488

adisky:
@gman0 Is it in final shape from your side??

gman0:
@adisky oops, forgot to post a comment...yes, this PR is done and waiting for LGTMs

PTAL @tsmetana @tombarron @gouthampacha

adisky:
considering above comments will be addressed in a separate PR
/lgtm

adisky:
@gman0 Please before continuing with features add minimal gate jobs

gman0:
@ramineni thanks for your feedback! @adisky will do, thanks!

gman0:
I'll do a more thorough manifest refactoring in a separate PR

PR #656, opened 2019-06-25
kubeadm documentation update

sfxworks:

This clarifies the kubeadm documentation for cloud-provider-openstack

This is based on troubleshooting assistance provided by a_sykim in the k8s slack channel #sig-cloud-provider 

- InitializerConfiguration was removed in v1.14 and is no longer referenced in the documentation linked. 
- A known issue related to a race condition causes external addresses to not be listed when checking on nodes. A notice was put in to prevent the user from wondering why this is occurring. https://github.com/kubernetes/kubernetes/pull/75229

### DIFF
--- a/docs/using-controller-manager-with-kubeadm.md
+++ b/docs/using-controller-manager-with-kubeadm.md
@@ -28,7 +28,7 @@
     ```
 
     You can find an example kubeadm.conf in `manifests/controller-manager/kubeadm.conf`. Follow the usual steps to install the network plugin and then bootstrap the other nodes using `kubeadm join`.
-    _Note, your nodes may not have an External IP address. This will cause logs and CNI to have have issues until the cluster provider is complete. Reference https://github.com/kubernetes/kubernetes/pull/75229 for further information._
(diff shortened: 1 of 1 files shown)


docs/using-controller-manager-with-kubeadm.md
```diff
@@ -28,6 +28,7 @@
     ```
 
     You can find an example kubeadm.conf in `manifests/controller-manager/kubeadm.conf`. Follow the usual steps to install the network plugin and then bootstrap the other nodes using `kubeadm join`.
+    _Note, your nodes may not have an External IP address. This will cause logs and CNI to have have issues until the cluster provider is complete. Reference https://github.com/kubernetes/kubernetes/pull/75229 for further information._
```
ramineni: could we add Note as seperate paragraph, right now its showing as combined.
you could start with 
> NOTE: 
to make it visible as seperate paragraph

adisky: +1

sfxworks: Sure! Looks better like that. Updated!

ramineni:
Thanks.
/lgtm

PR #657, opened 2019-06-26
Don't run jobs when .gitignore is updated

ramineni:
This commit makes sure not to run test jobs which irrelavent while
updating .gitignore

**What this PR does / why we need it**:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
```

PR #655, opened 2019-06-25
trivial: Add manila-csi-plugin to .gitignore

adisky:
**What this PR does / why we need it**:
Add manila-csi-plugin binary to gitignore

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
```
`NONE`

PR #680, opened 2019-07-04
Add manila-csi-plugin binary to .gitignore

tghartland:

manila-csi-plugin is missing from the .gitignore, all the other binaries are there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -145,6 +145,7 @@ zz_generated.openapi.go
 /octavia-ingress-controller
 /client-keystone-auth
 /manila-provisioner
+/manila-csi-plugin
 /barbican-kms-plugin
 /magnum-auto-healer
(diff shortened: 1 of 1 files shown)


tghartland:
Should the .gitignore be added to the zuul.yaml irrelevant files? I don't think the full set of tests needs to be run for a change in the .gitignore.

ramineni:
@tghatland Thanks for the patch . Both of these patches are proposed already, due to gate issue we are unable to merge the same. https://github.com/kubernetes/cloud-provider-openstack/pull/655 ,https://github.com/kubernetes/cloud-provider-openstack/pull/657

adisky:
/close
in favour of https://github.com/kubernetes/cloud-provider-openstack/pull/655

Issue #668, opened 2019-06-27
Build container images from distroless 


ramineni:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
>
> /kind bug
> /kind feature

**What happened**:
 The kubernetes project in general is moving towards distroless: https://github.com/kubernetes/enhancements/blob/master/keps/sig-release/20190316-rebase-images-to-distroless.md , explore the same to be supported in cpo
**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- openstack-cloud-controller-manager version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #681, opened 2019-07-05
Update base image to debian

ramineni:
This PR updates base image to debian 
Motivation:
Officilally supported distro by kubernetes and recommended .
Some of the functionalities like resize doesnt work properly on Alpine.

**Release note**:

```release-note
```

### DIFF
--- a/cluster/images/cinder-csi-plugin/Dockerfile
+++ b/cluster/images/cinder-csi-plugin/Dockerfile
@@ -1,9 +1,9 @@
-FROM alpine:3.7
+FROM k8s.gcr.io/debian-base-amd64:1.0.0
 LABEL maintainers="Kubernetes Authors"
 LABEL description="Cinder CSI Plugin"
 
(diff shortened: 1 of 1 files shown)


jichenjc:
should we consider move all Docker file to base distro? it's not directly related to this PR ...

ramineni:
> should we consider move all Docker file to base distro? it's not directly related to this PR ...

@jichenjc Thanks for the review. Ultimate goal is to move to distroless which would be best . As there needs to be some investigation needs to be done to do the same (https://github.com/kubernetes/cloud-provider-openstack/issues/668 ) as a stop gap I have added debian as recommended as it is needed by cinder-csi for some of the features and alpine is not supported. As other modules could wait, we can directly check and move to distroless instead of debian. I'll work on that in future PRs.

jichenjc:
works for me, thanks for the detail info

adisky:
@ramineni  I am also ok.
/lgtm

PR #682, opened 2019-07-08
Remove broken node before repairing

openstacker:
Removing the broken node from the current cluster before doing an actual repair which can avoid some potential conflicts on the cloud providers side.

Based on the current implementation for the OpenStack cloud provider, the new node may be failed to start due to a timing issue.

### DIFF
--- a/pkg/autohealing/controller/controller.go
+++ b/pkg/autohealing/controller/controller.go
@@ -317,6 +317,10 @@ func (c *Controller) repairNodes(unhealthyNodes []healthcheck.NodeInfo) {
 						log.Infof("Node %s is cordoned", nodeName)
 					}
 
+					// Remove the broken node from the cluster
+					if err := c.kubeClient.CoreV1().Nodes().Delete(nodeName, &metav1.DeleteOptions{}); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/autohealing/controller/controller.go
```diff
@@ -317,6 +317,10 @@ func (c *Controller) repairNodes(unhealthyNodes []healthcheck.NodeInfo) {
 						log.Infof("Node %s is cordoned", nodeName)
 					}
 
+					// Remove the broken node from the cluster
+					if err := c.kubeClient.CoreV1().Nodes().Delete(nodeName, &metav1.DeleteOptions{}); err != nil {
+						log.Errorf("Failed to remove the node %s from cluster, error: %v", unhealthyNodeNames.List(), err)
+					}
```
lingxiankong: If the node is going to be deleted, is it necessary to cordon the node first?

openstacker: Good question. I would  say no, but let's keep the cordon just in case. And given in the future, we may move this logic into cloud provider side, so let's keep the cordon because it's a common step for repairing. Cheers.

PR #688, opened 2019-07-09
Fix GetMaxVolLimit to return correct value

### LINKED ISSUE #633: [Cinder CSI] GetMaxVolLimit returns always max value.
**What happened**:
The function GetMaxVolLimit will always return max value when nodeserver calls it. 

This happens because nodeserver never calls function CreateOpenStackProvider what would imitialize cfg global variable. 

GetMaxVolLimit should return the value defined in configuration.

Set cloud-clonfig value "node-volume-attach-limit" to 1-255 and it will be 256.

Easiest fix would be just add check to the GetMaxVolLimit function if global variable cfg is initialized. 

For example something like this:
```
func GetMaxVolLimit() int64 {
  var err error
  if cfg.Global.AuthUrl == "" {
    cfg, _, err = GetConfigFromFile(configFile)
    if err != nil {
      klog.Errorf("Couldn't load config file!")
    }
  }
  if cfg.BlockStorage.NodeVolumeAttachLimit > 0 && cfg.BlockStorage.NodeVolumeAttachLimit <= 256 {
    return cfg.BlockStorage.NodeVolumeAttachLimit
  }
  return maxVol
}
```
Variable `cfg.Global.AuthUrl` is used to check initialization status as you can not have functional CSI it that value have not been initialized. 

**Environment**:
Cinder CSI master

/kind bug

ramineni:
Values of BlockStorageOpts set in config file are not getting
reflected as of now. This PR is to fix the same

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #633

**Release note**:

```release-note
```

### DIFF
--- a/pkg/csi/cinder/openstack/openstack.go
+++ b/pkg/csi/cinder/openstack/openstack.go
@@ -64,11 +64,13 @@ type IOpenStack interface {
 	WaitSnapshotReady(snapshotID string) error
 	GetInstanceByID(instanceID string) (*servers.Server, error)
 	ExpandVolume(volumeID string, size int) error
+	GetMaxVolLimit() int64
 }
 
 type OpenStack struct {
 	compute      *gophercloud.ServiceClient
 	blockstorage *gophercloud.ServiceClient
+	bsOpts       BlockStorageOpts
 }
 
 type BlockStorageOpts struct {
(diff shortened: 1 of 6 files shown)


pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -250,6 +250,16 @@ func (os *OpenStack) GetAttachmentDiskPath(instanceID, volumeID string) (string,
 	return "", fmt.Errorf("volume %s has no ServerId", volumeID)
 }
 
+//GetMaxVolLimit returns max vol limit
+func (os *OpenStack) GetMaxVolLimit() int64 {
+	if os.bsOpts.NodeVolumeAttachLimit > 0 && os.bsOpts.NodeVolumeAttachLimit <= 256 {
+		return os.bsOpts.NodeVolumeAttachLimit
+	}
+
+	return maxVol
```
jichenjc: since you are changing this, can we rename this param? maxVol to maxNodeVolumeAttach or simiar?

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -81,16 +83,17 @@ type Config struct {
...
+	klog.Infof("Username: %s", cfg.Global.Username)
+	klog.Infof("UserId: %s", cfg.Global.UserId)
+	klog.Infof("TenantId: %s", cfg.Global.TenantId)
+	klog.Infof("TenantName: %s", cfg.Global.TenantName)
+	klog.Infof("DomainName: %s", cfg.Global.DomainName)
+	klog.Infof("DomainId: %s", cfg.Global.DomainId)
+	klog.Infof("TrustID: %s", cfg.Global.TrustID)
+	klog.Infof("Region: %s", cfg.Global.Region)
+	klog.Infof("CAFile: %s", cfg.Global.CAFile)
+	klog.Infof("Block storage opts: %v", cfg.BlockStorage)
```
jichenjc: do we need any doc update for this block storage opts?

ramineni: Its already there

pkg/csi/cinder/openstack/openstack_volumes.go
ramineni: sure, will update it.

adisky:
Thanks for the fix!!
/approve

jichenjc:
/lgtm 
thanks

PR #693, opened 2019-07-11
Fix getStorageSizeInGiga

strigazi:
Corner cases like 1000G and 2000G were being translated
to 1G and 2G when sending to manila. Use ScaledValue to
round to Gigabytes.

Signed-off-by: Spyros Trigazis <spyridon.trigazis@cern.ch>

Corner cases like 1000G and 2000G were being translated
to 1G and 2G when sending to manila. 

reproducer:
```
package main

import (
	"fmt"
	"strconv"

	v1 "k8s.io/api/core/v1"
	"k8s.io/apimachinery/pkg/api/resource"
	"k8s.io/client-go/kubernetes/scheme"
)

var filecontent = `
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: deleteme
spec:
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: "1000G"
  storageClassName: meyrin-cephfs-hyperconverged
`

func main() {
	decode := scheme.Codecs.UniversalDeserializer().Decode
	obj, _, _ := decode([]byte(filecontent), nil, nil)


	pvc := obj.(*v1.PersistentVolumeClaim)
	storageSize, _ := pvc.Spec.Resources.Requests[v1.ResourceStorage]
	fmt.Println(storageSize)
	var buf []byte
	canonicalValue, _ := storageSize.AsScale(resource.Giga)
	fmt.Println("canonicalValue", canonicalValue)
	storageSizeAsByteSlice, _ := canonicalValue.AsCanonicalBytes(buf)
	fmt.Println("canonicalValue.AsCanonicalBytes(buf)", storageSizeAsByteSlice)
	res, _ := strconv.Atoi(string(storageSizeAsByteSlice))
	fmt.Println(res)
}
```

**Release note**:

```release-note
Fix getStorageSizeInGiga for manila provisioner
```

### DIFF
--- a/pkg/share/manila/util.go
+++ b/pkg/share/manila/util.go
@@ -18,7 +18,6 @@ package manila
 
 import (
 	"fmt"
-	"strconv"
 	"strings"
 
 	"github.com/gophercloud/gophercloud"
@@ -59,11 +58,7 @@ func getStorageSizeInGiga(pvc *v1.PersistentVolumeClaim) (int, error) {
 		return 0, fmt.Errorf("requested storage size must be greater than zero")
 	}
 
-	var buf []byte
-	canonicalValue, _ := storageSize.AsScale(resource.Giga)
-	storageSizeAsByteSlice, _ := canonicalValue.AsCanonicalBytes(buf)
-
-	return strconv.Atoi(string(storageSizeAsByteSlice))
+	return int(storageSize.ScaledValue(resource.Giga)), nil
 }
 
 // Chooses one ExportLocation according to the below rules:


adisky:
cc @gman0

gman0:
LGTM, thanks @strigazi ! At first I was unsure if the size is not supposed to be in GiB, but GB is actually correct https://developer.openstack.org/api-ref/shared-file-system/?expanded=create-share-detail#create-share

@adisky @dims

PR #704, opened 2019-07-19
csi-manila: roll-back storage resources on failure

gman0:
csi-manila currently leaks Manila resources if an error occurs during share/snapshot provisioning, they're simply left on the backend without any sort of clean up.

This PR first determines the type of the error that has occurred using the Status field of the share/snapshot and [User messages API](https://developer.openstack.org/api-ref/shared-file-system/#user-messages-since-api-2-37) (better error reporting, easier troubleshooting for the users!) and then attempts to clean up the failed resource.

User messages API requires minimum Manila microversion 2.37 (Queens)

**Release note**:

```release-note
CSI Manila now requires Manila version >=2.37
```

### DIFF
--- a/pkg/csi/manila/controllerserver.go
+++ b/pkg/csi/manila/controllerserver.go
@@ -262,8 +262,20 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		readyToUse = false
 	case snapshotAvailable:
 		readyToUse = true
+	case snapshotError:
+		manilaErrMsg, err := lastResourceError(snapshot.ID, manilaClient)
+		if err != nil {
+			res.err = fmt.Errorf("snapshot %s of share %s is in error state, error description could not be retrieved: %v", snapshot.ID, req.GetSourceVolumeId(), err)
+			return nil, status.Error(codes.Internal, res.err.Error())
+		}
+
+		// An error occurred, try to roll-back the snapshot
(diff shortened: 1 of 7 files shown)


pkg/csi/manila/controllerserver.go
```diff
@@ -262,8 +262,20 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
 		readyToUse = true
+	case snapshotError:
+		manilaErrMsg, err := lastResourceError(snapshot.ID, manilaClient)
+		if err != nil {
+			res.err = fmt.Errorf("snapshot %s of share %s is in error state, error description could not be retrieved: %v", snapshot.ID, req.GetSourceVolumeId(), err)
+			return nil, status.Error(codes.Internal, res.err.Error())
+		}
+
+		// An error occurred, try to roll-back the snapshot
+		tryDeleteSnapshot(snapshot, manilaClient)
```
tsmetana: I'm not quite sure about this piece... The `tryDeleteSnapshot` uses exponential backoff -- this means the call to CreateSnapshot might time-out. Wouldn't it be better to run the `tryDeleteSnapshot` in a goroutine so the caller gets an error immediately and the driver attempts to delete the broken snapshot in backgground?

tsmetana: We have clarified this with @gman0 on Slack: The timeouts should not cause problems in this case: the CO would retry until some other error is returned or the operation succeeds.

PR #708, opened 2019-07-29
Add CSIDriver object

adisky:
CSIDriver objects needs to created after v1.14 as
cluster-driver-registrar is deprecated

**Release note**:

```release-note
NONE.
```

### DIFF
--- a/docs/using-cinder-csi-plugin.md
+++ b/docs/using-cinder-csi-plugin.md
@@ -49,6 +49,31 @@ csi-cinder-controllerplugin         4/4     Running   0          29h
 csi-cinder-nodeplugin               2/2     Running   0          46h
 ```
 
+you can get information about CSI Drivers running in a cluster, using **CSIDriver** object    
+
+```
+$ kubectl get csidrivers.storage.k8s.io
+NAME                       CREATED AT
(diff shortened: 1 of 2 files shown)


ramineni:
@adisky I suppose we were not using cluster-driver-registrar before (1.14) also right? so it should not affect if its deprecated?

adisky:
> @adisky I suppose we were not using cluster-driver-registrar before (1.14) also right? so it should not affect if its deprecated?

@ramineni yes we are not affected by this, i have added it as it is good to know what drivers are running in a k8s cluster, from user's perspective, moreover it is required for inline volume support

ramineni:
> > @adisky I suppose we were not using cluster-driver-registrar before (1.14) also right? so it should not affect if its deprecated?
> 
> @ramineni yes we are not affected by this, i have added it as it is good to know what drivers are running in a k8s cluster, from user's perspective, moreover it is required for inline volume support

@adisky ok, could we also update the documentation for the same .

adisky:
@ramineni Done!!, doc updated

ramineni:
Thanks. 
/lgtm

jichenjc:
looks good~

Issue #714, opened 2019-08-02
[Manila CSI] add csi-sanity tests


gman0:
This is a tracker ticket for PRs needed to get csi-sanity tests in csi-manila.

gman0:
3 pending/to-be PRs:
- [x] add an abstract interface for the Manila client, move it into a separate package, change all references from `gophercloud.ServiceClient` to `manilaclient.Interface`:
**feature branch** https://github.com/gman0/cloud-provider-openstack/tree/csimanila-manilaclient
**PR** https://github.com/kubernetes/cloud-provider-openstack/pull/716
- [ ] fix the driver in accordance with the CSI specification in order to pass the csi-sanity test suite:
**feature branch** https://github.com/gman0/cloud-provider-openstack/tree/csimanila-csi-spec-compliance-fix
- [ ] implement fake Manila client based on the `manilaclient.Interface` interface, add sanity tests:
**feature branch** https://github.com/gman0/cloud-provider-openstack/tree/csimanila-sanity

Even though the latter two bullet points don't have a PR yet, they are finished and ready to go ASAP - as soon as the PR that's currently pending is finished with reviews and gets merged.

PR #716, opened 2019-08-06
[csi-manila] Manila client moved into a separate package

gman0:

Part of https://github.com/kubernetes/cloud-provider-openstack/issues/714

This PR:
1. moves the Manila client into a separate package. It implements `manilaclient.Interface` interface, making it possible to create other implementations of the client - fake Manila client in particular - a prerequisite for unit/sanity tests,
2. processing of cmd args was broken: changes introduced by #641 caused a bunch of cmd flags to not be recognized anymore. We're now using the Cobra library which solves the issue.

```release-note
NONE
```


gman0:
PTAL @tsmetana @gouthampacha @tombarron

cmd/manila-csi-plugin/main.go
```diff
@@ -17,38 +17,87 @@ limitations under the License.
...
+		},
 	}
 
-	manilaClientBuilder := &manilaclient.ClientBuilder{}
+	cmd.Flags().AddGoFlagSet(flag.CommandLine)
+
+	cmd.PersistentFlags().StringVar(&endpoint, "endpoint", "unix://tmp/csi.sock", "CSI endpoint")
+	cmd.MarkPersistentFlagRequired("endpoint")
+
+	cmd.PersistentFlags().StringVar(&driverName, "drivername", "manila.csi.openstack.org", "name of the driver")
```
adisky: shouldn't be there a default name supported here??

cmd/manila-csi-plugin/main.go
```diff
@@ -17,38 +17,87 @@ limitations under the License.
...
+	cmd.MarkPersistentFlagRequired("drivername")
+
+	cmd.PersistentFlags().StringVar(&nodeID, "nodeid", "", "this node's ID")
+	cmd.MarkPersistentFlagRequired("nodeid")
+
+	cmd.PersistentFlags().StringVar(&protoSelector, "share-protocol-selector", "", "specifies which Manila share protocol to use")
```
adisky: same here a default protocol??

cmd/manila-csi-plugin/main.go
gman0: The name defaults to `manila.csi.openstack.org`...is that ok or do you mean something else?

cmd/manila-csi-plugin/main.go
gman0: Protocol selector heavily depends on how Manila is configured and deployed and this in turn dictates what goes into `--fwdendpoint`...IMHO there's too many possibilities, making any assumptions on what should be the default would probably end up being wrong anyway. I think the safest thing to do here is to make the users explicitly set which share protocol they're interested in.

tombarron: I agree that we shouldn't set a default protocol selector, i.e. we should force the deployer to choose.  That requires some thought up front on the part of the deployer but will in the end be much less confusing than doing an apparently successful deployment and then discovering at time of PVC fulfillment that it does not work as intended.

adisky: got it, but can we specify some valid values here?? and have a validity check here only??

cmd/manila-csi-plugin/main.go
adisky: sorry missed it :)

cmd/manila-csi-plugin/main.go
gman0: @adisky fixed in https://github.com/kubernetes/cloud-provider-openstack/pull/716/commits/aaeb55134a5a67aa8b66f992ec77d191183d82fb

Also, I'm not too familiar with Cobra, but apparently even if a flag has a default value, it must not be marked as required and left unspecified - otherwise it complains that a required flag is missing (which, on second though, makes sense) => I've removed `MarkPersistentFlagRequired` for flags that have default values specified.

tsmetana:
The CI failures don't look to be related to the code and seem quite random. I think it's OK to merge now.

/lgtm

gman0:
@tsmetana thanks! @adisky PTAL, thank you!

gman0:
Also @ramineni, could you please take a look if this seems ok?

cmd/manila-csi-plugin/main.go
```diff
@@ -17,33 +17,103 @@ limitations under the License.
...
+				f2 := klogFlags.Lookup(f1.Name)
+				if f2 != nil {
+					value := f1.Value.String()
+					f2.Value.Set(value)
+				}
+			})
+		},
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := validateShareProtocolSelector(protoSelector); err != nil {
+				klog.Fatalf(err.Error())
```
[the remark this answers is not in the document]
gman0: `klog.Fatalf` exits with 255

https://github.com/kubernetes/klog/blob/3ca30a56d8a775276f9cdae009ba326fdc05af7f/klog.go#L1280-L1285

https://github.com/kubernetes/klog/blob/3ca30a56d8a775276f9cdae009ba326fdc05af7f/klog.go#L822-L847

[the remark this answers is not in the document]
ramineni: great , then sounds good :)

PR #721, opened 2019-08-13
Wrap Errors Cinder CSI

### LINKED ISSUE #717: CSI driver should wrap all errors
/kind bug

Looking at the CSI driver code, it often returns naked `fmt.Errorf()` or `errors.New()` to gRPC as CSI response.

All errors should be wrapped into `status.Error(codes.XXX, err.Error())`. CSI spec explicitly enumerates what errors can be returned from each call and in what cases.

adisky:
Wrapped all CSI errors to status.Error()

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #717 

None.

**Release note**:

```release-note
NONE.
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -17,7 +17,6 @@ limitations under the License.
 package cinder
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/golang/protobuf/ptypes"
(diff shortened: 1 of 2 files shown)


ramineni:
Thanks.
/lgtm

PR #723, opened 2019-08-14
Implement ValidateVolumeCapabilities

### LINKED ISSUE #636: [Cinder CSI] Implement `ValidateVolumeCapabilities`

/kind feature

**What happened**:
Implement `ValidateVolumeCapabilities`.
It is required and causing sanity failures.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

adisky:
Implement Volume Capabilities

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #636 
Currently we are advertising one mode only, `csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER`, so we are validating only this capability in `ValidateVolumeCapabilities`, but cinder can support multi attach in RO/RW mode, we need to add support for this later.

**Release note**:

```release-note
```
`NONE`

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -432,7 +432,44 @@ func (cs *controllerServer) ControllerGetCapabilities(ctx context.Context, req *
 }
 
 func (cs *controllerServer) ValidateVolumeCapabilities(ctx context.Context, req *csi.ValidateVolumeCapabilitiesRequest) (*csi.ValidateVolumeCapabilitiesResponse, error) {
-	return nil, status.Error(codes.Unimplemented, fmt.Sprintf("ValidateVolumeCapabilities is not yet implemented"))
+
+	reqVolCap := req.GetVolumeCapabilities()
+
+	if reqVolCap == nil || len(reqVolCap) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "ValidateVolumeCapabilities Volume Capabilities must be provided")
+	}
+	volumeID := req.GetVolumeId()
+
+	if len(volumeID) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "ValidateVolumeCapabilities Volume ID must be provided")
+	}
+
+	_, err := cs.Cloud.GetVolume(volumeID)
+	if err != nil {
+		if cpoerrors.IsNotFound(err) {
+			return nil, status.Error(codes.NotFound, fmt.Sprintf("ValidateVolumeCapabiltites Volume %s not found", volumeID))
+		}
+		return nil, status.Error(codes.Internal, fmt.Sprintf("ValidateVolumeCapabiltites %v", err))
+	}
+
+	for _, cap := range reqVolCap {
+		if cap.GetAccessMode().GetMode() != cs.Driver.vcap[0].Mode {
+			return &csi.ValidateVolumeCapabilitiesResponse{Message: "Requested Volume Capabilty not supported"}, nil
+		}
+	}
+
+	// Cinder CSI driver currently supports one mode only
+	resp := &csi.ValidateVolumeCapabilitiesResponse{
+		Confirmed: &csi.ValidateVolumeCapabilitiesResponse_Confirmed{
+			VolumeCapabilities: []*csi.VolumeCapability{
+				{
+					AccessMode: cs.Driver.vcap[0],
+				},
+			},
+		},
+	}
+
+	return resp, nil
 }
 
 func (cs *controllerServer) GetCapacity(ctx context.Context, req *csi.GetCapacityRequest) (*csi.GetCapacityResponse, error) {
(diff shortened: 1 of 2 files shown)


pkg/csi/cinder/controllerserver.go
```diff
@@ -415,7 +415,36 @@ func (cs *controllerServer) ControllerGetCapabilities(ctx context.Context, req *
...
+	volumeID := req.GetVolumeId()
+
+	if len(volumeID) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "ValidateVolumeCapabilities Volume ID must be provided")
+	}
+
+	_, err := cs.Cloud.GetVolume(volumeID)
+	if err != nil {
+		if cpoerrors.IsNotFound(err) {
+			return nil, status.Error(codes.NotFound, "ValidateVolumeCapabiltites Volume not found")
```
jichenjc: should we add the volume ID here?

jichenjc: others looks good

ramineni:
@adisky Thanks for the PR. Could we add unit test also for this.
And also, Issue ref seems to be missing in description, please update that as well.  https://github.com/kubernetes/cloud-provider-openstack/issues/636

pkg/csi/cinder/controllerserver.go
hogepodge: Being a bit more verbose on the problematic ID throughout seems like a good bet.

adisky: Done

adisky:
> @adisky Thanks for the PR. Could we add unit test also for this.
> And also, Issue ref seems to be missing in description, please update that as well. #636

@ramineni Done

pkg/csi/cinder/controllerserver.go
```diff
@@ -415,7 +415,37 @@ func (cs *controllerServer) ControllerGetCapabilities(ctx context.Context, req *
...
+	if err != nil {
+		if cpoerrors.IsNotFound(err) {
+			return nil, status.Error(codes.NotFound, fmt.Sprintf("ValidateVolumeCapabiltites Volume %s not found", volumeID))
+		}
+		return nil, status.Error(codes.Internal, fmt.Sprintf("ValidateVolumeCapabiltites %v", err))
+	}
+
+	// Cinder CSI driver currently supports one mode only
+	resp := &csi.ValidateVolumeCapabilitiesResponse{
+		Confirmed: &csi.ValidateVolumeCapabilitiesResponse_Confirmed{
```
ramineni: @adisky This should return confirmed only if all the volume capabilities specified in the request are supported , but here , I dont see given request is not validated at all. Could you clarify , am i missing something?

adisky: @ramineni list of supported capabilities to be returned in `confirmed` , since we are supporting only one capability, i am returning it in confirmed, ya i missed the check with request, will update the PR

adisky:
@ramineni PTAL, updated PR

pkg/csi/cinder/controllerserver.go
```diff
@@ -415,7 +415,42 @@ func (cs *controllerServer) ControllerGetCapabilities(ctx context.Context, req *
...
+
+	_, err := cs.Cloud.GetVolume(volumeID)
+	if err != nil {
+		if cpoerrors.IsNotFound(err) {
+			return nil, status.Error(codes.NotFound, fmt.Sprintf("ValidateVolumeCapabiltites Volume %s not found", volumeID))
+		}
+		return nil, status.Error(codes.Internal, fmt.Sprintf("ValidateVolumeCapabiltites %v", err))
+	}
+
+	for _, cap := range req.VolumeCapabilities {
```
ramineni: it should be either req.GetVolumeCapabilities() or req.volume_capabilities ?

pkg/csi/cinder/controllerserver.go
```diff
@@ -415,7 +415,42 @@ func (cs *controllerServer) ControllerGetCapabilities(ctx context.Context, req *
 }
 
 func (cs *controllerServer) ValidateVolumeCapabilities(ctx context.Context, req *csi.ValidateVolumeCapabilitiesRequest) (*csi.ValidateVolumeCapabilitiesResponse, error) {
-	return nil, status.Error(codes.Unimplemented, fmt.Sprintf("ValidateVolumeCapabilities is not yet implemented"))
+
+	if req.GetVolumeCapabilities() == nil || len(req.GetVolumeCapabilities()) == 0 {
```
ramineni: :nit: storing in variable would have been good, as its used multiple times below also

ramineni:
Thanks.
/lgtm

jichenjc:
@adisky need rebase ... others look good

PR #728, opened 2019-08-17
[csi-manila] CSI client moved into a separate package

gman0:

Part of https://github.com/kubernetes/cloud-provider-openstack/issues/714

Continuing in an effort of integrating the csi-sanity test suite into CSI Manila, this is the second patch out of 4:

This PR moves the CSI client into a separate package, which brings `Node`, `Identity` and `Builder` interfaces. This in turn makes it possible to have various implementations for CSI clients - e.g. creating a fake CSI client which is a prerequisite for sanity/unit tests.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/cmd/manila-csi-plugin/main.go
+++ b/cmd/manila-csi-plugin/main.go
@@ -25,6 +25,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"k8s.io/cloud-provider-openstack/pkg/csi/manila"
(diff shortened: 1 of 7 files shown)


adisky:
@gman0 is it good to go?? 
cc @tombarron @gouthampacha @tsmetana

gman0:
@adisky finished and ready for reviews :)

tsmetana:
It looks OK to me: as the description mentions -- this is only new interfaces that implement the existing logic. Most of the code has only been moved.
/lgtm

PR #730, opened 2019-08-20
[csi-manila] fixed incorrect behaviour discovered by csi-sanity test suite

gman0:

Part of #714

Continuing in an effort of integrating the csi-sanity test suite into CSI Manila, this is the third patch out of 4:

While testing the driver against the csi-sanity test suite, following errors were found:

* missing ValidateVolumeCapabilities in Controller Plugin
* validateNode{Unstage,Unpublish}Request functions were missing checks for empty staging/target paths
* verifySnapshotCompatibility's check for source volume ID was incorrect

this PR fixes the issues mentioned above and cleans up the codebase a bit

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/csi/manila/controllerserver.go
+++ b/pkg/csi/manila/controllerserver.go
@@ -230,7 +230,7 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 
 	if sourceShare, res.err = manilaClient.GetShareByID(req.GetSourceVolumeId()); res.err != nil {
 		if isManilaErrNotFound(res.err) {
-			return nil, status.Errorf(codes.InvalidArgument, "failed to create a snapshot (%s) for share %s because the share doesn't exist: %v", req.GetName(), req.GetSourceVolumeId(), err)
+			return nil, status.Errorf(codes.NotFound, "failed to create a snapshot (%s) for share %s because the share doesn't exist: %v", req.GetName(), req.GetSourceVolumeId(), err)
 		}
 
 		return nil, status.Errorf(codes.Internal, "failed to retrieve source share %s when creating a snapshot (%s): %v", req.GetSourceVolumeId(), req.GetName(), res.err)
@@ -251,7 +251,7 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		}
 
 		if isManilaErrNotFound(res.err) {
-			return nil, status.Errorf(codes.InvalidArgument, "failed to create a snapshot (%s) for share %s because the share doesn't exist: %v", req.GetName(), req.GetSourceVolumeId(), err)
+			return nil, status.Errorf(codes.NotFound, "failed to create a snapshot (%s) for share %s because the share doesn't exist: %v", req.GetName(), req.GetSourceVolumeId(), err)
 		}
 
 		return nil, status.Errorf(codes.Internal, "failed to create a snapshot (%s) of share %s: %v", req.GetName(), req.GetSourceVolumeId(), res.err)
@@ -378,7 +378,7 @@ func (cs *controllerServer) ValidateVolumeCapabilities(ctx context.Context, req 
 	}
 
 	if share.Status != shareAvailable {
-		return nil, status.Errorf(codes.InvalidArgument, "share %s is in an unexpected state: wanted %s, got %s", share.ID, shareAvailable, share.Status)
+		return nil, status.Errorf(codes.Internal, "share %s is in an unexpected state: wanted %s, got %s", share.ID, shareAvailable, share.Status)
 	}
 
 	if !compareProtocol(share.ShareProto, cs.d.shareProto) {
--- a/pkg/csi/manila/controllerserver.go
+++ b/pkg/csi/manila/controllerserver.go
@@ -229,6 +229,10 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 	// Retrieve the source share
 
 	if sourceShare, res.err = manilaClient.GetShareByID(req.GetSourceVolumeId()); res.err != nil {
+		if isManilaErrNotFound(res.err) {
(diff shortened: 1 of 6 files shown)


gman0:
PTAL @tsmetana @gouthampacha @tombarron @adisky

pkg/csi/manila/controllerserver.go
```diff
@@ -229,6 +229,10 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 	// Retrieve the source share
 
 	if sourceShare, res.err = manilaClient.GetShareByID(req.GetSourceVolumeId()); res.err != nil {
+		if isManilaErrNotFound(res.err) {
+			return nil, status.Errorf(codes.InvalidArgument, "failed to create a snapshot (%s) for share %s because the share doesn't exist: %v", req.GetName(), req.GetSourceVolumeId(), err)
```
adisky: `codes.InvalidArgument`? shouldn't it be` codes.NotFound`

pkg/csi/manila/controllerserver.go
```diff
@@ -246,10 +250,14 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 			return nil, status.Errorf(codes.DeadlineExceeded, "deadline exceeded while waiting for snapshot %s of share %s to become available", snapshot.ID, req.GetSourceVolumeId())
 		}
 
+		if isManilaErrNotFound(res.err) {
```
adisky: ditto

pkg/csi/manila/controllerserver.go
```diff
@@ -332,7 +340,58 @@ func (cs *controllerServer) ControllerGetCapabilities(ctx context.Context, req *
...
+	if err != nil {
+		if isManilaErrNotFound(err) {
+			return nil, status.Errorf(codes.NotFound, "share %s not found: %v", req.GetVolumeId(), err)
+		}
+
+		return nil, status.Errorf(codes.Internal, "failed to retrieve share %s: %v", req.GetVolumeId(), err)
+	}
+
+	if share.Status != shareAvailable {
+		return nil, status.Errorf(codes.InvalidArgument, "share %s is in an unexpected state: wanted %s, got %s", share.ID, shareAvailable, share.Status)
```
adisky: here also, shouldn't it be codes.Internal?

adisky: InvalidArgument, is sent when something is missing in request from CO or request arguments not adhering to CSI specs

gman0: Actually, `FAILED_PRECONDITION` or `UNAVAILABLE` seem to come even closer to describing this particular error. The status mismatch might be just a transient error (i.e. when the share is still in `creating` state).

https://github.com/grpc/grpc-go/blob/3bb34e5324978dca98dea479160fe5f434d0f960/codes/codes.go#L78-L97

gman0:
@adisky thank you for your feedback! I've left a comment on your review above but it doesn't make all that much of a difference so i've worked in all the changes you suggested.

adisky:
> @adisky thank you for your feedback! I've left a comment on your review above but it doesn't make all that much of a difference so i've worked in all the changes you suggested.

@gman0 actually based on return error codes CO performs next actions like retry/abort

gman0:
@adisky yup i know but i'd imagine `INTERNAL` and `FAILED_PRECONDITION` would have similar behavior (i.e. abort), likewise `INVALID_ARGUMENT` and `UNAVAILABLE` (retry when ready)...leaning on the conservative side, abort seems safer (internal/failed precondition)

gman0:
Although an explicit status check could be added

gman0:
@adisky PTAL

gman0:
When validating share/snapshot states, transient `creating` status will now cause [`UNAVAILABLE` error code](https://github.com/grpc/grpc-go/blob/3bb34e5324978dca98dea479160fe5f434d0f960/codes/codes.go#L133-L140), any other share/snapshot status will result in [`FAILED_PRECONDITION` error code](https://github.com/grpc/grpc-go/blob/3bb34e5324978dca98dea479160fe5f434d0f960/codes/codes.go#L78-L97).

Note that invalid states detected during share/snapshot creation are still considered as `INTERNAL` errors.

pkg/csi/manila/util.go
```diff
@@ -176,6 +177,26 @@ func lastResourceError(resourceID string, manilaClient manilaclient.Interface) (
 	return manilaErrorMessage{message: "unknown error"}, nil
 }
 
+func isManilaErrNotFound(err error) bool {
```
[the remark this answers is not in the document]
gman0: Thanks for pointing this out! Will update

pkg/csi/manila/nodeserver.go
```diff
@@ -240,12 +244,12 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
+		return nil, status.Errorf(codes.InvalidArgument, "invalid volume context: %v", err)
 	}
 
 	osOpts, err := options.NewOpenstackOptions(req.GetSecrets())
 	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "invalid volume secret: %v", err)
+		return nil, status.Errorf(codes.InvalidArgument, "invalid OpenStack secrets: %v", err)
 	}
 
 	volID := volumeID(req.GetVolumeId())
```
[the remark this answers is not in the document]
gman0: It's already being checked in `buildVolumeContext` function:
https://github.com/gman0/cloud-provider-openstack/blob/4d0ef767cfb92359a7e42e2ecb7eb70e8d55caf7/pkg/csi/manila/nodeserver.go#L49-L71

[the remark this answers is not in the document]
gman0: ...which returns `INVALID_ARGUMENT` if it fails to retrieve the share - this is of course not correct. I'll fix this as well

PR #748, opened 2019-09-04
Allow cinder volume live resize

kayrus:

Add capability to use Cinder V3 online resize feature: https://docs.openstack.org/cinder/rocky/contributor/api_microversion_history.html#id40

Cinder CSI plugin already introduced this feature, however not in the built-in plugin.

* You have to enable `allowVolumeExpansion: true` in the storage class.
* You have to enable the `ExpandInUsePersistentVolumes=true` feature gate in the kube-controller-manager

This PR depends on https://github.com/kubernetes/kubernetes/issues/82319, which allows kubelet to poll new block device geometry note**:

```release-note
Built-in Cinder plugin allows to resize in-use volumes
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack_volumes.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_volumes.go
@@ -236,7 +236,16 @@ func (volumes *VolumesV3) expandVolume(volumeID string, newSize int) error {
 	createOpts := volumeexpand.ExtendSizeOpts{
 		NewSize: newSize,
 	}
+
+	// save initial microversion
+	mv := volumes.blockstorage.Microversion
+	volumes.blockstorage.Microversion = "3.42"
+
 	err := volumeexpand.ExtendSize(volumes.blockstorage, volumeID, createOpts).ExtractErr()
+
+	// restore initial microversion
+	volumes.blockstorage.Microversion = mv
+
 	timeTaken := time.Since(startTime).Seconds()
 	recordOpenstackOperationMetric("expand_volume", timeTaken, err)
 	return err
@@ -346,10 +355,6 @@ func (os *OpenStack) ExpandVolume(volumeID string, oldSize resource.Quantity, ne
 	if err != nil {
 		return oldSize, err
 	}
-	if volume.Status != volumeAvailableStatus {
-		// cinder volume can not be expanded if its status is not available
-		return oldSize, fmt.Errorf("volume status is not available")
-	}
 
 	// Cinder works with gigabytes, convert to GiB with rounding up
 	volSizeGiB, err := volumeutil.RoundUpToGiBInt(newSize)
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/providers/openstack/openstack_volumes.go
```diff
@@ -368,6 +373,17 @@ func (os *OpenStack) ExpandVolume(volumeID string, oldSize resource.Quantity, ne
 		return oldSize, err
 	}
 
+	if volume.Status != volumeAvailableStatus {
+		if _, ok := volumes.(*VolumesV3); !ok {
+			// cinder volume can not be expanded if cinder API is not v3
+			return oldSize, fmt.Errorf("volume status is not available")
+		}
+		if volume.Status != volumeInUseStatus {
```
jichenjc: why you don't check at line 376?

pkg/cloudprovider/providers/openstack/openstack_volumes.go
```diff
@@ -236,7 +236,16 @@ func (volumes *VolumesV3) expandVolume(volumeID string, newSize int) error {
 	createOpts := volumeexpand.ExtendSizeOpts{
 		NewSize: newSize,
 	}
+
+	// save initial microversion
+	mv := volumes.blockstorage.Microversion
```
jichenjc: just curious, why we can't have a client with latest version so that we can always call this kind of version specific feature?

pkg/cloudprovider/providers/openstack/openstack_volumes.go
kayrus: On 376 we're checking the volumeAvailableStatus status. If it is available, then it doesn't matter whether it is V2 or V3. If the status is not available, then we're checking two things:
* if it is not V3, then fail
* if it is V3, then the status must be volumeInUseStatus

pkg/cloudprovider/providers/openstack/openstack_volumes.go
kayrus: because Cinder can have an old version and if we permanently set the microversion, then all further operations using the same client will fail.

jichenjc: yes, I fully understand the microversion concept
I mean, this kind of 3.42 hard coding why we not have a latest client object and 
use that object for *all* future features like this instead of hard code every where

kayrus: It is simpler to set the required microversion for an operation rather than set the latest one for all operations, because it can break stuff, which was available in lower microversions.

ramineni: @jichenjc @kayrus Even we have latest version, unless specific microversion is set , that feature is not enabled , In this case unless we set 3.42 microversion, live resize is not enabled

jichenjc: thanks, I am not aware that, at this is different to what I learn from nova side.. thanks for clarify this

ramineni:
Thanks. Looks good. Unless dependent PR is merged, live resize won't work properly right, so keeping hold until that PR is merged

kayrus:
@ramineni I'd not wait for the https://github.com/kubernetes/kubernetes/pull/82319, since it only complements this PR.

ramineni:
> @ramineni I'd not wait for the [kubernetes/kubernetes#82319](https://github.com/kubernetes/kubernetes/pull/82319), since it only complements this PR.

Ah, But I didnt face any issue while testing csi live resize , why the change is needed, where you were facing the issue?

/hold cancel

kayrus:
@ramineni I face this issue in Vmware hypervisors.

ramineni:
> @ramineni I face this issue in Vmware hypervisors.

ok.These changes seems ok to me.
/lgtm

kayrus:
@ramineni what is the merging process? I have 4 PRs pending and wonder whether they can be merged before the 1.6.x release.

jichenjc:
/lgtm

@adisky maybe you can help on this?

adisky:
@kayrus please rebase your PR, we would merge it before v1.16

kayrus:
@adisky why do you need to rebase this PR? There are no conflicts with the base branch.

adisky:
@kayrus we have disabled Octavia CI job, I would rebase it wont run on your PR, merging is blocked until all the test passes

kayrus:
@adisky  rebased

PR #756, opened 2019-09-11
Add info about v2 VPI not support in CSI

jichenjc:
Add no support in CSI cinder on cinder v2 API 

**Release note**:

```release-note
```

### DIFF
--- a/docs/provider-configuration.md
+++ b/docs/provider-configuration.md
@@ -45,6 +45,8 @@ Identity v2 API.
 
 ‡ Block Storage v3 API support was added in Kubernetes 1.9.
 
+  *NOTE*: Cinder CSI Supports Block Storage V3 API only.
+
 § Load Balancing v1 API support was removed in Kubernetes 1.9.
 
 Service discovery is achieved by listing the service catalog managed by


docs/provider-configuration.md
```diff
@@ -44,6 +44,7 @@ a future release. As of the "Queens" release, OpenStack no longer exposes the
 Identity v2 API.
 
 ‡ Block Storage v3 API support was added in Kubernetes 1.9.
+  For CSI cinder enablement, only cinder v3 API is supported currently.
```
adisky: NOTE: Cinder CSI Supports Block Storage V3 API only

docs/provider-configuration.md
```diff
@@ -44,6 +44,7 @@ a future release. As of the "Queens" release, OpenStack no longer exposes the
 Identity v2 API.
 
 ‡ Block Storage v3 API support was added in Kubernetes 1.9.
+  *NOTE*: Cinder CSI Supports Block Storage V3 API only.
```
ramineni: Isn't it better to add this info in cinder csi plugin doc https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/using-cinder-csi-plugin.md ?

And also formatting seems wrong (it is coming in same line as L46),  I guess there should be blank line above i guess .

adisky: we can add this in CSI docs also, but many people land on this page for provider configuration and assumes the same configuration for intree/CSI driver.

ramineni: Looks like we need to do some refactoring here,  for now this PR should be ok once formatting is fixed. Thanks

jichenjc: @ramineni updated just now

ramineni:
Thanks
/lgtm

Issue #741, opened 2019-09-02
Failed to get Barbican client Post x509: certificate signed by unknown authority


sfxworks:
/kind bug

When attempting to deploy the Barbican KMS container on the cluster while setting appropriate configurations in place, I get an error stating that the cert is invalid. However, our endpoint at  https://cloud.lax1.imhcloud.net:5000/v3/auth/tokens is not self-signed.

Likely similar to https://github.com/kubernetes/cloud-provider-openstack/issues/499 which was resolved by https://github.com/kubernetes/cloud-provider-openstack/pull/500 ?

**What you expected to happen**:
The Barbican KMS plugin to recognize that the OpenStack auth endpoint isn't self-signed and continue posting the secret.

**How to reproduce it (as minimally and precisely as possible)**:
1. Have an endpoint that isn't self-signed
2. Setup barbican
3. Post a secret

**Environment**:
- kubeadm + daemonset from build
```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: barbician-kms
  namespace: kube-system
  labels:
    component: kube-apiserver
    tier: control-plane
spec:
  selector:
    matchLabels:
      name: barbician-kms
  template:
    metadata:
      labels:
        name: barbician-kms
    spec:
      nodeSelector:
        role: master
      containers:
      - name: barbician-kms
        command: ["/bin/sh"]
        args: ["-c", "/bin/barbican-kms-plugin --socketpath /var/lib/kms/kms.sock --cloud-config /etc/kubernetes/cloud-kms.conf"]
        image: "git5.zswap.net:4567/quantomworks/barbican"
        volumeMounts:
        - name: k8s
          mountPath: /etc/kubernetes
        - name: varlibkms
          mountPath: /var/lib/kms
          mountPropagation: Bidirectional
        securityContext:
          privileged: true
      priorityClassName: system-cluster-critical
      volumes:
      - name: k8s
        hostPath:
          path: /etc/kubernetes
      - name: varlibkms
        hostPath:
          path: /var/lib/kms

```

```
NAME="Ubuntu"
VERSION="18.04.3 LTS (Bionic Beaver)"
ID=ubuntu
PRETTY_NAME="Ubuntu 18.04.3 LTS"
VERSION_ID="18.04"
```
```
Linux controller-1 4.15.0-58-generic #64-Ubuntu SMP Tue Aug 6 11:12:41 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```
[...]

PR #762, opened 2019-09-16
Switch dependency manager to go modules

### LINKED ISSUE #529: Switch to gomodules for dependency management
With introduction of gomodules in from GO 1.11, we should switch to go modules for dependency management.

kayrus:

Switch the dependency manager to go modules

Resolves #529

**Release note**:

```release-note
Switch dependency manager to go modules
```


lingxiankong:
@kayrus Hi, I found you usually submit multiple commits in a single PR which is not recommended for upstream contribution, at least in openstack-provider. 

If it's a big task(e.g. affects all the binaries) you are working on, I'd suggest splitting the task into multiple sub-tasks and submit multiple PRs while associating those PRs using the same issue number. Otherwise, please just submit the PR containing only one single commit.

I know it's painful compared to Gerrit, but this is what we have.

kayrus:
@lingxiankong thanks for the note. From my side am still quite confused of your requirement, because github supports [squash and merge](https://help.github.com/en/articles/about-pull-request-merges). If you have concerns about squashed commit message(s) - the PR topic has to represent the squashed commit.

As for the current PR state. It looks like there should be some modifications on the openlab side. Probably the golang version has to be bumped to 1.13 as well. I tested the compilation process in the clean docker environment and using go modules I was able to compile the package without installing 3rd-party dependency management software.

lingxiankong:
> because github supports [squash and merge](https://help.github.com/en/articles/about-pull-request-merges). If you have concerns about squashed commit message(s) - the PR topic has to represent the squashed commit.

Maybe I missed something, I know I can do `squash and merge` commits for the PRs in my own repo, but I am not sure what's the default behavior in openstack-provider when a PR with multiple commits is approved and how it's going to be merged. @adisky do you know if we could do `squash and merge` when approving PRs?

adisky:
@linxiankong we support auto squash in this repo too
@kayrus thanks for your PR, ya need to update version in openlab also, let's wait for next cycle to get this PR in
/hold

lingxiankong:
> @linxiankong we support auto squash in this repo too

Sounds good, that solves my concerns. Thanks!

.zuul.yaml
```diff
@@ -78,7 +78,7 @@
         - cloud-provider-openstack-acceptance-test-e2e-conformance:
             files:
               - ^pkg/cloudprovider/.*
-              - ^Gopkg.toml$
```
ramineni: we could add go.sum also here?

go.mod
```diff
@@ -0,0 +1,87 @@
+module k8s.io/cloud-provider-openstack
+
+go 1.13
+
+require (
+	github.com/NYTimes/gziphandler v1.1.1 // indirect
```
ramineni: Is this file automatically generated or any manual changes of versions done?

Makefile
```diff
@@ -35,10 +34,6 @@ TAGS      :=
 LDFLAGS   := "-w -s -X 'k8s.io/cloud-provider-openstack/pkg/version.Version=${VERSION}'"
 REGISTRY ?= k8scloudprovider
 
-ifneq ("$(DEST)", "$(PWD)")
```
ramineni: we still need to run make from the base dir of project , why is this removed?

kayrus: I doesn't matter anymore. `go` automatically detects the `go.mod` inside the current directory and builds packages independently on the path.

go.mod
ramineni: Ok, could you clarify one thing (sorry , still getting hang on using gomod), how the master branch things taken care, like we need to update from time to time , running go mod tidy or something ? 
And one more thing, doesnt vendor directory need to created again? because previous vendor was based on dep

Makefile
ramineni: But for make its required , the below error is for make right, for make to run we need to be in base dir

go.mod
kayrus: With go modules go automatically downloads all the dependencies into the go system cache, there is no need to maintain the `vendor` directory. `go mod vendor` makes sense only for projects, which commit the `vendor` dir inside the repo.

Makefile
kayrus: `make` command is usually executed within the current directory. This check is not needed, besides it [confuses end users](https://github.com/kubernetes/cloud-provider-openstack/issues/741#issuecomment-537359670), when they want to build the binary from source

.zuul.yaml
kayrus: Added

go.mod
kayrus: If you want to update dependencies, then you usually set the new dependency version in the `go.mod` and run `go mod download`. It is also possible to run `go mod download %dep_url%@version`, but sometimes it doesn't work. Then you run `go mod tidy`

go.mod
```diff
@@ -0,0 +1,87 @@
...
+	k8s.io/klog v0.4.0
+	k8s.io/kube-openapi v0.0.0-20190918143330-0270cf2f1c1d // indirect
+	k8s.io/kubernetes v1.16.0
+	k8s.io/utils v0.0.0-20190801114015-581e00157fb1
+	sigs.k8s.io/sig-storage-lib-external-provisioner v0.0.0-20190807214443-c525773885fc
+)
+
+replace (
+	k8s.io/api => k8s.io/kubernetes/staging/src/k8s.io/api v0.0.0-20190913145653-2bd9643cee5b
+	k8s.io/apiextensions-apiserver => k8s.io/kubernetes/staging/src/k8s.io/apiextensions-apiserver v0.0.0-20190913145653-2bd9643cee5b
```
adisky: @kayrus how the constraints and override like below taken care here, sorry i am also new to go modules
https://github.com/kubernetes/cloud-provider-openstack/blob/master/Gopkg.toml#L79

adisky:
@kayrus please check why `pull-cloud-provider-openstack-check` is failing? rest all jobs have now fix in openlab zuul

Makefile
```diff
@@ -47,123 +42,111 @@ $(GOBIN):
 
 work: $(GOBIN)
 
-depend: work
```
zetaab: why depend is removed? Imo we should have way to update components like "previously" see my PR https://github.com/kubernetes/cloud-provider-openstack/pull/816 about depend things

zetaab:
also this is quite old PR already. Could you please rebase against master that we can see that everything will build

kayrus:
> also this is quite old PR already. Could you please rebase against master that we can see that everything will build

@zetaab hence it is old, it is rebased when needed. Last time I rebased it yesterday.

Makefile
adisky: @kayrus removing depend causing changes at many place, @zetaab in your PR too depend-update is not doing the same thing, for go modules i don't think there is a requirement for separate update command

adisky: @kayrus with your PR we not creating any vendor directory before build? i see there is no requirement but @zetaab you are doing this any particular reason for that?

hack/verify-import-boss.sh
```diff
@@ -22,4 +22,7 @@ if ! which import-boss > /dev/null; then
   exit 1
 fi
 
-import-boss -i k8s.io/cloud-provider-openstack/... --verify-only
+# prepare the vendor dir
+go mod vendor
```
adisky: why using  `go mod vendor` here??

kayrus: because it required by `import-boss` and only during tests. I don't know what is the purpose of using the `import-boss`

Makefile
kayrus: @adisky there is no requirement in using the vendor dir, except the `import-boss`  for tests. more details are here: https://github.com/golang/go/wiki/Modules#can-i-control-when-gomod-gets-updated-and-when-the-go-tools-use-the-network-to-satisfy-dependencies

zetaab: I did vendor folder because that is how I have always done things. However, it is same to me is there vendor folder or not. The main thing is to get this PR forward. :)

adisky: @zetaab I am ok with the changes, let the test pass and also fine in both ways having depend or not.

adisky:
@kayrus Thanks!!
/lgtm
/assign @lingxiankong @ramineni @zetaab

kayrus:
Any other concerns? I tested this in docker container as well, and it works perfectly:

```
docker run -ti --rm -v "${PWD}":/go/src/k8s.io/cloud-provider-openstack -w /go/src/k8s.io/cloud-provider-openstack golang:latest make build
```

kayrus:

zetaab:
from my point of view it is enough if everything works :)
/approve

someone else can add lgtm

PR #776, opened 2019-09-23
Fix closing of dirs in doSafeMakeDir

odinuge:

This fixes the issue where "childFD" from syscall.Openat is assigned to
a local variable inside the for loop, instead of the correct one in the
function scope. This results in that when trying to close the "childFD"
in the function scope, it will be equal to "-1", instead of the correct
value.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*:

See: https://github.com/kubernetes/kubernetes/pull/79534, https://github.com/kubernetes/kubernetes/pull/82873 & https://github.com/kubernetes/kubernetes/pull/82980

**Release note**:

```release-note
Fix possible fd leak and closing of dirs in doSafeMakeDir
```

### DIFF
--- a/pkg/util/mount/mount_linux.go
+++ b/pkg/util/mount/mount_linux.go
@@ -1118,7 +1118,7 @@ func doSafeMakeDir(pathname string, base string, perm os.FileMode) error {
 			return fmt.Errorf("cannot create directory %s: %s", currentPath, err)
 		}
 		// Dive into the created directory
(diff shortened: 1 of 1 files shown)


ramineni:
Thanks . keystone job error not related to the PR.
/lgtm

PR #790, opened 2019-09-28
Auth: add endpoint type parameter support

### LINKED ISSUE #713: Ability to specify which interface to use for the cloud provider

/kind feature

**What happened**:
I couldn't specify which interface endpoints to use for the Openstack cloud provider. It defaults to the public interface from the gophercloud SDK.

A way to specify which interface to use in the cloud provider config:
e.g.
interface=internal or interface=admin
keep public by default if not specified.

Start any cluster with the openstack cloud provider and turn on verbose logging to see the requests going to the public endpoint.

**Environment**:
- openstack-cloud-controller-manager version: all
- OS (e.g. from /etc/os-release): all
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

kayrus:

It allows to specify a service endpoint type: e.g. admin, public, internal.

**Which issue this PR fixes**

fixes #713

**Release note**:

```release-note
Auth: add endpoint type parameter support
```

### DIFF
--- a/docs/provider-configuration.md
+++ b/docs/provider-configuration.md
@@ -121,6 +121,9 @@ with information about domain: `domain-id` or `domain-name`.
   connotation, a deployment can use a geographical name for a region identifier
   such as `us-east`. Available regions are found under the `/v3/regions`
   endpoint of the Keystone API.
+* `endpoint_type`: Specify which type of endpoint to use from the service
+  catalog. Can be `admin`, `internal` or `public`. If not set, public endpoint
+  is used.
 * `tenant-id`: Used to specify the ID of the project where you want
   to create your resources. When using Keystone V3 - which changed the
   identifier `tenant` to `project` - the `tenant-id` value is automatically
--- a/pkg/cloudprovider/providers/openstack/openstack.go
+++ b/pkg/cloudprovider/providers/openstack/openstack.go
@@ -170,6 +170,7 @@ type ServerAttributesExt struct {
 type OpenStack struct {
 	provider       *gophercloud.ProviderClient
 	region         string
+	endpointType   gophercloud.Availability
 	lbOpts         LoadBalancerOpts
 	bsOpts         BlockStorageOpts
 	routeOpts      RouterOpts
@@ -196,6 +197,12 @@ type AuthOpts struct {
 	Region           string `name:"os-region"`
 	CAFile           string `gcfg:"ca-file" mapstructure:"ca-file" name:"os-certAuthorityPath" value:"optional"`
 
+	EndpointType gophercloud.Availability `gcfg:"endpoint-type" mapstructure:"endpoint-type" name:"os-endpointType" value:"optional"`
+
+	ApplicationCredentialID     string `gcfg:"application-credential-id" mapstructure:"application-credential-id" name:"os-applicationCredentialID" value:"optional"`
+	ApplicationCredentialName   string `gcfg:"application-credential-name" mapstructure:"application-credential-name" name:"os-applicationCredentialName" value:"optional"`
+	ApplicationCredentialSecret string `gcfg:"application-credential-secret" mapstructure:"application-credential-secret" name:"os-applicationCredentialSecret" value:"optional"`
+
 	// Manila only options
 	TLSInsecure string `name:"os-TLSInsecure" value:"optional" dependsOn:"os-certAuthority|os-certAuthorityPath" matches:"^true|false$"`
 	// backward compatibility with the manila-csi-plugin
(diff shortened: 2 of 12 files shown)


pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -196,6 +197,12 @@ type AuthOpts struct {
 	Region           string `name:"os-region"`
 	CAFile           string `gcfg:"ca-file" mapstructure:"ca-file" name:"os-certAuthorityPath" value:"optional"`
 
+	EndpointType gophercloud.Availability `gcfg:"endpoint-type" mapstructure:"endpoint-type" name:"os-endpointType" value:"optional"`
```
ramineni: Should be ideally interface_type or interface ?

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -355,6 +359,15 @@ func ConfigFromEnv() Config {
 	cfg.Global.ApplicationCredentialName = os.Getenv("OS_APPLICATION_CREDENTIAL_NAME")
 	cfg.Global.ApplicationCredentialSecret = os.Getenv("OS_APPLICATION_CREDENTIAL_SECRET")
 
+	endpointType := os.Getenv("OS_ENDPOINT_TYPE")
+	if endpointType == "internal" || endpointType == "internalURL" {
```
ramineni: You have documented supported types are public, admin,internal, So I dont think its necessary to check the URL also?

pkg/volume/cinder/volumeservice/connection.go
```diff
@@ -88,10 +88,10 @@ func getKeystoneVolumeService(cfg cinderConfig) (*gophercloud.ServiceClient, err
 		return nil, err
 	}
 
-	volumeService, err := openstack.NewBlockStorageV2(provider,
```
ramineni: standalone cinder is deprecated, no need to update here

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -355,6 +359,15 @@ func ConfigFromEnv() Config {
...
+		cfg.Global.EndpointType = gophercloud.AvailabilityInternal
+	} else if endpointType == "admin" || endpointType == "adminURL" {
+		cfg.Global.EndpointType = gophercloud.AvailabilityAdmin
```
ramineni: can we just use Type Conversion here?
Eg:
cfg.Global.EndpointType = Availability(cfg.Global.EndpointType)

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -125,7 +125,8 @@ func CreateOpenStackProvider() (IOpenStack, error) {
 	}
 
 	epOpts := gophercloud.EndpointOpts{
-		Region: cfg.Global.Region,
+		Region:       cfg.Global.Region,
+		Availability: cfg.Global.EndpointType,
```
ramineni: Type Conversion missing here?

pkg/cloudprovider/providers/openstack/openstack.go
ramineni: And you are using this conversion only for env variables not for config file ones.. why?

pkg/cloudprovider/providers/openstack/openstack.go
kayrus: I thought about using a regular string, but it will require type conversion inside every package. Therefore I decided to keep the `gophercloud.Availability`.

pkg/cloudprovider/providers/openstack/openstack.go
kayrus: OpenStack python clients use `URL` suffix, gophercloud package doesn't use it. Extra options are needed to keep the compatibility.

pkg/csi/cinder/openstack/openstack.go
kayrus: no, it is not, since both `Availability` and `cfg.Global.EndpointType` are `gophercloud.Availability` https://github.com/gophercloud/gophercloud/blob/dee9bb4c60eccb45e80e3d2d0f4ced546f421ad6/endpoint_search.go

pkg/cloudprovider/providers/openstack/openstack.go
kayrus: > can we just use Type Conversion here?

Unfortunately no, since we have to make sure that the value is one of the supported.

> And you are using this conversion only for env variables not for config file ones.. why?

You're right, this requires custom unmarshaller with additional checks, or using the conversion func inside every dependent package. Probably it is better to create a conversion func and use the `string` type for the `EndpointType`.

pkg/volume/cinder/volumeservice/connection.go
kayrus: I prefer to keep the consistency within all packages.

docs/provider-configuration.md
```diff
@@ -121,6 +121,9 @@ with information about domain: `domain-id` or `domain-name`.
   connotation, a deployment can use a geographical name for a region identifier
   such as `us-east`. Available regions are found under the `/v3/regions`
   endpoint of the Keystone API.
+* `endpoint_type`: Specify which type of endpoint to use from the service
+  catalog. Can be `admin`, `internal` or `public`. If not set, public endpoint
+  is used.
```
lingxiankong: What if `endpoint_type: something-else`?

ramineni:
@kayrus Could you update this PR with changes requested ?

kayrus:
@ramineni will do a bit later

ramineni:
/close
please reopen if you would like to work on this again.

PR #792, opened 2019-10-03
Update share OWNERS file

tombarron:
**What this PR does / why we need it**: Update the new manila share OWNERS file with the current active set.

**Release note**:

```release-note
```

### DIFF
--- a/pkg/share/manila/OWNERS
+++ b/pkg/share/manila/OWNERS
@@ -1,5 +1,11 @@
 approvers:
 - tombarron
+- gouthampacha
+- gman0
+- tsmetana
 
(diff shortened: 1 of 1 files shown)


ramineni:
Only after added as members of the org, then only can be added here, thats the reason showing invalid-owners-file

tombarron:
> Only after added as members of the org, then only can be added here, thats the reason showing invalid-owners-file

@gouthamr and @tsmetana are already members and @gman0 is joining so hopefully this issue will resolve itself soon.

adisky:
Thanks!!
/approve

PR #801, opened 2019-10-16
Update octavia config

sfxworks:

It allows others to deploy octiava with the correct configurations

**Which issue this PR fixes**

Documentation / the template  ConfigMap octavia-ingress-controller-config for the latest octavia-ingress-controller image

Current template will fail as the old value with the latest build.

### DIFF
--- a/docs/using-octavia-ingress-controller.md
+++ b/docs/using-octavia-ingress-controller.md
@@ -93,7 +93,8 @@ data:
     cluster-name: ${cluster_name}
     openstack:
       auth-url: ${auth_url}
-      user-id: ${user_id}
+      domain-name: ${domain-name}
+      username: ${user_id}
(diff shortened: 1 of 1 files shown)


docs/using-octavia-ingress-controller.md
```diff
@@ -93,7 +93,8 @@ data:
     cluster-name: ${cluster_name}
     openstack:
       auth-url: ${auth_url}
-      user-id: ${user_id}
```
lingxiankong: we are using `user-id` in our production without any issue.

lingxiankong:
@sfxworks, we are deploying octavia-ingress-controller in our production with `user-id` and it's running well.

lingxiankong:
well, fair enough

/lgtm

PR #805, opened 2019-10-19
fix golint failures

FayerZhang:
/cleanup

fixes #28

**Release note**:
```NONE```

### DIFF
--- a/pkg/csi/cinder/mount/mount.go
+++ b/pkg/csi/cinder/mount/mount.go
@@ -87,10 +87,10 @@ func (m *Mount) GetHostUtil() hostutil.HostUtils {
 // probeVolume probes volume in compute
 func probeVolume() error {
 	// rescan scsi bus
-	scsi_path := "/sys/class/scsi_host/"
-	if dirs, err := ioutil.ReadDir(scsi_path); err == nil {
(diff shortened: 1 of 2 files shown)


kayrus:
Do these checks run automatically in tests?

ramineni:
@FayerZhang Thanks for the PR. Looks good. 
Could you also make this script up-to-date https://github.com/kubernetes/cloud-provider-openstack/blob/master/hack/verify-golint.sh , doesn't seem to work as expected. not sure if it is my env issue.

FayerZhang:
> @FayerZhang Thanks for the PR. Looks good.
> Could you also make this script up-to-date https://github.com/kubernetes/cloud-provider-openstack/blob/master/hack/verify-golint.sh , doesn't seem to work as expected. not sure if it is my env issue.
 please retry the script, and check my code again.  thanks!

ramineni:
@FayerZhang I have ran the script both on master branch and your branch (checkedout locally) , Im getting the o/p Congratulations!  All Go source files have been linted. But ideally on master branch it should throw lint errors. So thought script needs to be fixed. Let me know if you have different results when run on master.

But this is different issue , can be corrected in separate PR if required.

PR #733, opened 2019-08-22
Domain user name

kayrus:
**What this PR does / why we need it**:

It adds new auth options:
* `user-domain-id`
* `user-domain-name`

Aligns the order of the auth options to the one order.

Delegates the auth options verification to the `gophercloud/utils/openstack/clientconfig` package.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Release note**:

```release-note
```

PR #807, opened 2019-10-21
Auth: bring back identity v2 support

### LINKED ISSUE #803: Keystone auth v2 do not work
**What happened**:
In release 1.16 V2 keystone authentication stops working. 

Support for V2 should not be removed as still there are openstacks that use v2 for their authentication. 

Define V2 authentication for auth_url.

This was broken by commit https://github.com/kubernetes/cloud-provider-openstack/pull/733

Could be fixed by editing line 544 in pkg/cloudprovider/providers/openstack/openstack.go
from
`err = openstack.AuthenticateV3(provider, authOptsExt, gophercloud.EndpointOpts{})`
to
`err = openstack.Authenticate(provider, authOptsExt)`

/kind bug

kayrus:

Keystone V2 support was removed due to the auth code refactoring in #733. This PR should bring the support back.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #803

In #733 I accidentally used the `openstack.AuthenticateV3` instead of the `openstack.Authenticate` func. `openstack.Authenticate` is a wrapper around the `AuthenticateV2` and `AuthenticateV3`, therefore it should be fine to call it to keep the identity v2 support.

**Release note**:

```release-note
Fix the bug, when keystone v2 support was removed
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack.go
+++ b/pkg/cloudprovider/providers/openstack/openstack.go
@@ -547,7 +547,7 @@ func NewOpenStackClient(cfg *AuthOpts, userAgent string, extraUserAgent ...strin
 	}
 
 	opts := cfg.ToAuthOptions()
-	err = openstack.AuthenticateV3(provider, &opts, gophercloud.EndpointOpts{})
+	err = openstack.Authenticate(provider, opts)
 
 	return provider, err
 }
--- a/pkg/cloudprovider/providers/openstack/openstack_routes_test.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_routes_test.go
@@ -31,10 +31,11 @@ func TestRoutes(t *testing.T) {
 	const clusterName = "ignored"
 
 	cfg := ConfigFromEnv()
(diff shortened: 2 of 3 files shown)


ramineni:
> @ramineni k8s test server thinks that this is a unit test, but it is not. I now remember why I kept the `AuthenticateV3`, because it is simpler to track one exception in tests, rather than multiple without significantly changing the test logic. It looks like the logic has to be changed. I need some time to think about how exactly.

@kayrus ya, looks like ideally should be mocked. But, strange Authentciate is used before refactor also, then we haven't seen this issue.

kayrus:
@ramineni Authentciate was used before the refactoring, but there was different logic to check the auth parameters.

ramineni:
@kayrus Looks like you have skipped the failing tests?

kayrus:
@ramineni yes, I kept the same logic, which was before the refactoring. These tests run during functional tests in openlab, auth options unit tests run with no limitations.

kayrus:

PR #620, opened 2019-05-23
Add resize support for Cinder CSI 

ramineni:
**What this PR does / why we need it**:
This PR adds support for Volume expansion for CSI cinder Volumes.Supports both offline and Online Expansion.
Tested under debian OS.

TODO: Documentation. 

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #565

**Special notes for your reviewer**:

**Release note**:

```release-note
Enable Volume Expansion capability for Cinder CSI volumes
```

PR #808, opened 2019-10-21
Cinder: fix thread unsafe global client Microversion modification

### LINKED ISSUE #804: [Cinder CSI] ExpandVolume will eventyally change Microversion to 3.42
**What happened**:
As all CSI calls can be called simultaneously by multiple threads you should never modify global objects like os.blockstorage. As eventually some thread will fetch the temporary change as initial and then save it later. So then the "temporary" change will become permanent. 

If special client is needed it should be created for that call so that parallel calls would not mess up each others.

It was implement in https://github.com/kubernetes/cloud-provider-openstack/pull/620
and "fixed" in https://github.com/kubernetes/cloud-provider-openstack/pull/786

/kind bug

kayrus:

The #620 and #786 PRs haven't considered that modified variable is global, and this action is not thread-safe. This PR resolves unsafe global variable modification.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #804

@mape90 @ramineni @lingxiankong @adisky please review this PR and let me know if the bug has to be solved in another way

**Release note**:

```release-note
Cinder: fix thread unsafe global client Microversion modification
```


pkg/cloudprovider/providers/openstack/openstack_volumes.go
```diff
@@ -237,14 +239,17 @@ func (volumes *VolumesV3) expandVolume(volumeID string, newSize int) error {
 		NewSize: newSize,
 	}
 
-	// save initial microversion
-	mv := volumes.blockstorage.Microversion
-	volumes.blockstorage.Microversion = "3.42"
+	// Init a local thread safe copy of the Cinder ServiceClient
+	blockstorageClient, err := openstack.NewBlockStorageV3(volumes.blockstorage.ProviderClient, volumes.epOpts)
```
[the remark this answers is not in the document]
kayrus: Actually there is no need to do this, since a new copy of the volume client is already initialized [here](https://github.com/kubernetes/cloud-provider-openstack/blob/9774a6829cbc0c35b65824ff7fff82b936f3c51e/pkg/cloudprovider/providers/openstack/openstack_volumes.go#L371), I'll just put extra comments into the code and it should be enough.

pkg/cloudprovider/providers/openstack/openstack_volumes.go
```diff
@@ -237,15 +237,14 @@ func (volumes *VolumesV3) expandVolume(volumeID string, newSize int) error {
 		NewSize: newSize,
 	}
 
-	// save initial microversion
-	mv := volumes.blockstorage.Microversion
+	// cinder online resize is available since 3.42 microversion
+	// https://docs.openstack.org/cinder/latest/contributor/api_microversion_history.html#id40
+	// this operation is thread-safe, since a dedicated client is initialized before the
+	// "expandVolume" method call
 	volumes.blockstorage.Microversion = "3.42"
```
ramineni: @kayrus As you already have PR to fix resize, I think we have one more issue here that needs to be fixed. hardcoding microversion here might not be correct , as for offline resize its not needed . we need to check before , if its online/offline resize. Do you have the setup to test the resize operation again ?

kayrus: @ramineni good catch, thank you. fixed.

pkg/cloudprovider/providers/openstack/openstack_volumes.go
```diff
@@ -58,7 +58,7 @@ type volumeService interface {
 	createVolume(opts volumeCreateOpts) (string, string, error)
 	getVolume(volumeID string) (Volume, error)
 	deleteVolume(volumeName string) error
-	expandVolume(volumeID string, newSize int) error
+	expandVolume(volumeID string, newSize int, volumeStatus string) error
```
mape90: Everything in cloudprovider code related to volumes is dead code and it is here just because it was 1:1 copied from in-tree plugin that handled volumes. It might be good to delete whole openstack_volumes.go from cloudprovider as it is just confusing.

kayrus: @mape90 could you please elaborate? This code is used in openstack-cloud-controller-manager.

mape90: But openstack-cloud-controller-manager never manages volumes. So any code related to volumes is dead code. The CSI is managing all volume related actions.
In-tree version of openstack-cloud-controller-manager volumes where managed by it and thats why there is the code. However when this repo was done the code was just copy pasted from in-tree to here without any modifications. Then of course the volume control loops where removed. However the code base was never cleaned. There are also a lot of more dead code in openstack-cloud-controller-manager as it now behaves differently than it did when it was in-tree.

kayrus: Now I see, I had to recompile kubelet with the updated cloud-provider dependency to test the live volume resize with the 3.42 microversion. That is why I though that the code is used. When the `k8s.io/pkg/volume/cinder` is going to be removed then?

kayrus: Let's keep this as is and remove the code as a separate PR.

adisky: @kayrus agreed propose a separate  PR for removal, and also do not make any changes in the pkg `pkg/cloudprovider/providers/openstack/openstack_volumes.go`

kayrus:

ramineni:
Thanks.
/lgtm

kayrus:
The error message occurs only in golang 1.13.x, see https://github.com/golang/go/issues/31859

Pushed the fix.

kayrus:
Another issue https://github.com/kubernetes/kubernetes/pull/84662

kayrus:
go version was fixed, now there is another issue with go mod checksum: https://github.com/kubernetes-sigs/kubebuilder/issues/1169
@adisky is there an ability to merge this without waiting the dependency fix? this PR already contains the go 1.13 fix, which could be used in other PRs

kayrus:
@adisky @ramineni should be fine to merge

PR #820, opened 2019-11-06
Document for Inline Support

adisky:
Inline Volume support PR is in progress, it adds doc for the same

**Release note**:

```release-note
```
`NONE`

### DIFF
--- a/docs/using-cinder-csi-plugin.md
+++ b/docs/using-cinder-csi-plugin.md
@@ -281,10 +281,10 @@ Filesystem      Size  Used Avail Use% Mounted on
 /dev/vdb        2.0G   27M  1.97G   1% /var/lib/www/html
 ```
 ## Inline Volumes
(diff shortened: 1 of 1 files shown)


docs/using-cinder-csi-plugin.md
```diff
@@ -280,6 +280,35 @@ $ kubectl exec nginx -- df -h /var/lib/www/html
 Filesystem      Size  Used Avail Use% Mounted on
 /dev/vdb        2.0G   27M  1.97G   1% /var/lib/www/html
 ```
+## Inline Volumes
+This feature allows CSI volumes to be directly embedded in the Pod specification instead of a PersistentVolume. Volumes specified in this way are ephemeral and do not persist across Pod restarts. as of Kubernetes v1.16 this feature is beta so enabled by default. To enable this feature for CSI Driver, `volumeLifecycleModes` needs to be specified in CSIDriver object. The driver can run in `Persistent` mode, `Ephemeral` or in both modes. `podInfoOnMount` must be `true` to use this feature.
```
chrigl: Just a nit. "As of Kubernetes v.1.16 ..."

ramineni: Link to sample driver-yaml in our manifest would be good to point out here. If only Ephermal or both is added volumeLifecycleModes , then only inline volumes feature is enabled right? Only persistent is there , then it is like inline disabled? If yes, could we make that explicit as well?

adisky: @ramineni @chrigl Thanks for review, will update

adisky: @ramineni @chrigl  Done!!

adisky:
@ramineni Updated please review!!

PR #826, opened 2019-11-11
Remove loading config from Env Variables

### LINKED ISSUE #823: Remove loading config from Env variables

>
> /kind bug
> /kind feature

**What happened**:
Even though by default we load config options through secret in csi , there is fallback option to load from env if any errors. https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/openstack/openstack.go#L118 ,

With this issue, https://github.com/kubernetes/kubernetes/issues/81117#issuecomment-551403749 , its not recommended to load passwords from env , so plan is to remove the option to load from env . This shouldn't break anyone, as its not explicit option supported . Thoughts?
Use the by default supported option , using secret

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

ramineni:
This PR is to drop support to load config from Env .
**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #823 
https://github.com/kubernetes/kubernetes/issues/81117

**Release note**:

```release-note
Removed support to load config from env variables.
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack.go
+++ b/pkg/cloudprovider/providers/openstack/openstack.go
@@ -325,60 +325,12 @@ func (cfg AuthOpts) ToAuth3Options() tokens3.AuthOptions {
 	}
 }
 
-// ConfigFromEnv allows setting up credentials etc using the
-// standard OS_* OpenStack client environment variables.
-// TODO: Replace this with gophercloud upstream once community moves away from cloud.conf
-func ConfigFromEnv() Config {
-	var cfg Config
-
-	cfg.Global.AuthURL = os.Getenv("OS_AUTH_URL")
-	cfg.Global.UserID = os.Getenv("OS_USER_ID")
-	cfg.Global.Username = os.Getenv("OS_USERNAME")
-	cfg.Global.Password = os.Getenv("OS_PASSWORD")
-
-	cfg.Global.TenantID = os.Getenv("OS_TENANT_ID")
-	if cfg.Global.TenantID == "" {
-		cfg.Global.TenantID = os.Getenv("OS_PROJECT_ID")
-	}
-	cfg.Global.TenantName = os.Getenv("OS_TENANT_NAME")
-	if cfg.Global.TenantName == "" {
-		cfg.Global.TenantName = os.Getenv("OS_PROJECT_NAME")
-	}
-
-	cfg.Global.TrustID = os.Getenv("OS_TRUST_ID")
-	cfg.Global.DomainID = os.Getenv("OS_DOMAIN_ID")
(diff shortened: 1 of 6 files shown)


adisky:
Thanks!
/lgtm

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -323,60 +323,12 @@ func (cfg AuthOpts) ToAuth3Options() tokens3.AuthOptions {
...
-	cfg.Metadata.SearchOrder = fmt.Sprintf("%s,%s", metadata.ConfigDriveID, metadata.MetadataID)
-	cfg.Networking.IPv6SupportDisabled = false
-	cfg.Networking.PublicNetworkName = "public"
-	cfg.LoadBalancer.InternalLB = false
-
-	return cfg
-}
-
-// ReadConfig reads values from environment variables and the cloud.conf, prioritizing cloud-config
+// ReadConfig reads values from the cloud.conf, prioritizing cloud-config
```
lingxiankong: Could we remove 'prioritizing cloud-config' as well as we only use the config file now?

pkg/cloudprovider/providers/openstack/openstack_test.go
```diff
@@ -33,7 +33,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	"github.com/spf13/pflag"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
lingxiankong: I don't understand the change here

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -115,7 +115,8 @@ func CreateOpenStackProvider() (IOpenStack, error) {
 	// Get config from file
 	cfg, err := GetConfigFromFile(configFile)
 	if err != nil {
-		cfg = Config{Config: openstack_provider.ConfigFromEnv()}
+		klog.V(2).Infof("GetConfigFromFile %s failed with error: %v", configFile, err)
```
lingxiankong: I suggest we use ERROR log level here.

pkg/cloudprovider/providers/openstack/openstack.go
ramineni: Done

pkg/cloudprovider/providers/openstack/openstack_test.go
ramineni: Its automatically added I suppose , removed now.`

PR #845, opened 2019-11-24
[csi-manila] concurrency: replaced responsebroker with a try-lock

gman0:
This PR replaces csi-manila's response broker mechanism with a "try-lock by key" (`keyset` in `pkg/csi/manila/keyset`). If acquiring the lock fails, the operation in question (currently `CreateVolume` and `CreateSnapshot`) exits with `ABORTED` error code.

Snippet from the CSI spec [Concurrency](https://github.com/container-storage-interface/spec/blob/master/spec.md#concurrency):
> In general the Cluster Orchestrator (CO) is responsible for ensuring that there is no more than one call “in-flight” per volume at a given time. However, in some circumstances, the CO MAY lose state (for example when the CO crashes and restarts), and MAY issue multiple calls simultaneously for the same volume. The plugin SHOULD handle this as gracefully as possible. The error code ABORTED MAY be returned by the plugin in this case (see the Error Scheme section for details).

`ABORTED` error code has been supported by external-provisioner for some time now, csi-manila uses it to tell the CO whether the volume is ready to use or not. In a sense, this is similar to snapshot's [`ready_to_use` parameter](https://github.com/container-storage-interface/spec/blob/master/spec.md#the-ready_to_use-parameter).

**Motivation**:

By design, `responsebroker` blocks all future RPCs for the same resource (volume or snapshot identified by a CO-generated name) until the call that's currently being executed finishes. Response broker then either propagates successful response to all waiting calls, or retries with another waiting RPC on failure.

Such behavior is fine for reasonably fast request-response exchanges. Once the time spent in the call increases (by a large amount), open connections and goroutines will pile up over time. With the planned support for CephFS snapshot recovery in this CSI plugin, we anticipate that the time needed to recover a CephFS snapshot may be substantial, proportional to the amount of data in the snapshot.

Using `ABORTED` responses in conjunction with a try-lock mechanism ensures that the number of pending RPCs for any particular resource stays constant.

**Release note**:

```release-note
NONE
```


gman0:
PTAL @tsmetana @tombarron @gouthampacha @ramineni @adisky

pkg/csi/manila/keyset/keyset.go
```diff
@@ -0,0 +1,47 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package keyset
+
+import "sync"
+
+type KeySet struct {
```
kayrus: isn't it better to use https://golang.org/pkg/sync/atomic/#Value ?

gman0: Can you be please more specific? Instead of what?

gman0: What I'm confused about is that `Value` doesn't really seem to fit here. It's a single-value structure, whereas we need to store multiple keys (resource names in this case). It doesn't offer _compare-and-swap_ which is needed for checking and noting that an operation is in progress. Using it as a value inside the map wouldn't help either because the map itself must be protected against races when writing into it. `Value` wouldn't prevent from races here. Maybe you had a different idea?

gman0: `sync.Map` and its `LoadOrStore()` seems like a much better contender. Using that instead of `keyset` is possible.

gman0:
Closing in favor of https://github.com/kubernetes/cloud-provider-openstack/pull/881 which uses `sync.Map` instead.

Issue #856, opened 2019-11-28
New nodes always start off with a bunch of providerID-related issues


alvaroaleman:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
>
> /kind bug
> /kind feature

/kind bug

**What happened**:

A new node was added to the cluster. The openstack-cloud-controller-manager (`v1.16.0`) logs a bunch of
```
node_lifecycle_controller.go:154] error checking if node worker-6zq6t-56b66fc44-6zzvn is shutdown: ProviderID "" didn't match expected format "openstack:///InstanceID"
[...]
node_lifecycle_controller.go:171] error checking if node worker-6zq6t-56b66fc44-6zzvn exists: ProviderID "/1de1b490-4f9e-47e8-8dc0-81a47dd11a64" didn't match expected format "openstack:///InstanceID"
```

It eventually recovers and works, but there is quite a bit of time between "node joins" and "external CCM taint is removed", which I suspect is related to the errors above.

**What you expected to happen**:

No errors in the CCM

**How to reproduce it (as minimally and precisely as possible)**:

Add a new node to a cluster that uses the CCM, look at the CCM logs

**Anything else we need to know?**:

**Environment**:
- openstack-cloud-controller-manager version: v1.16.0
- OS (e.g. from /etc/os-release): ubuntu 18.04.2 LTS (Bionic Beaver)
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #857, opened 2019-11-28
Allow providerID without leading providerName

alvaroaleman:

Makes the cloudprovider cope with providerIDs that do not have a leading `provider://`, bandaid workaround for https://github.com/kubernetes/kubernetes/issues/85731 and works towards fixing https://github.com/kubernetes/cloud-provider-openstack/issues/856

**Release note**:

```release-note
none
```

/kind bug
/assign @chrigl

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack_instances.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_instances.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"strings"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/providers/openstack/openstack_instances.go
```diff
@@ -222,12 +223,18 @@ func srvInstanceType(srv *servers.Server) (string, error) {
 	return "", fmt.Errorf("flavor name/id not found")
 }
 
+// If Instances.InstanceID or cloudprovider.GetInstanceProviderID is changed, the regexp should be changed too.
+var providerIDRegexp = regexp.MustCompile(`^` + ProviderName + `:///([^/]+)$`)
```
alvaroaleman: I've moved this outside of the func to make sure we compile it only once and not everytime `instanceIDFromProviderID` is called, also has the advantage that should this ever be wrong, we panic on binary start rather than when `instanceIDFromProviderID` is called for the first time

chrigl:
Thanks @alvaroaleman . Could you please add a test case to https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/cloudprovider/providers/openstack/openstack_test.go#L878 ?

alvaroaleman:
@chrigl added a testcase, PTAL.

adisky:
Thanks!
/approve

Issue #859, opened 2019-12-02
Cinder CSI: wrong instanceID detection in OS with python based cloudinit


kayrus:
/kind bug

**What happened**:

Cinder CSI plugin detects invalid node ID:

```
I1202 13:49:57.542000       1 utils.go:79] GRPC call: /csi.v1.Identity/GetPluginInfo
I1202 13:49:57.542031       1 utils.go:80] GRPC request:
I1202 13:49:57.542113       1 identityserver.go:33] Using default GetPluginInfo
I1202 13:49:57.542119       1 utils.go:85] GRPC response: name:"cinder.csi.openstack.org" vendor_version:"1.1.0"
I1202 13:49:59.023895       1 utils.go:79] GRPC call: /csi.v1.Node/NodeGetInfo
I1202 13:49:59.023927       1 utils.go:80] GRPC request:
I1202 13:49:59.024432       1 mount.go:246] Got instance id from /var/lib/cloud/data/instance-id: i-0008fe64
I1202 13:49:59.298012       1 utils.go:85] GRPC response: node_id:"i-0008fe64" max_volumes_per_node:256 accessible_topology:<segments:<key:"topology.cinder.csi.openstack.org/zone" value:"region-3a" > >
```

OpenStack metadata service supports EC2 metadata API (e.g. `http://169.254.169.254/2009-04-04/meta-data/instance-id`), where nova [generates](https://github.com/openstack/nova/blob/56eac06561bc843dcf7497e8b7ac3027f62544c1/nova/api/metadata/base.py#L255) EC2 id, which doesn't have anything in common with a real OpenStack instance uuid. This EC2 id is used by python cloud-init and saved into the `/var/lib/cloud/data/instance-id`.

Legacy CCM code had an ability to define the [metadata order](https://github.com/kubernetes/cloud-provider-openstack/blob/f8ad69f78a71f945eef6c0858afecbe68f14c79d/docs/provider-configuration.md#metadata-optional-parameters), nevertheless it didn't use the `/var/lib/cloud/data/instance-id` file, but cloudinit drive, which is not really used in modern OpenStack deployments, therefore it always fell back to the [OpenStack metadata service](https://github.com/kubernetes/cloud-provider-openstack/blob/f8ad69f78a71f945eef6c0858afecbe68f14c79d/pkg/util/metadata/metadata.go#L42).

[...]

PR #862, opened 2019-12-03
Cinder CSI: add attach/detach timeouts

kayrus:

This PR adds sidecar timeouts to avoid error events like:

> connection.go:184] GRPC error: rpc error: code = DeadlineExceeded desc = context deadline exceeded

Default timeout is 15 seconds. 3 minutes should be enough to prepare a volume and attach it to a VM.

**Release note**:
```release-note
Cinder CSI: added attach/detach timeouts
```

### DIFF
--- a/manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml
+++ b/manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml
@@ -39,6 +39,7 @@ spec:
           args:
             - "--v=5"
             - "--csi-address=$(ADDRESS)"
+            - "--timeout=3m"
           env:
             - name: ADDRESS
               value: /var/lib/csi/sockets/pluginproxy/csi.sock
@@ -50,6 +51,7 @@ spec:
           image: quay.io/k8scsi/csi-provisioner:v1.4.0
           args:
             - "--csi-address=$(ADDRESS)"
+            - "--timeout=3m"
           env:
             - name: ADDRESS
               value: /var/lib/csi/sockets/pluginproxy/csi.sock


manifests/cinder-csi-plugin/cinder-csi-nodeplugin.yaml
```diff
@@ -70,9 +70,6 @@ spec:
             - name: kubelet-dir
               mountPath: /var/lib/kubelet
               mountPropagation: "Bidirectional"
-            - name: pods-cloud-data
-              mountPath: /var/lib/cloud/data
```
ramineni: @kayrus IMO, If its a valid and supported (not deprecated) way of retrieving data (eg: instanceid as mentioned in issue) . Removing is not a solution. I suggest we add a search order instead where we give user capability to choose from where he wants to retrieve the data . If its not really used (I dont have idea on that, others can chime in) , how about we deprecate this eventually and later be removed , documenting this way is not supported.

jichenjc: /var/lib/cloud/data seems to be a cloud-init thing? 
and agree that remove this seems too rush for now, maybe add some warning first. and if
someone has the /var/lib/cloud/data before, this will be a problem for them now

kayrus: agree, removed.

adisky:
@kayrus remove reference of #859 from description, update the heading too.

kayrus:
@adisky done

kayrus:
@lingxiankong @ramineni could you please review this?

jichenjc:
look good

Issue #863, opened 2019-12-03
CCM: Incorrect network interfaces order, when multiple interfaces are used


kayrus:
/kind bug

**What happened**:

When k8s compute nodes have multiple network interfaces, CCM constantly change them:

```sh
$ kubectl get nodes -o wide
NAME       STATUS   ROLES    AGE   VERSION   INTERNAL-IP   EXTERNAL-IP     OS-IMAGE                                      KERNEL-VERSION               CONTAINER-RUNTIME
node0012   Ready    <none>   18d   v1.16.3   10.128.0.34   192.1.106.99    Red Hat Enterprise Linux Server 7.7 (Maipo)   3.10.0-1062.1.2.el7.x86_64   docker://18.9.7
node0013   Ready    <none>   18d   v1.16.3   10.1.0.10     192.1.106.76    Red Hat Enterprise Linux Server 7.7 (Maipo)   3.10.0-1062.1.2.el7.x86_64   docker://18.9.7
$ sleep 5m
$ kubectl get nodes -o wide
NAME       STATUS   ROLES    AGE   VERSION   INTERNAL-IP   EXTERNAL-IP     OS-IMAGE                                      KERNEL-VERSION               CONTAINER-RUNTIME
node0012   Ready    <none>   18d   v1.16.3   10.1.0.34     192.1.106.99    Red Hat Enterprise Linux Server 7.7 (Maipo)   3.10.0-1062.1.2.el7.x86_64   docker://18.9.7
node0013   Ready    <none>   18d   v1.16.3   10.1.0.10     192.1.106.76    Red Hat Enterprise Linux Server 7.7 (Maipo)   3.10.0-1062.1.2.el7.x86_64   docker://18.9.7
```

This behavior causes issues, when kube-apiserver tries to connect to kubelet for `logs` or `exec` command, when `10.128.0.0/24` subnet is not available for kube-apiserver

And loadbalancer pool members change every sync loop:

```
I1203 19:28:31.845065       1 openstack_loadbalancer.go:1130] Ensured pool 0027f480-d407-44f3-bb5d-7d7c95fcbe2d has member for node0013 at 10.1.0.34
I1203 19:28:31.845271       1 openstack_loadbalancer.go:1135] Deleting obsolete member adfed073-bade-46cf-8b16-3bdfed2c7a62 for pool 0027f480-d407-44f3-bb5d-7d7c95fcbe2d address 10.128.0.34
```

**What you expected to happen**:

[...]

PR #864, opened 2019-12-04
CCM: Set permanent node addresses order

kayrus:

This PR sets permanent node IP addresses order, sorted by a network name.

This is a workaround for the #863 bug report. It doesn't help, when the first interface name appears at the very end after the alphabetical sorting. However this fix prevents LB members modification on each reconciliation loop.

**Release note**:

```release-note
CCM: Set permanent node addresses order, sort networks by name
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack.go
+++ b/pkg/cloudprovider/providers/openstack/openstack.go
@@ -29,6 +29,7 @@ import (
 	"os"
 	"reflect"
 	"regexp"
+	"sort"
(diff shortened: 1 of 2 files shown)


jichenjc:
/lgtm

as this is a workaround and helpful for some scenario

kayrus:
@lingxiankong @ramineni @adisky could you please review this PR?

lingxiankong:
/lgtm

thanks!

PR #877, opened 2019-12-14
Simplyfied dep management of k8s repos

chrigl:
All the repos from staging, now come with their own tags.
See #84608

**The binaries affected**:

- [x] cinder-csi-plugin
- [x] k8s-keystone-auth
- [x] client-keystone-auth
- [x] octavia-ingress-controller
- [x] manila-csi-plugin
- [x] manila-provisioner
- [x] magnum-auto-healer
- [x] barbican-kms-plugin

Changing hash based dependencies to tag based as mentioned in https://github.com/kubernetes/kubernetes/issues/84608. This simplifies future updates of kubernetes components.

fixes #

See https://github.com/kubernetes/kubernetes/issues/84608 for more information.

**Release note**:

```release-note
NONE
```


go.mod
```diff
@@ -45,12 +45,12 @@ require (
 	gopkg.in/square/go-jose.v2 v2.3.1 // indirect
 	gopkg.in/warnings.v0 v0.1.2 // indirect
 	gopkg.in/yaml.v2 v2.2.7
-	k8s.io/api v0.0.0
-	k8s.io/apimachinery v0.0.0
-	k8s.io/apiserver v0.0.0
-	k8s.io/client-go v0.0.0
-	k8s.io/cloud-provider v0.0.0
-	k8s.io/component-base v0.0.0
+	k8s.io/api v0.17.0
```
ramineni: Thanks, this makes lot more easier to understand. 
Could you help me understand how does it work with some intermediate releases, for example if 1.18.0.alpha1 ? will it be v0.18.0.alpha1 ?
I dont see this type of versioning in other k8s projects?

chrigl: It will be e.g. `v0.18.0-alpha.1` (which is already published by today). The publish-bot will create those tags on all repos corresponding to `k/k/staging/`. Even client-go adopts this schema.
See https://github.com/kubernetes/publishing-bot/pull/210

We still are able to follow individual commits in selected repos, e.g. if we need a fix that doesn't have a tagged release yet.

Because more and more go project will converge to use go modules, this schema will become more and more prominent. E.g. k3s already adopted this (well surprise, because they requested this change ;))

ramineni: Got it , Thanks . 
Ya, I see this type of versioning in latest tagged release ,https://github.com/kubernetes/apiserver/blob/v0.18.0-alpha.1/go.mod ,  I guess last time I checked for master branch , so couldnt find it ;) sorry about that :)

ramineni:
Thanks .
/lgtm

PR #881, opened 2019-12-16
[csi-manila] return ABORTED on simultaneous CreateVolume/Snapshot

gman0:
This PR replaces csi-manila's response broker mechanism with sync.Map as a means of
signaling whether there's already another operation processing the resource. If it is, CreateVolume/CreateSnapshot will now return `ABORTED` error code.

Snippet from the CSI spec [Concurrency](https://github.com/container-storage-interface/spec/blob/master/spec.md#concurrency):
> In general the Cluster Orchestrator (CO) is responsible for ensuring that there is no more than one call “in-flight” per volume at a given time. However, in some circumstances, the CO MAY lose state (for example when the CO crashes and restarts), and MAY issue multiple calls simultaneously for the same volume. The plugin SHOULD handle this as gracefully as possible. The error code ABORTED MAY be returned by the plugin in this case (see the Error Scheme section for details).

`ABORTED` error code has been supported by external-provisioner for some time now, csi-manila uses it to tell the CO whether the volume is ready to use or not. In a sense, this is similar to snapshot's [`ready_to_use` parameter](https://github.com/container-storage-interface/spec/blob/master/spec.md#the-ready_to_use-parameter).

**Motivation**:

By design, `responsebroker` blocks all future RPCs for the same resource (volume or snapshot identified by a CO-generated name) until the call that's currently being executed finishes. Response broker then either propagates successful response to all waiting calls, or retries with another waiting RPC on failure.

Such behavior is fine for reasonably fast request-response exchanges (even necessary, because external-provisioner didn't support ABORT back then). Once the time spent in the call increases, open connections and goroutines will pile up over time. With the planned support for CephFS snapshot recovery in this CSI plugin, we anticipate that the time needed to recover a CephFS snapshot may be substantial, proportional to the amount of data in the snapshot.

Bailing out with `ABORTED` on simultaneous calls ensures that there will always be only one goroutine waiting for a volume/snapshot to finish its preparation.

**Release note**:
```release-note
NONE
```


gman0:
PTAL @gouthampacha @tsmetana @tombarron

tombarron:
> I read the specification and I returning ABORT looks to be the right thing to do. I have no comments to the implementation.
Thanks for checking, I was wondering about that.

Code looks clean and the idea behind it seems sound.

I pulled the changes and went through our normal CRUD path and that works the same as without the changes.

I added a big time.Sleep() right before the csi.CreateVolumeResponse() call in pkg/csi/manila/controllerserver.go and was with a simple pvc creation able to see many returns of  ABORTs in the controller-plugin nfs-provisioner log.  As Robert says, these would otherwise turn into an pile of connections and goroutines and with long-running operations we would see them without the time.Sleep().

gouthampacha:
@adisky @ramineni  - can you help us merge this PR? Thank you! :)

PR #892, opened 2019-12-30
repair url in mysql-cinder-pd

zhanw15:
**The binaries affected**:
NONE

fixes #

**Release note**:

```release-note
NONE
```


ramineni:
@zhanw15  Thanks for the PR. Could you also update the description section under What this PR does / why we need it instead of leaving it blank.

PR #895, opened 2019-12-30
Remove chaosaffe from OWNERS

chaosaffe:
Removing myself from OWNERS as I am no longer actively participating

```release-note
NONE
```


RainbowMango:
@chaosaffe 
`at(@) mentions are not allowed in commit messages.`

PR #903, opened 2020-01-15
Multi region support

### LINKED ISSUE #885: Support for multiple OpenStack regions
/kind feature

From what I was able to see, there is no way to have multiple OpenStack regions in the same Kubernetes cluster... wouldn't this be a great addition to enable High-Availability ?

In my use-case, I have two regions corresponding to two datacenters that are about .600 ms away from each other and being able to run a single k8s cluster that span across both of them would be a great addition to our deployments !

cambierr:
This is a very drafty attempt to support multi-region in both the cloud provider and the CSI  cinder plugin as per #885 

**The binaries affected**:

- [x] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

This adds support for running cluster in multiple regions.

A common deployment topology with OpenStack deployment is to spawn multiple cinder/nova clusters with a single keystone one. In this case, this project doesn't work as we would expect it to work in a single region, multi AZ deployment.

fixes #885 

this is clearly a draft since 1. we are newbies with GO and 2. it's a first working version as per all the test we are running.

It's not ready for merge but before going any further we would like to see it there is any chance for this to get merged, and what would be your advice with it comes to good practices & general idea of our solution.

**Release note**:

```release-note
Added support for multi-region clusters in the CCM
Added support for multi-region clusters in the csi cinder plugin
```

### DIFF
--- a/examples/cinder-csi-plugin/nginx.yaml
+++ b/examples/cinder-csi-plugin/nginx.yaml
@@ -34,8 +34,8 @@ spec:
     - containerPort: 80
       protocol: TCP
     volumeMounts:
-    - mountPath: /var/lib/www/html
-      name: csi-data-cinderplugin
+      - mountPath: /var/lib/www/html
+        name: csi-data-cinderplugin
   volumes:
   - name: csi-data-cinderplugin
     persistentVolumeClaim:
--- a/examples/cinder-csi-plugin/nginx.yaml
+++ b/examples/cinder-csi-plugin/nginx.yaml
@@ -31,11 +31,11 @@ spec:
     imagePullPolicy: IfNotPresent
     name: nginx
     ports:
-      - containerPort: 80
-        protocol: TCP
+    - containerPort: 80
+      protocol: TCP
     volumeMounts:
-      - mountPath: /var/lib/www/html
-        name: csi-data-cinderplugin
+    - mountPath: /var/lib/www/html
+      name: csi-data-cinderplugin
   volumes:
   - name: csi-data-cinderplugin
     persistentVolumeClaim:
--- a/examples/cinder-csi-plugin/nginx.yaml
+++ b/examples/cinder-csi-plugin/nginx.yaml
@@ -14,7 +14,7 @@ metadata:
   name: csi-pvc-cinderplugin
 spec:
   accessModes:
-    - ReadWriteOnce
+  - ReadWriteOnce
   resources:
     requests:
       storage: 1Gi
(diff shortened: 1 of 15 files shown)


cambierr:
signed CLA

lingxiankong:
Hi @cambierr, thanks for proposing this patch, from the code I didn't see there is any change for creating Service? As I asked in the issue, how amphora in Octavia connect to VMs running in another region?

cambierr:
Hi @lingxiankong,

As described in the PR body, this is a very first draft limited to cinder-csi plugin, and CCM. This only adds the capacity to add the right region metadata and to provision volumes in the good region, as a first step.

For now, the other "features" of this cloud provider are left untouched as I'd like to first collect feedback on the general idea of the implementation ?

mikejoh:
Hi @cambierr , nice work with this feature! This PR has been closed for a while now, have you abandon it completely? The reason i ask is because we're moving into a multi region setup (with shared Keystone) and need support for this in CCM and Cinder. We're willing to help out!

jichenjc:
@mikejoh I assume you want to provide some insight and help on this PR so I reopened it 
please help to check whether you can contribute .. thanks

mikejoh:
> @mikejoh I assume you want to provide some insight and help on this PR so I reopened it
> please help to check whether you can contribute .. thanks

Thank you @jichenjc! We're about to test multi-region in our environments in the coming weeks and as soon as we have got something to test with we'll start working on this.

Sryther:
@mikejoh did you achieve to deploy in a multi-region environment? I'm interested too in this feature. I don't know if I could help on this since I never wrote in Go but I can give it a try.

Sryther:
@jichenjc could we assume that the scope of this PR is too wide? I mean the work already done here seems to be the minimum valuable product about the multi-region handling.

cambierr:
@Sryther if, like me, you need this to work, here is a 1.18 build that is running on our productions for a while: https://hub.docker.com/r/alphanetworkstv/openstack-cloud-controller-manager-amd64/tags?page=1&ordering=last_updated

We are now investigating the next step; either fork this repository and follow our own way (that's sad but we need this to work) or just build a proxy for the OpenStack API to "fake" replies

Sryther:
@cambierr, really kind of you!

I saw you referenced directly the regions you're using but I'm looking for a more generic way to support regions. I tried to improve your work by using an environment variable to set the different regions (`REGIONS` with space or comma separated values). It's not perfect but it could help others 😄.

I would like to retrieved values directly from the cloud.yml file but I think it's a bad idea to add custom properties in a worlwide used configuration format. What do you think about that?

Would you prefer I contribute to your branch or to a personal fork?

cambierr:
I just noticed the v18 isn't pushed to github, which means that indeed, regions are hardcoded. In my 1.18 versions, I added a `regions` parameters which is a coma-separated list... very close from what you just said, indeed ! 

Would you be interested in that ? if so, I can pushed the 1.18_multi_region branch so that anyone can benefit from that work and maybe extend it even

Issue #904, opened 2020-01-15
csi-manila: CreateVolume fails on GetExtraSpecs


gman0:
- [ ] openstack-cloud-controller-manager
- [ ] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] octavia-ingress-controller
- [x] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

**What happened**:

csi-manila tries to retrieve share type extra specs (used for determining manila caps), but those require admin rights. Regular users attempting to read extra specs will get following error:

```
E0115 19:46:31.312037       1 driver.go:282] [ID:10] GRPC error: rpc error: code = Internal desc = failed to get Manila capabilities for share type cephfstype: failed to retrieve extra specs: Request forbidden: [GET http://192.168.122.3/share/v2/83de2eb324e748018c5a9c47c2ed05a8/types/cephfstype/extra_specs], error message: {"forbidden": {"message": "Access was denied to this resource.", "code": 403}}
```

**How to reproduce it**:
Create a volume with a non-admin user.

/kind bug
/assign

PR #908, opened 2020-01-16
[manila-csi-plugin]: retrieve only public share type extra specs

gman0:
**The binaries affected**:

- [ ] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [x] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

csi-manila uses share type extra specs to determine which features (like snapshot support) should the plugin provide. Up until now, the plugin was querying for extra specs that were accessible to admin users only.

This PR makes it so that only public extra specs are queried (those contain all the info needed anyway), which are accessible to regular users as well.

Also, after a discussion with @tombarron , we've decided it's safer to not cache the results from "GET share types" due to possible cache collisions

fixes https://github.com/kubernetes/cloud-provider-openstack/issues/904

**Release note**:

```release-note
NONE
```


pkg/csi/manila/capabilities/manilacapabilities.go
```diff
@@ -33,23 +32,27 @@ const (
...
+	}
 
-	manilaCapabilitiesByShareTypeIDMtx.RLock()
-	caps, ok := manilaCapabilitiesByShareTypeID[shareType]
-	manilaCapabilitiesByShareTypeIDMtx.RUnlock()
+	return nil, fmt.Errorf("unknown share type %s", shareType)
+}
 
+func readManilaCaps(extraSpecs map[string]interface{}) ManilaCapabilities {
 	strToBool := func(ss interface{}) bool {
```
tombarron: Not for this PR but this could become a more general utility function that could be used e.g. in pkg/ingress/controller/controller.go and pkg/autohealing/cloudprovider/openstack/provider.go as well.

PR #917, opened 2020-01-29
Always retrieve node instance type from flavor.original_name

### LINKED ISSUE #916: CCM node instance type should always use flavor.original_name
/kind bug

**The binaries affected**:

**What happened**:

Node label `node.kubernetes.io/instance-type` was set to `2004`, a flavor id.

Node label `node.kubernetes.io/instance-type` should be `ug2.2xlarge`, the flavor name for id `2004`.

**How to reproduce it**:

Use the CCM against an Openstack Compute API that defaults to a micro-version older than 2.47.

The instance type uses one of `name`, `id` or `original_name`, whichever is found first in the server response under the `flavor` object.

If the compute service defaults to a micro-version earlier than 2.47, only `flavor.id` will be included (eg. `"2004"`), which is not terribly useful to have as a node label in Kubernetes.

The client micro-version should be fixed to 2.47 to guarantee the presence of `flavor.original_name`.

**Environment**:
- openstack-cloud-controller-manager version: v1.14.0

bgagnon:
To ensure `flavor.original_name` is included in the response from the compute API, set the micro version to `2.47`, where this property first appeared:
https://docs.openstack.org/api-ref/compute/?expanded=show-server-details-detail

This PR also removes support for `flavor.id` and `flavor.name`. The field `flavor.id` is no longer included in micro versions 2.47 and newer, and I cannot find any documented reference to `flavor.name` in the compute API; I don't believe this field ever existed.

**The binaries affected**:
-  openstack-cloud-controller-manager

This ensures `flavor.original_name` is consistently used as the node instance type label.

fixes #916 

**Release note**:
```release-note
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack_instances.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_instances.go
@@ -41,7 +41,8 @@ type Instances struct {
 }
 
 const (
-	instanceShutoff = "SHUTOFF"
+	instanceShutoff       = "SHUTOFF"
+	instanceTypeFlavorKey = "original_name"
 )
 
 // Instances returns an implementation of Instances for OpenStack.
@@ -53,6 +54,11 @@ func (os *OpenStack) Instances() (cloudprovider.Instances, bool) {
 		klog.Errorf("unable to access compute v2 API : %v", err)
 		return nil, false
 	}
+
+	// Server flavor details (flavor.original_name) requires
+	// micro-version 2.47. Prior to this, only flavor.id is
+	// returned, which is less useful as a node label.
+	compute.Microversion = "2.47"
 
 	klog.V(4).Info("Claiming to support Instances")
 
@@ -215,17 +221,14 @@ func (i *Instances) InstanceType(ctx context.Context, name types.NodeName) (stri
 }
 
 func srvInstanceType(srv *servers.Server) (string, error) {
-	keys := []string{"name", "id", "original_name"}
-	for _, key := range keys {
-		val, found := srv.Flavor[key]
-		if found {
-			flavor, ok := val.(string)
-			if ok {
-				return flavor, nil
-			}
+	val, found := srv.Flavor[instanceTypeFlavorKey]
+	if found {
+		flavor, ok := val.(string)
+		if ok {
+			return flavor, nil
 		}
 	}
-	return "", fmt.Errorf("flavor name/id not found")
+	return "", fmt.Errorf("flavor.%s not found", instanceTypeFlavorKey)
 }
 
 // If Instances.InstanceID or cloudprovider.GetInstanceProviderID is changed, the regexp should be changed too.


bgagnon:
CLA signed.

pkg/cloudprovider/providers/openstack/openstack_instances.go
```diff
@@ -54,6 +55,11 @@ func (os *OpenStack) Instances() (cloudprovider.Instances, bool) {
 		return nil, false
 	}
 
+	// Server flavor details (flavor.original_name) requires
+	// micro-version 2.47. Prior to this, only flavor.id is
+	// returned, which is less useful as a node label.
+	compute.Microversion = "2.47"
```
lingxiankong: This change brings incompatible change to the openstack cloud which doesn't support compute service version 2.47.

bgagnon: That's true. Do you think it should fallback to using `flavor.id` if this micro-version is not available? What is the strategy for the other services that also set a micro version?

kayrus: You can try to fetch supported microversions before setting it directly. I guess you can use this package https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/apiversions/requests.go against compute API, because urls are the same https://docs.openstack.org/api-guide/compute/microversions.html

See links below with similar issues:
https://github.com/kubernetes/cloud-provider-openstack/pull/786/files
https://github.com/kubernetes/cloud-provider-openstack/pull/808/files
Also be aware about the shared vars and threads.

lingxiankong: If the original requirement is to get a human-friendly flavor name, why not just change the order of keys in `srvInstanceType` function?

bgagnon: @lingxiankong sorry, that wasn't clear... The problem is that `id` and `original_name` are mutually exclusive in the response and depend on the micro-version selection. As such, changing the order does not make any difference.

Basically, you get either `id` _or_ `original_name`, and this depends on the negotiated micro-version. At and above 2.47, you get `original_name`, and below, you get `id`.

On our Openstack installation, the default micro version (if not specified) is less than 2.47, even though the compute service supports micro-versions up to 2.60.

I will try to make this clearer in the code and PR.

pkg/cloudprovider/providers/openstack/openstack_instances.go
```diff
@@ -215,17 +221,14 @@ func (i *Instances) InstanceType(ctx context.Context, name types.NodeName) (stri
 }
 
 func srvInstanceType(srv *servers.Server) (string, error) {
-	keys := []string{"name", "id", "original_name"}
```
jichenjc: I believe you don't want to change this logic if we are going to negotiate microversion?
as we might get name,id or original_name based on the API request we send?

pkg/cloudprovider/providers/openstack/openstack_instances.go
jichenjc: the way send to network API is doable but it's a little bit hacking..
can we create a followup PR to gophercloud to add microversion negoation to compute side?

lingxiankong: Before supporting micro-version negotiation in gophercloud, we need something like https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/util/openstack/loadbalancer.go#L77

lingxiankong:
PR not updated for almost 3 months. Close now.

PR #919, opened 2020-01-30
nit , cinder csi: remove default verbose flag

jichenjc:
during some debug found some logs are not enabled by default
and not consistent for all containers as well, add this PR
to make them verbose and consistent across pod

**The binaries affected**:

- [x] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml
+++ b/manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml
@@ -37,7 +37,6 @@ spec:
         - name: csi-attacher
           image: quay.io/k8scsi/csi-attacher:v2.0.0
           args:
-            - "--v=5"
             - "--csi-address=$(ADDRESS)"
             - "--timeout=3m"
           env:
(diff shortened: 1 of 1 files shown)


manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml
```diff
@@ -90,6 +92,7 @@ spec:
             - "--endpoint=$(CSI_ENDPOINT)"
             - "--cloud-config=$(CLOUD_CONFIG)"
             - "--cluster=$(CLUSTER_NAME)"
+            - "--v=5"
```
ramineni: For production deployments , I don't think enabling verbose by default is recommended , as right now we dont have different release and development manifest

jichenjc: ok, I will remove this one

adisky:
@jichenjc There is no need to enable verbose in any sidecars/cinder driver container, as these are default manifests, for your dev enviroment you can enable them locally.

jichenjc:
@adisky I agree it might be added locally, so I removed existing 2 flags to make consist..

adisky:
> @adisky I agree it might be added locally, so I removed existing 2 flags to make consist..
@jichenjc  Thanks!
/approve

PR #920, opened 2020-02-03
[cinder-csi-plugin] Add metadata search order

### LINKED ISSUE #859: Cinder CSI: wrong instanceID detection in OS with python based cloudinit

jichenjc:
**The binaries affected**:

- [x] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

fixes #859 

**Release note**:

```release-note
for CSI cinder, the search-order defined in cloud.conf is honored , so you can
provide a string like `configDrive,metadataService` or `metadataService,configDrive` to
define the order to get meta from. 
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/cloud-provider-openstack/920)

### DIFF
--- a/docs/using-cinder-csi-plugin.md
+++ b/docs/using-cinder-csi-plugin.md
@@ -41,6 +41,12 @@ Encode your ```$CLOUD_CONFIG``` file content using base64.
 Update ```cloud.conf``` configuration in ```manifests/cinder-csi-plugin/csi-secret-cinderplugin.yaml``` file
 by using the result of the above command.
 
+> NOTE: In OpenStack, the compute instance uses either config drive or metadata service to retrieve instance-specific data. As the cluster administrator, you are able to config the order in the cloud config file for cinder-csi-plugin, the default configuration is as follows:
+```
+[Metadata]
+search-order = configDrive,metadataService
+```
+
 > NOTE: if your openstack cloud has cert (which means you already has cafile definition in cloud-config), please make sure that you also updated the volumes list of `cinder-csi-controllerplugin.yaml` and `cinder-csi-nodeplugin.yaml` to include the cacert. e.g following sample then mount the volume to the pod as well.
 
 ```
--- a/pkg/csi/cinder/nodeserver.go
+++ b/pkg/csi/cinder/nodeserver.go
@@ -147,7 +147,7 @@ func nodePublishEphermeral(req *csi.NodePublishVolumeRequest, ns *nodeServer) (*
 
 	// attach volume
 	// for attach volume we need to have information about node.
-	nodeID, err := getNodeID(ns.Mount, ns.Metadata)
+	nodeID, err := getNodeID(ns.Mount, ns.Metadata, ns.Cloud.GetMetadataOpts().SearchOrder)
 	if err != nil {
 		klog.V(3).Infof("Ephermal Volume Attach: Failed to get Instance ID: %v", err)
 		return nil, status.Error(codes.Internal, fmt.Sprintf("Ephermal Volume Attach: Failed to get Instance ID: %v", err))
@@ -448,7 +448,7 @@ func (ns *nodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
 
 func (ns *nodeServer) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoRequest) (*csi.NodeGetInfoResponse, error) {
 
-	nodeID, err := getNodeID(ns.Mount, ns.Metadata)
+	nodeID, err := getNodeID(ns.Mount, ns.Metadata, ns.Cloud.GetMetadataOpts().SearchOrder)
 	if err != nil {
 		return nil, status.Error(codes.Internal, fmt.Sprintf("NodeGetInfo failed with error %v", err))
 	}
@@ -523,6 +523,7 @@ func getNodeIDMountProvider(m mount.IMount) (string, error) {
 		klog.V(3).Infof("Failed to GetInstanceID: %v", err)
 		return "", err
 	}
+	klog.V(5).Infof("getNodeIDMountProvider return node id %s", nodeID)
 
 	return nodeID, nil
 }
@@ -533,6 +534,7 @@ func getNodeIDMetdataService(m openstack.IMetadata) (string, error) {
 	if err != nil {
 		return "", err
 	}
+	klog.V(5).Infof("getNodeIDMetdataService return node id %s", nodeID)
 	return nodeID, nil
 }
 
@@ -545,17 +547,28 @@ func getAvailabilityZoneMetadataService(m openstack.IMetadata) (string, error) {
 	return zone, nil
 }
 
-func getNodeID(mount mount.IMount, metadata openstack.IMetadata) (string, error) {
-	// First try to get instance id from mount provider
-	nodeID, err := getNodeIDMountProvider(mount)
-	if err == nil {
-		return nodeID, nil
+func getNodeID(mount mount.IMount, iMetadata openstack.IMetadata, order string) (string, error) {
+	elements := strings.Split(order, ",")
+
+	var nodeID string
+	var err error
+	for _, id := range elements {
+		id = strings.TrimSpace(id)
+		switch id {
+		case metadata.ConfigDriveID:
+			nodeID, err = getNodeIDMountProvider(mount)
+		case metadata.MetadataID:
+			nodeID, err = getNodeIDMetdataService(iMetadata)
+		default:
+			err = fmt.Errorf("%s is not a valid metadata search order option. Supported options are %s and %s", id, metadata.ConfigDriveID, metadata.MetadataID)
+		}
+		if err == nil {
+			break
+		}
 	}
 
-	klog.V(3).Infof("Trying to GetInstanceID from metadata service")
-	nodeID, err = getNodeIDMetdataService(metadata)
 	if err != nil {
-		klog.V(3).Infof("Failed to GetInstanceID from metadata service: %v", err)
+		klog.Errorf("Failed to GetInstanceID from config drive or metadata service: %v", err)
 		return "", err
 	}
 	return nodeID, nil
--- a/pkg/csi/cinder/openstack/openstack.go
+++ b/pkg/csi/cinder/openstack/openstack.go
@@ -17,6 +17,7 @@ limitations under the License.
 package openstack
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/gophercloud/gophercloud"
@@ -27,6 +28,7 @@ import (
 	"github.com/spf13/pflag"
 	gcfg "gopkg.in/gcfg.v1"
 	openstack_provider "k8s.io/cloud-provider-openstack/pkg/cloudprovider/providers/openstack"
+	md "k8s.io/cloud-provider-openstack/pkg/util/metadata"
 	"k8s.io/klog"
 )
 
@@ -58,6 +60,7 @@ type IOpenStack interface {
 	GetInstanceByID(instanceID string) (*servers.Server, error)
 	ExpandVolume(volumeID string, size int) error
 	GetMaxVolLimit() int64
+	GetMetadataOpts() openstack_provider.MetadataOpts
 }
 
 type OpenStack struct {
@@ -65,6 +68,7 @@ type OpenStack struct {
 	blockstorage *gophercloud.ServiceClient
 	bsOpts       BlockStorageOpts
 	epOpts       gophercloud.EndpointOpts
+	metadataOpts openstack_provider.MetadataOpts
 }
 
 type BlockStorageOpts struct {
@@ -142,12 +146,18 @@ func CreateOpenStackProvider() (IOpenStack, error) {
 		return nil, err
 	}
 
+	// if no search order given, use default
+	if len(cfg.Config.Metadata.SearchOrder) == 0 {
+		cfg.Config.Metadata.SearchOrder = fmt.Sprintf("%s,%s", md.ConfigDriveID, md.MetadataID)
+	}
+
 	// Init OpenStack
 	OsInstance = &OpenStack{
 		compute:      computeclient,
 		blockstorage: blockstorageclient,
 		bsOpts:       cfg.BlockStorage,
 		epOpts:       epOpts,
+		metadataOpts: cfg.Config.Metadata,
 	}
 
 	return OsInstance, nil
--- a/pkg/csi/cinder/openstack/openstack_instances.go
+++ b/pkg/csi/cinder/openstack/openstack_instances.go
@@ -2,6 +2,7 @@ package openstack
 
 import (
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
+	cpo "k8s.io/cloud-provider-openstack/pkg/cloudprovider/providers/openstack"
 )
 
 // GetInstanceByID returns server with specified instanceID
@@ -11,4 +12,9 @@ func (os *OpenStack) GetInstanceByID(instanceID string) (*servers.Server, error)
 		return nil, err
 	}
 	return server, nil
+}
+
+// GetInstanceByID returns server with specified instanceID
+func (os *OpenStack) GetMetadataOpts() cpo.MetadataOpts {
+	return os.metadataOpts
 }
--- a/pkg/csi/cinder/openstack/openstack_mock.go
+++ b/pkg/csi/cinder/openstack/openstack_mock.go
@@ -21,6 +21,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	"github.com/stretchr/testify/mock"
+	cpo "k8s.io/cloud-provider-openstack/pkg/cloudprovider/providers/openstack"
 )
 
 var fakeVol1 = volumes.Volume{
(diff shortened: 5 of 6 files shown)


jichenjc:
@ramineni  @adisky can you help to provide some guide on this 
openlab/cloud-provider-openstack-e2e-test-csi-cinder  failed but I didn't know how to debug and what's wrong ..

pkg/csi/cinder/nodeserver.go
```diff
@@ -545,17 +547,33 @@ func getAvailabilityZoneMetadataService(m openstack.IMetadata) (string, error) {
 	return zone, nil
 }
 
-func getNodeID(mount mount.IMount, metadata openstack.IMetadata) (string, error) {
+func getNodeID(mount mount.IMount, iMetadata openstack.IMetadata, order string) (string, error) {
 	// First try to get instance id from mount provider
-	nodeID, err := getNodeIDMountProvider(mount)
-	if err == nil {
-		return nodeID, nil
+	if len(order) == 0 {
```
ramineni: It would be good if we could set the default order somewhere where config is initialized , so that it could be reused.

ramineni:
please add the release note as well , Support for 'search-order' Metadata opts added in cinder csi and also please update the same in docs as well

bzhaoopenstack:
recheck

bzhaoopenstack:
recheck

ramineni:
> recheck

Its retest in this repo :)

bzhaoopenstack:
@ramineni Thanks. ;-)

ramineni:
Thanks.
/lgtm

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -74,6 +83,7 @@ type BlockStorageOpts struct {
 type Config struct {
 	openstack_provider.Config
 	BlockStorage BlockStorageOpts
+	Metadata     MetadataOpts
```
lingxiankong: This option is already included in `openstack_provider.Config`.

docs/using-cinder-csi-plugin.md
```diff
@@ -41,6 +41,13 @@ Encode your ```$CLOUD_CONFIG``` file content using base64.
 Update ```cloud.conf``` configuration in ```manifests/cinder-csi-plugin/csi-secret-cinderplugin.yaml``` file
 by using the result of the above command.
 
+> NOTE: openstack supports get metadata data for new deployed vm through both config drive and metadata service methods. If you need modify the search order, please add Metadata section in your cloud.conf which is openstack cloud definition. For example, below sample makes CSI pod first search metadata service then use config drive. If you don't provide the value, then default search order `configDrive, metadataService` will be used.
+
+```
+[Metadata]
+search-order = metadataService,configDrive
```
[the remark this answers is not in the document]
jichenjc: ok

pkg/csi/cinder/openstack/openstack.go
ramineni: @lingxiankong But metadata opts in cloudprovider config includes additional options as well which is not yet supported by csi cinder, so isnt it better to override?
https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/cloudprovider/providers/openstack/openstack.go#L164

jichenjc: maybe later on when we do the refactory, we can consolidate then?

jichenjc:
not related failure ..

2020-02-11 04:04:40.306994 | ubuntu-xenial-citynetwork | --- FAIL: TestRoutes (17.21s)
2020-02-11 04:04:40.307139 | ubuntu-xenial-citynetwork |     openstack_routes_test.go:77: 10.164.2.0/24 via ubuntu-bionic-citynetwork-openlab-0000117831
2020-02-11 04:04:40.307225 | ubuntu-xenial-citynetwork |     openstack_routes_test.go:82: DeleteRoute error: failed to find object

pkg/csi/cinder/openstack/openstack.go
lingxiankong: > @lingxiankong But metadata opts in cloudprovider config includes additional options as well which is not yet supported by csi cinder, so isnt it better to override?

If ignoring the extra options doesn't break cinder-csi, that would be ok. The config options for those services should be well documented.

lingxiankong: > maybe later on when we do the refactory, we can consolidate then?

But if we add `Metadata` option here instead of re-using https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/cloudprovider/providers/openstack/openstack.go#L181, it will be hard for us to refactor in the future because we will introduce incompatible config change.

jichenjc: I tested the configuration ,with existing code add cloud.conf by adding [metadata] section
and works fine, I guess we didn't introduce incompatible config change?

my env is down now ,but I am sure it's something like

```
[Global]
username=user
password=pass
auth-url=https://<keystone_ip>/identity/v3
tenant-id=c869168a828847f39f7f06edd7305637
domain-id=2a73b8f597c04551a0fdc8e95544be8a

[metadata]
searchOrder = configDrive,metadata
```

docs/using-cinder-csi-plugin.md
```diff
@@ -41,6 +41,13 @@ Encode your ```$CLOUD_CONFIG``` file content using base64.
 Update ```cloud.conf``` configuration in ```manifests/cinder-csi-plugin/csi-secret-cinderplugin.yaml``` file
 by using the result of the above command.
 
+> NOTE: openstack supports get metadata data for new deployed vm through both config drive and metadata service methods. If you need modify the search order, please add Metadata section in your cloud.conf which is openstack cloud definition. For example, below sample makes CSI pod first search metadata service then use config drive. If you don't provide the value, then default search order `configDrive, metadataService` will be used.
```
lingxiankong: My suggestion:

```
> NOTE: In OpenStack, the compute instance uses either config drive or metadata service to retrieve instance-specific data. As the cluster administrator, you are able to config the order in the cloud config file for cinder-csi-plugin, the default configuration is as follows:
```

pkg/csi/cinder/openstack/openstack.go
lingxiankong: What I mean is, is `Metadata` already defined in `openstack_provider.Config`, isn't it?

docs/using-cinder-csi-plugin.md
jichenjc: ok

ramineni:
Strange , Im not able to comment in code , so adding here.
@lingxiankong I believe it should ignore extra options, ya proper documentation should avoid confusion on what all supported options. Even though, I think it wont be incompatible change as we are using the same group and parameter name, reusing would be better option as less hassle when we refactor.
@jichenjc I suppose what @lingxiankong meant is remove the line here https://github.com/kubernetes/cloud-provider-openstack/pull/920/files#diff-252011069a949a55c80b81de97af21f9R86 , as Metadata is already defined as part of openstack_provider.Config, so we could reuse the same instead of redefining the struct again.

lingxiankong:
> @lingxiankong I believe it should ignore extra options, ya proper documentation should avoid confusion on what all supported options. Even though, I think it wont be incompatible change as we are using the same group and parameter name, reusing would be better option as less hassle when we refactor.

If most of us prefer to add `Metadata` field explicitly, I am fine with that. @jichenjc could you please update the doc then I think the PR is good to go.

jichenjc:
um... I give another thought and seems the reuse of the param is a better option per suggested by @lingxiankong , I will update the PR..

docs/using-cinder-csi-plugin.md
lingxiankong: I didn't see the description is changed.

pkg/csi/cinder/nodeserver.go
```diff
@@ -545,17 +547,28 @@ func getAvailabilityZoneMetadataService(m openstack.IMetadata) (string, error) {
...
+		if err == nil {
+			break
+		}
 	}
 
-	klog.V(3).Infof("Trying to GetInstanceID from metadata service")
-	nodeID, err = getNodeIDMetdataService(metadata)
 	if err != nil {
-		klog.V(3).Infof("Failed to GetInstanceID from metadata service: %v", err)
+		klog.V(3).Infof("Failed to GetInstanceID from config drive or metadata service: %v", err)
```
lingxiankong: For error message, can you use `klog.Errorf()`?

docs/using-cinder-csi-plugin.md
lingxiankong: I didn't see the description is changed as suggested.

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -27,6 +28,7 @@ import (
 	"github.com/spf13/pflag"
 	gcfg "gopkg.in/gcfg.v1"
 	openstack_provider "k8s.io/cloud-provider-openstack/pkg/cloudprovider/providers/openstack"
+	md "k8s.io/cloud-provider-openstack/pkg/util/metadata"
```
lingxiankong: why not just use `metadata` as package name?

pkg/csi/cinder/openstack/openstack_instances.go
```diff
@@ -2,6 +2,7 @@ package openstack
 
 import (
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
+	openstack_provider "k8s.io/cloud-provider-openstack/pkg/cloudprovider/providers/openstack"
```
lingxiankong: Package name should be in lower case, with no under_scores or mixedCaps, so `cpo` could be used as in other files.

pkg/csi/cinder/openstack/openstack_mock.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	"github.com/stretchr/testify/mock"
+	openstack_provider "k8s.io/cloud-provider-openstack/pkg/cloudprovider/providers/openstack"
```
lingxiankong: Ditto

pkg/csi/cinder/sanity/fakecloud.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/snapshots"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
+	openstack_provider "k8s.io/cloud-provider-openstack/pkg/cloudprovider/providers/openstack"
```
lingxiankong: Ditto

docs/using-cinder-csi-plugin.md
```diff
@@ -41,6 +41,13 @@ Encode your ```$CLOUD_CONFIG``` file content using base64.
 Update ```cloud.conf``` configuration in ```manifests/cinder-csi-plugin/csi-secret-cinderplugin.yaml``` file
 by using the result of the above command.
 
+> NOTE: openstack supports get metadata data for new deployed vm through both config drive and metadata service methods. If you need modify the search order, please add Metadata section in your cloud.conf which is openstack cloud definition. For example, below sample makes CSI pod first search config drive then metadata service. If you don't provide the value, then default search order `configDrive, metadataService` will be used.
```
lingxiankong: Maybe you didn't see my previous suggestion, which I think is more clear and succinct

> NOTE: In OpenStack, the compute instance uses either config drive or metadata service to retrieve instance-specific data. As the cluster administrator, you are able to config the order in the cloud config file for cinder-csi-plugin, the default configuration is as follows:

pkg/csi/cinder/openstack/openstack.go
lingxiankong: Change missing?

jichenjc: I think I added a comment ,why it's not showing? werid..

I can't do so because if I move to metadata, it will be duplicate to openstack_metadata.go
the unit will fail

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/coreos/go-systemd v0.0.0-20190620071333-e64a0ec8b42a // indirect
 	github.com/emicklei/go-restful v2.9.6+incompatible // indirect
 	github.com/evanphx/json-patch v4.5.0+incompatible // indirect
+	github.com/go-logfmt/logfmt v0.4.0 // indirect
```
chrigl: Could you try to run `go mod tidy`? No idea where this comes from, because you didn't change any dependencies.

Everything else lgtm.

jichenjc: done, thanks

ramineni:
Thanks.
/lgtm

lingxiankong:
/lgtm
/approve

Thanks, @jichenjc

PR #925, opened 2020-02-07
csi-cinder-controller: changed NotFound behaviour of ControllerUnpublishVolume

chrigl:
**The binaries affected**:

- [x] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

At the moment we have a couple of flaky test in the CSIMigration effort.

During investigation, I found out that we have a race condition in the code in case of unattach a volume from a node. When either the OpenStack VM or Volume is gone, the `ControllerUnpublishVolume` function returns `csi.NotFound`. In case of CSIMigration this means being recalled over and over again with always the same result... `csi.NotFound`.

This PR changes: When either the volume or the instance in openstack is
already gone, the ControllerUnpublishVolume assumes the volume is not
attached any more.

This is is the behaviour DO and the GCP controller implement as well, and a sane implementation, because between any of the code steps, an out of cluster process could have delete the volume. E.g. a test case, where the internal teardown takes longer than reaching out for cinder delete in the cleanup step. We also can't assume to be called only once for a `ControllerUnpublishVolume` at the time.

fixes https://github.com/kubernetes/kubernetes/issues/87882

Like mentioned above: The DO and GCP plugins implement this in the similar way. See
* https://github.com/digitalocean/csi-digitalocean/blob/2bb4825cbf9e0fac34a3566e19b939ab378f91e7/driver/controller.go#L425
* https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/04e403a9ce733b9a9cc0545f434db0d8db8bc51a/pkg/gce-pd-csi-driver/controller.go#L387

The new implementation also aligns with my understanding of the spec
* https://github.com/container-storage-interface/spec/blob/master/spec.md#controllerunpublishvolume-errors

Conditions:
* Volume does not exist and volume not assumed ControllerUnpublished from node
* Node does not exist and volume not assumed ControllerUnpublished from node

... because we can assume that the volume is not attached if it (or the node) do not exist in OpenStack any more.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -214,17 +214,19 @@ func (cs *controllerServer) ControllerUnpublishVolume(ctx context.Context, req *
 	_, err := cs.Cloud.GetInstanceByID(instanceID)
 	if err != nil {
 		if cpoerrors.IsNotFound(err) {
-			return nil, status.Error(codes.NotFound, "ControllerUnpublishVolume Instance not found")
+			klog.V(3).Infof("ControllerUnpublishVolume assuming volume %s is detached, because node %s does not exist", volumeID, instanceID)
+			return &csi.ControllerUnpublishVolumeResponse{}, nil
 		}
 		return nil, status.Error(codes.Internal, fmt.Sprintf("ControllerUnpublishVolume GetInstanceByID failed with error %v", err))
 	}
 
 	err = cs.Cloud.DetachVolume(instanceID, volumeID)
 	if err != nil {
-		klog.V(3).Infof("Failed to DetachVolume: %v", err)
 		if cpoerrors.IsNotFound(err) {
-			return nil, status.Error(codes.NotFound, "ControllerUnpublishVolume Volume not found")
+			klog.V(3).Infof("ControllerUnpublishVolume assuming volume %s is detached, because it does not exist", volumeID)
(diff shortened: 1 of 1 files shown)


chrigl:
Running these flaky tests ...

```
[Fail] [sig-storage] In-tree Volumes [Driver: cinder] [Testpattern: Dynamic PV (default fs)] subPath [It] should support restarting containers using file as subpath [Slow][LinuxOnly]
/workspace/anago-v1.17.2-beta.0.2+59603c6e503c87/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/storage/testsuites/subpath.go:180
[Fail] [sig-storage] In-tree Volumes [Driver: cinder] [Testpattern: Dynamic PV (default fs)] subPath [It] should support existing directories when readOnly specified in the volumeSource
/workspace/anago-v1.17.2-beta.0.2+59603c6e503c87/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/storage/testsuites/subpath.go:180
```

...  over and over again over the entire weekend didn't show any failure.

lingxiankong:
Good catch! Thanks for the fix!

/lgtm

PR #927, opened 2020-02-08
[octavia-ingress-controller] Support TLS in Ingress

### LINKED ISSUE #287: TLS support for Octavia ingress controller
Feature Request
>
> /kind bug
/kind feature

**What happened**:
OpenStack Octavia ingress controller should support TLS now as there is Barbican support in gopher cloud/openstack cloud provider.

Annotations for TLS support need to be implemented.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

lingxiankong:
**The binaries affected**:

- [ ] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [x] octavia-ingress-controller
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

Support to create TLS Ingress.

This implementation for octavia-ingress-controller requires the TLS secret must exist when creating the Ingress. In the future, we could consider to create the secret on the fly with integration of e.g. cert-manager.

The TLS support requires OpenStack Barbican deployed in the cloud, TLS Ingress creation will fail if Barbican is not available.

fixes #287

Steps to test:

1. Preprae for the TLS secret, make sure to specify the server name in the CN field of the certificate, in this example, we use `www.example.com` as the server name.

    ```
    $ ll
    total 32
    -rw-rw-r-- 1 ubuntu ubuntu 1346 Feb  7 11:01 ca.crt
    -rw-rw-r-- 1 ubuntu ubuntu 1038 Feb  4 21:39 www.example.com.crt
    -rw-rw-r-- 1 ubuntu ubuntu  887 Feb  4 21:39 www.example.com.key
    $ openssl x509 -noout -text -in www.example.com.crt
    Certificate:
    ...
            Subject: C = NZ, ST = Wellington, L = Wellington, O = Catalyst, OU = Lingxian, CN = www.example.com
    ...
    $ kubectl create secret tls tls-secret --cert www.example.com.crt --key www.example.com.key
    $ kubectl describe secret tls-secret
    Name:         tls-secret
    Namespace:    default
    Labels:       <none>
    Annotations:  <none>

    Type:  kubernetes.io/tls

    Data
    ====
    tls.key:  887 bytes
    tls.crt:  1038 bytes
    ```
2. Create the Ingress

    ```
    cat <<EOF | kubectl apply -f -
    ---
    apiVersion: networking.k8s.io/v1beta1
    kind: Ingress
    metadata:
      name: test-octavia-ingress
      annotations:
        kubernetes.io/ingress.class: "openstack"
        octavia.ingress.kubernetes.io/internal: "false"
    spec:
      backend:
        serviceName: default-http-backend
        servicePort: 80
      tls:
        - secretName: tls-secret
      rules:
      - host: www.example.com
        http:
          paths:
          - path: /hostname
            backend:
              serviceName: hostname
              servicePort: 8080
    EOF
    ```

3. Check a secret created in OpenStack Barbican.

    ```
    $ openstack secret list
    +------------------------------------------------------------------------------------+------------------------------------------------------------------+---------------------------+--------+-----------------------------------------+-----------+------------+-------------+------+------------+
    | Secret href                                                                        | Name                                                             | Created                   | Status | Content types                           | Algorithm | Bit length | Secret type | Mode | Expiration |
    +------------------------------------------------------------------------------------+------------------------------------------------------------------+---------------------------+--------+-----------------------------------------+-----------+------------+-------------+------+------------+
    | http://128.110.154.166/key-manager/v1/secrets/c1b1ac4c-eced-4dee-8d61-efaaf13c4639 | kube_ingress_lingxiank8s_default_test-octavia-ingress_tls-secret | 2020-02-09T10:11:12+00:00 | ACTIVE | {'default': 'application/octet-stream'} | aes       |        256 | opaque      | cbc  | None       |
    +------------------------------------------------------------------------------------+------------------------------------------------------------------+---------------------------+--------+-----------------------------------------+-----------+------------+-------------+------+------------+
    ```

4. Verify https connection

    ```
    $ kubectl get ing
    NAME                   HOSTS             ADDRESS        PORTS     AGE
    test-octavia-ingress   www.example.com   172.24.5.218   80, 443   102s
    $ ip=172.24.5.218
    $ curl --cacert ca.crt --resolve www.example.com:443:$ip https://www.example.com/hostname
    hostname-544845f9c6-t9tnv
    ```

[Here](https://asciinema.org/a/299553?speed=2&cols=180&rows=40) is a live demo.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/cloud-provider-openstack/927)

### DIFF
--- a/pkg/ingress/controller/controller.go
+++ b/pkg/ingress/controller/controller.go
@@ -17,6 +17,11 @@ limitations under the License.
 package controller
 
 import (
+	"crypto"
+	"crypto/rand"
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/pem"
 	"fmt"
 	"reflect"
 	"strconv"
@@ -42,10 +47,12 @@ import (
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
+	pkcs12 "software.sslmate.com/src/go-pkcs12"
 
 	"k8s.io/cloud-provider-openstack/pkg/ingress/config"
 	"k8s.io/cloud-provider-openstack/pkg/ingress/controller/openstack"
 	"k8s.io/cloud-provider-openstack/pkg/ingress/utils"
+	openstackutil "k8s.io/cloud-provider-openstack/pkg/util/openstack"
 )
 
 const (
@@ -57,6 +64,13 @@ const (
 	defaultBurst = 1e6
 
 	maxRetries = 5
+
+	// CreateEvent event associated with new objects in an informer
+	CreateEvent EventType = "CREATE"
+	// UpdateEvent event associated with an object update in an informer
+	UpdateEvent EventType = "UPDATE"
+	// DeleteEvent event associated when an object is removed from an informer
+	DeleteEvent EventType = "DELETE"
 
 	// IngressKey picks a specific "class" for the Ingress.
 	// The controller only processes Ingresses with this annotation either
@@ -78,19 +92,18 @@ const (
 
 	// IngressControllerTag is added to the related resources.
 	IngressControllerTag = "octavia.ingress.kubernetes.io"
+
+	// IngressSecretCertName is certificate key name defined in the secret data.
+	IngressSecretCertName = "tls.crt"
+	// IngressSecretKeyName is private key name defined in the secret data.
+	IngressSecretKeyName = "tls.key"
+
+	// BarbianSecretNameTemplate is the name format string to create Barbican secret.
+	BarbianSecretNameTemplate = "kube_ingress_%s_%s_%s_%s"
 )
 
 // EventType type of event associated with an informer
 type EventType string
-
-const (
-	// CreateEvent event associated with new objects in an informer
-	CreateEvent EventType = "CREATE"
-	// UpdateEvent event associated with an object update in an informer
-	UpdateEvent EventType = "UPDATE"
-	// DeleteEvent event associated when an object is removed from an informer
-	DeleteEvent EventType = "DELETE"
-)
 
 // Event holds the context of an event
 type Event struct {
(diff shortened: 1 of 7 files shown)


lingxiankong:
cc @ramineni @chrigl

pkg/ingress/controller/controller.go
```diff
@@ -558,7 +627,26 @@ func (c *Controller) ensureIngress(ing *nwv1beta1.Ingress) error {
 		log.WithFields(log.Fields{"sgID": sgID, "ingress": key}).Info("ensured security group")
 	}
 
-	listener, err := c.osClient.EnsureListener(name, lb.ID)
+	// Convert kubernetes secrets to barbican ones
+	var secretRefs []string
+	for _, tls := range ing.Spec.TLS {
+		if c.osClient.Barbican == nil {
```
chrigl: If I understand this correct, this should go in form of
```
if len(ing.Spec.TLS) > 0 && c.osClient.Barbican == nil {
}
```
To the very beginning of the function.

There is no need to `EnsureLoadBalancer` and Managing Security Groups if this code check will fail.

pkg/ingress/controller/openstack/client.go
```diff
@@ -69,11 +70,21 @@ func NewOpenStack(cfg config.Config) (*OpenStack, error) {
 		return nil, fmt.Errorf("failed to find compute v2 endpoint for region %s: %v", cfg.OpenStack.Region, err)
 	}
 
+	// Get barbican service client.
+	var barbican *gophercloud.ServiceClient
+	barbican, err = openstack.NewKeyManagerV1(provider, gophercloud.EndpointOpts{
+		Region: cfg.OpenStack.Region,
+	})
+	if err != nil {
+		log.Warn("Barbican not suppported.")
```
chrigl: I think you should set `barbican = nil` explicit here. `openstack.NewKeyManagerV1` will also return a `ServiceClient`  in case of an error.
See https://github.com/gophercloud/gophercloud/blob/master/openstack/client.go#L465

But at other locations in the code you check for `os.Barbican == nil`.

pkg/ingress/controller/controller.go
lingxiankong: Make sense, thanks!

pkg/ingress/controller/openstack/client.go
lingxiankong: Agree

PR #930, opened 2020-02-11
[openstack-cloud-controller-manager] Delete the floating IP only when needed

lingxiankong:
**The binaries affected**:

- [ ] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

Delete the floating IP only if it was created dynamically by the controller manager.

The openstack-cloud-controller-manager now supports to specify an existing floating IP in `Spec.LoadBalancerIP` field, but that IP address is deleted by openstack-cloud-controller-manager when the Service is deleted which is not the expected behavior.

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
```diff
@@ -1839,9 +1840,17 @@ func (lbaas *LbaasV2) EnsureLoadBalancerDeleted(ctx context.Context, clusterName
...
-					return fmt.Errorf("failed to delete floating IP %s for loadbalancer VIP port %s: %v", fip.FloatingIP, portID, err)
+				matched, err := regexp.Match("Floating IP for Kubernetes external service", []byte(fip.Description))
+				if err != nil {
+					return err
+				}
+				if matched {
+					if err := floatingips.Delete(lbaas.network, fip.ID).ExtractErr(); err != nil {
+						return fmt.Errorf("failed to delete floating IP %s for loadbalancer VIP port %s: %v", fip.FloatingIP, portID, err)
+					}
 				}
```
jichenjc: should we add some log in !matched flag ?

lingxiankong: We could, but I didn't see it's very useful, it's obvious that only the floating IP whose description not satisfies the regex will not be touched. The condition answer is yes or no, the log is useful only if there are multiple conditions.

lingxiankong: Hi @jichenjc, are you ok if we merge this PR?

ramineni:
cc @chrigl

jichenjc:
yes, I think it's good to merge

PR #937, opened 2020-02-17
[cinder-flexvolume-driver] Disable the CI job

lingxiankong:
**The binaries affected**:

- [ ] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

As Cinder CSI Plugin has more advantages over the FlexVolume plugin. Although FlexVolume plugin is still maintained by Kubernetes developers, new functionality will only be added to CSI, not to FlexVolume.

This PR disables the CI job `cloud-provider-openstack-acceptance-test-flexvolume-cinder`.

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -85,15 +85,16 @@
               - ^pkg/util/.*
               - ^go.mod$
               - ^go.sum$
-    cloud-provider-openstack-acceptance-test-flexvolume-cinder:
-      jobs:
-        - cloud-provider-openstack-acceptance-test-flexvolume-cinder:
-            irrelevant-files:
-              - ^docs/.*$
-              - ^.*\.md$
-              - ^OWNERS$
-              - ^SECURITY_CONTACTS$
-              - ^.gitignore$
+    # Disable Cinder FlexVolume Driver Job as it is deprecated in cloud-provider-openstack repo.
+    # cloud-provider-openstack-acceptance-test-flexvolume-cinder:
+    #   jobs:
+    #     - cloud-provider-openstack-acceptance-test-flexvolume-cinder:
+    #         irrelevant-files:
+    #           - ^docs/.*$
+    #           - ^.*\.md$
+    #           - ^OWNERS$
+    #           - ^SECURITY_CONTACTS$
+    #           - ^.gitignore$
     cloud-provider-openstack-acceptance-test-csi-manila:
       jobs:
         - cloud-provider-openstack-acceptance-test-csi-manila:
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 Thank you for visiting the `Cloud Provider OpenStack` repository!
 
-This Repository hosts various plugins relavant to OpenStack and Kubernetes Integration
+This Repository hosts various plugins relevant to OpenStack and Kubernetes Integration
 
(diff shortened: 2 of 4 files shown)


.zuul.yaml
```diff
@@ -75,6 +75,12 @@
     cloud-provider-openstack-acceptance-test-flexvolume-cinder:
       jobs:
         - cloud-provider-openstack-acceptance-test-flexvolume-cinder:
+            files:
+              - ^cmd/cinder-flex-volume-driver/.*
+              - ^pkg/flexvolume/.*
```
ramineni: pkg/cloudprovider/ is used as well

ramineni:
@lingxiankong Is this driver used / maintained?

.zuul.yaml
lingxiankong: Yeah, just saw that.

BTW, do you know if there is anyone using the cinder flex driver? As CSI support has been GA since v1.13, do you think it's necessary for us to maintain this driver?

jichenjc: maybe you want to ask in some mail list about this ? not sure if someone is running this in their production/test env..

lingxiankong: Disabling CI job doesn't mean we will remove the code in repo nor images on docker hub, it means we don't have any contributor interested in maintaining the driver.

jichenjc: I am ok to do so,  as this will prevent quality control of the spcified code, e.g will someone get some image from docker hub that doesn't have the CI run for the driver so they don't know it's been tested and start to use it?
if that's not gonna happen, then we can disable CI ..

lingxiankong: 1. The latest image tag is not supposed to be production-ready.
2. We won't do release for Cinder FlexVolume Plugin from the next release (I am the person to do the release, so it's on my release note now).

lingxiankong: > I am ok to do so, as this will prevent quality control of the spcified code, e.g will someone get some image from docker hub that doesn't have the CI run for the driver so they don't know it's been tested and start to use it?
> if that's not gonna happen, then we can disable CI ..

However, I would be very happy if you are interested in maintaining this plugin and wanna be the owner

jichenjc: no, I have other items so can't do the job :)
the reason I am having this kind of question is to avoid break anyone who is using this
so as you mentioned, we will never have further release, I agree we need stop CI

thanks

README.md
```diff
@@ -14,6 +14,7 @@ This Repository hosts various plugins relavant to OpenStack and Kubernetes Integ
 * [Manila CSI Plugin](/docs/using-manila-csi-plugin.md/)
 * [Manila Provisioner](/docs/using-manila-provisioner.md/)
 * [Barbican KMS Plugin](/docs/using-barbican-kms-plugin.md/)
+* [Magnum Auto Healer](/docs/using-magnum-auto-healer.md/)
```
ramineni: @lingxiankong should we add flexvolume here and mark as deprecated explicitly?

adisky:
Agreed to disable the CI job and marking it as deprecated, from long time we have not seen any maintainer/user for the flex volume plugin
/lgtm

README.md
lingxiankong: Done

PR #939, opened 2020-02-17
[manila-csi-plugin] added AZ/topology support

### LINKED ISSUE #938: [manila-csi-plugin] add availability zones / topology support

/kind feature

**The binaries affected**:

- [ ] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [x] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

CSI Manila lacks any way of passing availability zones to Manila when provisioning a new share. Moreover, Manila AZs may not be equally accessible from all compute nodes of the cluster.

We need to add support for specifying Manila availability zones along with passing in information about the topology of compute nodes in case it's needed.

gman0:
**The binaries affected**:

- [ ] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [x] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

This PR adds support for Manila AZs and topology.

Documentation, examples, Helm chart and the sanity test have been updated as well.

fixes #938 

Our take on availability zones is two-fold:
* we provide users with a way of specifying Manila AZs via a storage class parameter
* this parameter is completely opaque to the CO and doesn't contribute to scheduling decisions on compute nodes
* this is sufficient because usually all compute nodes should have connectivity to the storage
* not always, though: if the nodes are partitioned into their own compute AZs and only certain compute AZs are able to reach a subset of Manila AZs, the cluster operator is able to deploy csi-manila with topology-awareness enabled and define the mapping between compute and storage AZs in StorageClasses

A couple of user-facing changes:
* added new StorageClass parameter `availability` : specifies AZ for Manila side of things
* added `--with-topology` cmd flag: enables topology-awareness of the driver (i.e. enables advertising of `VOLUME_ACCESSIBILITY_CONSTRAINTS` CSI capability)
* added `--nodeaz` cmd flag: sets the AZ of the node
* Helm chart:
  * added `csimanila.enableTopologyAwareness` into values.yaml: passes `--with-topology` and `--nodeaz`  to the manila-csi-plugin executable
  * added `csimanila.nodeAZ` into values.yaml: this value is sources into the `--nodeaz` cmd arg

```release-note
manila-csi-plugin: added support for Manila availability zones and topology-aware dynamic provisioning
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/cloud-provider-openstack/939)

### DIFF
--- a/cmd/manila-csi-plugin/main.go
+++ b/cmd/manila-csi-plugin/main.go
@@ -36,6 +36,8 @@ var (
 	endpoint              string
 	driverName            string
 	nodeID                string
+	nodeAZ                string
+	withTopology          bool
 	protoSelector         string
 	fwdEndpoint           string
 	userAgentData         []string
@@ -124,7 +126,21 @@ func main() {
 			manilaClientBuilder := &manilaclient.ClientBuilder{UserAgent: "manila-csi-plugin", ExtraUserAgentData: userAgentData}
 			csiClientBuilder := &csiclient.ClientBuilder{}
 
-			d, err := manila.NewDriver(nodeID, driverName, endpoint, fwdEndpoint, protoSelector, manilaClientBuilder, csiClientBuilder, compatOpts)
+			d, err := manila.NewDriver(
+				&manila.DriverOpts{
+					DriverName:          driverName,
+					NodeID:              nodeID,
+					NodeAZ:              nodeAZ,
+					WithTopology:        withTopology,
+					ShareProto:          protoSelector,
+					ServerCSIEndpoint:   endpoint,
+					FwdCSIEndpoint:      fwdEndpoint,
+					ManilaClientBuilder: manilaClientBuilder,
+					CSIClientBuilder:    csiClientBuilder,
+					CompatOpts:          compatOpts,
+				},
+			)
+
 			if err != nil {
 				klog.Fatalf("driver initialization failed: %v", err)
 			}
@@ -141,6 +157,10 @@ func main() {
 
 	cmd.PersistentFlags().StringVar(&nodeID, "nodeid", "", "this node's ID")
 	cmd.MarkPersistentFlagRequired("nodeid")
+
+	cmd.PersistentFlags().StringVar(&nodeAZ, "nodeaz", "", "this node's availability zone")
+
+	cmd.PersistentFlags().BoolVar(&withTopology, "with-topology", false, "cluster is topology-aware")
 
 	cmd.PersistentFlags().StringVar(&protoSelector, "share-protocol-selector", "", "specifies which Manila share protocol to use. Valid values are NFS and CEPHFS")
 	cmd.MarkPersistentFlagRequired("share-protocol-selector")
--- a/examples/manila-csi-plugin/helm-deployment/values.yaml
+++ b/examples/manila-csi-plugin/helm-deployment/values.yaml
@@ -9,11 +9,17 @@ shareProtocols:
       sockFile: csi.sock
 #  - protocolSelector: CEPHFS
 #    fwdNodePluginEndpoint:
-#      dir: /var/lib/kubelet/plugins/csi-cephfsplugin
+#      dir: /var/lib/kubelet/plugins/cephfs.csi.ceph.com
 #      sockFile: csi.sock
 
-# CSI Manila image spec
+# CSI Manila spec
 csimanila:
+  # Set topologyAwarenessEnabled to true to enable topology-awareness
+  topologyAwarenessEnabled: false
+  # Availability zone for each node. topologyAwarenessEnabled must be set to true for this option to have any effect.
+  # If your Kubernetes cluster runs atop of Nova and want to use Nova AZs as AZs for the nodes of the cluster, uncomment the line below:
+  # nodeAZ: "$(curl http://169.254.169.254/openstack/latest/meta_data.json | jq .availability_zone)"
+  # Image spec
   image:
     repository: manila-csi-plugin
     tag: latest
--- a/examples/manila-csi-plugin/helm-deployment/values.yaml
+++ b/examples/manila-csi-plugin/helm-deployment/values.yaml
@@ -14,7 +14,7 @@ shareProtocols:
 
 # CSI Manila spec
 csimanila:
-  # Set topologyAwarenessEnabled to true to enable topology-awareness
+  # Set topologyAwarenessEnabled to true to enable topology awareness
   topologyAwarenessEnabled: false
   # Availability zone for each node. topologyAwarenessEnabled must be set to true for this option to have any effect.
   # If your Kubernetes cluster runs atop of Nova and want to use Nova AZs as AZs for the nodes of the cluster, uncomment the line below:
--- a/examples/manila-csi-plugin/helm-deployment/values.yaml
+++ b/examples/manila-csi-plugin/helm-deployment/values.yaml
@@ -18,7 +18,7 @@ csimanila:
   topologyAwarenessEnabled: false
   # Availability zone for each node. topologyAwarenessEnabled must be set to true for this option to have any effect.
   # If your Kubernetes cluster runs atop of Nova and want to use Nova AZs as AZs for the nodes of the cluster, uncomment the line below:
-  # nodeAZ: "$(curl http://169.254.169.254/openstack/latest/meta_data.json | jq .availability_zone)"
+  # nodeAZ: "$(curl http://169.254.169.254/openstack/latest/meta_data.json | jq -r .availability_zone)"
   # Image spec
   image:
     repository: manila-csi-plugin
--- a/examples/manila-csi-plugin/nfs/topology-aware/storageclass.yaml
+++ b/examples/manila-csi-plugin/nfs/topology-aware/storageclass.yaml
@@ -1,0 +1,37 @@
+# Topology constraints example:
+#
+# Let's have two Manila AZs: manila-zone-1, manila-zone-2
+# Let's have six Nova AZs: nova-{1..6}
+#
+# manila-zone-1 is accessible from nodes in nova-{1,2,3} only
+# manila-zone-2 is accessible from nodes in nova-{4,5,6} only
+#
+# We're provisioning into manila-zone-1
+# allowedTopologies reflects the topology constraints
+
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: csi-manila-nfs-az
+provisioner: nfs.manila.csi.openstack.org
+parameters:
+  type: default
+  availability: manila-zone-1
+  
+  csi.storage.k8s.io/provisioner-secret-name: csi-manila-secrets
+  csi.storage.k8s.io/provisioner-secret-namespace: default
+  csi.storage.k8s.io/node-stage-secret-name: csi-manila-secrets
+  csi.storage.k8s.io/node-stage-secret-namespace: default
+  csi.storage.k8s.io/node-publish-secret-name: csi-manila-secrets
+  csi.storage.k8s.io/node-publish-secret-namespace: default
+allowedTopologies:
+  - matchLabelExpressions:
+    - key: topology.manila.csi.openstack.org/zone
+      values:
+        - nova-1
+        - nova-2
+        - nova-3
+    # ...or you may use other node labels:
+    # - key: my-zone-label
+    #   values:
+    #     - nodes-that-can-reach-manila-az-1
--- a/pkg/csi/manila/controllerserver.go
+++ b/pkg/csi/manila/controllerserver.go
@@ -153,9 +153,9 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 
 	var accessibleTopology []*csi.Topology
 	if cs.d.withTopology {
-		// All requisite topologies are considered valid. Nodes from those zones are required to be able to reach the storage.
+		// All requisite/preferred topologies are considered valid. Nodes from those zones are required to be able to reach the storage.
 		// The operator is responsible for making sure that provided topology keys are valid and present on the nodes of the cluster.
-		accessibleTopology = req.GetAccessibilityRequirements().GetRequisite()
+		accessibleTopology = req.GetAccessibilityRequirements().GetPreferred()
 	}
 
 	return &csi.CreateVolumeResponse{
--- a/pkg/csi/manila/controllerserver.go
+++ b/pkg/csi/manila/controllerserver.go
@@ -151,11 +151,19 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		}
 	}
 
+	var accessibleTopology []*csi.Topology
+	if cs.d.withTopology {
+		// All requisite topologies are considered valid. Nodes from those zones are required to be able to reach the storage.
+		// The operator is responsible for making sure that provided topology keys are valid and present on the nodes of the cluster.
+		accessibleTopology = req.GetAccessibilityRequirements().GetRequisite()
+	}
+
 	return &csi.CreateVolumeResponse{
 		Volume: &csi.Volume{
-			VolumeId:      share.ID,
-			ContentSource: req.GetVolumeContentSource(),
-			CapacityBytes: int64(sizeInGiB) * bytesInGiB,
+			VolumeId:           share.ID,
+			ContentSource:      req.GetVolumeContentSource(),
+			AccessibleTopology: accessibleTopology,
+			CapacityBytes:      int64(sizeInGiB) * bytesInGiB,
 			VolumeContext: map[string]string{
 				"shareID":        share.ID,
 				"shareAccessID":  accessRight.ID,
--- a/cluster/images/manila-csi-plugin/Dockerfile.build
+++ b/cluster/images/manila-csi-plugin/Dockerfile.build
@@ -3,4 +3,6 @@ FROM ${ALPINE_ARCH}/alpine:3.11
 
 ARG ARCH=amd64
 
+RUN apk add --no-cache ca-certificates jq curl
+
 ADD manila-csi-plugin-${ARCH} /bin/manila-csi-plugin
--- a/docs/using-manila-csi-plugin.md
+++ b/docs/using-manila-csi-plugin.md
@@ -13,7 +13,7 @@ The CSI Manila driver is able to create and mount OpenStack Manila shares. Snaps
 * [Deployment](#deployment)
   * [Kubernetes 1.15+](#kubernetes-115)
     * [Verifying the deployment](#verifying-the-deployment)
-    * [Enabling topology-awareness](#enabling-topology-awareness)
+    * [Enabling topology awareness](#enabling-topology-awareness)
 * [Share protocol support matrix](#share-protocol-support-matrix)
 * [For developers](#for-developers)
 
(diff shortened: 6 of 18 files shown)


examples/manila-csi-plugin/helm-deployment/values.yaml
```diff
@@ -9,11 +9,17 @@ shareProtocols:
...
 #      sockFile: csi.sock
 
-# CSI Manila image spec
+# CSI Manila spec
 csimanila:
+  # Set topologyAwarenessEnabled to true to enable topology-awareness
+  topologyAwarenessEnabled: false
+  # Availability zone for each node. topologyAwarenessEnabled must be set to true for this option to have any effect.
+  # If your Kubernetes cluster runs atop of Nova and want to use Nova AZs as AZs for the nodes of the cluster, uncomment the line below:
+  # nodeAZ: "$(curl http://169.254.169.254/openstack/latest/meta_data.json | jq .availability_zone)"
```
gouthampacha: Neat! 👍

gman0: `jq .availability_zone` returns `"some-zone"` - quotation marks included. This is undesirable but easily fixable by asking for a raw string using the `-r` flag.

gman0:
I think this is ready for review, PTAL @gouthampacha @tombarron @tsmetana @ramineni @adisky

tombarron:
/lgtm
And I've been running locally with topologyAwarenessEnabled and haven't hit any "surprises".

gman0:
The CI job `cloud-provider-openstack-acceptance-test-csi-manila` is broken after the recent multi-arch changes, i'll work on a fix.

lingxiankong:
> The CI job `cloud-provider-openstack-acceptance-test-csi-manila` is broken after the recent multi-arch changes, i'll work on a fix.

I think @ricolin is working on that, just FYI to avoid duplicating efforts.

gman0:
Some CI jobs are not triggering, I'll try to rebase.

gman0:
@tombarron @gouthampacha please reapply your LGTMs
@ramineni please if you could offer feedback and/or help us merge this

cmd/manila-csi-plugin/main.go
```diff
@@ -142,6 +158,10 @@ func main() {
 	cmd.PersistentFlags().StringVar(&nodeID, "nodeid", "", "this node's ID")
 	cmd.MarkPersistentFlagRequired("nodeid")
 
+	cmd.PersistentFlags().StringVar(&nodeAZ, "nodeaz", "", "this node's availability zone")
```
ramineni: @gman0 availability zone of the node could be directly retrieved from the metadata service, is it required to have  input here?

pkg/csi/manila/controllerserver.go
```diff
@@ -151,11 +151,19 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		}
 	}
 
+	var accessibleTopology []*csi.Topology
+	if cs.d.withTopology {
+		// All requisite topologies are considered valid. Nodes from those zones are required to be able to reach the storage.
+		// The operator is responsible for making sure that provided topology keys are valid and present on the nodes of the cluster.
+		accessibleTopology = req.GetAccessibilityRequirements().GetRequisite()
```
ramineni: Attempt should be made to provision volume accessible from preferred topology right ? not seeing any zone info being passed while creating, Im not sure if applicable for manila though.

cmd/manila-csi-plugin/main.go
```diff
@@ -142,6 +158,10 @@ func main() {
...
+
+	cmd.PersistentFlags().BoolVar(&withTopology, "with-topology", false, "cluster is topology-aware")
```
ramineni: Im wondering if this key is actually required? If no topology info are passed , the AccessiblityRequirements would be nil , so no harm done in that case? why is it needed to explicitly set this?

cmd/manila-csi-plugin/main.go
tombarron: I was thinking this is an optional flag for use when e.g. the OpenStack cloud isn't running the metadata service.  I may be confused though.

gouthampacha: Or when this CSI plugin is used on worker nodes that aren't OpenStack/nova VMs

tombarron: Right, people are working on k8s on baremetal instances in OpenStack clouds ...

ramineni: ok, since its optional flag , incase user doesnt provide it , there should be provision to retrieve automatically, if there are large number of nodes in cluster, adding AZ manually for all the nodes wouldn't be feasible?

gman0: I second what the guys have said :) !

> ok, since its optional flag , incase user doesnt provide it , there should be provision to retrieve automatically

The problem with defaulting to querying the Metadata Service is that in case it's not present (i.e.  _http://169.254.169.254/openstack/latest/meta_data.json_ is unreachable), it would first wait for a 30s timeout in [`http.Get()`](https://github.com/kubernetes/cloud-provider-openstack/blob/8c02748fab095675514861628ea57317bb36c263/pkg/csi/cinder/openstack/openstack_metadata.go#L39) and then it would return an error.

This error is ambiguous - we wouldn't be able to distinguish between:

(a) the plugin really is supposed to query the Metadata Service and the fact that `http.Get()` returned an error means there's a real problem in our cloud => the plugin should return an error and fail
(b) there's no Metadata Service in our cloud => ignore this error

By having no default value here, we're moving the responsibility of knowing about the presence of the Metadata Service to the cluster admins, which clears this ambiguity.

> adding AZ manually for all the nodes wouldn't be feasible?

What we're suggesting is that the admins would feed the output of `$(curl http://169.254.169.254/openstack/latest/meta_data.json | jq .availability_zone)` into the `--nodeaz` argument when deploying the plugin. It's mentioned in the documentation [here](https://github.com/kubernetes/cloud-provider-openstack/blob/fd6aa3342fb1d917a5fa79e02176075dc9cc5f7d/docs/using-manila-csi-plugin.md#enabling-topology-awareness)

gman0: Oh and the docs are missing `-r` flag for `jq`, I'll need to fix this

cmd/manila-csi-plugin/main.go
gman0: The reasoning behind having an explicit flag for enabling topology awareness is:

Linking these documents for reference:
* https://kubernetes-csi.github.io/docs/topology.html#overview
* https://github.com/kubernetes-csi/external-provisioner#topology-support

Let's say there's only one Manila AZ `shared-storage-a` and it is accessible from all nodes of the cluster. Those are spread across Nova zones `nova-1`, `nova-2`, `nova-3`. If

pkg/csi/manila/controllerserver.go
gman0: > not seeing any zone info being passed while creating, Im not sure if applicable for manila though.

It's here:

https://github.com/kubernetes/cloud-provider-openstack/blob/fd6aa3342fb1d917a5fa79e02176075dc9cc5f7d/pkg/csi/manila/volumesource.go#L40-L48

We've added a new storage class parameter `availability` where the operators will specify the Manila AZ.

We're using the storage class parameter AND topology awareness in conjunction because Manila may have its own set of storage AZs, and the k8s cluster may also have its own separate set of compute AZs and that those two sets may or may not overlap, so we've left this as a provisioning detail for the cluster admins: the admin has to know which [`allowedTopologies`](https://kubernetes.io/docs/concepts/storage/storage-classes/#allowed-topologies) are able to  access the Manila AZ specified in the `availability` parameter and define the StorageClass accordingly.

> Attempt should be made to provision volume accessible from preferred topology right ?

Yes you're right, according to the spec we should choose zone(s) from the preferred topologies and fallback to requisite topologies and if none of them are applicable then the provisioning should fail.

The problem is that in case of csi-manila, AFAIK there's no reliable way to check whether a certain storage zone is accessible from a certain compute zone. So I can change this to Preferred topologies, but it won't make much difference, because all the listed topology terms will be used for provisioning anyway.

cmd/manila-csi-plugin/main.go
gman0: > If no topology info are passed , the AccessiblityRequirements would be nil

If [`allowedTopologies`](https://kubernetes.io/docs/concepts/storage/storage-classes/#allowed-topologies) are not defined in the StorageClass, the external-provisioner will actually aggregate all available topologies of the cluster and continue from that set - if that's what you meant? [external-provisioner: Topology support](https://github.com/kubernetes-csi/external-provisioner#topology-support) for reference.

My reasoning behind this flag was that by default, all compute nodes have accessibility to the Manila-backed storage - so no topology-resolution is needed. In those "rare" cases where this is not true, admins need to explicitly state this fact and choose `allowedTopologies` carefully.

@tombarron @gouthampacha would you agree to get rid of this flag? This would mean that topology awareness would be mandatory (i.e. `--with-topology` as a permanent option), provisioned volumes would be tagged with:
* if `allowedTopologies == nil`: all topology terms from the whole cluster (for the topology key `topology.csi.manila.openstack.org/zone`)
* if `allowedTopologies != nil`: topology terms from `allowedTopologies`

Looking back, it does seem a bit superfluous.

examples/manila-csi-plugin/nfs/topology-aware/storageclass.yaml
```diff
@@ -0,0 +1,37 @@
...
+  csi.storage.k8s.io/node-publish-secret-name: csi-manila-secrets
+  csi.storage.k8s.io/node-publish-secret-namespace: default
+allowedTopologies:
+  - matchLabelExpressions:
+    - key: topology.manila.csi.openstack.org/zone
+      values:
+        - nova-1
+        - nova-2
+        - nova-3
+    # ...or you may use other node labels:
```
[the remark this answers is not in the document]
gman0: As I've said in the other comments, due to this loose coupling between Manila and compute AZs, the selected topology is not used during the actual storage provisioning. So there's not much point in using `WaitForFirstConsumer` binding mode...unless I'm missing something?

cmd/manila-csi-plugin/main.go
gouthampacha: --with-topology kind of shadows the options you'd set on the external provisioner itself (feature gates, Topology=true and `--strict-topology`), but, how do we declare the capability `VOLUME_ACCESSIBILITY_CONSTRAINTS` without something like this? If `nodeAZ` is set? Or will it always be declared?

cmd/manila-csi-plugin/main.go
ramineni: hmm, Im not sure distinguishing of error will be of that importance , we can log warning/error stating that unable to get AZ of node, need to set manually or so? 
In the doc anyway you are mentioning to set using metadata service , if that is accessible from node we could set it automatically only.  if for some issue , we got error on that then asking to set manually is good.
@gman0 @gouthampacha @tombarron Thoughts?

cmd/manila-csi-plugin/main.go
gman0: Yes, `VOLUME_ACCESSIBILITY_CONSTRAINTS` would be declared always.

`--nodeaz` is another thing though: previously I thought the value part of the topology term is allowed to be empty (so that we'd just do `--nodeaz=""` in the case where our cluster doesn't care about topology), now I'm not so sure:

An excerpt from [CreateVolume CSI spec](https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume) - down at the very bottom of the section:
```
// Each value (topological segment) MUST contain 1 or more strings.
// Each string MUST be 63 characters or less and begin and end with an
// alphanumeric character with '-', '_', '.', or alphanumerics in
// between.
message Topology {
  map<string, string> segments = 1;
}
```

ramineni: @gouthampacha Capability only tells that this driver supports topology. If user wants to use it or not, it depends on him. We dont need variable to advertise capability. Thats the feature we are providing.

examples/manila-csi-plugin/nfs/topology-aware/storageclass.yaml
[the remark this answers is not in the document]
ramineni: @gman0 @gouthampacha sorry about the questions, trying to understand lil better here..
'availability' is manila AZ where volume is provisioned
Allowed Topologies , say what all node AZ accessible with manila AZ provided, Am I right?
So when pod is scheduled , you would want to still take into account, it should be scheduled where manila AZ is accessible right , that is, it should scheduled on nodes where node AZ , nova1, nova2 or nova3 right?, so that will be taken care only when u specify delayed binding mode AFAIK . please correct if Im wrong

cmd/manila-csi-plugin/main.go
gman0: @ramineni csi-cinder returns an error on `getAvailabilityZoneMetadataService()` failure:

https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/nodeserver.go#L458-L461

not acting on this error + having to fail a timeout doesn't seem very nice to be honest :P

The difference between Cinder and Manila in this regard is that Cinder will always need Nova so it's guaranteed to also have the metadata svc, whereas Manila doesn't.

Just to clarify, we're not asking users to set the node AZs manually, it's set automatically during the initialization of the container using that little shell script.

pkg/csi/manila/controllerserver.go
ramineni: @gman0 @gouthampacha Thanks for the detailed explanation. I understand better now.
I think if delayed binding is not enabled, I agree this preferred/requisite wont make a difference. See my question on above comment , if my understanding correct on delayed binding needed or not.

cmd/manila-csi-plugin/main.go
gouthampacha: @ramineni: I prefer this approach over coding the metadata service lookup inside the node server... it wouldn't really be a concern for operators in my opinion because manila-csi will be deployed/managed with some automation for the most part (helm, operators, ansible etc), so it'll be easy to configure per the example shared if such a metadata lookup is desired.

gouthampacha: > Oh and the docs are missing `-r` flag for `jq`, I'll need to fix this

You have it fixed in https://github.com/kubernetes/cloud-provider-openstack/commit/3d16188b74866e7719778ef5c0bace651206415a - you're missing it in https://github.com/kubernetes/cloud-provider-openstack/pull/939/files#diff-2a40081587db138d3f23b73622df5608R167 :)

pkg/csi/manila/controllerserver.go
ramineni: And also , to add my reason why the 'withTopology' flag not needed , even if condition is not there , if Topology not enabled in provisoner, AccessibilityRequirements would be nil. If topology enabled in provisioner , it gets all aggregate topologies , which should be still fine , I think only field will be populated , and there wont be any difference in functionality . Please correct me if Im wrong.

ramineni:
@gman0  @gouthampacha Added my responses. We can get this into 1.18 if it is required by you guys. But please check on my comments on having additional flag to enable topology or not. IMO its not needed . Other things like automation of getting nodeAZ can be added later also in followup PR if needed and agreed, so that shouldn't be a blocker for this PR.

cmd/manila-csi-plugin/main.go
tombarron: Avoiding the timeout is a big improvement for operators and they can use automation such as the supplied shell script to set the node AZs.

ramineni: @tombarron @gouthampacha @gman0 ok, got it . `nodeAZ` it is :) . Lets go with this only as all are agreement with this flag :)
@gman0 do you mean retry would be better in case of cinder-csi?

cmd/manila-csi-plugin/main.go
gouthampacha: Ah, that makes sense. I was looking at the [code as it is written](https://github.com/kubernetes/cloud-provider-openstack/pull/939/files#diff-2220be4cd1a6283cbdae8088c8359a57R61-R69) and thinking we'd need to selectively advertise that capability. 

@gman0, ouch - does that mean that it has to always have a default/non-empty string? perhaps a default like "manila" will be harmless?

examples/manila-csi-plugin/nfs/topology-aware/storageclass.yaml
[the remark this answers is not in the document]
gouthampacha: @ramineni - not so, because _we_ have no programmatic way to tell if manila AZ (ex: "manila-zone-1") is actually reachable by any of the nova AZs (ex: "nova1", "nova2", "nova3"). This reachability will only be known to the deployer/operator. 

So, both delayed binding or immediate binding will have the same effect - the onus is on the deployer to set these selectors correctly, because both AZs are something they conceive (or know about through other means).

pkg/csi/manila/controllerserver.go
gman0: @ramineni 

> if Topology not enabled in provisoner, AccessibilityRequirements would be nil. If topology enabled in provisioner , it gets all aggregate topologies , which should be still fine , I think only field will be populated

That's correct!

But so far we've identified a couple of issues with having topology always enabled - all of those seem to be revolving around that Manila may be running outside of Nova environment, or that the (kubernetes) cluster may be consuming Manila storage without actually it being run in OpenStack (i.e. on baremetal or whatever):
* we'll get a silent timeout each time the driver is initializing
* in case we're actually relying on the data from the Metadata Service, we won't be able to fail in case the Metadata svc is (temporarily) unavailable
* if we don't care about topology (i.e. all compute nodes are able to access the Manila-backed storage), we still need to fill `--nodeaz` with something, otherwise we seem to be in conflict with the CSI spec? This could be solved with some placeholder value...although this feels a bit hack-y

cmd/manila-csi-plugin/main.go
gman0: @ramineni 

> do you mean retry would be better in case of cinder-csi?

IMHO the way csi-cinder is doing is the correct way. In case of a failure, the CO should do the retry.

examples/manila-csi-plugin/nfs/topology-aware/storageclass.yaml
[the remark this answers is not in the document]
tombarron: With manila, volume reachability is more reflective of network topology than compute  topology, compute failure domains.  It doesn't really map to neutron AZs [1] directly either because these reflect availability of neutron agents like DHCP, L3 agent, etc.

[1] https://docs.openstack.org/neutron/latest/admin/config-az.html

gouthampacha:
> @gman0 @gouthampacha Added my responses. We can get this into 1.18 if it is required by you guys. But please check on my comments on having additional flag to enable topology or not. IMO its not needed . Other things like automation of getting nodeAZ can be added later also in followup PR if needed and agreed, so that shouldn't be a blocker for this PR.

@ramineni - thank you. We'd like this in, and I agree that the code and minor doc fixes suggested can come as enhancements later... It'd be valuable to eliminate the redundancy in the flags, but, I agree this isn't a blocker. Thank you for reviewing this diligently..

examples/manila-csi-plugin/nfs/topology-aware/storageclass.yaml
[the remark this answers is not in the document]
ramineni: @gouthampacha Im not saying programmatic way , but if operator specify in AllowedTopologies that means manila AZ is reachable from them I suppose. But if you see there are three options given by operator nova1, nova2, nova3 in the above example saying they are reachable to manila AZ , so you also want pod to be scheduled in either 3 of them right to be able to access storage if Im not wrong . If pod gets scheduled independent of these it might go into pending state . IIUC delayed binding takes care of that by taking into account the topology common to both. May be Im missing something in what you have mentioned.

ramineni:
After discussion with @gman0 on slack, I think this should be good to go.
/approve

PR #947, opened 2020-02-28
[cinder-csi-plugin] Make Keystone probe period configurable separetely

takmatsu:
**The binaries affected**:
- [x] cinder-csi-plugin

* First motivation is we want to use livenessProbe: https://github.com/kubernetes-csi/livenessprobe for checking nodeplugin status.
* livenessProbe uses Probe(). In cinder-csi-plugin, it calls Keystone request every time.
* We want to check gRPC server itself every 2 secs or something, but we don't want to create requests to Keystone in same frequency.
* new flag `keystone-probe-period` make possible to use another period for checking keystone

**Release note**:

```release-note
Add new flag `keystone-probe-period`
```

### DIFF
--- a/cmd/cinder-csi-plugin/main.go
+++ b/cmd/cinder-csi-plugin/main.go
@@ -20,6 +20,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -31,10 +32,11 @@ import (
 )
 
 var (
-	endpoint    string
-	nodeID      string
-	cloudconfig string
-	cluster     string
+	endpoint            string
+	nodeID              string
+	cloudconfig         string
+	cluster             string
+	keystoneProbePeriod string
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


jichenjc:
https://github.com/kubernetes/cloud-provider-openstack/pull/952 maybe need align with this PR
as both seems want to improve probe function .. @ramineni

lingxiankong:
> * We want to check gRPC server itself every 2 secs or something, but we don't want to create requests to Keystone in same frequency.

@takmatsu If you don't want to make too much pressure to Keystone, why not just decrease the `periodSeconds` for `livenessProbe`? The `Probe()` should do the right thing and it should be up to the caller for the controls.

Any reason do you think `periodSeconds` param is not an option?

/hold

takmatsu:
actually making new keystone token doesn't make sense because it is not required to actual operation, especially for node plugin.
But if #952 is merged, it make sense for me.

ramineni:
#952 is merged, so closing this.
Feel free to reopen, if you feel still necessary
/close

PR #949, opened 2020-03-02
[cinder-csi-plugin] Update side car container version

jichenjc:
CSI side car has some update which fix some bugs
and something like metrics are added, so we can consume it

**The binaries affected**:

- [x ] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

update external attacher to 2.1.1
https://github.com/kubernetes-csi/external-attacher/releases/tag/v2.1.1
https://github.com/kubernetes-csi/external-attacher/blob/v2.1.0/CHANGELOG-2.1.md
there is no break changes

update external resizer to 0.4.0
https://github.com/kubernetes-csi/external-resizer/blob/v0.4.0/CHANGELOG-0.4.md
there is no breaking changes

update csi-node-driver-registrar to 1.2.0
https://github.com/kubernetes-csi/node-driver-registrar/blob/v1.2.0/CHANGELOG-1.2.md
there is no breaking changes

other incompatible changes will be added later

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml
+++ b/manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml
@@ -35,7 +35,7 @@ spec:
       serviceAccount: csi-cinder-controller-sa
       containers:
         - name: csi-attacher
-          image: quay.io/k8scsi/csi-attacher:v2.0.0
+          image: quay.io/k8scsi/csi-attacher:v2.1.1
           args:
             - "--csi-address=$(ADDRESS)"
             - "--timeout=3m"
@@ -70,7 +70,7 @@ spec:
             - mountPath: /var/lib/csi/sockets/pluginproxy/
               name: socket-dir
         - name: csi-resizer
-          image: quay.io/k8scsi/csi-resizer:v0.3.0
+          image: quay.io/k8scsi/csi-resizer:v0.4.0
           args:
             - "--csi-address=$(ADDRESS)"
           env:
--- a/manifests/cinder-csi-plugin/cinder-csi-nodeplugin.yaml
+++ b/manifests/cinder-csi-plugin/cinder-csi-nodeplugin.yaml
@@ -19,7 +19,7 @@ spec:
       hostNetwork: true
       containers:
         - name: node-driver-registrar
-          image: quay.io/k8scsi/csi-node-driver-registrar:v1.1.0
+          image: quay.io/k8scsi/csi-node-driver-registrar:v1.2.0
           args:
             - "--csi-address=$(ADDRESS)"
             - "--kubelet-registration-path=$(DRIVER_REG_SOCK_PATH)"


manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml
```diff
@@ -59,7 +59,7 @@ spec:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
         - name: csi-snapshotter
-          image: quay.io/k8scsi/csi-snapshotter:v1.2.2
+          image: quay.io/k8scsi/csi-snapshotter:v2.0.1
```
ramineni: If we are updating to 2.0.1 , I believe we need snapshot-controller also to be deployed?

jichenjc: the reason I made this PR is because I found some metrics can only be showed in latest
e.g 2.0.0, but I am not aware the snapshot-controller need to be there
so some homework need to be done.. thanks for pointing that

jichenjc: ok, I found 2.0.0 requires 1.17 k8s and controller to be deployed
so it might takes time for us to upgrade to there

anyway, will open an issue on that

ramineni: @jichenjc the metrics is part of which sidecar? provisioner?

jichenjc: @ramineni the metrics is introduced at 2.0.0, so because 2.0.0 has the controller 
requirement, then we have to first have the controller first

ramineni: @jichenjc ok, please update the description as well as now this PR is not updating snapshot sidecar, metrics is not applicable still , right.

manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml
```diff
@@ -47,7 +47,7 @@ spec:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
         - name: csi-provisioner
-          image: quay.io/k8scsi/csi-provisioner:v1.4.0
+          image: quay.io/k8scsi/csi-provisioner:v1.5.0
```
ramineni: @jichenjc v1alpha1 snapshot CRDs are not supported with this new version, but our examples still use it , make sure all gets updated as well.

please recheck all the breaking changes with respect to new versions of all sidecars .

lingxiankong:
@jichenjc besides the `metrics` addition, what else key features were added?

manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml
jichenjc: ok, will do ,thanks for the reminder

jichenjc:
> @jichenjc besides the `metrics` addition, what else key features were added?

@lingxiankong I will update commit message after solve the question from @ramineni 
the metrics thing which supposed to be done in the PR need defer to another one because I have to introduce another side car container...

jichenjc:
@ramineni @lingxiankong can you help to take a look? Thanks

manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml
```diff
@@ -35,7 +35,7 @@ spec:
       serviceAccount: csi-cinder-controller-sa
       containers:
         - name: csi-attacher
-          image: quay.io/k8scsi/csi-attacher:v2.0.0
+          image: quay.io/k8scsi/csi-attacher:v2.1.0
```
ramineni: latest release for attacher is 2.1.1?

jichenjc: yes, updates along with commit msg

PR #952, opened 2020-03-03
[cinder-csi-plugin] Improve Probe() call

### LINKED ISSUE #637: Improve CSI Probe Check

 /kind feature

**What happened**:
Currently in CSI Probe check, we are calling `CreateOpenStackProvider`
https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/identityserver.go#L49

which reads the config file, creates all the client, As Probe function can be called by CO multiple times, we can optimize this call to a smaller check, maybe calling a minimal Cinder API

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

ramineni:
**The binaries affected**:

- [x] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

This PR improves probe functionality 

fixes #637 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/identityserver.go
+++ b/pkg/csi/cinder/identityserver.go
@@ -47,11 +47,15 @@ func (ids *identityServer) GetPluginInfo(ctx context.Context, req *csi.GetPlugin
 }
 
 func (ids *identityServer) Probe(ctx context.Context, req *csi.ProbeRequest) (*csi.ProbeResponse, error) {
-
-	_, err := openstack.CreateOpenStackProvider()
+	klog.V(5).Infof("Probe() called with req %+v", req)
+	oProvider, err := openstack.GetOpenStackProvider()
 	if err != nil {
-		klog.V(3).Infof("Failed to CreateOpenStackProvider: %v", err)
-		return nil, status.Error(codes.FailedPrecondition, "Failed to communicate with openstack")
+		klog.Errorf("Failed to GetOpenStackProvider: %v", err)
+		return nil, status.Error(codes.FailedPrecondition, "Failed to retrieve openstack provider")
+	}
+	if err := oProvider.CheckBlockStorageAPI(); err != nil {
+		klog.Errorf("Failed to query blockstorage API: %v", err)
+		return nil, status.Error(codes.FailedPrecondition, "Failed to communicate with Openstack BlockStorage API")
 	}
 	return &csi.ProbeResponse{}, nil
 }
--- a/pkg/csi/cinder/openstack/openstack.go
+++ b/pkg/csi/cinder/openstack/openstack.go
@@ -41,6 +41,7 @@ func AddExtraFlags(fs *pflag.FlagSet) {
 }
 
 type IOpenStack interface {
+	CheckBlockStorageAPI() error
 	CreateVolume(name string, size int, vtype, availability string, snapshotID string, sourcevolID string, tags *map[string]string) (*volumes.Volume, error)
 	DeleteVolume(volumeID string) error
 	AttachVolume(instanceID, volumeID string) (string, error)
--- a/pkg/csi/cinder/openstack/openstack_mock.go
+++ b/pkg/csi/cinder/openstack/openstack_mock.go
@@ -116,6 +116,11 @@ func (_m *OpenStackMock) GetVolume(volumeID string) (*volumes.Volume, error) {
 	return &fakeVol1, nil
 }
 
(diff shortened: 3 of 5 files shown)


pkg/csi/cinder/identityserver.go
```diff
@@ -47,11 +47,15 @@ func (ids *identityServer) GetPluginInfo(ctx context.Context, req *csi.GetPlugin
...
-	_, err := openstack.CreateOpenStackProvider()
+	klog.V(5).Infof("Probe() called with req %+v", req)
+	oProvider, err := openstack.GetOpenStackProvider()
 	if err != nil {
-		klog.V(3).Infof("Failed to CreateOpenStackProvider: %v", err)
-		return nil, status.Error(codes.FailedPrecondition, "Failed to communicate with openstack")
+		klog.V(3).Infof("Failed to GetOpenStackProvider: %v", err)
+		return nil, status.Error(codes.FailedPrecondition, "Failed to retrieve openstack provider")
+	}
+	if _, err := oProvider.CheckBlockStorageAPI(); err != nil {
```
jichenjc: sorry, from bug description I got this
multiple times, we can optimize this call to a smaller check, maybe calling a minimal Cinder API
but seem this only add a call instead optimize? maybe I missed something..thanks

mtanino: Hi @ramineni @jichenjc 
IIUC, this change reduces number of call for `CreateOpenStackProvider()` because `CreateOpenStackProvider()` is called only one time when kubelet or control manager start up, right?
But I suppose there is a case that Keystone authentication token in OpenStack Client will be expired if we don't renew `NewOpenStackClient()` after passing certain period.

ramineni: @mtanino right, CreateOpenStackProvider() is called at startup , so every probe call doesnt need to create it again.
>> But I suppose there is a case that Keystone authentication token in OpenStack Client will >> be expired if we don't renew NewOpenStackClient() after passing certain period.
I believe automatic renewal of session should be possible, but Ill look into to it again. Thanks for bringing it up.

ramineni: @jichenjc hope your question is answered now.

jichenjc: ok, thanks for the info @ramineni

mtanino: @ramineni @jichenjc 
> I believe automatic renewal of session should be possible, but Ill look into to it again. 
> Thanks for bringing it up.
Thank you for the comment. For the purpose of automatic renewal of session, I think we can use following PR to configure probe period of keystone.
Could you also consider if this works for your improvement of Probe() func?

https://github.com/kubernetes/cloud-provider-openstack/pull/947

ramineni: @mtanino it should automatically renew token, https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/cloudprovider/providers/openstack/openstack.go#L331

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -39,6 +39,7 @@ func AddExtraFlags(fs *pflag.FlagSet) {
 }
 
 type IOpenStack interface {
+	CheckBlockStorageAPI() (bool, error)
```
lingxiankong: What's the intention of this `bool` type value? I didn't see it's used anywhere. Isn't `error` sufficient?

ramineni: you are right, not used. Ill update

pkg/csi/cinder/identityserver.go
```diff
@@ -47,11 +47,15 @@ func (ids *identityServer) GetPluginInfo(ctx context.Context, req *csi.GetPlugin
...
+		klog.V(3).Infof("Failed to query blockstorage API: %v", err)
```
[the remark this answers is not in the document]
ramineni: ok

pkg/csi/cinder/identityserver.go
mtanino: Thanks for the info. This param works for me.
```
	// Persistent service, so we need to be able to renew tokens.
	ao.AllowReauth = true
```

PR #953, opened 2020-03-03
[k8s-keystone-auth] Support role mapping

lingxiankong:
**The binaries affected**:

- [ ] cinder-csi-plugin
- [x] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

Support conversion from Keystone roles to Kubernetes users or groups during authentication, so that the Kubernetes cluster admin could config RBAC based on Keystone roles. Please see more details in the doc.

fixes #

Other changes in this PR:

- Fetch user's groups from Keystone.
- Use kebab-case for the config options
- Fix the import alias
- Rename pkg/identity/keystone/syncconfig.yaml to pkg/identity/keystone/sync_test.yaml because it's only used for UT.

**Release note**:
```release-note
NONE
```


pkg/util/util.go
```diff
@@ -25,3 +25,13 @@ func StringListEqual(list1, list2 []string) bool {
 
 	return s1.Equal(s2)
 }
+
+// Contains searches if a string list contains the given string or not.
+func Contains(list []string, strToSearch string) bool {
```
kayrus: there is already a defined func with the same functionality:
https://github.com/kubernetes/cloud-provider-openstack/blob/d38032ff1a7da28a0648d8f6e3ac5d3750f1a53f/pkg/autohealing/utils/utils.go#L20

lingxiankong: Thanks @kayrus, the plan is to move the common utility functions to a single place gradually. That said, we should remove `Contains` function for autohealing and use this one instead, but that's another PR.

lingxiankong:
The job cloud-provider-openstack-acceptance-test-k8s-cinder has problem that is going to be fixed in https://github.com/theopenlab/openlab-zuul-jobs/pull/809

lingxiankong:
/override cloud-provider-openstack-acceptance-test-k8s-cinder

just a test

lingxiankong:
/override openlab/cloud-provider-openstack-acceptance-test-k8s-cinder

This patch has nothing to do with in-tree cinder, so override the job here.

go.sum
```diff
@@ -545,6 +545,8 @@ github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXf
 github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
 github.com/stretchr/testify v1.4.0 h1:2E4SXV/wtOkTonXsotYi4li6zVWxYlZuYNCXe9XRJyk=
```
ramineni: @lingxiankong looks like old dependencies still exist, please run go mod tidy to cleanup the same.

lingxiankong: weird, I scrolled up a little bit, found there are many modules that have different versions, did you happen to know the reason?

ramineni: ya, strange , I thought go mod tidy should do required cleanup of the older versions?
One reason may be older versions are some indirect dependency of some package, but not sure

ramineni:
cc @chrigl

go.sum
kayrus: it is only a hashsum, not a direct dependency. if you remove go.sum and run `go mod tidy` again, it should cleanup old dep hashes.

lingxiankong: @kayrus Do you mean we need to remove go.sum file every time before modifying go.mod in order to remove those hashsum? Does the hashsum bring any effect to the build process?

kayrus: > Does the hashsum bring any effect to the build process?
no
> Do you mean we need to remove go.sum file every time before modifying go.mod in order to remove those hashsum?
I'd recommend to ignore this next time

Issue #883, opened 2019-12-19
Add arm64 images to docker.io/k8scloudprovider


xin3liang:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind feature

**What happened**:
There is no openstack cloud provider images for arm64 arch on  docker.io/k8scloudprovider.

**What you expected to happen**:
As arm64 is becoming a more and more popular arch. Arm64 images should be built and put into docker.io/k8scloudprovider. 
So that the users and downstream project (such as magnum) can use them easily on arm64.

**How to reproduce it**:

**Anything else we need to know?**:

**Environment**:
- openstack-cloud-controller-manager version:
- OpenStack version: master (ussuri)
- Others:

ricolin:
@xin3liang  are you plan to work on this or you wish others to?:)

xin3liang:
> @xin3liang are you plan to work on this or you wish others to?:)

I'm happy to do this work if I know how to do it :-).
IMO, here are some things might needed to do:
1, Make [dockerfiles](https://github.com/kubernetes/cloud-provider-openstack/tree/master/cluster/images) can be built on other archs, such as arm64;
2, Some CI integration work might need to do, if images are pushed to docker.io/k8scloudprovider automatically by CI;
3, Add tests for other archs?

ricolin:
@xin3liang I updated the patch out since that's happened to relate to some of my tasks. Please check the pull request.

xin3liang:
> @xin3liang I updated the patch out since that's happened to relate to some of my tasks. Please check the pull request.

@ricolin That really great! Thanks for you work.

lingxiankong:
Re-open this issue to track other tasks relating to CI jobs, etc.

lingxiankong:
@ricolin Please re-use this issue for other tasks.

ramineni:
@ricolin This seems to breaking the CI, https://logs.openlabtesting.org/logs/52/952/fdfdde5f8469461c67a4c76aac190af1db86d535/cloud-provider-openstack-e2e-test-csi-cinder/cloud-provider-openstack-e2e-test-csi-cinder/2854b4e/job-output.txt.gz

+ docker build -t docker.io/k8scloudprovider/cinder-csi-plugin:latest cluster/images/cinder-csi-plugin
[...]

PR #962, opened 2020-03-08
[OCCM] Improve documentation

lingxiankong:
**The binaries affected**:

- [ ] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

- Merge several files for openstack-cloud-controller-manager to make it easy to read and find useful information in a single place.
- Add ToC in the markdown file.

fixes #

**Release note**:
```release-note
NONE
```


docs/provider-configuration.md
```diff
@@ -179,8 +179,21 @@ These configuration options for the OpenStack provider pertain to the load
 balancer and should appear in the `[LoadBalancer]` section of the `$CLOUD_CONFIG`
 file.
 
+Although the openstack-cloud-controller-manager was initially implemented with
+Neutron-LBaaS support, Octavia is recommended now because Neutron-LBaaS has
+been deprecated since Queens OpenStack release cycle and not longer accepted
```
ramineni: s/not longer expected/no longer accepts

docs/provider-configuration.md
```diff
@@ -179,8 +179,21 @@ These configuration options for the OpenStack provider pertain to the load
...
+based on Octavia enabled OpenStack environment. Functionalities are not
+guaranteed if using Neutron-LBaaS.
+
 ##### Load Balancer Optional Parameters
 
+* `use-octavia`: Used to determine whether to look for and use an
+  Octavia LBaaS V2 service catalog endpoint. Valid values are `true` or `false`.
+  Where `true` is specified and an Octavia LBaaS V2 entry can not be found, the
+  provider will fall back and attempt to find a Neutron LBaaS V2 endpoint
+  instead. The default value is `false`.
```
ramineni: @lingxiankong so by default code uses neutron Lbaas? shouldnt we change to default to octavia as the prior one is deprecated?

lingxiankong: Yeah, sounds good to me, we are still getting some that related to Neutron-LBaaS which is annoying. I will submit a patchset with that information.

jichenjc: +1 to default to Octavia
and just to confirm (not sure ) we don't have anything that only on neutron-lbaas but not on octavia, correct?

lingxiankong: No, we don't

jichenjc: ok, +1 to me ,thanks

lingxiankong:
@ramineni @chrigl

chrigl:
@lingxiankong I guess, defaulting to Octavia will surprise people. Should we maybe highlight this in README.md?

lingxiankong:
> @lingxiankong I guess, defaulting to Octavia will surprise people. Should we maybe highlight this in README.md?

Sure, I will add some description in `using-openstack-cloud-controller-manager.md`

lingxiankong:

This job needs to be fixed in openlab as part of https://github.com/kubernetes/cloud-provider-openstack/issues/883

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -365,6 +365,11 @@ func ReadConfig(config io.Reader) (Config, error) {
 		return Config{}, fmt.Errorf("no OpenStack cloud provider config file given")
 	}
 	var cfg Config
+
+	// Set default values
+	cfg.LoadBalancer.UseOctavia = true
+	cfg.LoadBalancer.InternalLB = false
```
jichenjc: I understand this is improve doc so I think we might want to separate those 2 
one is to set default value (code change) and one is overall doc update.

lingxiankong: This is part of the intentional change, as we said in doc we will support octavia by default.

jichenjc: ok, no strong opinion , but usually refactory the doc and do function changes should be separated

ramineni: @lingxiankong As its breaking change for existing users of Neutron Lbaas, IMO, separate PR would be good , where we can add release note mentioning the same, Add PR description/title accordingly, would be easy to track which PR introduced the change.
And also, if they still want to use, they just need to change the flag to false right , and we can remove the full support after 2 releases or so once we update the documentation on deprecation. Thoughts?

lingxiankong: ok, thanks @jichenjc @ramineni

lingxiankong:
/approve

Will submit another PR to make use-octavia=true the default option.

PR #972, opened 2020-03-13
Fix the file filtering in CI jobs

lingxiankong:
**The binaries affected**:

- Do not run cloud-provider-openstack-unittest and cloud-provider-openstack-format for shell script
- Disable the job for Kubernetes in-tree cinder function. I don't know when and why it was added, but don't think it's needed for CPO.
- Add files filtering for the job cloud-provider-openstack-acceptance-test-csi-cinder
- Do not use `^` in the files regular expression for some jobs, I found sometimes that's not working

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -8,19 +8,20 @@
               - ^OWNERS$
               - ^SECURITY_CONTACTS$
               - ^.gitignore$
+              - ^.*\.sh$
     cloud-provider-openstack-acceptance-test-lb-octavia:
       jobs:
         - cloud-provider-openstack-acceptance-test-lb-octavia:
             files:
-              - ^cmd/openstack-cloud-controller-manager/.*
-              - ^pkg/cloudprovider/.*
-              - ^pkg/util/.*
-              - ^tests/e2e/cloudprovider/.*
+              - cmd/openstack-cloud-controller-manager/.*
+              - pkg/cloudprovider/.*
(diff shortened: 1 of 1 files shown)


lingxiankong:
@ramineni @chrigl

.zuul.yaml
```diff
@@ -50,28 +52,37 @@
...
+    #           - ^docs/.*$
+    #           - ^.*\.md$
+    #           - ^OWNERS$
+    #           - ^SECURITY_CONTACTS$
+    #           - ^.gitignore$
     cloud-provider-openstack-acceptance-test-csi-cinder:
       jobs:
         - cloud-provider-openstack-acceptance-test-csi-cinder:
+            files:
+              - cmd/cinder-csi-plugin/.*
```
ramineni: pkg/cloudprovider/providers/openstack is also needed

lingxiankong: I can't find it's used, cinder-csi has a copy of openstack package.

ramineni: ya, inside that its used :) , https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/openstack/openstack.go#L30

lingxiankong: such a bad design :-(

ramineni: ya, refactor is necessary

PR #974, opened 2020-03-16
[cinder-csi-plugin] upgrade snapshot to v2.1.1 

jichenjc:
**The binaries affected**:

IMPORTANT: Please also add the binary name in the title, e.g.
`[openstack-cloud-controller-manager]: Add UDP protocol support`
unless the PR affects multiple binaries.

- [ ] openstack-cloud-controller-manager
- [x] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] octavia-ingress-controller
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

**What this PR does / why we need it**:
update snapshot side car container to v2.1.1

upgrade snapshot to v2.1.1

    https://github.com/kubernetes-csi/external-snapshotter/releases
    has a broken update when upgrade to 2.0.0

    so the upgrade of the side car container including:
    1) upgrade the version
    2) honor the split of snapshot controller and snapshotter
    3) update rbac
    4) update example according to the v1beta1 definition

**Which issue this PR fixes**:
xref #951 

**Special notes for reviewers**:

**Release note**:

```release-note
[cinder-csi-plugin] updated external-snapshotter to v2.1.1
```

PR #976, opened 2020-03-17
[cinder-csi-pugin] Support reconcile volume attachements

takmatsu:
**The binaries affected**:

- [x] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

For supporting VolumeAttachment reconcile, adding capability.
Related: https://github.com/kubernetes-csi/external-attacher/pull/184

**Release note**:

```release-note
[cinder-csi-pugin] Support LIST_VOLUMES_PUBLISHED_NODES capability.
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/cloud-provider-openstack/976)

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -269,6 +269,13 @@ func (cs *controllerServer) ListVolumes(ctx context.Context, req *csi.ListVolume
 				CapacityBytes: int64(v.Size * 1024 * 1024 * 1024),
 			},
 		}
+
+		status := &csi.ListVolumesResponse_VolumeStatus{}
+		for _, attachment := range v.Attachments {
+			status.PublishedNodeIds = append(status.PublishedNodeIds, attachment.ServerID)
+		}
+		ventry.Status = status
+
 		ventries = append(ventries, &ventry)
 	}
 	return &csi.ListVolumesResponse{
--- a/pkg/csi/cinder/controllerserver_test.go
+++ b/pkg/csi/cinder/controllerserver_test.go
@@ -327,12 +327,16 @@ func TestListVolumes(t *testing.T) {
 					VolumeId:      FakeVol1.ID,
 					CapacityBytes: int64(FakeVol1.Size * 1024 * 1024 * 1024),
 				},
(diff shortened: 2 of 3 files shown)


pkg/csi/cinder/controllerserver.go
```diff
@@ -261,6 +261,13 @@ func (cs *controllerServer) ListVolumes(ctx context.Context, req *csi.ListVolume
 				CapacityBytes: int64(v.Size * 1024 * 1024 * 1024),
 			},
 		}
+
+		status := &csi.ListVolumesResponse_VolumeStatus{}
+		for _, attachment := range v.Attachments {
```
jichenjc: just curious , this attachments means we considered multi attach feature?
one volume attached to multiple node?

takmatsu: I also think normally block storage is attached to one node.
But cinder response supports multiple attachments in API level.
https://docs.openstack.org/api-ref/block-storage/v3/index.html?expanded=list-accessible-volumes-with-details-detail#list-accessible-volumes-with-details
and actually some driver may support multi-attach:
https://docs.openstack.org/cinder/latest/admin/blockstorage-volume-multiattach.html

lingxiankong:
Does this require a minimal version of external-attacher in order to support this capability?

jichenjc:
good point, 
we are using 1.4.0
https://github.com/kubernetes/cloud-provider-openstack/pull/974/files#diff-b50205da73c11460aff7f62130649d99R50

and the feature is after that, maybe 2.0.0 or later, so we should incresae the attacher version
I have a https://github.com/kubernetes/cloud-provider-openstack/pull/974
which do the snapshot version increase, maybe create a new one for provisioner or combine those 2 together

lingxiankong:
@jichenjc so could we hold this one until yours are merged?

jichenjc:
@lingxiankong thanks ,did so just now

takmatsu:
@jichenjc it's ok to remove hold now?

ramineni:
@takmatsu pls rebase on latest master

PR #985, opened 2020-03-24
[magnum-auto-healer] Support updating cluster health status

### LINKED ISSUE #984: [magnum-auto-healer] Support updating cluster health status
/kind feature

**The binaries affected**:

- [ ] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [x] magnum-auto-healer
- [ ] barbican-kms-plugin

Magnum supports private cluster which means the Magnum control plane can reach the k8s cluster to get the cluster health status. As a result, Magnum cannot update the cluster health status. Aa a controller running inside of the cluster, magnum-auto-healer can get a good insight into the cluster health status. So it would be nice if the magnum-auto-healer can support this.

openstacker:
- [ ] All
- [ ] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [x] magnum-auto-healer
- [ ] barbican-kms-plugin

fixes #984 

**Release note**:
```release-note
[magnum-auto-healer] Support updating the health status for Magnum cluster.
```

### DIFF
--- a/pkg/autohealing/cloudprovider/openstack/provider.go
+++ b/pkg/autohealing/cloudprovider/openstack/provider.go
@@ -18,6 +18,7 @@ package openstack
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"strings"
@@ -161,6 +162,26 @@ func (provider OpenStackCloudProvider) waitForServerPoweredOff(serverID string, 
 	return err
 }
 
+// waitForClusterStatus checks periodically to see if the cluster has entered a given status.
+// Returns when the status is observed or the timeout is reached.
+func (provider OpenStackCloudProvider) waitForClusterComplete(clusterID string, timeout time.Duration) error {
+	log.V(2).Infof("Waiting for cluster %s in complete status", clusterID)
+
+	err := wait.Poll(3*time.Second, timeout,
+		func() (bool, error) {
+			cluster, err := clusters.Get(provider.Magnum, clusterID).Extract()
+			log.V(5).Infof("Cluster %s in status %s", clusterID, cluster.Status)
+			if err != nil {
+				return false, err
+			}
+			if strings.HasSuffix(cluster.Status, "_COMPLETE") {
+				return true, nil
+			}
+			return false, nil
+		})
+	return err
+}
+
 // For master nodes: Soft deletes the VMs, marks the heat resource "unhealthy" then trigger Heat stack update in order to rebuild
 // the VMs. The information this function needs:
 //     - Nova VM IDs
@@ -171,13 +192,26 @@ func (provider OpenStackCloudProvider) Repair(nodes []healthcheck.NodeInfo) erro
 		return nil
 	}
 
+	masters := []healthcheck.NodeInfo{}
+	workers := []healthcheck.NodeInfo{}
+
 	clusterName := provider.Config.ClusterName
+	isWorkerNode := nodes[0].IsWorker
+	if isWorkerNode {
+		workers = nodes
+	} else {
+		masters = nodes
+	}
+
+	err := provider.UpdateHealthStatus(masters, workers)
+	if err != nil {
+		return fmt.Errorf("Failed to update the helath status of cluster %s, error: %v", clusterName, err)
+	}
+
 	cluster, err := clusters.Get(provider.Magnum, clusterName).Extract()
 	if err != nil {
 		return fmt.Errorf("failed to get the cluster %s, error: %v", clusterName, err)
 	}
-
-	isWorkerNode := nodes[0].IsWorker
 
 	if isWorkerNode {
 		for _, n := range nodes {
@@ -304,6 +338,66 @@ func (provider OpenStackCloudProvider) getNodeGroup(clusterName string, node hea
 	return ng, fmt.Errorf("failed to find node group")
 }
 
+// UpdateHealthStatus can update the cluster health status to reflect the
+// real-time health status of the k8s cluster.
+func (provider OpenStackCloudProvider) UpdateHealthStatus(masters []healthcheck.NodeInfo, workers []healthcheck.NodeInfo) error {
+	log.Infof("start to update cluster health status.")
+	clusterName := provider.Config.ClusterName
+
+	healthStatus := "UNHEALTHY"
+	healthStatusReasonMap := make(map[string]string)
+	healthStatusReasonMap["updated_at"] = time.Now().String()
+
+	if len(masters) == 0 && len(workers) == 0 {
+		// No unhealthy node passed in means the cluster is healthy
+		healthStatus = "HEALTHY"
+		healthStatusReasonMap["api"] = "ok"
+		healthStatusReasonMap["nodes"] = "ok"
+	} else {
+		if len(workers) > 0 {
+			for _, n := range workers {
+				// TODO: Need to figure out a way to reflect the detailed error information
+				healthStatusReasonMap[n.KubeNode.Name+"."+n.FailedCheck] = "error"
+			}
+		} else {
+			// TODO: Need to figure out a way to reflect detailed error information
+			healthStatusReasonMap["api"] = "error"
+		}
+	}
+
+	jsonDumps, err := json.Marshal(healthStatusReasonMap)
+	if err != nil {
+		return fmt.Errorf("failed to build health status reason for cluster %s, error: %v", clusterName, err)
+	}
+	healthStatusReason := strings.Replace(string(jsonDumps), "\"", "'", -1)
+
+	updateOpts := []clusters.UpdateOptsBuilder{
+		clusters.UpdateOpts{
+			Op:    clusters.ReplaceOp,
+			Path:  "/health_status",
+			Value: healthStatus,
+		},
+		clusters.UpdateOpts{
+			Op:    clusters.ReplaceOp,
+			Path:  "/health_status_reason",
+			Value: healthStatusReason,
+		},
+	}
+
+	log.Infof("updating cluster health status as %s for reason %s.", healthStatus, healthStatusReason)
+	res := clusters.Update(provider.Magnum, clusterName, updateOpts)
+
+	if res.Err != nil {
+		return fmt.Errorf("failed to update the health status of cluster %s error: %v", clusterName, res.Err)
+	}
+
+	if err := provider.waitForClusterComplete(clusterName, 30*time.Second); err != nil {
+		log.Warningf("failed to wait the cluster %s in complete status, error: %v", clusterName, err)
+	}
+
+	return nil
+}
+
 // Enabled decides if the repair should be triggered.
 // There are  two conditions that we disable the repair:
 // - The cluster admin disables the auto healing via OpenStack API.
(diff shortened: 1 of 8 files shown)


openstacker:
This PR depends on the work in Magnum https://review.opendev.org/#/c/710384/

pkg/autohealing/cloudprovider/openstack/provider.go
```diff
@@ -168,13 +193,21 @@ func (provider OpenStackCloudProvider) Repair(nodes []healthcheck.NodeInfo) erro
...
+	if err != nil {
+		return fmt.Errorf("Failed to update the helath status of cluster %s, error: %v", clusterName, err)
+	}
+
 	cluster, err := clusters.Get(provider.Magnum, clusterName).Extract()
 	if err != nil {
 		return fmt.Errorf("failed to get the cluster %s, error: %v", clusterName, err)
 	}
 
+	if err := provider.waitForClusterComplete(cluster.UUID, 30*time.Second); err != nil {
```
lingxiankong: Why we need to call this function here? Shouldn't it be part of checks in `Enabled()` function?

pkg/autohealing/controller/controller.go
```diff
@@ -340,6 +344,8 @@ func (c *Controller) startMasterMonitor(wg *sync.WaitGroup) {
 		return
 	}
 
+	totalUnhealthyNodes = append(totalUnhealthyNodes, unhealthyNodes...)
```
lingxiankong: `append` function is not goroutine-safe.

pkg/autohealing/controller/controller.go
```diff
@@ -371,6 +379,10 @@ func (c *Controller) startWorkerMonitor(wg *sync.WaitGroup) {
 	log.V(3).Info("Finished checking worker nodes.")
 }
 
+func (c *Controller) updateClusterHealthStatus() {
+	c.provider.UpdateHealthStatus(totalUnhealthyNodes)
```
lingxiankong: We don't need an extra function here.

Why do we need to call UpdateHealthStatus function three times (master+worker+total)?

lingxiankong:
> This PR depends on the work in Magnum https://review.opendev.org/#/c/710384/

Please mark this PR WIP until the magnum patch merged and backported to stable branch.

lingxiankong:
Suggested release note (guide: https://github.com/kubernetes/community/blob/master/contributors/guide/release-notes.md):

```
magnum-auto-healer: Added health status update for the cluster before repair.
```

pkg/autohealing/cloudprovider/openstack/provider.go
openstacker: Yep, `Enabled()` could be a better place. But no matter we put the wait here or in Enabled(), we do need to have this check because the health status update won't trigger the Heat stack update, so the Heat statck status won't be changed. That means the existing stack status check in Enabled is not enough.

I will move the check to Enabled().

pkg/autohealing/controller/controller.go
openstacker: Good point. I will think about how to workaround it.

pkg/autohealing/controller/controller.go
openstacker: Yep, we should be able to drop the `updateClusterHealthStatus()` for Controller and then call the provider's `UpdateClusterHealthStatus()` directly.

The `UpdateHealthStatus()` could be dropped from the Repair()  function and only keep the one in Controller. I will test it and fix it in the next commit.

openstacker:
> Suggested release note (guide: https://github.com/kubernetes/community/blob/master/contributors/guide/release-notes.md):
> 
> ```
> magnum-auto-healer: Added health status update for the cluster before repair.
> ```

openstacker:
The API change in Magnum has been merged.

pkg/autohealing/cloudprovider/openstack/provider.go
```diff
@@ -262,6 +296,66 @@ func (provider OpenStackCloudProvider) Repair(nodes []healthcheck.NodeInfo) erro
...
+	healthStatusReasonMap := make(map[string]string)
+	healthStatusReasonMap["updated_at"] = time.Now().String()
+
+	if len(masters) == 0 && len(workers) == 0 {
+		// No unhealthy node passed in means the cluster is healthy
+		healthStatus = "HEALTHY"
+		healthStatusReasonMap["api"] = "ok"
+		healthStatusReasonMap["nodes"] = "ok"
+	} else {
+		healthStatus = "UNHEALTHY"
```
lingxiankong, line 314: Duplicate with Line 304

pkg/autohealing/cloudprovider/openstack/provider.go
```diff
@@ -262,6 +296,66 @@ func (provider OpenStackCloudProvider) Repair(nodes []healthcheck.NodeInfo) erro
 	return nil
 }
 
+// UpdateHealthStatus can update the cluster health status by
```
lingxiankong, line 299: Incomplete doc string?

pkg/autohealing/cloudprovider/openstack/provider.go
```diff
@@ -262,6 +296,66 @@ func (provider OpenStackCloudProvider) Repair(nodes []healthcheck.NodeInfo) erro
...
+			}
+		} else {
+			// TODO: Need to figure out a way to reflect detailed error information
+			healthStatusReasonMap["api"] = "error"
+		}
+	}
+
+	jsonDumps, err := json.Marshal(healthStatusReasonMap)
+	if err != nil {
+		return fmt.Errorf("Failed to build health status reason for cluster %s, error: %v", clusterName, err)
```
lingxiankong, line 328: Error string should not be capitalized, so s/Failed/failed, please change others as well.

pkg/autohealing/controller/controller.go
```diff
@@ -65,6 +65,11 @@ const (
 	LabelNodeRoleMaster = "node-role.kubernetes.io/master"
 )
 
+var (
+	masterUnhealthyNodes []healthcheck.NodeInfo
+	workerUnhealthyNodes []healthcheck.NodeInfo
+)
+
```
lingxiankong, line 72: Do we really need those two vars here given they are defined in L392?

pkg/autohealing/controller/controller.go
```diff
@@ -392,6 +403,7 @@ func (c *Controller) Start(ctx context.Context) {
 			}
 
 			wg.Wait()
+			c.provider.UpdateHealthStatus(masterUnhealthyNodes, workerUnhealthyNodes)
```
lingxiankong, line 406: It would be better to update cluster status before the repair action is triggered, the repair action is a long time operation, I know it's limited by the current goroutine repair design, do you think it's acceptable for now?

pkg/autohealing/healthcheck/healthcheck.go
```diff
@@ -76,6 +79,7 @@ func CheckNodes(checkers []HealthCheck, nodes []NodeInfo, controller NodeControl
 	for _, node := range nodes {
 		for _, checker := range checkers {
 			if !checker.Check(node, controller) {
+				node.FailedCheck = reflect.TypeOf(checker).String()
```
lingxiankong, line 82: I would rather we add an extra method(e.g. `GetName()`) in HealthCheck interface to get a meaningful checker type or name.

pkg/autohealing/cloudprovider/openstack/provider.go
```diff
@@ -34,6 +35,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
```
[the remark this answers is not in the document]
openstacker: Will do.

pkg/autohealing/cloudprovider/openstack/provider.go
```diff
@@ -171,13 +193,25 @@ func (provider OpenStackCloudProvider) Repair(nodes []healthcheck.NodeInfo) erro
 	}
 
 	clusterName := provider.Config.ClusterName
+	masters := nodes
+	workers := nodes
+	isWorkerNode := nodes[0].IsWorker
+	if isWorkerNode {
+		masters = []healthcheck.NodeInfo{}
+	} else {
+		workers = []healthcheck.NodeInfo{}
```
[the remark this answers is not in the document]
openstacker: I'm using 2 vars to avoid the thread-safe issue of list appending.

pkg/autohealing/cloudprovider/openstack/provider.go
openstacker: Will fix in next commit.

pkg/autohealing/cloudprovider/openstack/provider.go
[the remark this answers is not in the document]
openstacker: Will do.

pkg/autohealing/cloudprovider/openstack/provider.go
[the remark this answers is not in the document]
openstacker: Right, it's a bit confusing. My original purpose is initializing them at the same time then at line 200 and line 202 setting either masters or workers as EMPTY. I will see if there is a better way to express this.

pkg/autohealing/cloudprovider/openstack/provider.go
openstacker: Will fix it in the next commit.

pkg/autohealing/cloudprovider/openstack/provider.go
openstacker: Will remove it in next commit.

pkg/autohealing/cloudprovider/openstack/provider.go
openstacker: No problem. Will do.

pkg/autohealing/controller/controller.go
openstacker: They're global variables, called at line 348 and 372.

pkg/autohealing/healthcheck/healthcheck.go
openstacker: OK, I will do that in next commit.

pkg/autohealing/controller/controller.go
openstacker: I can see your point. My understanding is, given the repairing is a time-consuming operation, so it should be acceptable to do the status update here. I will do more testing about this. Thanks.

pkg/autohealing/controller/controller.go
lingxiankong: Right, they are defined here but re-initialized in each loop, I got it.

brtknr:
Ping @openstacker

brtknr:
Which tag will this be released under?

PR #986, opened 2020-03-24
[OCCM] Add annotation for creating health monitor

lingxiankong:
**The binaries affected**:

- [ ] All
- [x] openstack-cloud-controller-manager(OCCM)
- [ ] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

Add service annotation `loadbalancer.openstack.org/enable-health-monitor` to define whether or not to create health monitor for the load balancer pool, if not specified, use 'create-monitor' config. The health monitor can be created or deleted dynamically.

The scenaro to use this annotation is to deal with the pod that exposes TCP via SNI (e.g. Envoy) and is performing client certificate validation, in this case the health monitor created will fail. We need provide the capability to customize.

fixes #

**Release note**:

```release-note
openstack-cloud-controller-manager: Added a service annotation `loadbalancer.openstack.org/enable-health-monitor` to define whether or not to create health monitor for the load balancer pool.
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack.go
+++ b/pkg/cloudprovider/providers/openstack/openstack.go
@@ -366,9 +366,16 @@ func ReadConfig(config io.Reader) (Config, error) {
 	}
 	var cfg Config
 
-	// Set default values
+	// Set default values explicitly
 	cfg.LoadBalancer.UseOctavia = true
 	cfg.LoadBalancer.InternalLB = false
+	cfg.LoadBalancer.LBProvider = "amphora"
+	cfg.LoadBalancer.LBMethod = "ROUND_ROBIN"
+	cfg.LoadBalancer.CreateMonitor = false
+	cfg.LoadBalancer.ManageSecurityGroups = false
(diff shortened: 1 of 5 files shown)


pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -471,21 +478,6 @@ func readInstanceID(searchOrder string) (string, error) {
 
 // check opts for OpenStack
 func checkOpenStackOpts(openstackOpts *OpenStack) error {
-	lbOpts := openstackOpts.lbOpts
-
-	// if need to create health monitor for Neutron LB,
```
ramineni: Just to be sure, neutron LB still works if we remove this right? We are still supporting neutron LB , but just not the default option right

lingxiankong: The check here is removed because the default value is specified. It doesn't affect neutron-lbaas.

PR #1002, opened 2020-04-01
[cinder-csi-plugin] Allow force create snapshot

### LINKED ISSUE #832: Improve CSI Cinder Snapshot: add --force flag to create snapshots of in-use volumes
/kind feature

**What happened**:

The current implementation of snapshots only allows us to create snapshots of volumes which are not in-use.

It is possible to use a --force flag to create snapshots of volumes which are attached to an instance as described in https://docs.openstack.org/python-openstackclient/pike/cli/command-objects/volume-snapshot.html

Create a snapshot of an in-use volume

jichenjc:
**The binaries affected**:

- [ ] All
- [x] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] manila-csi-plugin
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

fixes #832 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/examples/cinder-csi-plugin/snapshot/example.yaml
+++ b/examples/cinder-csi-plugin/snapshot/example.yaml
@@ -11,6 +11,8 @@ kind: VolumeSnapshotClass
 metadata:
   name: csi-cinder-snapclass
 snapshotter: cinder.csi.openstack.org
+parameters:
+  snapshot-inuse-volumes: "false"
 
 ---
 
--- a/pkg/csi/cinder/openstack/openstack_snapshots.go
+++ b/pkg/csi/cinder/openstack/openstack_snapshots.go
@@ -20,6 +20,7 @@ package openstack
 
 import (
 	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/snapshots"
@@ -34,15 +35,26 @@ const (
 	snapReadySteps      = 10
 
 	snapshotDescription = "Created by OpenStack Cinder CSI driver"
+	snapshotForceCreate = "snapshot-inuse-volumes"
 )
 
 // CreateSnapshot issues a request to take a Snapshot of the specified Volume with the corresponding ID and
 // returns the resultant gophercloud Snapshot Item upon success
 func (os *OpenStack) CreateSnapshot(name, volID string, tags *map[string]string) (*snapshots.Snapshot, error) {
+
+	force := false
+	// if no flag given, then force will be false by default
+	// if flag it given , check it
+	if item, ok := (*tags)[snapshotForceCreate]; ok {
+		force, _ = strconv.ParseBool(item)
+		delete(*tags, snapshotForceCreate)
+	}
+	// Force the creation of snapshot even the Volume is in in-use state
 	opts := &snapshots.CreateOpts{
 		VolumeID:    volID,
 		Name:        name,
 		Description: snapshotDescription,
+		Force:       force,
 	}
 	if tags != nil {
 		opts.Metadata = *tags
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -25,7 +25,7 @@ fixes #
 **Release note**:
 <!--
 1. Release note is required if a significant change is introduced, otherwise please keep this section as is.
-2. Release note is in Markdown format and should begin with the binary name unless multiple binaries are affected, e.g. openstack-cloud-controller-manager: Deprecate Neutron-LBaaS support.
+2. Release note is in Markdown format and should begin with the binary name unless multiple binaries are affected, e.g. [openstack-cloud-controller-manager] Deprecate Neutron-LBaaS support.
 3. Instructions for writing a release note: https://git.k8s.io/community/contributors/guide/release-notes.md
 -->
 ```release-note
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -85,6 +85,18 @@
               - go.mod$
               - go.sum$
               - Makefile$
+    cloud-provider-openstack-multinode-csi-migration-test:
+      jobs:
+        - cloud-provider-openstack-multinode-csi-migration-test:
+            files:
+              - cmd/cinder-csi-plugin/.*
+              - manifests/cinder-csi-plugin/.*
+              - pkg/csi/cinder/.*
+              - pkg/util/.*
+              - go.mod$
+              - go.sum$
+              - Makefile$
+              - .zuul.yaml$
     cloud-provider-openstack-acceptance-test-csi-manila:
       jobs:
         - cloud-provider-openstack-acceptance-test-csi-manila:
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,7 @@ endif
 import-boss:
 ifndef HAS_IMPORT_BOSS
 	echo "installing import-boss"
-	GO111MODULE=off go get -u k8s.io/code-generator/cmd/import-boss
+	go get -u k8s.io/code-generator/cmd/import-boss
 endif
 	hack/verify-import-boss.sh
 
--- a/docs/getting-started-provider-dev.md
+++ b/docs/getting-started-provider-dev.md
@@ -251,7 +251,7 @@ cloud-controller-manager-roles.yaml
 you need use following command to create ClusterRole and ClusterRoleBinding
 otherwise the cloud-controller-manager is not able to access k8s API.
 ```
-./cluster/kubectl.sh create -f cluster/addons/rbac/
+./cluster/kubectl.sh create -f $working_dir/cloud-provider-openstack/cluster/addons/rbac/
 ```
 
 Have a good time with OpenStack and Kubernetes!
--- a/docs/using-cinder-csi-plugin.md
+++ b/docs/using-cinder-csi-plugin.md
@@ -32,7 +32,7 @@ Check [kubernetes CSI Docs](https://kubernetes-csi.github.io/docs/) for flag det
 
 ### Deploy
 
-If you already created the `cloud-config` secret used by the [cloud-controller-manager](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/using-controller-manager-with-kubeadm.md#steps), remove the file ```manifests/cinder-csi-plugin/csi-secret-cinderplugin.yaml``` and then jump directly to the `kubectl apply ...` command.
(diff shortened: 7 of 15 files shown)


lingxiankong:
Thanks for the PR, some questions:
- Do we know if there is any Cinder backend doesn't support creating snapshot when the volume in `in-use` status?
- How about changing the title to `[cinder-csi-plugin] Allow force create snapshot`?
- A release note should be added.

jichenjc:
> Thanks for the PR, some questions:
> 
> * Do we know if there is any Cinder backend doesn't support creating snapshot when the volume in `in-use` status?

this is a good question, I noticed the api doc is not correct then check the code, at least
I didn't see anything mentioned in the doc or code comments
I will add more info here after dig

pkg/csi/cinder/openstack/openstack_snapshots.go
```diff
@@ -39,10 +39,12 @@ const (
 // CreateSnapshot issues a request to take a Snapshot of the specified Volume with the corresponding ID and
 // returns the resultant gophercloud Snapshot Item upon success
 func (os *OpenStack) CreateSnapshot(name, volID string, tags *map[string]string) (*snapshots.Snapshot, error) {
+	// Force the creation of snapshot even the Volume is in in-use state
 	opts := &snapshots.CreateOpts{
 		VolumeID:    volID,
 		Name:        name,
 		Description: snapshotDescription,
+		Force:       true,
```
ramineni, line 47: @jichenjc instead of adding it by default, we have option in CreateSnapshotRequest which allows parameters which are opaque to CO but can be used by driver, we could use that  I suppose, if allow_inuse_volume_snapshot: true, then add force:true, something like that  https://github.com/container-storage-interface/spec/blob/master/spec.md

vkamra: @ramineni - most (if not all) use cases that involve snapshots (at least in Kubernetes) will target volumes that are in-use (PVC is bound and used by a pod).

Given that - I do think the default behavior should be to allow in-use and if for some reason we want to change it - we can leverage the opaque parameters.

ramineni: @vkamra we need to go by default behavior given by cinder , by default cinder doesnt allow creating snapshots for inuse volume, so the same behaviour should reflect here to avoid confusion.

vkamra: @ramineni - will defer to you and others here. I do expect most users in Kubernetes to want to enable this option.

@ramineni @jichenjc - would love to see this move forward. With the current limitation we're unable to leverage snapshots because it is not possible for us to scale down running applications in order to snapshot data.

lingxiankong: > @jichenjc instead of adding it by default, we have option in CreateSnapshotRequest which allows parameters which are opaque to CO but can be used by driver, we could use that I suppose, if allow_inuse_volume_snapshot: true, then add force:true, something like that https://github.com/container-storage-interface/spec/blob/master/spec.md

Do you mean the user needs to add `allow_inuse_volume_snapshot: true` explicitly in the request?

lingxiankong: > @ramineni - most (if not all) use cases that involve snapshots (at least in Kubernetes) will target volumes that are in-use (PVC is bound and used by a pod).
> 
> Given that - I do think the default behavior should be to allow in-use and if for some reason we want to change it - we can leverage the opaque parameters.

OpenStack Cinder may have different behaviors for different backends (still waiting for the answer from @jichenjc ), if that's the case, I'd rather we add a config option, leaving decision to the cluster admin.

jichenjc: I didn't find detail doc on the flag and from code
https://github.com/openstack/cinder/blob/master/cinder/volume/api.py#L879
looks like here will validate the flag 

then here will not handle this flag to lower layer
https://github.com/openstack/cinder/blob/master/cinder/volume/api.py#L864


so seems only check at API layer and not related to backend...

so maybe we can follow the way @lingxiankong mentioned,
add a configuration here to allow such behavior to be disbaled/enabled ? and previously we disable this so how about default to disable and allow admin to change to enable?

ramineni: >> Do you mean the user needs to add allow_inuse_volume_snapshot: true explicitly in the request? 
@lingxiankong yes

jichenjc: > CreateSnapshotRequest

yes, I think it's a reasonable request, I will follow up the way

jichenjc:
@ramineni @lingxiankong please help to take a quick look at new code
I didn't run FVT now just want to get rough understanding on this idea
BTW we used params for tags previously, now we will hack some fixed wording such as
allow_snapshot_inuse_volume ,maybe we should remove that from tags?

examples/cinder-csi-plugin/snapshot/example.yaml
```diff
@@ -11,6 +11,8 @@ kind: VolumeSnapshotClass
 metadata:
   name: csi-cinder-snapclass
 snapshotter: cinder.csi.openstack.org
+parameters:
+  - allow_inuse_volume_snapshot: false
```
ramineni, line 15: adding documentation of this parameter would be good.

jichenjc:
> > @ramineni @lingxiankong please help to take a quick look at new code
> > I didn't run FVT now just want to get rough understanding on this idea
> > BTW we used params for tags previously, now we will hack some fixed wording such as
> > allow_snapshot_inuse_volume ,maybe we should remove that from tags?
> 
> @jichenjc code looks ok to me.
> I didnt understand your question here?
> 
> > BTW we used params for tags previously, now we will hack some fixed wording such as
> > allow_snapshot_inuse_volume ,maybe we should remove that from tags?
> 
> I believe parameters is the only way to add options, is there any other way of passing allow_snapshot_inuse_volume ?

@ramineni here's the question
https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/openstack/openstack_snapshots.go#L52
will create tags from input
https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/openstack/openstack_snapshots.go#L47

and now we add this flag here, so we should not create tags or we need create tags for allow_snapshot_inuse_volume? both seems fine but need some agreement
both looks go

ramineni:
> @ramineni here's the question
> https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/openstack/openstack_snapshots.go#L52
> will create tags from input
> https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/openstack/openstack_snapshots.go#L47
> 
> and now we add this flag here, so we should not create tags or we need create tags for allow_snapshot_inuse_volume? both seems fine but need some agreement
> both looks go

ok, got it . ya , as you mentioned , adding this as part of metadata doesnt look good. IMO removing that before updating it to metadata would be good. 
If we are changing the map, better we dont pass the pointer here , https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/openstack/openstack_snapshots.go#L41 ,

pkg/csi/cinder/openstack/openstack_snapshots.go
lingxiankong: > > > Do you mean the user needs to add allow_inuse_volume_snapshot: true explicitly in the request?
> 
> @lingxiankong yes

I don't think it's a good idea, we shouldn't allow the user to enable/disable a feature that depends on the backend configuration.

jichenjc: this is not through backend, instead, through configuration 
which is volumestorageclass, will that be reasonable thing to do?

ramineni: @lingxiankong its independent of backend, and by default cinder doesnt allow creating snapshots for in-use volumes irrespective of backend.

examples/cinder-csi-plugin/snapshot/example.yaml
vkamra: @jichenjc - this should be camelCase with no underscores to align with Kubernetes naming conventions (see https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions). Maybe `allowInUseVolumeSnapshot`

ramineni: @vkamra that is for GO field names.
@jichenjc But , ya, we should not use `_` , it should be `-` instead . For example, https://github.com/kubernetes-csi/docs/blob/master/book/src/secrets-and-credentials-volume-snapshot-class.md#L43 .
Im ok with name `snapshot-inuse-volumes` too.

vkamra: @ramineni - you're right. `-` is the way to go.

jichenjc: ok, that's reasonable , will change accordingly

jichenjc:
my shared openstack env is broken , have to wait for a while to do test on this... code refreshed

lingxiankong:
I would suggest we use `force_create` in the parameters and remove the field before sending request to Cinder.

jichenjc:
I tested locally, by setting the value to false and true, got following and checked in GUI side 
if force is false (default) then create snapshot will fail while force is true is able to create snapshot

```
 Event(v1.ObjectReference{Kind:"VolumeSnapshot", Namespace:"default", Name:"new-snapshot-demo", UID:"e79c1732-ae03-4411-8ddf-cba7f264bf02", APIVersion:"snapshot.storage.k8s.io/v1alpha1", ResourceVersion:"838", FieldPath:""}): type: 'Warning' reason: 'SnapshotCreationFailed' Failed to create snapshot: failed to take snapshot of the volume, pvc-d4bfc1bd-5b96-40bf-99cb-67fa8bfc648e: "rpc error: code = Internal desc = CreateSnapshot failed with error Bad request with: [POST https://hdc-prod-cloud.hursley.ibm.com:8776/v3/07962130d7044e3c84e1825859d5bef9/snapshots], error message: {\"badRequest\": {\"message\": \"Invalid volume: Volume 18056a4d-9115-4415-9214-29d2149ac193 status must be available, but current status is: in-use.\", \"code\": 400}


I0417 04:01:35.494658       1 snapshot_controller.go:615] createSnapshot: Creating snapshot default/new-snapshot-demo through the plugin ...
```

jichenjc:
oops, git issue 
https://github.com/kubernetes/cloud-provider-openstack/pull/1034
repushed and will close this one

PR #1008, opened 2020-04-03
Bump Go version in make.sh helper script

### LINKED ISSUE #1007: Helper make.sh script fails due to the use of Go 1.12

/kind bug

**What happened**:
When running the helper script `./hack/make.sh` for building and compiling the `openstack-cloud-controller-manager` binary, via a Docker container, the following error occurs:

`/go/pkg/mod/k8s.io/apimachinery@v0.18.0/pkg/util/errors/errors.go:99:10: undefined: errors.Is` 

This `errors.Is()` function was added to the `errors` package in Go `1.13` and at the moment the `make.sh` is hardcoded to use Go `1.12`. Note that apimachinery started to use `errors.Is` as of `v0.18.0`, basically the latest version.

Here's the `make.sh` file:
https://github.com/kubernetes/cloud-provider-openstack/blob/466040afb25d04184267b13da183f42d9a923eed/hack/make.sh#L23

That there's no errors when compiling using the helper script.

**How to reproduce it**:
On e.g. release branch `release-1.18` running `./hack/make.sh openstack-cloud-controller-manager`.

I've not tested to build the rest of the binaries using the helper script, we could consider adding the Go version as a parameter. I can create a PR for this since it's a easy fix to bump the Go version but i'm not sure if this script is used in the build system and something might break there.

mikejoh:
See #1007

Fixes #1007 

We could consider adding a argument to pass in the Go-version, not sure how this script is used out there though. I'm using it occasionally.

**Release note**:
```release-note
None
```

### DIFF
--- a/hack/make.sh
+++ b/hack/make.sh
@@ -20,4 +20,4 @@
 # GOPATH and handle all of the Go ENV stuff for you.  All you need is Docker
 docker run -it -v "$PWD":/go/src/k8s.io/cloud-provider-openstack:z \
 	-w /go/src/k8s.io/cloud-provider-openstack \
-	golang:1.12 make $1
+	golang:1.13 make $1


lingxiankong:
can we use `Bump Go version to 1.13 in make.sh helper script` for more explicit?

mikejoh:
Of course, i've changed the release note!

ramineni:
I dont think release note is needed in this case @lingxiankong ?
/lgtm

lingxiankong:
> I dont think release note is needed in this case @lingxiankong ?
> /lgtm

I agree @ramineni, considering release notes are required for any pull request with user-visible changes, I will remove this one and merge.

PR #1016, opened 2020-04-09
Update the doc of getting-started-provider-dev

sxmatch:
Update the instruction "./cluster/kubectl.sh create -f cluster/addons/rbac/"
to avoid error arised.

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/getting-started-provider-dev.md
+++ b/docs/getting-started-provider-dev.md
@@ -251,7 +251,7 @@ cloud-controller-manager-roles.yaml
 you need use following command to create ClusterRole and ClusterRoleBinding
 otherwise the cloud-controller-manager is not able to access k8s API.
 ```
-./cluster/kubectl.sh create -f cluster/addons/rbac/
+./cluster/kubectl.sh create -f $working_dir/cloud-provider-openstack/cluster/addons/rbac/
 ```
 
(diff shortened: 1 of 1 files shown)


docs/getting-started-provider-dev.md
```diff
@@ -251,7 +251,8 @@ cloud-controller-manager-roles.yaml
 you need use following command to create ClusterRole and ClusterRoleBinding
 otherwise the cloud-controller-manager is not able to access k8s API.
 ```
-./cluster/kubectl.sh create -f cluster/addons/rbac/
+./cluster/kubectl.sh create -f https://raw.githubusercontent.com/kubernetes/cloud-provider-openstack/master/cluster/addons/rbac/cloud-controller-manager-roles.yaml
```
ramineni, line 254: @sxmatch generally it is better to use the yamls of that particular version they have deployed not of master everytime. So, instead of adding pointing to master branch files, to make it more clear , how about we could add something like below :
`  $ cd $working_dir/cloud-provider-openstack
   $ kubectl create -f cluster/addons/rbac/ `

sxmatch: sure, this is better, thanks

Issue #765, opened 2019-09-17
Cinder CSI not availability zone aware


piwi91:
/kind bug

**What happened**:

When moving from in-tree `kubernetes.io/cinder` to `csi-cinderplugin`, the Cinder CSI plugin doesn't respect the availability zone of the statefulset. When `kubernetes.io/cinder` is used, this is not a problem.

**What you expected to happen**:

I would expect that the CSI Cinder plugin forces the volume and statefulset pods to be in the same availability zone.

**How to reproduce it (as minimally and precisely as possible)**:

```
---
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: csi-archive
parameters:
  fsType: ext4
  type: archive
provisioner: csi-cinderplugin
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: my-app
spec:
  serviceName: my-app
  replicas: 1
  selector:
    matchLabels:
      app: my-app
  template:
    metadata:
      labels:
        app: my-app
    spec:
      containers:
      - image: nginx
        imagePullPolicy: IfNotPresent
        name: nginx
        ports:
        - containerPort: 80
          protocol: TCP
        volumeMounts:
        - mountPath: /var/lib/www/html
          name: my-app-vol
  volumeClaimTemplates:
  - metadata:
      name: my-app-vol
    spec:
      accessModes:
        - ReadWriteOnce
      resources:
        requests:
          storage: 1Gi
      storageClassName: csi-archive
```

**Environment**:
- openstack-cloud-controller-manager version:
`1.15.0`

- OS (e.g. from /etc/os-release): 
```
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"
```

- Kernel (e.g. `uname -a`):
[...]

PR #1019, opened 2020-04-09
[cinder-csi-plugin] enable availability zone topology awareness

kayrus:
* cinder-csi-plugin

Enable `--feature-gates=Topology=true` for cinder CSI provisioner by default

Though https://kubernetes-csi.github.io/docs/topology.html states that this feature gate is enabled by default, in my 1.17 cluster I had to explicitly enable it.

#765

**Release note**:

```release-note
[cinder-csi-controller] enable topology feature gate by default
```

### DIFF
--- a/manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml
+++ b/manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml
@@ -50,6 +50,7 @@ spec:
           image: quay.io/k8scsi/csi-provisioner:v1.6.0
           args:
             - "--csi-address=$(ADDRESS)"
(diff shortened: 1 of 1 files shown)


ramineni:
@kayrus its not added intentionally though , as anyone require would add it explicitly. But I dont see any harm in adding by default , may be we can update it here as well, https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/using-cinder-csi-plugin.md#enable-topology-aware-dynamic-provisioning-for-cinder-volumes , stating added by default in manifests , but can be disabled if not required.

ramineni:
@kayrus could you do quick update in doc so we can merge this

ramineni:
/close
No update

PR #1021, opened 2020-04-09
[cinder-csi-plugin] Allow disabling particular driver services

rfranzke:
This PR adds two new command line flags `--run-controller-service` and `--run-node-service` (both default to `true`) which allow to particularly disable individual services if not required.

This enables the use-case of running the CSI controllers (csi-provisioner, csi-attacher, etc., + the driver controller) separately outside of the cluster they are serving, and not necessarily on an OpenStack nova VM. The node service part of the CSI driver will still inside the cluster for mounting/unmounting the volume, etc.

**Release note**:

```release-note
For the cinder CSI plugin, it is now possible to disable the controller service by setting `--run-controller-service=false`. Similarly, it is possible to disable the node service by setting `--run-node-service=false`. The latter enables running the controller server of the Cinder CSI driver separately/outside of the cluster it is serving.
```

### DIFF
--- a/cmd/cinder-csi-plugin/main.go
+++ b/cmd/cinder-csi-plugin/main.go
@@ -21,20 +21,23 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/spf13/cobra"
-	"github.com/spf13/pflag"
 	"k8s.io/cloud-provider-openstack/pkg/csi/cinder"
 	"k8s.io/cloud-provider-openstack/pkg/csi/cinder/openstack"
 	"k8s.io/cloud-provider-openstack/pkg/util/mount"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 	"k8s.io/component-base/logs"
 	"k8s.io/klog"
 )
 
 var (
-	endpoint    string
-	nodeID      string
-	cloudconfig string
-	cluster     string
+	cloudconfig          string
+	cluster              string
+	endpoint             string
+	nodeID               string
+	runControllerService bool
+	runNodeService       bool
 )
 
 func init() {
@@ -73,16 +76,20 @@ func main() {
 
 	cmd.Flags().AddGoFlagSet(flag.CommandLine)
 
-	cmd.PersistentFlags().StringVar(&nodeID, "nodeid", "", "node id")
-	cmd.MarkPersistentFlagRequired("nodeid")
-
-	cmd.PersistentFlags().StringVar(&endpoint, "endpoint", "", "CSI endpoint")
-	cmd.MarkPersistentFlagRequired("endpoint")
-
 	cmd.PersistentFlags().StringVar(&cloudconfig, "cloud-config", "", "CSI driver cloud config")
 	cmd.MarkPersistentFlagRequired("cloud-config")
 
 	cmd.PersistentFlags().StringVar(&cluster, "cluster", "", "The identifier of the cluster that the plugin is running in.")
+
+	cmd.PersistentFlags().StringVar(&endpoint, "endpoint", "", "CSI endpoint")
+	cmd.MarkPersistentFlagRequired("endpoint")
+
+	cmd.PersistentFlags().StringVar(&nodeID, "nodeid", "", "node id")
+	cmd.MarkPersistentFlagRequired("nodeid")
+
+	cmd.PersistentFlags().BoolVar(&runControllerService, "run-controller-service", true, "If set to false then the CSI driver does not activate its controller service (default: true)")
+
+	cmd.PersistentFlags().BoolVar(&runNodeService, "run-node-service", true, "If set to false then the CSI driver does not activate its node service (default: true)")
 
 	openstack.AddExtraFlags(pflag.CommandLine)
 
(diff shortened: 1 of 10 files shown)


lingxiankong:
> This enables the use-case of running the CSI controllers (csi-provisioner, csi-attacher, etc., + the driver controller) separately outside of the cluster they are serving, and not necessarily on an OpenStack nova VM.

cinder-csi-plugin already supports this without this PR, it's running a gPRC server with all the interfaces implemented, we don't need to add two extra parameters explicitly.

/hold

rfranzke:
> cinder-csi-plugin already supports this without this PR, it's running a gPRC server with all the interfaces implemented

@jianglingxia Can you explain more? As of today, when starting the cinder CSI driver it unconditionally sets up both the controller and the node server, see https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/driver.go#L138-L144. I don't see how is it supported already today?

lingxiankong:
> Can you explain more? As of today, when starting the cinder CSI driver it unconditionally sets up both the controller and the node server, see https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/driver.go#L138-L144. I don't see how is it supported already today?

The code you mentioned is just registering services and its implementation to the gRPC server, they are not the 'real' server running behind, after registering, you still have only one server running, but can serve different types of gRPC calling.

What problem do you want to solve? What problems are you facing currently? If you do see an issue caused by the current implementation, it would be great if you could submit an issue before actually submitting a PR.

rfranzke:
> The code you mentioned is just registering services and its implementation to the gRPC server, they are not the 'real' server running behind, after registering, you still have only one server running, but can serve different types of gRPC calling.

This is understood. However, I want to optionally enable that either the controller or the node service can be disabled from being registered to the gRPC server.

> What problem do you want to solve?

I added docs and a description for the problem in the PR. Concretely, I want to enable the use-case where the CSI controllers are NOT running inside the cluster they are serving but outside in a different cluster. Imagine cluster `A` that runs the CSI provisioner+attacher + Cinder CSI driver (with only controller service enabled) **for** cluster `B`. In cluster `B` the CSI driver (with only node service enabled) is ran (+driver-registrar and liveness probe).

Also, it should be made possible that cluster `A` is not necessarily backed by OpenStack VMs. The today's implementation assumes that the controller service can talk to the OpenStack metadata service which is not possible in this case.

>  What problems are you facing currently? 

The above mentioned scenarios are not possible to run today because it's not possible to cleanly separate the driver services.

> If you do see an issue caused by the current implementation, it would be great if you could submit an issue before actually submitting a PR.

Well, if you prefer I can open an issue now and describe the above problems + reference this PR as solution for it. Shall I really do it?

FYI: We have done similar enhancements to the AWS EBS CSI driver (https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/438) and to the GCE PD CSI driver (https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/449) already.

lingxiankong:
Hi @rfranzke I'm still confused, I totally understand want you want do do with this PR, but I don't understand why you can't run controller services and node services separately? If you look at https://github.com/kubernetes/cloud-provider-openstack/tree/master/manifests/cinder-csi-plugin, as example, we are actually running controller and node plugin at different places using different yaml files.

Maybe I'm wrong or I didn't fully understand you use case, but can you paste some error logs with your preferable deployment methods?

If it's possible, please create an issue with the steps and error log, we could move discussion there. Thanks for your patience.

@ramineni @jichenjc @chrigl Please also join us for this discussion if you could.

ramineni:
> Also, it should be made possible that cluster `A` is not necessarily backed by OpenStack VMs. The today's implementation assumes that the controller service can talk to the OpenStack metadata service which is not possible in this case.

Controller service doesnt depend on Openstack metatdata service, it never queries it as per https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/controllerserver.go  
https://github.com/kubernetes/cloud-provider-openstack/blob/master/cmd/cinder-csi-plugin/main.go#L111 , this only creates an object , doesnt actually query, so if only controller plugin manifest is deployed on node which is not backed by Openstack VMs, this should ideally still work? Are you facing issue with that?

> Imagine cluster `A` that runs the CSI provisioner+attacher + Cinder CSI driver (with only controller service enabled) **for** cluster `B`. In cluster `B` the CSI driver (with only node service enabled) is ran (+driver-registrar and liveness probe).
> 

As per current implementation, even if we have both node service and controller service enabled in plugin, unless we run node-driver-registrar sidecar, NodeGetInfo, NodeStageVolume, NodePublishVolume are not possible , so no effect of node service if only controller plugin manifest is deployed on the node ?

If the above doesnt solve your usecase, could you elaborate on issues you are facing with both enabled and having different manifests doesnt solve your usecase

rfranzke:
@lingxiankong @ramineni Thanks for your views! I was trying it out again and thoroughly followed the example manifests, and indeed, it works as you expected. Probably I was working with outdated manifests earlier or did something else wrong. Anyways, I guess the PR can be closed then as the scenario is already supported. Thanks again for your help and sorry for taking your time with this.

/close

lingxiankong:
@rfranzke thanks for the followup

PR #1023, opened 2020-04-10
[occm] Support multiple network names

zuzzas:

In some intricate network designs, it is possible to have multiple networks on different hosts. This PR allows the user to specify multiple network names that may be considered public or private on cluster's nodes.

I've tried to maintain complete backward compatibility. Due to the way gcfg works, it is possible to specify arrays via repeated use of the same key.

**Release note**:

```release-note
[openstack-cloud-controller-manager] Support multiple network names for config options `internal-network-name` and `public-network-name`.
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack.go
+++ b/pkg/cloudprovider/providers/openstack/openstack.go
@@ -34,6 +34,8 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/cloud-provider-openstack/pkg/util"
+
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/attachinterfaces"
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -150,9 +150,11 @@ type BlockStorageOpts struct {
 
 // NetworkingOpts is used for networking settings
 type NetworkingOpts struct {
-	IPv6SupportDisabled bool   `gcfg:"ipv6-support-disabled"`
-	PublicNetworkName   string `gcfg:"public-network-name"`
-	InternalNetworkName string `gcfg:"internal-network-name"`
+	IPv6SupportDisabled  bool     `gcfg:"ipv6-support-disabled"`
+	PublicNetworkName    []string `gcfg:"public-network-name"`
+	InternalNetworkName  []string `gcfg:"internal-network-name"`
+	publicNetworkNames   map[string]struct{}
```
lingxiankong, line 156: I'd suggest to use https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/util/util.go#L30 rather than adding two extra fields here.

zuzzas: You are absolutely right. Such a minor performance improvement does not warrant such an ugly solution.

Fixed!

zuzzas:
@lingxiankong, is that ^ something I can fix?

lingxiankong:
The pull-cloud-provider-openstack-check job issue should be fixed, if it fails again, please rebase the code.

PR #1025, opened 2020-04-12
[manila-csi-plugin] report fully qualified version

gman0:

Currently, manila-csi-plugin is reporting only the driver's semantic version. This PR makes it so that the cloud-provider-openstack's version is part of the reported version too. This will help identify the exact version of the driver e.g. when troubleshooting a problem.

The format used for the fully qualified version is `{driver version}@{CPO version}`, where "CPO version" is passed in via linker flags in Makefile as a git hash:

https://github.com/kubernetes/cloud-provider-openstack/blob/b2335ad887a06e28acdf56e2d0933289bcb0ddfe/Makefile#L34-L35

It's advertised in:
* logs during initialization
* identity service's `GetPluginInfoResponse`

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/manila/driver.go
+++ b/pkg/csi/manila/driver.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/cloud-provider-openstack/pkg/csi/manila/csiclient"
 	"k8s.io/cloud-provider-openstack/pkg/csi/manila/manilaclient"
 	"k8s.io/cloud-provider-openstack/pkg/csi/manila/options"
+	"k8s.io/cloud-provider-openstack/pkg/version"
 	"k8s.io/klog"
 )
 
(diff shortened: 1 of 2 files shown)


gman0:
@ramineni by the way, I've noticed cinder-csi-plugin is advertising only semantic version of the driver too. Would advertising CPO version there make sense?

ramineni:
> @ramineni by the way, I've noticed cinder-csi-plugin is advertising only semantic version of the driver too. Would advertising CPO version there make sense?

@gman0 IIUC, here CPO version is mostly occm version right, means if cpo version is 1.18 , doesnt mean that cinder-csi-plugin deployed is of 1.18 , it depends on the what version they are using in manifest, so both are independent?

PR #1027, opened 2020-04-14
[cinder-csi-plugin] Add version into cinder CSI driver info

jichenjc:
this is inspired by https://github.com/kubernetes/cloud-provider-openstack/pull/1025
Cinder CSI better to contain those info for further usage

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/driver.go
+++ b/pkg/csi/cinder/driver.go
@@ -24,6 +24,7 @@ import (
 	"google.golang.org/grpc/status"
 	"k8s.io/cloud-provider-openstack/pkg/csi/cinder/openstack"
 	"k8s.io/cloud-provider-openstack/pkg/util/mount"
+	"k8s.io/cloud-provider-openstack/pkg/version"
 	"k8s.io/klog"
 )
 
@@ -33,13 +34,15 @@ const (
 )
 
 var (
-	version = "1.2.0"
+	specVersion = "1.2.0"
+	// we used to use spec version as driver version, now separate them
+	Version = "1.2.0"
 )
 
 type CinderDriver struct {
 	name        string
 	nodeID      string
-	version     string
+	fqVersion   string //Fully qualified version in format {Version}@{CPO version}
 	endpoint    string
 	cloudconfig string
 	cluster     string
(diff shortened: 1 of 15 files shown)


ramineni:
@jichenjc please check my question here , https://github.com/kubernetes/cloud-provider-openstack/pull/1025#issuecomment-612837586

ramineni:
> @jichenjc please check my question here , [#1025 (comment)](https://github.com/kubernetes/cloud-provider-openstack/pull/1025#issuecomment-612837586)

/hold

gman0:
@jichenjc: @ramineni and i had a discussion about how to advertise versions better. Since this repository is moving at a different speed than Cinder/Manila CSI drivers, having a separate driver version accompanied by `@{CPO version}` seems like the way to go.

Having the CSI spec version as part of the driver's version doesn't convey much information about the actual state of the driver - so the spec version could be logged, but maybe shouldn't be part of the driver version.

Another benefit of having a separate driver version is that we give guarantees to the users about backwards compatibility between minor versions, and possible breaking changes when bumping major versions.

---

So we're thinking the version should be in the format of `{driver version}@{CPO version}`, and the CSI spec version may be logged.

jichenjc:
we need CPO version, cinder CSI driver version and CSI Spec version

@ramineni any comments above from @gman0 ? the only question I have is Cinder CSI
seems doesn't have a version defined, I can certainly define one such as 1.0.0 but that will be weird
because we don't have version before.. any comments?

ramineni:
@jichenjc we have maintained the driver version and spec version on par till now. This is the driver version , https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/driver.go#L36 . But driver version could be updated every release (depending on changes) whereas spec version need not be , so they can be different in future. You could define one more variable for  spec version and keep the driver version as it is.

jichenjc:
@ramineni understood, I thought the 1.2.0 is for CSI Spec.. will do the update~

jichenjc:
tested locally

```
# kubectl logs csi-cinder-controllerplugin-0 -n kube-system -c cinder-csi-plugin
I0416 02:33:05.938343       1 driver.go:69] Driver: cinder.csi.openstack.org
I0416 02:33:05.938409       1 driver.go:70] Driver version: 1.2.0@4ddd97cd
I0416 02:33:05.938415       1 driver.go:71] CSI Spec version: 1.2.0


```

pkg/csi/cinder/driver.go
```diff
@@ -33,13 +34,13 @@ const (
 )
 
 var (
-	version = "1.2.0"
+	specVersion = "1.2.0"
```
ramineni, line 37: @jichenjc as mentioned before , this should be driver version . you can add one more variable for spec version

pkg/csi/cinder/driver.go
```diff
@@ -54,15 +55,18 @@ type CinderDriver struct {
 }
 
 func NewDriver(nodeID, endpoint, cluster string) *CinderDriver {
-	klog.Infof("Driver: %v version: %v", driverName, version)
 
 	d := &CinderDriver{}
 	d.name = driverName
 	d.nodeID = nodeID
-	d.version = version
+	// cinder CSI don't have a driver version now, so use spec CSI version instead
```
ramineni, line 62: @jichenjc the version field specifies the driver version, not the spec version. we just maintain it on par doesnt necessarily mean its spec versio

jichenjc:
weird, I think I added the code.. why not pushed upstream
will check again, thanks

jichenjc:
due to my git problem, the files are corrupted ,so I have to resubmit here again
https://github.com/kubernetes/cloud-provider-openstack/pull/1035

PR #1026, opened 2020-04-14
Fix the release note guide

lingxiankong:
**What this PR does / why we need it**:
To keep consistent with the PR title format to avoid confusion.

**Which issue this PR fixes(if applicable)**:
fixes #

**Special notes for reviewers**:

**Release note**:

```release-note
NONE
```

PR #1028, opened 2020-04-14
[All] upgrade klog to klog/v2 

jichenjc:

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,10 +33,11 @@ require (
 	github.com/spf13/viper v1.4.0
 	github.com/stretchr/testify v1.5.1
 	golang.org/x/crypto v0.0.0-20200220183623-bac4c82f6975
-	golang.org/x/net v0.0.0-20191126235420-ef20fe5d7933
+	golang.org/x/net v0.0.0-20200226121028-0de0cce0169b
 	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d // indirect
 	golang.org/x/sys v0.0.0-20200122134326-e047566fdf82
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0 // indirect
+	golang.org/x/tools v0.0.0-20200414032229-332987a829c3 // indirect
 	google.golang.org/appengine v1.6.2 // indirect
 	google.golang.org/grpc v1.26.0
 	gopkg.in/gcfg.v1 v1.2.3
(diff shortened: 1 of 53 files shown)


ramineni:
@jichenjc hope you are working on this , still all files needs to be updated to use klog v2 . We have a broken CI now, appreciate if you can complete this ASAP

jichenjc:
@ramineni not sure I am using right approach
I thought klog v2 is missing so we need take them in go.mod? 
or you are saying we have to switch to klog v2 for *all* the files?

ramineni:
@jichenjc It is better to switch to klogv2 in our repo too, no point in maintaining/adding 2 versions?

jichenjc:
you are right, let me check how to proceed as soon as I can

jichenjc:
@ramineni not sure the import-boss change is the right approach... need further check as well
any comments please advice, thanks

ramineni:
@jichenjc this seems to fix the issue, https://github.com/kubernetes/cloud-provider-openstack/pull/1026

jichenjc:
ok, thanks for the info 
anyway, I am trying to upgrade to klog v2

lingxiankong:
Hi @jichenjc , I think we don't need this as the previous issue was fixed in https://github.com/kubernetes/cloud-provider-openstack/pull/1026

jichenjc:
@lingxiankong do you think we can switch to klog v2? just like 
https://github.com/kubernetes/gengo/commit/565683fea29eeb99987bd45fd485f148eda97471 did?

lingxiankong:
> @lingxiankong do you think we can switch to klog v2? just like
> [kubernetes/gengo@565683f](https://github.com/kubernetes/gengo/commit/565683fea29eeb99987bd45fd485f148eda97471) did?

What's the benefit can we get from klog v2?

jichenjc:
as more projects move to v2, we may need doing so to avoid klog/klogv2 both?
and seems it's provide more reasonable return functions (see V(2).Enabled() change in the code)

lingxiankong:
> as more projects move to v2, we may need doing so to avoid klog/klogv2 both?
> and seems it's provide more reasonable return functions (see V(2).Enabled() change in the code)

Awesome, thanks, please go ahead

go.mod
```diff
@@ -48,8 +49,11 @@ require (
 	k8s.io/apiserver v0.18.0
 	k8s.io/client-go v0.18.0
 	k8s.io/cloud-provider v0.18.0
+	k8s.io/code-generator v0.18.1 // indirect
 	k8s.io/component-base v0.18.0
+	k8s.io/gengo v0.0.0-20200413195148-3a45101e95ac // indirect
 	k8s.io/klog v1.0.0
```
ramineni, line 55: no need of klog anymore?

jichenjc: others might still need it?

ramineni: not required I suppose , if you remove it and run go mod tidy , it would be added automatically as indirect dependency if required.

ramineni:
@jichenjc This dependency shouldnt be required anymore, 	k8s.io/klog v1.0.0 , please remove the same.

jichenjc:
@ramineni can you help to take a look?

PR #1034, opened 2020-04-17
[cinder-csi-plugin] Allow force create snapshot

### LINKED ISSUE #832: Improve CSI Cinder Snapshot: add --force flag to create snapshots of in-use volumes

jichenjc:
**The binaries affected**:

- [ ] All
- [x] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] manila-csi-plugin
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

fixes #832 

**Release note**:

```release-note
[cinder-csi-plugin] Added `force-create` parameter in VolumeSnapshotClass to support creating snapshot for a volume in in-use status.
```

### DIFF
--- a/docs/using-cinder-csi-plugin.md
+++ b/docs/using-cinder-csi-plugin.md
@@ -32,7 +32,7 @@ Check [kubernetes CSI Docs](https://kubernetes-csi.github.io/docs/) for flag det
 
 ### Deploy
 
-If you already created the `cloud-config` secret used by the [cloud-controller-manager](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/using-controller-manager-with-kubeadm.md#steps), remove the file ```manifests/cinder-csi-plugin/csi-secret-cinderplugin.yaml``` and then jump directly to the `kubectl apply ...` command.
+If you already created the `cloud-config` secret used by the [cloud-controller-manager](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/using-openstack-cloud-controller-manager.md#steps), remove the file ```manifests/cinder-csi-plugin/csi-secret-cinderplugin.yaml``` from [manifests](https://github.com/kubernetes/cloud-provider-openstack/tree/master/manifests/cinder-csi-plugin) and then jump directly to the `kubectl apply ...` command.
 
 Encode your ```$CLOUD_CONFIG``` file content using base64.
 
@@ -58,7 +58,7 @@ volumes:
     type: Directory
 ```
 
-Then call following command by using existing manifests:
+Then call following command by using existing [manifests](https://github.com/kubernetes/cloud-provider-openstack/tree/master/manifests/cinder-csi-plugin):
 
 ```kubectl -f manifests/cinder-csi-plugin apply```
 
@@ -100,7 +100,7 @@ Events:               <none>
 ### Example Nginx application usage
 
 After performing above steps, you can try to create StorageClass, PersistentVolumeClaim and pod to consume it.
-Try following command:
+Try following command by using [examples](https://github.com/kubernetes/cloud-provider-openstack/blob/master/examples/cinder-csi-plugin/nginx.yaml):
 
 ```kubectl -f examples/cinder-csi-plugin/nginx.yaml create```
 
--- a/docs/using-cinder-csi-plugin.md
+++ b/docs/using-cinder-csi-plugin.md
@@ -197,8 +197,10 @@ Following prerequisite needed for volume snapshot feature to work.
 1. Enable `--feature-gates=VolumeSnapshotDataSource=true` in kube-apiserver
 2. Make sure, your csi deployment contains external-snapshotter sidecar container, external-snapshotter sidecar container will create three crd's for snapshot management VolumeSnapshot,VolumeSnapshotContent, and VolumeSnapshotClass. external-snapshotter is a part of `csi-cinder-controllerplugin`
 
-For Snapshot Creation and Volume Restore, please follow below steps:
+Supported parameters for VolumeSnapshotClass:
+* `force-create`: Support creating snapshot for a volume in in-use status.
 
+For Snapshot Creation and Volume Restore, please follow below steps:
 * Create Storage Class, Snapshot Class and PVC    
 ```
 $ kubectl -f examples/cinder-csi-plugin/snapshot/example.yaml create
@@ -206,6 +208,7 @@ storageclass.storage.k8s.io/csi-sc-cinderplugin created
 volumesnapshotclass.snapshot.storage.k8s.io/csi-cinder-snapclass created
 persistentvolumeclaim/pvc-snapshot-demo created
 ```
+
 * Verify that pvc is bounded
 ``` 
 $ kubectl get pvc --all-namespaces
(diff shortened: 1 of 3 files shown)


jichenjc:
I tested locally, by setting the value to false and true, got following and checked in GUI side 
if force is false (default) then create snapshot will fail while force is true is able to create snapshot

So I think it's ready for further review and merge

```
 Event(v1.ObjectReference{Kind:"VolumeSnapshot", Namespace:"default", Name:"new-snapshot-demo", UID:"e79c1732-ae03-4411-8ddf-cba7f264bf02", APIVersion:"snapshot.storage.k8s.io/v1alpha1", ResourceVersion:"838", FieldPath:""}): type: 'Warning' reason: 'SnapshotCreationFailed' Failed to create snapshot: failed to take snapshot of the volume, pvc-d4bfc1bd-5b96-40bf-99cb-67fa8bfc648e: "rpc error: code = Internal desc = CreateSnapshot failed with error Bad request with: [POST https://hdc-prod-cloud.hursley.ibm.com:8776/v3/07962130d7044e3c84e1825859d5bef9/snapshots], error message: {\"badRequest\": {\"message\": \"Invalid volume: Volume 18056a4d-9115-4415-9214-29d2149ac193 status must be available, but current status is: in-use.\", \"code\": 400}


I0417 04:01:35.494658       1 snapshot_controller.go:615] createSnapshot: Creating snapshot default/new-snapshot-demo through the plugin ...
```

examples/cinder-csi-plugin/snapshot/example.yaml
```diff
@@ -11,6 +11,8 @@ kind: VolumeSnapshotClass
 metadata:
   name: csi-cinder-snapclass
 snapshotter: cinder.csi.openstack.org
+parameters:
+  snapshot-inuse-volumes: "false"
```
lingxiankong, line 15: As suggested in my previous review, I suggest we use `force-create: "false"` for brevity.

pkg/csi/cinder/openstack/openstack_snapshots.go
```diff
@@ -34,15 +35,26 @@ const (
 	snapReadySteps      = 10
 
 	snapshotDescription = "Created by OpenStack Cinder CSI driver"
+	snapshotForceCreate = "snapshot-inuse-volumes"
```
lingxiankong, line 38: I prefer we use `force-create`

pkg/csi/cinder/openstack/openstack_snapshots.go
```diff
@@ -34,15 +35,26 @@ const (
...
 
 // CreateSnapshot issues a request to take a Snapshot of the specified Volume with the corresponding ID and
 // returns the resultant gophercloud Snapshot Item upon success
 func (os *OpenStack) CreateSnapshot(name, volID string, tags *map[string]string) (*snapshots.Snapshot, error) {
+
+	force := false
+	// if no flag given, then force will be false by default
+	// if flag it given , check it
+	if item, ok := (*tags)[snapshotForceCreate]; ok {
+		force, _ = strconv.ParseBool(item)
```
lingxiankong, line 49: Please catch the error and use False if any error occurs.

lingxiankong:
@jichenjc @ramineni How about we use `force-create` which is more simple and readable?

jichenjc:
thanks for the reminder, will update soon

examples/cinder-csi-plugin/snapshot/example.yaml
```diff
@@ -11,6 +11,8 @@ kind: VolumeSnapshotClass
 metadata:
   name: csi-cinder-snapclass
 snapshotter: cinder.csi.openstack.org
+parameters:
+  force-create: "false"
```
adisky, line 15: going the k8s way, `forceCreate` would be better.

jichenjc: this has been discussed above...

jichenjc:
@lingxiankong can you help to check the latest changes?thanks

examples/cinder-csi-plugin/snapshot/example.yaml
adisky: I mean prefer camel case in yaml, rather than kebab case.

ramineni: @jichenjc discussion is lost because it is seperate PR now . @adisky pls check here , https://github.com/kubernetes/cloud-provider-openstack/pull/1002 , camel case is for GO lang not for yamls

jichenjc: my bad... I forgot this totally, @ramineni thanks for the reminder

adisky: got it, thanks.

pkg/csi/cinder/openstack/openstack_snapshots.go
```diff
@@ -34,15 +35,31 @@ const (
...
 func (os *OpenStack) CreateSnapshot(name, volID string, tags *map[string]string) (*snapshots.Snapshot, error) {
+
+	force := false
+	// if no flag given, then force will be false by default
+	// if flag it given , check it
+	if item, ok := (*tags)[snapshotForceCreate]; ok {
+		var err error
+		force, err = strconv.ParseBool(item)
+		if err != nil {
+			force = false
```
ramineni, line 52: :nit: This if condition is not required I suppose. It would be false anyway .

jichenjc:
@ramineni can you help to check again? Thanks a lot

lingxiankong:
please update release note to reflect the latest change.

jichenjc:
@lingxiankong thanks for the reminder, update~

ramineni:
@jichenjc could you do quick update in docs as well on supported flag here , https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/using-cinder-csi-plugin.md#example-snapshot-create-and-restore

ramineni:
/lgtm cancel
just we dont miss the doc change on supported flags .

jichenjc:
@ramineni please let me know your thoughts on the doc update, thanks

docs/using-cinder-csi-plugin.md
```diff
@@ -206,6 +206,8 @@ storageclass.storage.k8s.io/csi-sc-cinderplugin created
 volumesnapshotclass.snapshot.storage.k8s.io/csi-cinder-snapclass created
 persistentvolumeclaim/pvc-snapshot-demo created
 ```
+Note: `force-create` can be used to force create the snapshot when the volume is `in-use` state.
```
lingxiankong, line 209: Please use a plain paragraph rather than a note:

```
Supported parameters for VolumeSnapshotClass:

* `force-create`: Support creating snapshot for a volume in in-use status.
```

lingxiankong:
Release note updated.

docs/using-cinder-csi-plugin.md
jichenjc: updated , tried several times and seems current position is best to add those info, thanks

lingxiankong:
/lgtm

thanks!

PR #1036, opened 2020-04-18
[occm] Remove volume related code from occm

lingxiankong:
Remove volume related code from openstack-cloud-controller-manager. The code was initially moved from k8s in-tree code, but the volume related functions were never provided by occm, especially the cinder-csi-plugin is the way to go.

What's more, removing the code also deletes the volume related unit test from cloud-provider-openstack-unittest job, which always causes CI issues.

fixes #

**Release note**:

```release-note
NONE
```


lingxiankong:
@ramineni ready for review

ramineni:
Thanks.
/lgtm

ramineni:
To fix quota issues , merging this
/approve

PR #1037, opened 2020-04-20
[cinder-csi-plugin] Add instance ID validation

takmatsu:
* cinder-csi-plugin

`/var/lib/cloud/data/instance-id` is not always reliable. 
we need to check its format because sometimes it may be EC2 style (i-******) one when OpenStack DataSource cloud-init is failed.
and this check also make possible to fallback to metadata provider if the id is invalid.

fixes https://github.com/kubernetes/cloud-provider-openstack/issues/859

**Release note**:

```release-note
NONE
```


pkg/util/mount/mount.go
```diff
@@ -287,9 +288,10 @@ func (m *Mount) GetInstanceID() (string, error) {
 		instanceID := string(idBytes)
 		instanceID = strings.TrimSpace(instanceID)
 		klog.V(3).Infof("Got instance id from %s: %s", instanceIDFile, instanceID)
-		if instanceID != "" {
+		if instanceID != "" && util.IsValidUUID(instanceID) {
```
jichenjc, line 291: why this happens... it's a cloud-init used file so why uuid not valid..

takmatsu: In our investigation:
* actually OpenStack provides 2 types of metadata
  * AWS EC2: http://169.254.169.254/2009-04-04/meta-data/
     * it includes i-**** instance id
  * OpenStack: http://169.254.169.254/openstack
     * it includes UUID
* cloud-init try to execute all type of supported data sources by default
   ```
   2020-04-17 08:04:32,983 - __init__.py[DEBUG]: Looking for data source in: ['NoCloud', 'ConfigDrive', 'OpenNebula', 'DigitalOcean', 'Azure', 'AltCloud', 'OVF', 'MAAS', 'GCE', 'OpenStack', 'AliYun', 'Ec2', 'CloudSigma', 'CloudStack', 'SmartOS', 'Bigstep', 'Scaleway', 'Hetzner', 'IBMCloud', 'Oracle', 'None']
   ```
If OpenStack DataSource is (partially) failed, Ec2 one will be applied.
that's why sometimes we're facing the issues like #859 , and then we have EC2 style instance-id.

honestly, this is just from OpenStack metadata API stability.
but it is difficult to detect on OpenStack API side because creation of instance itself is succeed.

jichenjc: ok, understand better ... if it's just for stable of metadata, then looks fine to me 
as this is anyway an improvement

lingxiankong:
> Hi folks! I don't work on this project anymore. Could you please remove me from the reviewers list? Thanks :)

Hi Flavio, please feel free to submit a PR to remove yourself from OWNERS file.

pkg/util/mount/mount.go
ramineni: @takmatsu IIUC the issue, it could also be solved by changing the search-order to query metadata first for retrieving instance-id , https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/openstack/openstack.go#L154 ?

takmatsu: @ramineni yes, but I think this function is expected as returning correct instance-id. and actually we want to use ConfigDrive as a first because it just reading from local filesystem.

lingxiankong: @takmatsu 

> If OpenStack DataSource is (partially) failed

May I know the reason why it's failed?

I would prefer we refactor the function by using `https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/util/metadata/metadata.go#L225`

takmatsu: > May I know the reason why it's failed?

this is just from metadata-api(or proxy) stability.

> I would prefer we refactor the function by using https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/util/metadata/metadata.go#L225

yes, I also think we don't have to maintain 2 functions for same objective.
I fixed my patch

celiawa:
In one of our user case, metadata service is disabled in the lab and /var/lib/cloud/data/instance-id is not correct with style (i-******) . We had workaround to get the instance-id from "kubectl get node <node> -o yaml | grep systemUUID".

takmatsu:
> We had workaround to get the instance-id from "kubectl get node -o yaml | grep systemUUID"

maybe we can use it as a new metadata provider

ramineni:
@takmatsu In PR https://github.com/kubernetes/cloud-provider-openstack/pull/1096, instead of retrieving instance-id from cloud-init , changes have been to made to get directly get from configdrive or metadata service which would guarantee to retrieve only openstack id , this should fix the issue of invalid instance id.

PR #977, opened 2020-03-18
[OCCM] Set default value for some options

lingxiankong:
**The binaries affected**:

- [x] openstack-cloud-controller-manager
- [ ] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] octavia-ingress-controller
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

**What this PR does / why we need it**:
Support Octavia by default as load balancer implementation in openstack-cloud-controller-manager.

**Which issue this PR fixes**:
fixes #

**Special notes for reviewers**:
Document has already been changed in PR #962

**Release note**:
```release-note
openstack-cloud-controller-manager: Support OpenStack Octavia by default as load balancer implementation
```

PR #1038, opened 2020-04-20
[OCCM] Print error messages, when OpenStack client returns error

kayrus:

Now OCCM prints error message, when a required endpoint cannot be found.

Due to #977 OCCM stopped to work silently, there were no errors showed in logs. It took a while to debug the root cause.

**Release note**:

```release-note
None
```

/cc @rfranzke
/assign @lingxiankong

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack.go
+++ b/pkg/cloudprovider/providers/openstack/openstack.go
@@ -865,16 +865,19 @@ func (os *OpenStack) LoadBalancer() (cloudprovider.LoadBalancer, bool) {
 
 	network, err := os.NewNetworkV2()
 	if err != nil {
+		klog.Errorf("Failed to create an OpenStack Network client: %v", err)
 		return nil, false
 	}
 
 	compute, err := os.NewComputeV2()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


ramineni:
@kayrus release note is not required for all PRs , required only for features and imp bugs . For this PR , release note can be NONE

kayrus:
@lingxiankong tests pass

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -865,16 +865,19 @@ func (os *OpenStack) LoadBalancer() (cloudprovider.LoadBalancer, bool) {
 
 	network, err := os.NewNetworkV2()
 	if err != nil {
+		klog.Errorf("Failed to create an OpenStack Network client: %s", err)
```
lingxiankong, line 868: The convention of including an error text msg is to use `%v`

kayrus: fixed

lingxiankong:
release note was deleted.
/lgtm
/approve

PR #1040, opened 2020-04-20
[occm] make cascade deletion configurable

hamzazafar:
Make cascade deletion of load balancers configurable

If octavia is used, occm tries to perform cascade deletion of load balancers. Some Octavia provider drivers might not support cascade deletion, therefore this PR makes cascade deletion option configurable. By default, cascade deletion is enabled.

Cascade deletion can be enabled/disabled by setting `cascade-delete` option in cloud-config file.

**Release note**:

```release-note
[openstack-cloud-controller-manager] Make cascade deletion configurable when Octavia is enabled.
```

### DIFF
--- a/docs/using-openstack-cloud-controller-manager.md
+++ b/docs/using-openstack-cloud-controller-manager.md
@@ -173,7 +173,8 @@ Although the openstack-cloud-controller-manager was initially implemented with N
   Deprecated.
 * `internal-lb`
   Determines whether or not to create an internal load balancer (no floating IP) by default. Default: false.
-
+* `cascade-delete`
+  Determines whether or not to perform cascade deletion of load balancers. Default: true.
 ### Metadata
 
 * `search-order`
--- a/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
@@ -1857,7 +1857,7 @@ func (lbaas *LbaasV2) EnsureLoadBalancerDeleted(ctx context.Context, clusterName
 	}
 
 	// delete the loadbalancer and all its sub-resources.
-	if lbaas.opts.UseOctavia {
+	if lbaas.opts.UseOctavia && lbaas.opts.CascadeDelete {
 		deleteOpts := loadbalancers.DeleteOpts{Cascade: true}
 		if err := loadbalancers.Delete(lbaas.lb, loadbalancer.ID, deleteOpts).ExtractErr(); err != nil {
 			return fmt.Errorf("failed to delete loadbalancer %s: %v", loadbalancer.ID, err)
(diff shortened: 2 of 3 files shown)


lingxiankong:
> Cascade deletion can be enabled/disabled by setting `cascade-delete` option in cloud-config file or by using annotation `loadbalancer.openstack.org/cascade-delete`

Can we only use config option? I don't think the service user should be able to config that.

hamzazafar:
> Can we only use config option? I don't think the service user should be able to config that.

I have dropped the annotation.

pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
```diff
@@ -1857,7 +1857,7 @@ func (lbaas *LbaasV2) EnsureLoadBalancerDeleted(ctx context.Context, clusterName
 	}
 
 	// delete the loadbalancer and all its sub-resources.
-	if lbaas.opts.UseOctavia {
+	if lbaas.opts.UseOctavia && lbaas.opts.CascadeDelete {
```
ramineni, line 1860: @hamzazafar @lingxiankong  one question here, IIUC if cascadeDelete is false , dont we suppose to proceed with this operation with deleteOpts := loadbalancers.DeleteOpts{Cascade: false} ?
something like this instead
`if lbaas.opts.UseOctavia`
` deleteOpts := loadbalancers.DeleteOpts{Cascade: lbaas.opts.CascadeDelete} `

docs/using-openstack-cloud-controller-manager.md
```diff
@@ -173,7 +173,8 @@ Although the openstack-cloud-controller-manager was initially implemented with N
   Deprecated.
 * `internal-lb`
   Determines whether or not to create an internal load balancer (no floating IP) by default. Default: false.
-
+* `cascade-delete`
```
adisky, line 176: shouldn't it be `cascade-delete-lb`, as it is specific to load balancers.??

pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
hamzazafar: @ramineni if we set `Cascade: false` then only the load balancer resource will get deleted, listeners, health-monitors, pools and members will be left behind. 

In current implementation, if cascade delete is false, then load balancer and related resources are deleted one by one [ref](https://github.com/kubernetes/cloud-provider-openstack/blob/3c1dcc8b7083e0d6d0749cb003cff3df7a468fad/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go#L1865).

docs/using-openstack-cloud-controller-manager.md
hamzazafar: I wasn't aware of this naming convention, most of the other options related to load balancer (create-monitor, monitor-delay , monitor-max-retries etc) don't have `-lb` appended to them.

I am happy to change `cascade-delete` -> `cascade-delete-lb`. Please let me know @adisky @lingxiankong

pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
ramineni: ok. Got it

docs/using-openstack-cloud-controller-manager.md
lingxiankong: @adisky I don't think `-lb` is needed, all the options here are in the `LoadBalancer` config section, adding `-lb` seems redundant.

ramineni:
@hamzazafar Thanks.
/lgtm

PR #1043, opened 2020-04-23
Bump gophercloud dependencies

kayrus:

Bump gophercloud dependencies with a better connection reuse logic. All components will now reuse an existing HTTP connection instead of establishing a new connection per each HTTP request.

**Release note**:

```release-note
Bump gophercloud to v0.10.1 version
```

/assign @lingxiankong @ramineni @jichenjc


ramineni:
@kayrus release note mentioned is of gophercloud enhancement not our repo right . So here it would be reasonable to add just `Bump gophercloud to xxx version`

PR #1045, opened 2020-04-26
[cinder-csi-plugin] Support image-csi-plugin target in Makefile

lingxiankong:
Add target `image-csi-plugin` in Makefile which is used by CI jobs against previous branches (e.g. release-1.17).

fixes #

As a follow-up, I will change the openlab CI job to use `image-csi-plugin` so we don't break CI of previous branches.

**Release note**:

```release-note
NONE
```


ramineni:
/approve
Thanks.

PR #1063, opened 2020-05-11
[manila-csi-plugin] Make `availability` optional

### LINKED ISSUE #1062: [manila-csi-plugin] Manila share provisioning fails if `availability` is not set in the StorageClass
/kind bug

**What happened**:
If I do not specify `availability` in the StorageClass description, provisioning fails with next error:

> I0508 01:31:25.628972       1 volumesource.go:38] creating a new share (pvc-dbff138f-5e4e-4fb8-8ad3-8eb43e630c38) in AZ <none>
> E0508 01:31:26.862467       1 driver.go:313] [ID:149] GRPC error: rpc error: code = ResourceExhausted desc = failed to create a share (pvc-dbff138f-5e4e-4fb8-8ad3-8eb43e630c38): share da6ab1cd-fed9-4d16-8f15-c5003ba7c552 is in error state: allocate host: No storage could be allocated for this share request, AvailabilityZone filter didn't succeed.

But the documentation says this parameter is optional:
https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/using-manila-csi-plugin.md#controller-service-volume-parameters

Share provisioning succeeds.

**How to reproduce it**:

Try to create a dynamic PV with a StorageClass without `avalability` parameter. For example: https://github.com/kubernetes/cloud-provider-openstack/blob/master/examples/manila-csi-plugin/nfs/dynamic-provisioning/storageclass.yaml

Fedosin:

Now if you do not specify `availability` in the StorageClass, provisioning always fails, despite the fact that the parameter is optional.

This commit makes the plugin to ignore the parameter if its value's not set, as it's said in the documentation.

fixes #1062 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/csi/manila/volumesource.go
+++ b/pkg/csi/manila/volumesource.go
@@ -35,16 +35,23 @@ type volumeCreator interface {
 type blankVolume struct{}
 
 func (blankVolume) create(req *csi.CreateVolumeRequest, shareName string, sizeInGiB int, manilaClient manilaclient.Interface, shareOpts *options.ControllerVolumeContext) (*shares.Share, error) {
-	klog.V(4).Infof("creating a new share (%s) in AZ %s", shareName, coalesceValue(shareOpts.AvailabilityZone))
+	if shareOpts.AvailabilityZone != "" {
+		klog.V(4).Infof("creating a new share (%s) in AZ %s", shareName, coalesceValue(shareOpts.AvailabilityZone))
+	} else {
+		klog.V(4).Infof("creating a new share (%s)", shareName)
+	}
 
 	createOpts := &shares.CreateOpts{
-		AvailabilityZone: shareOpts.AvailabilityZone,
-		ShareProto:       shareOpts.Protocol,
-		ShareType:        shareOpts.Type,
-		ShareNetworkID:   shareOpts.ShareNetworkID,
-		Name:             shareName,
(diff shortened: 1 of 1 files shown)


gman0:
@Fedosin huh, this is really strange. The (possibly empty) string from `shareOpts.AvailabilityZone` is passed here
 https://github.com/gophercloud/gophercloud/blob/822b5cd65fc83eaf2bf5ba859000069e2b838ccb/openstack/sharedfilesystems/v2/shares/requests.go#L48

which is also empty by default.

Could you please check that this error is really caused by passing in an empty string from `shareOpts.AvailabilityZone`, and not something else? It's weird because it shouldn't have any effect, if it's in fact empty.

Fedosin:
@gman0 yeah, it's weird... The full log is here: http://paste.openstack.org/show/793450/
I'm going to put this patch on hold for further investigation.

kayrus:
@Fedosin you can enable JSON request/response logs by adding `--v=6`

kayrus:
From my perspective the patch is useless since `availability_zone` has the `omitempty` tag. I'd take a look on the actual JSON requests.

Fedosin:
Sorry for the buzz folks, that was a problem in our cloud. I close the pr and the issue

gman0:
@Fedosin glad you got it working!

PR #1067, opened 2020-05-14
[cinder-csi-plugin][manila-csi-plugin] Add Github workflow for releasing charts

### LINKED ISSUE #1066: [cinder-csi-plugin] [manila-csi-plugin] Release helm charts
Will the two charts be released in a repository?
I think the best option is github package like ingress-nginx. I am not sure how compatible is with this git repo that has many components.

https://github.com/kubernetes/ingress-nginx/tree/master/charts/ingress-nginx#tldr

/kind feature

brtknr:

This PR proposes using Github action to release Cinder CSI charts.

fixes #1066

For testing, reviewers may use this repo url instead as this PR will need to be merged for this to take effect:

    helm repo add  cpo https://brtknr.github.io/cloud-provider-openstack/

**Release note**:

```release-note
NONE
```


ramineni:
@brtknr never used helm, could you explain what is being checked in here and also ad some description in PR as well

charts/cinder-csi-plugin/templates/controllerplugin-statefulset.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
     {{- include "cinder-csi.controllerplugin.labels" . | nindent 4 }}
 spec:
   replicas: 1
+  serviceName: {{ include "cinder-csi.name" . }}-controllerplugin
```
lingxiankong, line 9: The `imagePullPolicy` below is duplicated, do you mind helping to fix?

lingxiankong:
@brtknr How cinder-csi-1.1.0.tgz is gonna be maintained?

@ramineni please help to check if the version 1.1.0 is correct or not. Thanks

charts/cinder-csi-plugin/templates/controllerplugin-statefulset.yaml
brtknr: Done

brtknr:
> @brtknr How cinder-csi-1.1.0.tgz is gonna be maintained?
>
I am not sure, are you asking for someone to volunteer to become a maintainer?
> 
> @ramineni please help to check if the version 1.1.0 is correct or not. Thanks

lingxiankong:
> I am not sure, are you asking for someone to volunteer to become a maintainer?

No, I mean there should be some mechanism to update that file (e.g. adding a new make target).

strigazi:
FYI
https://github.com/helm/chart-releaser-action
https://github.com/helm/chart-testing-action
https://github.com/helm/charts-repo-actions-demo

lingxiankong:
hmm, still don't think putting tar files in the master branch is a good idea. We have two options:
1. Contribute cinder-csi-plugin helm chart to https://hub.helm.sh/charts
2. Use GitHub Pages as described [here](https://helm.sh/docs/topics/chart_repository/#github-pages-example)

Thoughts?

strigazi:
> hmm, still don't think putting tar files in the master branch is a good idea. We have two options:

1. and 2. are essentially the same option.

> 
> 1. Contribute cinder-csi-plugin helm chart to https://hub.helm.sh/charts

hub.helm.sh is just a catalog. It is not for hosting.

> 2. Use GitHub Pages as described [here](https://helm.sh/docs/topics/chart_repository/#github-pages-example)

>>  FYI
>> https://github.com/helm/chart-releaser-action
>> https://github.com/helm/chart-testing-action
>> https://github.com/helm/charts-repo-actions-demo

I posted the detailed version in my comment above (it is using github pages. chart-releaser-action does the job).  I would say that the docs of helm above are a bit outdated, since the charts-releaser is creating a github release  which can be browesed in the releases page:
https://github.com/helm/chart-releaser-action#example-workflow
```
This uses @helm/chart-releaser-action to turn your GitHub project into a self-hosted Helm chart repo. It does this – during every push to master – by checking each chart in your project, and whenever there's a new chart version, creates a corresponding GitHub release named for the chart version, adds Helm chart artifacts to the release, and creates or updates an index.yaml file with metadata about those releases, which is then hosted on GitHub Pages
```

> 
> Thoughts?

lingxiankong:
> hub.helm.sh is just a catalog. It is not for hosting.

Yes, after double-check, you are right.

> I posted the detailed version in my comment above (it is using github pages. chart-releaser-action does the job). I would say that the docs of helm above are a bit outdated, since the charts-releaser is creating a github release which can be browesed in the releases page:
> https://github.com/helm/chart-releaser-action#example-workflow
> 
> ```
> This uses @helm/chart-releaser-action to turn your GitHub project into a self-hosted
> Helm chart repo. It does this – during every push to master – by checking each chart
> in your project, and whenever there's a new chart version, creates a corresponding
> GitHub release named for the chart version, adds Helm chart artifacts to the release,
> and creates or updates an index.yaml file with metadata about those releases, which
> is then hosted on GitHub Pages
> ```

This approach sounds like a plan.

charts/manila-csi-plugin/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: v1.18.0
+description: Manila CSI Plugin for OpenStack
+name: manila-csi
+version: 1.2.0
```
gman0, line 5: Not sure what version the `version` field refers to. If it's supposed to be the actual version of the plugin, it's actually `0.9.0`:

https://github.com/kubernetes/cloud-provider-openstack/blob/3909137e68519733d9832ef2dabd812550ca6884/pkg/csi/manila/driver.go#L85-L89

charts/manila-csi-plugin/README.md
```diff
@@ -0,0 +1,14 @@
+# Manila CSI provisioner
```
gman0, line 1: Just a nit here: neither Cinder or Manila CSI plugins are just provisioners, they have other competencies too. I realize that Cinder's helm chart readme was used as a template for Manila's, but it's factually not correct. This is not a blocker though, we can cook up better docs for both later.

charts/manila-csi-plugin/Chart.yaml
gman0: Note that the `v1.18.0` specifies the CPO version, and is decoupled from the plugin's version `0.9.0`. The plugin's development is moving at a different velocity than the whole CPO umbrella repo, so that's why it's beneficial to mention both.

charts/manila-csi-plugin/README.md
brtknr: I am going by the docker image tag. I dont see any images tagged 0.9.0 here: https://hub.docker.com/r/k8scloudprovider/manila-csi-plugin/tags

gman0: @brtknr right, but there's no `1.2.0` either ;) That's why I was asking what does `version: 1.2.0` relate to. If that's the chart version, then it seems a bit arbitrary. Maybe the plugin's current version (0.9.0) could pose as a starting point for chart's versioning.

brtknr: Oh chart version 1.0.0 == tag v1.16.0, Oh chart version 1.1.0 == tag v1.17.0 and 1.2.0 == tag v1.18.0.

brtknr: Normally chart version gets bumped up based on reconfiguration of the chart, e.g. new default value include new default appVersion. appVersion corresponds to the app underneath, which i presumed was the manila_csi_tag

gman0: @brtknr Alright, fair enough.

strigazi:
> > hub.helm.sh is just a catalog. It is not for hosting.
> 
> Yes, after double-check, you are right.
> 
> > I posted the detailed version in my comment above (it is using github pages. chart-releaser-action does the job). I would say that the docs of helm above are a bit outdated, since the charts-releaser is creating a github release which can be browesed in the releases page:
> > https://github.com/helm/chart-releaser-action#example-workflow
> > ```
> > This uses @helm/chart-releaser-action to turn your GitHub project into a self-hosted
> > Helm chart repo. It does this – during every push to master – by checking each chart
> > in your project, and whenever there's a new chart version, creates a corresponding
> > GitHub release named for the chart version, adds Helm chart artifacts to the release,
> > and creates or updates an index.yaml file with metadata about those releases, which
> > is then hosted on GitHub Pages
> > ```
> 
> This approach sounds like a plan.

@brtknr status?

brtknr:
> > > hub.helm.sh is just a catalog. It is not for hosting.
> > 
> > 
> > Yes, after double-check, you are right.
> > > I posted the detailed version in my comment above (it is using github pages. chart-releaser-action does the job). I would say that the docs of helm above are a bit outdated, since the charts-releaser is creating a github release which can be browesed in the releases page:
> > > https://github.com/helm/chart-releaser-action#example-workflow
> > > ```
> > > This uses @helm/chart-releaser-action to turn your GitHub project into a self-hosted
> > > Helm chart repo. It does this – during every push to master – by checking each chart
> > > in your project, and whenever there's a new chart version, creates a corresponding
> > > GitHub release named for the chart version, adds Helm chart artifacts to the release,
> > > and creates or updates an index.yaml file with metadata about those releases, which
> > > is then hosted on GitHub Pages
> > > ```
> > 
> > 
> > This approach sounds like a plan.
> 
> @brtknr status?

I will see if I can use the release on my fork this weekend or the following week.

charts/cinder-csi-plugin/templates/controllerplugin-statefulset.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   labels:
     {{- include "cinder-csi.controllerplugin.labels" . | nindent 4 }}
 spec:
+  serviceName: {{ include "cinder-csi.name" . }}-controller-service
```
strigazi, line 8: maybe add a dummy service too as it is done in the manila chart?

stateful sets need  a service https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#statefulset-v1-apps
```
serviceName is the name of the service that governs this StatefulSet. This service must exist before the StatefulSet, and is responsible for the network identity of the set. Pods get DNS/hostnames that follow the pattern: pod-specific-string.serviceName.default.svc.cluster.local where "pod-specific-string" is managed by the StatefulSet controller.
```

strigazi: I broke it out here: https://github.com/kubernetes/cloud-provider-openstack/pull/1106

hoping for a fast merge :)

strigazi: this is done in #1105

charts/manila-csi-plugin/README.md
ramineni: @brtknr IMO appVersion :1.18.0 is bit confusing , we are working on master branch, so it should be latest , on the release-branch 1.18 we can update the appversion as 1.18.0 , that maintains consistency with the manifests (and can be updated on par when we support new sidecars) . wdyt?

brtknr: Hi @ramineni, thanks for the review. It is difficult to release the `latest` tag because it will quickly go out of date. This pattern of appVersion matching the latest stable version is what a lot of projects do. If a user wants to, they can easily replace the image tag with latest via Helm CLI but it makes more sense for the default behaviour to be stable. Here is the index.yaml produced by chart-releaser action on this PR: https://brtknr.github.io/cloud-provider-openstack/index.yaml as an example of what it looks like. The `latest` tag is a moving target, it is better to be explicit about the version.

brtknr:
> @brtknr How cinder-csi-1.1.0.tgz is gonna be maintained?

@lingxiankong looks like if we use the helm chart releaser github action, the chart gets packaged up as a release artifact: see https://brtknr.github.io/cloud-provider-openstack/index.yaml

charts/manila-csi-plugin/README.md
ramineni: @brtknr How do you suggest to maintain consistency between manifests and helm values . Whenever we do support new sidecar versions on master branch , we cant update the corresponding values in helm charts because that relies on 1.18 (or stable version) . And in future if we would like to use the helm in CI as well , we wont be able to do that , as that needs latest?
What we follow as of now , when we cut a stable branch (release-1.18), we update the tags to 1.18 to reflect the latest stable version and master will be always on latest.

brtknr: Hi @ramineni I have updated the chart so that new releases will use the "latest" tag. I am not sure myself whether this is good practice but hey ho. I have not checked whether stable branches can have their own workflow. We may be talking about different things here. The aim of this PR is to use Github actions to release a chart. At the moment, there is no relationship between chart version and app version. Therefore I am not convinced that updating the values in the stable branch will help us here. However this approach means that if users want to deploy a particular version of cinder/manila-csi-plugin container, they will be able to override the imageTag value to the version they choose instead of running the latest.

brtknr: @

> @brtknr Can you update? File moves etc are fine. We just miss the CI stuff.
> We can mimic [kubernetes/autoscaler#3341](https://github.com/kubernetes/autoscaler/pull/3341) .

Done @strigazi. Do you agree with the latest changes?

ramineni: @brtknr I was more of talking about compatibility of sidecars we support in actual and that of helm would be different if we make it 1.18.0 . Never mind, this PR is for github workflows , so this shouldn't be blocker.

ramineni:
Thanks for the PR. Looks good to me
Not an expert of helm, so leaving it to @strigazi @gman0 @lingxiankong for further review
/lgtm

lingxiankong:
> Thanks for the PR. Looks good to me
> Im not an expert of helm, so leaving it to @strigazi @gman0 @lingxiankong for further review
> /lgtm

Me neither, but I don't think this will be harmful to the whole repo, let's not block this. Leave it to the interested parties for any potential bugfix.

/lgtm
/approve

brtknr:
@lingxiankong So the Github action is currently failing because this change depends on an orphan branch called `gh-pages`, e.g. https://github.com/brtknr/cloud-provider-openstack/tree/gh-pages. Additionally, the settings of this repository needs to reflect that gh-pages branch is used as a website. Are you able to make these changes or ask somebody who is able to? Cheers.

lingxiankong:
> @lingxiankong So the Github action is currently failing because this change depends on an orphan branch called `gh-pages`, e.g. https://github.com/brtknr/cloud-provider-openstack/tree/gh-pages which needs to be manually created. Additionally, the settings of this repository needs to reflect that gh-pages branch is used as a website. Are you able to make these changes or ask somebody who is able to? Cheers.

Hi @brtknr, what's the requirement for the branch? Should it be created based on current master branch? Does it need to be rebased frequently?

brtknr:
Hi @lingxiankong i think the branch needs to be an orphan with a dummy index.html with maybe install instructions like the example above. All the updates are done automatically by GitHub  actions, I.e. publish index.yaml used by helm.

brtknr:
At least the release part seems to have worked: https://github.com/kubernetes/cloud-provider-openstack/releases

lingxiankong:
@brtknr The branch created https://github.com/kubernetes/cloud-provider-openstack/tree/gh-pages

PR #1073, opened 2020-05-22
[cinder-csi-plugin] support "multiattach" volume type 

### LINKED ISSUE #729: Support Multinode attachement Cinder CSI

 /kind feature

**What happened**:
Cinder Volume can be attached to multiple nodes, with each attachment specifying RO/RW modes, which corresponds to following CSI caps
VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY
VolumeCapability_AccessMode_MULTI_NODE_SINGLE_WRITER
VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER  

Intially we will add support for first two modes. 

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

ramineni:
When the type of the volume is of "multiattach", it can be attached to multiple instances.
Currently type of the volume can be specified as part of parameters['type'] of storageclass , https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/controllerserver.go#L63
This PR enables the same for multiattach type as well .
fixes #729 

Specify the 'type' as 'multiattach' in storageclass as specified in example doc.
type should exist in backend.

**Release note**:

```release-note
[cinder-csi-plugin] Support to attach a volume to multiple instances if ‘multiattach’ flag is enabled in Cinder volume type.
```

### DIFF
--- a/docs/using-cinder-csi-plugin.md
+++ b/docs/using-cinder-csi-plugin.md
@@ -436,6 +436,25 @@ Following prerequisites needed for volume cloning to work :
 
 Sample yamls can be found [here](https://github.com/kubernetes/cloud-provider-openstack/tree/master/examples/cinder-csi-plugin/clone)
 
+### Multi-Attach Volumes
+
+To avail the multiattach feature of cinder, specify the ID/name of cinder volume type that includes an extra-spec capability setting of `multiattach=<is> True` in storage class parameters as shown below.
+
+> Note: This volume type must exist in cinder already (`openstack volume type list`)
+
+This should enable to attach a volume to multiple hosts/servers simultaneously.
+
+Example:
+
+```
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: csi-sc-cinderplugin
+provisioner: cinder.csi.openstack.org
+parameters:
+  type: <multiattach-volume-type>
+
 ## Running Sanity Tests
 
 Sanity tests create a real instance of driver and fake cloud provider.
--- a/pkg/csi/cinder/openstack/openstack_volumes.go
+++ b/pkg/csi/cinder/openstack/openstack_volumes.go
@@ -59,6 +59,7 @@ func (os *OpenStack) CheckBlockStorageAPI() error {
 
 // CreateVolume creates a volume of given size
 func (os *OpenStack) CreateVolume(name string, size int, vtype, availability string, snapshotID string, sourcevolID string, tags *map[string]string) (*volumes.Volume, error) {
+
 	opts := &volumes.CreateOpts{
 		Name:             name,
 		Size:             size,
@@ -145,20 +146,31 @@ func (os *OpenStack) GetVolume(volumeID string) (*volumes.Volume, error) {
 
 // AttachVolume attaches given cinder volume to the compute
 func (os *OpenStack) AttachVolume(instanceID, volumeID string) (string, error) {
+	computeServiceClient := os.compute
+
 	volume, err := os.GetVolume(volumeID)
 	if err != nil {
 		return "", err
 	}
 
-	if len(volume.Attachments) > 0 {
-		if instanceID == volume.Attachments[0].ServerID {
+	for _, att := range volume.Attachments {
+		if instanceID == att.ServerID {
 			klog.V(4).Infof("Disk %s is already attached to instance %s", volumeID, instanceID)
 			return volume.ID, nil
 		}
-		return "", fmt.Errorf("disk %s is attached to a different instance (%s)", volumeID, volume.Attachments[0].ServerID)
 	}
 
-	_, err = volumeattach.Create(os.compute, instanceID, &volumeattach.CreateOpts{
+	if volume.Multiattach {
+		// For multiattach volumes, supported compute api version is 2.60
+		// Init a local thread safe copy of the compute ServiceClient
+		computeServiceClient, err = openstack.NewComputeV2(os.compute.ProviderClient, os.epOpts)
+		if err != nil {
+			return "", err
+		}
+		computeServiceClient.Microversion = "2.60"
+	}
+
+	_, err = volumeattach.Create(computeServiceClient, instanceID, &volumeattach.CreateOpts{
 		VolumeID: volume.ID,
 	}).Extract()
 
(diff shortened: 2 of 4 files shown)


pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -145,20 +146,32 @@ func (os *OpenStack) GetVolume(volumeID string) (*volumes.Volume, error) {
...
 // AttachVolume attaches given cinder volume to the compute
 func (os *OpenStack) AttachVolume(instanceID, volumeID string) (string, error) {
+	computeServiceClient := os.compute
+
 	volume, err := os.GetVolume(volumeID)
 	if err != nil {
 		return "", err
 	}
 
 	if len(volume.Attachments) > 0 {
```
jichenjc, line 156: this check not needed?

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -205,22 +218,16 @@ func (os *OpenStack) DetachVolume(instanceID, volumeID string) error {
 		return nil
 	}
 
-	if volume.Status != VolumeInUseStatus {
-		return fmt.Errorf("can not detach volume %s, its status is %s", volume.Name, volume.Status)
-	}
```
jichenjc, line 210: should we check MultiAttach flag ? otherwise, for single attach, we remove that check?

ramineni: Right, this check can be kept as it is. Will update.

docs/using-cinder-csi-plugin.md
```diff
@@ -434,6 +434,24 @@ Following prerequisites needed for volume cloning to work :
...
+Example:
+
+```
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: csi-sc-cinderplugin
+provisioner: cinder.csi.openstack.org
+parameters:
+  type: multiattach
```
lingxiankong, line 451: I'm wondering why it's related to volume type? I thought it's related to `accessModes` when defining PersistentVolumeClaim.

lingxiankong (commented):
Do we need to change here https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/driver.go#L84? If yes, how we could give the csi driver deployer the ability to define if the MULTI_NODE_MULTI_WRITER access mode is supported?

docs/using-cinder-csi-plugin.md
ramineni: @lingxiankong right , there are 2 ways of doing it (either way creating volume type is required if we need to attach to more than one node) 
1. Support automatically create volume type of multiattach , depending on the access modes that are specified in pvc (ReadWriteMany, ReadOnlyMany..).
2. Use the existing field of volume type we already support and create volume of that type. 

So, in the initial support we provide in k8s, I thought it would be better to leave it for operator if they really would like to create the volume of type multiattach,(understanding their limitations ) instead of creating it automatically for them which might cause issues later on . 

https://docs.openstack.org/cinder/latest/admin/blockstorage-volume-multiattach.html#ro-rw-caveats-the-secondary-rw-attachment-issue .

jichenjc: I think leave to operator to decide is a better option here..
if customer has requirement on option 1) above, we can consider it later on

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -259,13 +269,15 @@ func (os *OpenStack) GetAttachmentDiskPath(instanceID, volumeID string) (string,
...
-		if instanceID == volume.Attachments[0].ServerID {
-			return volume.Attachments[0].Device, nil
+	if len(volume.Attachments) > 0 {
+		for _, att := range volume.Attachments {
+			if att.ServerID == instanceID {
+				return att.Device, nil
+			}
 		}
-		return "", fmt.Errorf("disk %q is attached to a different compute: %q, should be detached before proceeding", volumeID, volume.Attachments[0].ServerID)
+		return "", fmt.Errorf("disk %q is not attached to compute: %q", volumeID, instanceID)
```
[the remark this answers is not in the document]
ramineni: right, its compute node

lingxiankong (commented):
IIUC, supporting multi-attach is kind of transparent to the end users, right? The requirements are:

* The openstack cloud admin creates a multiattach enabled volume type.
* The k8s cluster admin creates a storage class with that volume type.

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -209,18 +221,16 @@ func (os *OpenStack) DetachVolume(instanceID, volumeID string) error {
 		return fmt.Errorf("can not detach volume %s, its status is %s", volume.Name, volume.Status)
 	}
 
+	// Incase volume is of type multiattach, it could be attached to more than one instance
 	if len(volume.Attachments) > 0 {
-		if volume.Attachments[0].ServerID != instanceID {
-			return fmt.Errorf("disk: %s is not attached to compute: %s", volume.Name, instanceID)
-		}
 		err = volumeattach.Delete(os.compute, instanceID, volume.ID).ExtractErr()
```
lingxiankong, line 226: We still need to check if the volume has been attached to the instance.

docs/using-cinder-csi-plugin.md
```diff
@@ -434,6 +434,24 @@ Following prerequisites needed for volume cloning to work :
 
 Sample yamls can be found [here](https://github.com/kubernetes/cloud-provider-openstack/tree/master/examples/cinder-csi-plugin/clone)
 
+### Multi-Attach Volumes
+
+To avail the multiattach feature of cinder, specify the 'type' of the volume as 'multiattach' in storageclass definition.
```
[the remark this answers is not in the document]
jichenjc: this is a good suggestion, 

`Starting from the Queens release the ability to attach a volume to multiple hosts/servers requires that the volume is of a special type that includes an extra-spec capability setting of multiattach=<is> True. You can create the volume type the following way:` reflect your suggestion

pkg/csi/cinder/openstack/openstack_volumes.go
ramineni: ya, I suppose it could be done, loop through attachments.  I felt it should be handled by delete call anyway. I can add the check

ramineni:
> IIUC, supporting multi-attach is kind of transparent to the end users, right? The requirements are:
> 
> * The openstack cloud admin creates a multiattach enabled volume type.
> * The k8s cluster admin creates a storage class with that volume type.

Correct

ramineni:
@jichenjc PTAL

ekeih:
@ramineni Thanks for implementing this. As you did this, I assume that you also use this in production in some capacity. Are you able to share some insights how you manage a cluster filesystem ontop of this?

baurmatt:
Did anyone managed to run this in production with k8s creating the cluster filesystem on the attached volume?

Issue #1080, opened 2020-05-27
[occm] Investigate to remove vendored code from occm


mikejoh:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind feature

I've created this issue to discuss and investigate if we can switch to fully use the k/k `cloud-controller-manager` code as seen [here](https://github.com/kubernetes/kubernetes/blob/master/cmd/cloud-controller-manager/controller-manager.go), and not use our own implementation of parts of the k/k `cloud-controller-manager` in [this](https://github.com/kubernetes/cloud-provider-openstack/blob/master/cmd/openstack-cloud-controller-manager/main.go) part of the `openstack-cloud-controller-manager`. 

I've been reading up on and looking a lot on the different external cloud controller managers out there as well as the packages and source code that makes up the cloud-controller-manager and it's internals.

Today the `main()` entrypoint looks like [this](https://github.com/kubernetes/cloud-provider-openstack/blob/master/cmd/openstack-cloud-controller-manager/main.go). My concern is that we have vendored code within this repository that we actually should be used by consuming the `k8s.io/kubernetes/cmd/cloud-controller-manager/app`. If i compare the code of the init function of the different controllers within the cloud-controller-manager it has actually diverged already:

https://github.com/kubernetes/kubernetes/blob/356c121c6d16240b4f5ad7dc58c3065b9adca28e/cmd/cloud-controller-manager/app/core.go#L105

vs

https://github.com/kubernetes/cloud-provider-openstack/blob/3909137e68519733d9832ef2dabd812550ca6884/cmd/openstack-cloud-controller-manager/main.go#L227-L259

I'm _almost_ convinced that we don't need more than the [k/k cloud-controller-manager `main.go`](https://github.com/kubernetes/kubernetes/blob/master/cmd/cloud-controller-manager/controller-manager.go) code to run our (`openstack-cloud-controller-manager`). Of course there's parts that we'll need to add but the majority of the code can be removed. So basically there's no need for **us** to maintain that code!

Questions:
* Does this make any sense?
* Is there any good reasons for keeping that code vendored instead of consuming the upstream? 

**What happened**:
N/A

**What you expected to happen**:
N/A

**How to reproduce it**:
N/A

**Anything else we need to know?**:
N/A

**Environment**:
N/A

mikejoh:
[...]

PR #1081, opened 2020-05-27
[WIP] [occm] Code removal and fully consume k/k cloud-controller-manager

mikejoh:
Part of the discussion and investigation in #1080 

This is a *really* rough WIP PR that just shows how much we possibly could strip off the occm in regards to LoC. At the same time also consume the k/k cloud-controller-manager utility and boilerplate code an not have parts of it maintaind in our source.

Please note that the part of the code where the `openstack.ProviderName` is internally set on the `--cloud-provider` flag can be skipped but it's a nice way of enforcing that this cloud-controller-manager can only be started as a cloud provider of `openstack`. [Inspired by how the DigitalOcean CCM is doing this](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/cloud-controller-manager/cmd/digitalocean-cloud-controller-manager/main.go).

I've remove loads of flag and argument handling, some of that might be needed, i have added comments to the PR commit to elaborate a bit more and mark parts that needs to stay.

I haven't had time to test more than building and testing the binary (outside of a k8s cluster on OpenStack). I will try to run it within a OpenStack k8s cluster this week. Maybe the CI will catch anything obvious.

Do not *merge*! :+1: 

WIP fix for #1080 

**Release note**:
```release-note
NONE
```


cmd/openstack-cloud-controller-manager/main.go
```diff
@@ -20,240 +20,56 @@ limitations under the License.
...
-
-	fs := command.Flags()
-	namedFlagSets := s.Flags(KnownControllers(), ControllersDisabledByDefault.List())
-	for _, f := range namedFlagSets.FlagSets {
-		fs.AddFlagSet(f)
-	}
-
-	fs.BoolVar(&versionFlag, "version", false, "Print version and exit")
-
-	openstack.AddExtraFlags(pflag.CommandLine)
```
mikejoh, line 124: This would need to stay, it's adding this command line flag:

```
// AddExtraFlags is called by the main package to add component specific command line flags
func AddExtraFlags(fs *pflag.FlagSet) {
	fs.StringArrayVar(&userAgentData, "user-agent", nil, "Extra data to add to gophercloud user-agent. Use multiple times to add more than one component.")
}
```

cmd/openstack-cloud-controller-manager/main.go
```diff
@@ -20,240 +20,56 @@ limitations under the License.
...
 	// normalize func and add the go flag set by hand.
-	pflag.CommandLine.SetNormalizeFunc(flag.WordSepNormalizeFunc)
-	pflag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 	// utilflag.InitFlags()
 	logs.InitLogs()
 	defer logs.FlushLogs()
 
-	klog.V(1).Infof("openstack-cloud-controller-manager version: %s", version.Version)
-
-	s.KubeCloudShared.CloudProvider.Name = openstack.ProviderName
```
mikejoh, line 137: I'm a bit unsure here, but i think this is set while using the `--cloud-provider` flag, which in this commit is done by parsing all command line flags and if `cloud-provider` is found i set that to `openstack` using the `openstack.ProviderName` variable.

cmd/openstack-cloud-controller-manager/main.go
```diff
@@ -20,240 +20,56 @@ limitations under the License.
...
-		},
-	}
-
-	fs := command.Flags()
-	namedFlagSets := s.Flags(KnownControllers(), ControllersDisabledByDefault.List())
-	for _, f := range namedFlagSets.FlagSets {
-		fs.AddFlagSet(f)
-	}
-
-	fs.BoolVar(&versionFlag, "version", false, "Print version and exit")
```
mikejoh, line 122: This would also need to stay in some form.

mikejoh:
Instead of commenting and leaving this PR kind of half done i'll redo the work in another PR, starting with a draft.

PR #1093, opened 2020-06-09
Update using-auth-data-synchronization.md

wangxy518:
blacklist->blocklist


docs/using-auth-data-synchronization.md
```diff
@@ -36,11 +36,11 @@ The config options:
 
   To correctly create the *rolebindings*, the cluster admin should create the *clusterrole* (same with Keystone role name) first. For example, if the user *alice* has a role assignment *member* in Keystone, when *alice* accesses the cluster, before Kubernetes actually does authorization, the webhook should create a new *rolebinding* in the new namespace with the *clusterrole* name *member* and the user *alice*. As a result, the user *alice* should have some pre-defined resource permissions even it's the first time to access the cluster.
 
-* **projects-blacklist**
+* **projects-blocklist**
```
ramineni, line 39: @wangxy518 blacklist is the right word used here , which means to avoid/ exclude

lingxiankong: Agree

PR #1094, opened 2020-06-09
Update sync.go

wangxy518:
blacklist->blocklist


pkg/identity/keystone/sync.go
```diff
@@ -59,10 +59,10 @@ type syncConfig struct {
 	NamespaceFormat string `yaml:"namespace-format"`
 
 	// List of project ids to exclude from syncing.
-	ProjectBlackList []string `yaml:"projects-blacklist"`
+	ProjectBlockList []string `yaml:"projects-blacklist"`
```
jichenjc, line 62: it should be blacklist ...

ramineni:
blacklist is the right word used 
/close

PR #1095, opened 2020-06-09
Update sync_test.yaml

wangxy518:
blacklist->blocklist


jichenjc:
it should be blacklist instead of blocklist..

ramineni:
blacklist is the right word used, so closing it
/close

PR #1099, opened 2020-06-11
[occm] Add error message in RegisterCloudProvider()

oomichi:

RegisterCloudProvider() calls NewOpenStack() but we cannot see the root cause if an error happens due to lack of actual error message.
This adds the error message in log output.

Now I am facing an issue failed to deploy occm from Kubespray and it is hard to debug it because of this lack of the error message.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack.go
+++ b/pkg/cloudprovider/providers/openstack/openstack.go
@@ -293,7 +293,7 @@ func init() {
 		}
 		cloud, err := NewOpenStack(cfg)
 		if err != nil {
-			klog.V(1).Infof("New openstack client created failed with config")
+			klog.Warningf("New openstack client created failed with config: %v", err)
 		}
 		return cloud, err
 	})


oomichi:
I guess I am facing the issue of https://github.com/kubernetes-sigs/kubespray/pull/6262 on Kubespray.
This PR will be helpful to investigate this kind of issues.

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -293,7 +293,7 @@ func init() {
 		}
 		cloud, err := NewOpenStack(cfg)
 		if err != nil {
-			klog.V(1).Infof("New openstack client created failed with config")
+			klog.Warningf("New openstack client created failed with config: %s", err)
```
ramineni, line 296: s/ %s / %v

zetaab: `			klog.Warningf("New openstack client created failed with config: %v", err)`

mikejoh: Out of curiosity, why `%v`? Is it in case of  `err` being of a type other than a `string`?

oomichi: Thanks, done.

mikejoh:
The conformance test failure(s) we see here have reported: https://github.com/kubernetes/kubernetes/issues/91739

jichenjc:
/lgtm

seems such change don't need wait conformance test but better to have

oomichi:
/cc @dims 

Please take a look :-)

dims:
/approve

next time please ping @ramineni :)

oomichi:
@dims Thanks!
OK, I will do it next time.

jichenjc:
seems https://github.com/kubernetes/kubernetes/issues/91739 still under discussion 
so maybe need wait more time on it..

lingxiankong:

PR #1102, opened 2020-06-22
[occm] improve logic for floating IP creation 

hamzazafar:
**Refactor/Improve logic for Floating IP creation**:

The PR fixes the following problems: 
* "service.beta.kubernetes.io/openstack-internal-load-balancer" annotation is ignored by occm:
If floating network ID and floating subnet ID is not found in load balancer class then occm sets the internal loadbalancer setting to True, it simply ignores the annotation set by the user [ref](https://github.com/hamzazafar/cloud-provider-openstack/blob/a7e33e5a83027f87c9781cb2cb21ab423fad99a4/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go#L860). openstack-internal-load-balancer annotation should be used as a source of truth.

* occm tries to fetch floating network and subnet ID from loadbalancer class in the beginning of EnsureLoadbalancer function [ref](https://github.com/hamzazafar/cloud-provider-openstack/blob/a7e33e5a83027f87c9781cb2cb21ab423fad99a4/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go#L852), it then creates listeners, pools members etc and finally tries to create a floating IP. If Load balancer class is not set then it tries to read floating network information from annotations or from cloud-config file [ref](https://github.com/hamzazafar/cloud-provider-openstack/blob/a7e33e5a83027f87c9781cb2cb21ab423fad99a4/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go#L1296). If floating network information is not found it throws an error. We have 2 problems here:
1. Floating network information shouldn't be processed at the end of Ensureloadbalancer func, errors/warnings should be thrown before creating any lb resources (pools, members, etc)
2. If floating network information is not found in lb class then occm tries to ensure an internal loadbalancer [ref](https://github.com/hamzazafar/cloud-provider-openstack/blob/a7e33e5a83027f87c9781cb2cb21ab423fad99a4/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go#L859), when lb class is not used and floating network information is not provided then occm throws an error [ref](https://github.com/hamzazafar/cloud-provider-openstack/blob/a7e33e5a83027f87c9781cb2cb21ab423fad99a4/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go#L1310), this is clearly an inconsistent behavior of occm.

* If the internal loadbalancer annotation is set by user, occm still tries to get the floating IP attached to load balancer's VIP port. This is completely unnecessary in case of internal lbs [ref](https://github.com/hamzazafar/cloud-provider-openstack/blob/a7e33e5a83027f87c9781cb2cb21ab423fad99a4/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go#L1251). 

This PR fixes the above mentioned issues and improves the code quality for floating ip creation.
**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
@@ -901,6 +901,18 @@ func (lbaas *LbaasV2) EnsureLoadBalancer(ctx context.Context, clusterName string
 				floatingSubnetID = getStringFromServiceAnnotation(apiService, ServiceAnnotationLoadBalancerFloatingSubnetID, lbaas.opts.FloatingSubnetID)
 			}
 		}
+
+		// check subnets belongs to network
+		if floatingNetworkID != "" && floatingSubnetID != "" {
+			subnet, err := subnets.Get(lbaas.network, floatingSubnetID).Extract()
+			if err != nil {
+				return nil, fmt.Errorf("Failed to find subnet %q: %v", floatingSubnetID, err)
+			}
+
+			if subnet.NetworkID != floatingNetworkID {
+				return nil, fmt.Errorf("FloatingSubnet %q doesn't belong to FloatingNetwork %q", floatingSubnetID, floatingSubnetID)
+			}
+		}
 	} else {
 		klog.V(4).Infof("Ensure an internal loadbalancer service.")
 	}
--- a/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
@@ -845,52 +845,64 @@ func (lbaas *LbaasV2) EnsureLoadBalancer(ctx context.Context, clusterName string
 		return nil, fmt.Errorf("no ports provided to openstack load balancer")
 	}
 
-	internalAnnotation := false
+	internalAnnotation, err := getBoolFromServiceAnnotation(apiService, ServiceAnnotationLoadBalancerInternal, lbaas.opts.InternalLB)
+	if err != nil {
+		return nil, err
+	}
+
 	var lbClass *LBClass
-	var err error
-	class := getStringFromServiceAnnotation(apiService, ServiceAnnotationLoadBalancerClass, "")
-	if class != "" {
-		lbClass = lbaas.opts.LBClasses[class]
-		if lbClass == nil {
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
```diff
@@ -845,52 +845,64 @@ func (lbaas *LbaasV2) EnsureLoadBalancer(ctx context.Context, clusterName string
...
+				lbSubnet, err := lbaas.getSubnet(floatingSubnetName)
+				if err != nil || lbSubnet == nil {
+					klog.Warningf("Failed to find floating-subnet-id for Service %s: %v", serviceName, err)
+				} else {
+					floatingSubnetID = lbSubnet.ID
+				}
+			}
+
+			if floatingSubnetID == "" {
+				floatingSubnetID = getStringFromServiceAnnotation(apiService, ServiceAnnotationLoadBalancerFloatingSubnetID, lbaas.opts.FloatingSubnetID)
```
lingxiankong, line 901: Can we check if the subnet ID is in the floating network?

hamzazafar: sure, that's a good check to make. I'll update the PR

hamzazafar: I have pushed a new [commit](https://github.com/kubernetes/cloud-provider-openstack/pull/1102/commits/039f611cc1c6d37a7e72b76408991e617d5e82f9) to check if subnet belongs to the network. Can you review again? Thanks

lingxiankong:
I have to re-installed my dev environment to verify this PR.

lingxiankong:

/lgtm

dr4Ke:
I believe this merge has introduced the need to specify both `floating-network-id` and `floating-subnet-id` in the loadbalancer configuration (as annotation or in the global configuration).

It wasn't the case before and the documentation doesn't mention the `floating-subnet-id`.

Am I wrong?

hamzazafar:
@dr4Ke Thanks for pointing that out. You are correct, floating subnet ID was optional before this merge. 

I'll open a new PR to make it optional and improve the docs.

hamzazafar:
@dr4Ke I have opened to a [PR](https://github.com/kubernetes/cloud-provider-openstack/pull/1130) to make floating subnet optional.

PR #1105, opened 2020-06-25
Add serviceName to cinder-csi-plugin/templates/controllerplugin-statefulset.yaml

possani:
**What this PR does / why we need it**:

It fixes the following error:

helm install cinder-csi cloud-provider-openstack/charts/cinder-csi-plugin
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(StatefulSet.spec): missing required field "serviceName" in io.k8s.api.apps.v1.StatefulSetSpec

**Which issue this PR fixes(if applicable)**:
fixes #

**Special notes for reviewers**:

**Release note**:

```release-note
NONE
```

PR #1106, opened 2020-06-25
[cinder-csi-plugin] Add serviceName in statefulset

strigazi:
serviceName is required for stateful sets
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#statefulset-v1-apps

Add a dummy service like in manila-csi so that serviceName points to something.

Since there a change in the chart, make a point release.

```release-note
[cinder-csi-plugin] Add serviceName in chart's statefulset chart version: 1.1.1  
```


brtknr:
+1

ramineni:
@strigazi Thanks for adding this . But there seems to be already PR raised for this , https://github.com/kubernetes/cloud-provider-openstack/pull/1105 . Its duplicate.

Issue #932, opened 2020-02-11
[occm] Support to use an existing loadbalancer for service type loadbalancer


hamzazafar:
/kind feature

**The binaries affected**:

- [x] openstack-cloud-controller-manager
- [ ] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] octavia-ingress-controller
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

**What happened**:
There is a one-to-one mapping of k8s services and external loadbalancers. Creating a new loadbalancer in cloud for each k8s service can be costly for customers. [octavia-ingress-controller](https://github.com/kubernetes/cloud-provider-openstack/blob/08824aaabec43b3ede9a8af3e9aed897fa3bf054/docs/using-octavia-ingress-controller.md) already solves this problem by creating an external loadbalancer and setting some L7 policies to support multiple k8s services behind the same loadbalancer.

**Problem**
Not all octavia provider drivers support L7 policies.

**Solution**
Add support in OCCM to use an existing loadbalancer instead of creating a new one.

hamzazafar:
User can create a loadbalancer using OpenStack CLI and then specify the ID of loadbalancer in an annotation while creating a k8s service. 

I'll be happy to provide a patch if there is interest in this feature

rochaporto:
Adding here... the goal would be to add serviceType: LoadBalancer instances into existing load balancers, and handle the deletion in the same way (drop the members only, not the whole LB instance).

Looking at the code, is the best way to do this to:
1. try to refactor the ensureloadbalancer code
2. hook one more condition into both ensureloadbalancer and ensureloadbalancerdelete

lingxiankong:
Hi, thanks for proposing this feature request. However, personally, it doesn't make sense for me to re-use load balancer in the cloud for k8s services. As mentioned in the description, if the end users have concerns about the cloud resources consumption, Ingress is what they should use.

rochaporto:
@lingxiankong can you expand on why this doesn't make sense? It isn't just for optimizing resource usage, we have multiple use cases where services have endpoints in both VMs and Kubernetes clusters and need to share an external load balancer.

lingxiankong:
[...]

PR #1118, opened 2020-07-14
[occm] Support for adding members to an existing pool

### LINKED ISSUE #933: [occm] Allow ip address sharing for services of type loadbalancers
>
> /kind bug
/kind feature

**The binaries affected**:

- [ ] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

**What happened**:

* Created two services of `type: loadBalancer` having set the same `svc.spec.loadBalancerIP`
* No listener, pool, healthmonitor or members are created for the second service:
  * kubectl describe svc shows:
    `Warning  SyncLoadBalancerFailed  2m32s (x6 over 5m9s)   service-controller  Error syncing load balancer: failed to ensure load balancer: floating IP X.X.X.X is not available`

* Controller will add a new listener, pool, healthmonitor and members for the additional port/nodeport to the existing loadbalancer having floating IP X.X.X.X

**How to reproduce it**:

```
$ cat service.yaml 
apiVersion: v1
kind: Service
metadata:
  name: kuard-1
spec:
  type: LoadBalancer
  loadBalancerIP: 192.168.200.5
  ports:
  - port: 80
    protocol: TCP
    targetPort: 8080
  selector:
    app: kuard
$ cat service2.yaml 
apiVersion: v1
kind: Service
metadata:
  name: kuard-2
spec:
  type: LoadBalancer
  loadBalancerIP: 192.168.200.5
  ports:
  - port: 8080
    protocol: TCP
    targetPort: 8080
  selector:
    app: kuard
$ kubectl apply -f service.yaml -f service2.yaml
```

* Metall LB implements this feature: [docs](https://metallb.universe.tf/usage/#ip-address-sharing)
* We could implement it similar to Metall LB and toggle it via an annotation on services analog to `metallb.universe.tf/allow-shared-ip`

What get's touched by implementing it:
* The creation and reconcile of a loadbalancer by the service name may not work.
  * The above exampe creates the following loadbalancers:
    ```
    +--------------------------------------+-----------------------------------------+-------------+---------------------+------------+
    | id                                   | name                                    | vip_address | provisioning_status | provider   |
    +--------------------------------------+-----------------------------------------+-------------+---------------------+------------+
    | cd117dcf-518e-4a34-8613-d2ba29b3f9e2 | kube_service_kubernetes_default_kuard-1 | 10.6.0.32   | ACTIVE              | vmwareedge |
    | df2281b6-6c0b-46a2-bfbd-6566b5697372 | kube_service_kubernetes_default_kuard-2 | 10.6.0.10   | ACTIVE              | vmwareedge |
    ```
    We would need to adjust the following code to have a common name for mutliple services for the loadBalancer, or detect the loadBalancer by other loadBalancer metadata: https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go#L520
  
  We could enforce the need to set a loadBalancerIP when the `ip-address-sharing` annotation is set and resolve the name for the loadBalancer by using the loadBalancerIP by using for example something like `kube_service_x-x-x-x` for the loadBalancer name

* We would need some additional validation for e.g.:
  * do not allow multiple use of the external port
  * potential other invalid options when merging multiple services of type loadBalancer (e.g. annotations)

There may be other things I didn't consider here.

**Environment**:
- OpenStack version: Queens
- Others: Kubernetes v1.16

hamzazafar:

fixes # [932](https://github.com/kubernetes/cloud-provider-openstack/issues/932)
fixes #933

This PR brings support for adding members to an existing pool. When `loadbalancer.openstack.org/unmanaged-pool-id` annotation is set, then cloud provider will simply add or remove members from the given pool. Following requirements should be met:
* User has to make sure that the port number specified in service spec matches with the port number of listener attached to the pool.
* User is expected to create loadbalancer, listener, pool and healthmonitor resources beforehand using openstack CLI. 
* Only one port could be specified in service spec because a single pool is referenced in `loadbalancer.openstack.org/unmanaged-pool-id` annotation and a pool is attached to a single listener. 

Some other stuff:
* When a service with annotation `loadbalancer.openstack.org/unmanaged-pool-id` is deleted then cloud provider will only delete the members.
* When nodes are added or removed from the k8s cluster, then cloud provider will automatically add or remove members from the pool respectively.
* **Only new functionality has been added. Existing features of cloud provider will continue working as before**

With this PR, occm will be able to support the following usecases:

* Using a single loadbalancer to expose services running on multiple k8s clusters
* Using a single loadbalancer to expose multiple services running on the same k8s cluster
* Using a single loadbalancer to serve traffic on a combination of VMs and k8s services
* Minimize blast radius of applications running on k8s:
  For HA purposes, multiple instances of application can be deployed in different K8s clusters (running in different availability zones). With this feature, a user can expose multiple instances of application using the same loadbalancer. If one k8s cluster goes down (let's say from power outage) then users will experience a reduced capacity instead of a complete downtime for application.  
* Save cost associated with public IP addresses by exposing multiple services behind a single loadbalancer. 

**Release note**:

```release-note
support for adding members to an existing pool
```

### DIFF
--- a/docs/expose-applications-using-loadbalancer-type-service.md
+++ b/docs/expose-applications-using-loadbalancer-type-service.md
@@ -147,6 +147,10 @@ Request Body:
 
   Defines whether or not to create health monitor for the load balancer pool, if not specified, use `create-monitor` config. The health monitor can be created or deleted dynamically.
 
+- `loadbalancer.openstack.org/unmanaged-pool-id`
+
+  Cloud provider will simply add or remove members from the pool. Creation and deletion of pool, listener, healthmonitor and loadbalancer will not be performed by cloudprovider---user is expected to create these resources beforehand using openstack CLI. With this feature services running on different k8s clusters can be served from the same loadbalancer.
+  
 ### Switching between Floating Subnets by using preconfigured Classes
 
 If you have multiple `FloatingIPPools` and/or `FloatingIPSubnets` it might be desirable to offer the user logical meanings for `LoadBalancers` like `internetFacing` or `DMZ` instead of requiring the user to select a dedicated network or subnet ID at the service object level as an annotation.
--- a/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
@@ -220,7 +220,7 @@ func getListenerForPort(existingListeners []listeners.Listener, port corev1.Serv
 	return nil
 }
 
-// Get listener for a pool. A pool is always assosiated with a single listener
+// Get listener for a pool. A pool is always associated with a single listener
 func getListenerByPoolID(client *gophercloud.ServiceClient, poolID string) (*listeners.Listener, error) {
 	pool, err := v2pools.Get(client, poolID).Extract()
 	if err != nil {
@@ -241,7 +241,7 @@ func getListenerByPoolID(client *gophercloud.ServiceClient, poolID string) (*lis
 	return listener, nil
 }
 
-// Get loadbalancer assosiated with a pool.
+// Get loadbalancer associated with a pool.
 func getLoadBalancerByPoolID(client *gophercloud.ServiceClient, poolID string) (*loadbalancers.LoadBalancer, error) {
 	pool, err := v2pools.Get(client, poolID).Extract()
 	if err != nil {
@@ -249,7 +249,7 @@ func getLoadBalancerByPoolID(client *gophercloud.ServiceClient, poolID string) (
 	}
 
 	if len(pool.Loadbalancers) == 0 {
-		return nil, fmt.Errorf("No Loadbalancer is assosiated with pool %q", poolID)
+		return nil, fmt.Errorf("No Loadbalancer is associated with pool %q", poolID)
 	}
 
 	loadbalancerID := pool.Loadbalancers[0].ID
--- a/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
@@ -98,6 +98,10 @@ const (
 	// ServiceAnnotationLoadBalancerEnableHealthMonitor defines whether or not to create health monitor for the load balancer
 	// pool, if not specified, use 'create-monitor' config. The health monitor can be created or deleted dynamically.
 	ServiceAnnotationLoadBalancerEnableHealthMonitor = "loadbalancer.openstack.org/enable-health-monitor"
+	// When ServiceAnnotationLoadBalancerUnmanagedPoolID annotation is set, cloud provider will simply add/remove members
+	// from the pool. Creation and deletion of pool, listener, healthmonitor and loadbalancer will not be performed by
+	// cloudprovider. User has to create a pool beforehand using openstack CLI.
+	ServiceAnnotationLoadBalancerUnmanagedPoolID = "loadbalancer.openstack.org/unmanaged-pool-id"
 )
 
 // LbaasV2 is a LoadBalancer implementation for Neutron LBaaS v2 API
@@ -214,6 +218,47 @@ func getListenerForPort(existingListeners []listeners.Listener, port corev1.Serv
 	}
 
 	return nil
+}
+
+// Get listener for a pool. A pool is always assosiated with a single listener
+func getListenerByPoolID(client *gophercloud.ServiceClient, poolID string) (*listeners.Listener, error) {
+	pool, err := v2pools.Get(client, poolID).Extract()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get pool %q: %v", poolID, err)
+	}
+
+	if len(pool.Listeners) == 0 {
+		return nil, fmt.Errorf("No Listener is attached to pool %q", poolID)
+	}
+
+	listenerID := pool.Listeners[0].ID
+
+	listener, err := listeners.Get(client, listenerID).Extract()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get listener %q: %v", listenerID, err)
+	}
+
+	return listener, nil
+}
+
+// Get loadbalancer assosiated with a pool.
+func getLoadBalancerByPoolID(client *gophercloud.ServiceClient, poolID string) (*loadbalancers.LoadBalancer, error) {
+	pool, err := v2pools.Get(client, poolID).Extract()
+	if err != nil {
+		return nil, err
+	}
+
+	if len(pool.Loadbalancers) == 0 {
+		return nil, fmt.Errorf("No Loadbalancer is assosiated with pool %q", poolID)
+	}
+
+	loadbalancerID := pool.Loadbalancers[0].ID
+	loadbalancer, err := loadbalancers.Get(client, loadbalancerID).Extract()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get loadbalancer %q: %v", loadbalancerID, err)
+	}
+
+	return loadbalancer, nil
 }
 
 // Get pool for a listener. A listener always has exactly one pool.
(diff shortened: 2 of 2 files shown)


hamzazafar:
> * TIMED_OUT



> /test cloud-provider-openstack-acceptance-test-e2e-conformance

Following two e2e tests have failed:
- Services should be able to change the type from ExternalName to NodePort [Conformance]
- Services should be able to create a functioning NodePort service [Conformance]

These tests are not related to service type loadbalancers. Can you re-trigger e2e conformance tests?

lingxiankong:
@hamzazafar I know, I was using this PR to debug e2e test CI job, sorry for annoying, have been working on this for a long time. I only have bandwidth to review all the non-trivial PRs until solving the CI issue.

hamzazafar:
> @hamzazafar I know, I was using this PR to debug e2e test CI job, sorry for annoying, have been working on this for a long time. I only have bandwidth to review all the non-trivial PRs until solving the CI issue.

@lingxiankong  no problem at all. Take your time :)

lingxiankong:
Finally, I have a chance to look at this PR.

The idea sounds a little bit hacking to me, as the end user needs to create most of the load balancer resources manually before creating a LoadBalancer type Service in Kubernetes, which bring the users too much overhead.

By re-using load balancer in the cloud, why not just adding a new annotation e.g. `loadbalancer.openstack.org/shared-loadbalancer`, so services with the same annotation could share a single load balancer in Octavia but using different listeners (services must use different ports). What do you think?

hamzazafar:
I agree that this feature brings a lot of overhead for users to create loadbalancer resources. 

We want to support a special use case of adding services (running of different k8s clusters) with the same port to a shared loadbalancer. In order to keep things simple, we have added the requirement to create loadbalancer pool by hand. If we automate this using cloud provider then maintaining the ownership of resources will be a challenge. 

For services with different port we can automate the creation of listeners, pools and members! Let me know if i should proceed with this change.

lingxiankong:
I think with an annotation like `loadbalancer.openstack.org/shared-loadbalancer`, even different Services in different clusters can share a single load balancer, which could satisfy your requirement.

> maintaining the ownership of resources will be a challenge.

Could you please elaborate more on this?

lingxiankong:
> Let me know if i should proceed with this change.

Let's have this fully discussed before drawing a conclusion. Thanks for your patience BTW.

hamzazafar:
> > maintaining the ownership of resources will be a challenge.
> 
> Could you please elaborate more on this?

Let's consider an example in which we have 2 k8s clusters running, we want want to expose services **(with same port numbers)** using a shared loadbalancer.
- Cluster one - service one for port 80 is created (listener with port 80 and a pool will be created by cloud provider)
- Cluster two - service two (replica of service one) for port 80 is created for HA purposes. Cloud provider is expected to simply add members to the pool attached to listener (created earlier by service one on cluster one) for port 80 
- Cluster one - delete service one, members from service two on cluster two exist in a pool created by service one. This is the main conflict here. When service one gets deleted, we should only remove members of service one and leave all of the other resources (listener for port 80, pool and healthmonitor) behind because we still want to serve traffic using loadbalancer for service two. Another problem occurs if we finally delete service two, now service two will be responsible for cleaning up the resources (listener, pool and healthmonitor) created by service one. Ideally, the external resource created by a cluster should be managed by itself. Therefore we have added a requirement for creating listener and pool manually, cloud provider will be responsible for adding and removing members to a pool. This makes things less complicated and it's a cleaner approach in my opinion. 

We have two cases to handle:
- **Add services for same port to a shared loadbalancer**: For this we can specify the annotation `loadbalancer.openstack.org/unmanaged-pool-id` in service spec, pool should be created beforehand by the user. This PR implements this feature.

- **Add services with different ports to a shared loadbalancer**: For this was can specify the annotation `loadbalancer.openstack.org/shared-loadbalancer` in service spec. Cloud provider will be responsible for creating listeners, pools, healthmonitors and members. We have to implement this feature, maybe in this PR or in a new one.

lingxiankong:
Thanks for the detailed explanation.

> Cluster two - service two (replica of service one) for port 80 is created for HA purposes.

I don't fully understand the "HA purpose" here, the Kubernetes Service itself is designed for HA implementation of Pods, why we need to re-implement the HA across multiple clusters in the same region?

strigazi:
> > Cluster two - service two (replica of service one) for port 80 is created for HA purposes.
> 
> I don't fully understand the "HA purpose" here, the Kubernetes Service itself is designed for HA implementation of Pods, why we need to re-implement the HA across multiple clusters in the same region?

There are two points I want to mention that this PR touches.

1. Regarding HA
This feature will allow users to do more advanced architectures on how they send traffic in their clusters. In an openstack region users can spread to availability zones for HA, as yout mention , which covers components that run on the Hypervisors (VMs, network components like ovs, linux-bridge etc). But, the kubernetes control plane (even if it spread across many nodes) is a failure domain too, an uprgade that went wrong, a cluster is compromised, a config that hit the entire cluster by a human-error etc. Octavia supports different backends (apart from the amphora reference implementation) [0] which can run in the same region or not, but let's leave outside this discussion the multi-region case.

2.1 Migration from "traditional" VM based infrastructure
Another big selling point is that feature will allow users to migrate, with zero downtime, their applications from (ansible, puppet managed etc) VMs to kubernetes. Users that run VMs behind an LB can add one or many k8s clusters in the same pool and eventually delete or just reduce the VM capacity. 

2.2 Migration from one cluster to another
Users will be able to refresh their k8s infra not only by upgrading clusters but moving to new ones. This would be the pattern "kubernetes clusters as cattle". A few big clusters that can be renewed, many small clusters or just a big one that some times is migrated to the next big cluster.

Being able to refresh kubernetes clusters is very useful for deployment that use cluster-api, magnum, kubespray or in general have a fast automated way to provision new clusters. In cluster-api terminology, we could have more than workload clusters [1] service the same application, of course the same can be done with magnum, kubespray, rancher etc.

My apologies for the long comment, I tried tell the full story.

[0] https://docs.openstack.org/octavia/latest/admin/providers/index.html
[1] https://cluster-api.sigs.k8s.io/user/concepts.html?search=#workload-cluster

lingxiankong:
I think either 2.1 or 2.2 could be solved using DNS based traffic management. However, I agree we could implement that in OCCM as another option.

Instead of adding two extra annotations for load balancer sharing, I'd rather we support only one. Below is the draft design:

1. When creating Service-1 with annotation `loadbalancer.openstack.org/shared-loadbalancer: ""`, indicating there is no load balancer created yet, so OCCM will create a new one (e.g. LB-1) as before. OCCM also adds the Service-1 identifier into the tag of related Octavia resources: loadbalancer, listener.

2. When creating Service-2 with `loadbalancer.openstack.org/shared-loadbalancer: "LB-1"`,

    2.1 If Service-2 defines different ports with Service-1, OCCM will create different listeners in the load balancer with Service-2 identifier in the tag.

    2.2 If Service-2 defines the same ports with Service-1, OCCM will just add members to the existing pool, but add Service-2 identifier in the loadbalancer and listener tag.

    No matter either cases, OCCM also adds the Service-2 identifier into the load balancer tag.

3. When deleting a Service-N, OCCM checks the resource tag to decide if the loadbalancer or listener should be deleted, if not, remove Service-N identifier from the tag.

With the above design, suppose we are creating 3 Services in two clusters:

* Service-1 with `loadbalancer.openstack.org/shared-loadbalancer: ""` in cluster-1, port 80
* Service-2 with `loadbalancer.openstack.org/shared-loadbalancer: "LB-1"` in cluster-1, port 8080
* Service-3 with `loadbalancer.openstack.org/shared-loadbalancer: "LB-1"` in cluster-2, port 80. The functions of Service-3 and Service-1 are identical.

In the end, the Octavia resources created are:

* One loadbalancer with tag: `['Service-1', 'Service-2', 'Service-3']`
* Two listeners, one with tag: `['Service-1', 'Service-3']`, the other one with tag: `['Service-2']`
* Two pools, one with members of cluster-1 nodes (for Service-2), the other one with members of both cluster-1 and cluster-2 nodes (for Service-1 and Service-3).

The Service deletion processes:

* After deleting Service-1, Octavia resources are changed to:

  * One loadbalancer with tag: `['Service-2', 'Service-3']`
  * Two listeners, one with tag: `['Service-2']`, the other one with tag: `['Service-3']`
  * Two pools, one with members of cluster-1 nodes (for Service-2), the other one with members of cluster-2 nodes (for Service-3).

* After deleting Service-2, Octavia resources are changed to:

  * One loadbalancer with tag: `['Service-3']`
  * One listener with tag: `['Service-3']`
  * One pool with members of cluster-2 nodes (for Service-3)

* After deleting Service-3, all the above Octavia resources should be deleted.


Notes:

* Still need a mapping function from Service to Octavia loadbalancer.
* The new feature shouldn't affect the current behavior (i.e. creating Service without `loadbalancer.openstack.org/shared-loadbalancer` annotation).
* The number of Services sharing the same load balancer should be limited.
* Octavia service should support resource tag.
* The tag modification should be implemented with lock.

lingxiankong:
Let's continue the discussion here, but before we starting implementation, I need to finish the `EnsureLoadBalancer` function refactoring first. The current function code is tedious and hard to maintain, I will split the logic for octavia and neutron-lbaas support.

rochaporto:
Hi @lingxiankong .

We want to reuse an existing pool and have each kubernetes cluster managing only the members it owns - giving us automation of pool member management (which is the most cumbersome part to do manually) while giving us flexibility to move services around VMs, kubernetes, etc.

@hamzazafar stated a couple use cases for this in his initial description.

The LB and listeners already exist and are created either manually or relying on other automation tools - this is fine and we do it already. We do not want kubernetes to manage those in these cases.

This is what we're doing today and having an annotation "unmanaged-pool-id" as mentioned by @hamzazafar matches this exact use case, with no additional logic or complexity on the occm code or tags required on the octavia side.

Your suggestion to have each service instance creating/deleting its own pool was also considered - we chose to also support a single pool to better control the LB algorithm if/when needed. We could reevaluate the impact of split pools in the algorithm choice for our users, i can check my notes from when reviewing this a couple months ago.

One key thing is that we want to make sure when operating in this mode the OCCM will never touch existing LB instances, listeners and other pool members. Having a dedicated and clear annotation in the kubernetes resource and a very simple/obvious logic in the code sounded like a good way to guarantee this.

rochaporto:
> I think either 2.1 or 2.2 could be solved using DNS based traffic management. However, I agree we could implement that in OCCM as another option.

The drawbacks of our dns based solution are the main reason to have moved to a solution like this.

JohnGarbutt:
@rochaporto agreed with you on this approach vs DNS. Disposable K8s clusters for production workloads seemed very popular during the recent OpenDev discussions. Having this feature would make that much more viable than existing slow DNS draining of old clusters.

We are really interested to help moving this forward, @brtknr was thinking about trying this out internally on 1.18. I guess it needs rebasing now to merge here.

PR #1120, opened 2020-07-15
[manila-csi-plugin] Fix volume in nodeplugin

strigazi:
Make hostpath volumes DirectoOrCreate instead of Directory
for a more resilient installation.

```release-note
manila-csi-plugin chart v0.1.1 Change to DirectoryOrCreate for nodeplugin volume.
```

### DIFF
--- a/charts/manila-csi-plugin/templates/nodeplugin-daemonset.yaml
+++ b/charts/manila-csi-plugin/templates/nodeplugin-daemonset.yaml
@@ -110,7 +110,7 @@ spec:
         - name: {{ .protocolSelector | lower }}-fwd-plugin-dir
           hostPath:
             path: {{ .fwdNodePluginEndpoint.dir }}
-            type: Directory
+            type: DirectoryOrCreate
         {{- end }}
     {{- if .Values.nodeplugin.affinity -}}
       affinity:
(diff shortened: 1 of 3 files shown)


manifests/manila-csi-plugin/csi-nodeplugin.yaml
```diff
@@ -88,13 +88,13 @@ spec:
         - name: registration-dir
           hostPath:
             path: /var/lib/kubelet/plugins_registry
-            type: Directory
+            type: DirectoryOrCreate
```
ramineni, line 91: this should be Directory only

charts/manila-csi-plugin/templates/nodeplugin-daemonset.yaml
```diff
@@ -101,7 +101,7 @@ spec:
         - name: registration-dir
           hostPath:
             path: /var/lib/kubelet/plugins_registry
-            type: Directory
+            type: DirectoryOrCreate
```
ramineni, line 104: should be Directory only

strigazi:
> Are there any differences in permissions or ownership if manila-csi creates these paths rather than the node-plugin doing so?

Everything seems to be owned by root.

strigazi:
@ramineni @tombarron is this patch ok?

ramineni:
Thanks
/lgtm

strigazi:
> Thanks
> /lgtm

Thank you!

Who needs to approve?

ramineni:
@strigazi Lets get +1 from manila-csi core team , then it can be approved :)

ramineni:
cc @gman0 @tombarron  @gouthampacha

tombarron:
OK, I tested this locally with the NFS nodeplugin making sure to deploy manila-csi plugin before I brought up the nfs node plugin (normally I do it in the opposite order) and it works OK.

/lgtm
/approve

PR #1125, opened 2020-07-18
Add `auth-type` as input param

### LINKED ISSUE #1124: [all] no AuthType provided in openstack connection
/kind feature

**What happened**:
need honor `auth-type` so that we can change method to auth from cloud provider to openstack

**How to reproduce it**:

**Environment**:
- OpenStack version:
- Others:

jichenjc:

fixes #1124 

**Release note**:

```release-note
auth-type can be input as param to auth from provider to openstack 
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack.go
+++ b/pkg/cloudprovider/providers/openstack/openstack.go
@@ -187,21 +187,22 @@ type OpenStack struct {
 }
 
 type AuthOpts struct {
-	AuthURL          string `gcfg:"auth-url" mapstructure:"auth-url" name:"os-authURL" dependsOn:"os-password|os-trustID"`
-	UserID           string `gcfg:"user-id" mapstructure:"user-id" name:"os-userID" value:"optional" dependsOn:"os-password"`
-	Username         string `name:"os-userName" value:"optional" dependsOn:"os-password"`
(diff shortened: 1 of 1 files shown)


lingxiankong:
Hi, what if we don't specify `auth-type` but provide configs like `application-credential-id` and `application-credential-secret`, will the authentication will fail?

kayrus:
> Hi, what if we don't specify auth-type but provide configs like application-credential-id and application-credential-secret, will the authentication will fail?

@lingxiankong no, it won't

lingxiankong:
> > Hi, what if we don't specify auth-type but provide configs like application-credential-id and application-credential-secret, will the authentication will fail?
> 
> @lingxiankong with the current master branch, no, it won't. with this change - probably yes.

Thanks @kayrus, so is it necessary to add the extra `auth-type` config option?

jichenjc:
@kayrus @lingxiankong from my test, if didn't change auth-type, the auth will fail
maybe I used wrong configuration before... so I will redo my test and paste result here

kayrus:
@lingxiankong I never had issues with auth. app creds works, password auth works.

jichenjc:
I think I misunderstood the logic

gophercloud handles the logic for us https://github.com/gophercloud/gophercloud/blob/master/auth_options.go#L224
so we should close this PR

PR #1130, opened 2020-07-23
[occm] make floating subnet optional

hamzazafar:
In [PR ](https://github.com/kubernetes/cloud-provider-openstack/pull/1102) floating subnet was made mandatory, for backward compatibility we should make it optional again. I have also added docs to clearly specify that floating-subnet is an optional argument. 

**Release note**:

```release-note
NONE
```


lingxiankong:

PR #1144, opened 2020-08-01
[occm] Split Octavia related functions

lingxiankong:
Refactor loadbalancer related functions, split the code when Octavia is enabled.

fixes #

Now, `ensureOctaviaLoadBalancer()` and `updateOctaviaLoadBalancer()` should be the two functions we need to maintain. Legacy code should be removed at some point in the future.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/util/openstack/loadbalancer.go
+++ b/pkg/util/openstack/loadbalancer.go
@@ -29,7 +29,7 @@ import (
 	"github.com/gophercloud/gophercloud/pagination"
 	version "github.com/hashicorp/go-version"
 	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/klog/v2"
+	klog "k8s.io/klog/v2"
 
 	cpoerrors "k8s.io/cloud-provider-openstack/pkg/util/errors"
 )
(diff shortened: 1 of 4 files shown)


lingxiankong:
@ramineni @zetaab @jichenjc please help to review this PR, thanks!

pkg/util/openstack/loadbalancer.go
```diff
@@ -257,6 +257,42 @@ func GetPoolByName(client *gophercloud.ServiceClient, name string, lbID string)
...
+	if err != nil {
+		if cpoerrors.IsNotFound(err) {
+			return nil, ErrNotFound
+		}
+		return nil, err
+	}
+
+	if len(listenerPools) == 0 {
+		return nil, ErrNotFound
+	} else if len(listenerPools) > 1 {
```
ramineni, line 289: Is this check required here again? with L275 , if its multiple results, already might have handled by L280 ?

lingxiankong: This is copy from the old version, but I can fix this.

ramineni:
Thanks.
/lgtm

lingxiankong:
/approve

Merge this PR so we could continue with the improvement.

PR #1155, opened 2020-08-11
[occm] fix bug for network-id annotation

hamzazafar:
This bug was introduced in PR: https://github.com/kubernetes/cloud-provider-openstack/pull/1144

How to reproduce?
Set `loadbalancer.openstack.org/network-id` annotation in service template. Cloud provider will not consume the value mentioned in the annotation!

**Release note**:
```release-note
NONE
```


lingxiankong:
/ok-to-test

Thanks for the fix!

Issue #1136, opened 2020-07-29
[Manila-CSI-Plugin] Unable to mount NFS volume after dynamic provision


JPRbrs:
> /kind Feature

**What happened**:
When trying to use the examples in the dynamic-provisioning folder, the NFS is correctly created in OpenStack but it is not mounted in the pod. The cause is that Openstack provides several export locations (IP + path) but Manila CSI is selecting one with an IP which does not belong to the Kubernetes Cluster Network Interface.

**What you expected to happen**:
NFS volume is mounted in the node and then in the pod

**How to reproduce it**:
Deploy CSI NFS driver (https://github.com/kubernetes-csi/csi-driver-nfs) and Manila NFS plugin. Deploy manifests in examples/nfs/dynamic-provisioning

**Anything else we need to know?**:
Is there anyway we can tell Manila CSI which subnet to use when choosing the export location? I've tried with nfs-shareClient and shareNetworkID but neither of them worked

tombarron:
Thank you for reporting this issue, which seems to me genuine.  Do you need selection from export locations in different address families (IPv6, IPv4) or is the address-family the same?  We need to figure out what information is needed to choose, who has that information, and whether it would be the manila-csi plugin itself or its helper node-plugin (in this case nfs-csi) that would be doing the selection.

JPRbrs:
Thanks for the answer @tombarron, I only need to use IPv4 IPs. After looking into the code base, I reckon passing the subnet in CIDR notation as an optional parameter to ChooseExportLocation function in pkg/csi/manila/util/manila.go  would be ideal. Perhaps adding an extra parameter to the Controller Service volume parameters (storage class for dynamically provisioned volumes) would be the perfect place to add it. What do you think?

JPRbrs:
Hello again @tombarron Would it make sense to add a new parameter to NodeVolumeContext struct in pkg/csi/manila/options/shareoptions.go with the CIDR formatted IP for the subnet so it can be used by buildVolumeContext function in pkg/csi/manila/nodeserver.go and in turn it can be passed to ChooseExportLocation through GetChosenExportLocation in pkg/csi/manila/util/manila.go?

The only problem I see is that I cannot find any reference to the IP where the NFS share is exposed...

gman0:
[...]

PR #1157, opened 2020-08-13
[manila-csi-plugin] Replaced ChooseExportLocation with FindExportLocation

gman0:

This PR replaces `ChooseExportLocation` with `FindExportLocation` and adds a unit test for this function.

`FindExportLocation` retains the behavior of its predecessor, but adds the ability to further constrain the result by using a filter function that's supplied to `FindExportLocation` as an argument.

We'll need this for the NFS share adapter (at least), where we'll need to filter out unsuitable export locations (e.g. those that are not from a certain subnet).

xref https://github.com/kubernetes/cloud-provider-openstack/issues/1136

**Release note**:

```release-note
NONE
```


gman0:
PTAL @tsmetana @tombarron @gouthampacha @ramineni

gman0:
FYI @JPRbrs there will be one or (at most) two more PRs till we get a fix for https://github.com/kubernetes/cloud-provider-openstack/issues/1136 in

JPRbrs:
Thanks Robert!

pkg/csi/manila/util/exportlocation.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+// Predicate type for filtering out export locations.
+// It's supplied with the index into the export location slice that's being scanned.
+// Should return (true, nil) when search conditions are satisfied,
+// and (false, nil) when they are not.
+type ExportLocationPredicate func(index int) (match bool, err error)
+
+// Predicate matching any export location
+func AnyExportLocation(int) (bool, error) { return true, nil }
```
ramineni, line 30: Are any other predicates supported now?

pkg/csi/manila/util/exportlocation.go
```diff
@@ -0,0 +1,68 @@
...
+// Searches for an export location.
+// Returns index of an export location from the `locs` slice that satisfies following rules:
+// 1. Location is not admin-only and is not empty
+// 2. Location satisfies ExportLocationPredicate
+// Search is biased:
+// 1. Location.Preferred == true is preferred over Location.Preferred == false
+// 2. Locations with lower index are preferred over those with higher index
+func FindExportLocation(locs []shares.ExportLocation, pred ExportLocationPredicate) (index int, err error) {
+	const invalidIdx = -1
+	firstMatchNotPreferred := invalidIdx
```
ramineni, line 41: confused, why is the name firstMatchNotPreferred , I guess this stores, first preferred match on contrary

pkg/csi/manila/util/exportlocation.go
gman0: Not yet, we're planning two more for now: 1) exact match, and 2) subnet match.

pkg/csi/manila/util/exportlocation.go
gman0: A preferred location is `locs[i].Preferred==true`, which makes the function return early:

https://github.com/gman0/cloud-provider-openstack/blob/26613136736920cd667fa28fdc56d8b103cd19fc/pkg/csi/manila/util/exportlocation.go#L51-L55

When not preferred (`Preferred==false`), we keep going through the list, searching for a preferred location. If none is found, we return the first matching "not preferred" location.

gman0:
@ramineni this is in preparation for an upcoming PR that will fix the issue

gman0:
@ramineni I see [gouthampacha's "approve"](https://github.com/kubernetes/cloud-provider-openstack/pull/1157#issuecomment-673720413) didn't trigger merge, even though he is an approver for `pkg/csi/manila`, and this PR already had a LGTM label at that time. Any idea why this might be?

pkg/csi/manila/util/exportlocation.go
ramineni: Ah, got it, you are returning if preferred.

ramineni:
@gman0 This PR is merged.

PR #1158, opened 2020-08-16
[octavia-ingress-controller] Use REGEX compare type for HOST_NAME l7 rule type

lingxiankong:
When a client sends HTTP/HTTPS request to an Ingress, sometimes the client could add port number to Host HTTP header (see [1]), e.g. when the host in the Ingress is "foo.bar.com", the Host header could be "foo.bar.com:80". This PR adds support for both "foo.bar.com" and "foo.bar.com:80".

[1]: https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/using-octavia-ingress-controller.md
+++ b/docs/using-octavia-ingress-controller.md
@@ -210,28 +210,52 @@ Wait until the StatefulSet is up and running.
 ## Setting up HTTP Load Balancing with Ingress
 
 ### Create a backend service
-Create a simple service(echo hostname) that is listening on a HTTP server on port 8080.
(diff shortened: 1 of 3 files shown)


ramineni:
@lingxiankong Thanks. Looks good.  
Doesn't it suppose to trigger the lb-octavia job for this change?

lingxiankong:
> Doesn't it suppose to trigger the lb-octavia job for this change?

No, this is for octavia-ingress-controller, lb-octavia job is for occm

ramineni:
Ok, got it

/lgtm

PR #1160, opened 2020-08-17
[occm] add octavia loadbalancer flavor option

phiphi282:
This pr adds an option to specify the flavor-id used for the octavia loadbalancer.
The flavor-id can be set in the config or via annotation.

fixes #

Flavors were being introduced in octavia version 4.0.0 if I am not mistaken.
I haven't found a way to check the octavia version from within the occm so if a check is needed here I would appreciate if someone can point me in the right direction here.

**Release note**:

```release-note
[openstack-cloud-controller-manager] Add support to set octavia loadbalancer flavor-id.
```

### DIFF
--- a/docs/expose-applications-using-loadbalancer-type-service.md
+++ b/docs/expose-applications-using-loadbalancer-type-service.md
@@ -153,6 +153,10 @@ Request Body:
 - `loadbalancer.openstack.org/enable-health-monitor`
 
   Defines whether or not to create health monitor for the load balancer pool, if not specified, use `create-monitor` config. The health monitor can be created or deleted dynamically.
+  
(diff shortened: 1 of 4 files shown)


phiphi282:
Tests are working now again :)

lingxiankong:
Flavor was introduced in Octavia since version v2.6 (on 2019-01-25), for those deployments using Octavia prior to v2.6, load balancer creation will fail as the parameter `flavor_id` is not recognized.

@phiphi282 Please submit another PR to add the version check (example https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/util/openstack/loadbalancer.go#L88), otherwise, we have to revert this PR.

PR #1165, opened 2020-08-19
[occm] Add version check for lb flavors support

phiphi282:
I previously created #1160 
There a version check was missing to find out if the used octavia version supports flavors already.
This would cause the controller to throw an error if a user specified a flavor-id on a loadbalancer with an old octavia version (< v2.6).

This pr creates a version check so that the flavor-id option gets ignored if the octavia version is too old

fixes #

Thank you @lingxiankong for pointing out how to create a version check.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
@@ -462,7 +462,10 @@ func (lbaas *LbaasV2) createOctaviaLoadBalancer(service *corev1.Service, name, c
 		Name:        name,
 		Description: fmt.Sprintf("Kubernetes external service %s/%s from cluster %s", service.Namespace, service.Name, clusterName),
 		Provider:    lbaas.opts.LBProvider,
-		FlavorID:    svcConf.flavorID,
+	}
+
+	if openstackutil.IsOctaviaFeatureSupported(lbaas.lb, openstackutil.OctaviaFeatureFlavors) {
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
```diff
@@ -462,7 +462,10 @@ func (lbaas *LbaasV2) createOctaviaLoadBalancer(service *corev1.Service, name, c
 		Name:        name,
 		Description: fmt.Sprintf("Kubernetes external service %s/%s from cluster %s", service.Namespace, service.Name, clusterName),
 		Provider:    lbaas.opts.LBProvider,
-		FlavorID:    svcConf.flavorID,
+	}
+
+	if openstackutil.IsOctaviaFeatureSupported(lbaas.lb, openstackutil.OctaviaFeatureFlavors) {
+		createOpts.FlavorID = svcConf.flavorID
 	}
```
lingxiankong, line 469: Could you please move this to `checkService()` function? See `svcConf.allowedCIDR` for an example. The only difference is, by default `FlavorID` should be nil.

phiphi282: Moved the check. I am not sure if the `svcConf.flavorID != ""` check I changed this to is really needed since the default is an empty string anyways, but I find it more readable this way

lingxiankong:
/lgtm
/approve

Thanks!

PR #1170, opened 2020-08-25
[cinder-csi-plugin]update chart cacert location to /etc/caert

jichenjc:
per doc for chart and our previous manifest, cacert is located
at /etc/cacert

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/cinder-csi-plugin/templates/controllerplugin-statefulset.yaml
+++ b/charts/cinder-csi-plugin/templates/controllerplugin-statefulset.yaml
@@ -101,12 +101,12 @@ spec:
             - name: socket-dir
               mountPath: /csi
             - name: cacert
-              mountPath: /etc/kubernetes
+              mountPath: /etc/cacert
               readOnly: true
       volumes:
         - name: cacert
           hostPath:
-            path: /etc/kubernetes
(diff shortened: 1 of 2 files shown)


charts/cinder-csi-plugin/templates/controllerplugin-statefulset.yaml
```diff
@@ -101,12 +101,12 @@ spec:
             - name: socket-dir
               mountPath: /csi
             - name: cacert
-              mountPath: /etc/kubernetes
+              mountPath: /etc/cacert
```
ramineni, line 104: Actually it is updated in docs , that it expects under /etc/kubernetes , https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/using-cinder-csi-plugin.md#using-the-helm-chart

jichenjc: nope, doc said it's /etc/cacert
the cloud conf is /etc/kubernetes

ramineni: Ah!!

PR #1077, opened 2020-05-25
[cinder-csi-plugin] Update debian-base image

ialidzhikov:
**What this PR does / why we need it**:
The PR updates the base image with us.gcr.io/k8s-artifacts-prod/build-image/debian-base-amd64:v2.1.0 as minimal base - ref https://github.com/kubernetes/kubernetes/tree/master/build/debian-base

**Which issue this PR fixes(if applicable)**:
fixes #

**Special notes for reviewers**:

**Release note**:

```release-note
NONE
```

PR #1180, opened 2020-09-01
[cinder-csi-plugin]  Update debian base image to v2.1.3

ialidzhikov:
Update debian base image to v2.1.3 to address CVEs.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster/images/cinder-csi-plugin/Dockerfile
+++ b/cluster/images/cinder-csi-plugin/Dockerfile
@@ -13,7 +13,7 @@
 ARG DEBIAN_ARCH=amd64
 # We not using scratch because we need to keep the basic image information
 # from parent image
-FROM us.gcr.io/k8s-artifacts-prod/build-image/debian-base-${DEBIAN_ARCH}:v2.1.0
+FROM k8s.gcr.io/build-image/debian-base-${DEBIAN_ARCH}:v2.1.3
 
 ARG ARCH=amd64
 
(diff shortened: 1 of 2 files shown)


ramineni:
@ialidzhikov looks like its unable to find the version 2.1.3 , https://logs.openlabtesting.org/logs/80/1180/68a6e132902bde8c056a5ed1f4bc96b2f9c3d26e/cloud-provider-openstack-acceptance-test-csi-cinder/cloud-provider-openstack-acceptance-test-csi-cinder/d7c92f6/job-output.json.gz

ialidzhikov:
@ramineni , @jichenjc , sorry I had a typo in the base image. It should be good with the latest force push - ref  https://github.com/kubernetes/cloud-provider-openstack/compare/68a6e132902bde8c056a5ed1f4bc96b2f9c3d26e..95108bc09bcfd463b3ee3298543f775ccc6656c6.

> why not only update with last version directly instead of change repo?

@jichenjc , `k8s.io.gcr` is a vanity domain that will resolve to `{eu,us,...}.gcr.io/k8s-artifacts-prod/`. The last time we did similar base image update (ref https://github.com/kubernetes/cloud-provider-openstack/pull/1077), I believe the [vanity domain flip](https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/Vanity-Domain-Flip.md) was in process and it was possible to use only the `{eu,us,...}.gcr.io/k8s-artifacts-prod/` images.

jichenjc:
@ialidzhikov thanks for the info , the upgrade is reasonable to me ~
/lgtm

PR #1188, opened 2020-09-07
[cinder-csi-plugin]add e2e test when cloudprovider changes

jichenjc:

https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/openstack/openstack.go#L30
uses cloud provider, so better to run cinder e2e when anything changed in that folder
fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -80,6 +80,7 @@
               - cmd/cinder-csi-plugin/.*
               - manifests/cinder-csi-plugin/.* 
               - pkg/csi/cinder/.*
+              - pkg/cloudprovider/.*
               - pkg/util/.*
               - tests/e2e/csi/cinder/.*
               - cmd/tests/.*


.zuul.yaml
```diff
@@ -80,6 +80,7 @@
               - cmd/cinder-csi-plugin/.*
               - manifests/cinder-csi-plugin/.* 
               - pkg/csi/cinder/.*
+              - pkg/cloudprovider/.*
```
lingxiankong, line 83: There are several basic structs that are inherited from those in pkg/cloudprovider, considering the changes we've made are mostly happening in service loadbalancer, I'd prefer we remove this dependency in CI for now.

If we really have concerns, the recommended approach would be removing the dependencies to pkg/cloudprovider.

jichenjc: ok, so I think it takes time to split the code ... let me see what I can do

PR #1191, opened 2020-09-07
Remove index.yaml and add README.md

brtknr:
GH action fails when there is an empty index.yaml file.  It is better to
have none at all so that it can create a valid index.yaml file for us.

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,14 @@
+# Cloud Provider OpenStack Helm Chart Repository
+
+## Add repo
+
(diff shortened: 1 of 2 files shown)


brtknr:
@lingxiankong @jichenjc

jichenjc:
/ok-to-test
/lgtm

thanks for the update

brtknr:
Hmm there is no OWNERS file in this branch which is probably why the CI is ignoring your command.

PR #1198, opened 2020-09-11
[occm] Fix bug where pointer from local loop variable instead of array variable is used

phiphi282:
range copies the value as a local variable in the loop.
Getting the pointer from it makes all the elements in the map point to
the last value that the range variable copied.
This means that only the last port in the array will ever be updated.

Change to get value with an index from the array directly.

fixes #

Testing:
Create a service/ingress with multiple ports and check if failures regarding creating listeners will show up.

**Release note**:

```release-note
NONE
```


lingxiankong:
Nice catch! 
/lgtm

PR #1190, opened 2020-09-07
Bump up chart version to trigger chart release

brtknr:
**What this PR does / why we need it**:

**Which issue this PR fixes(if applicable)**:
fixes #1193

**Special notes for reviewers**:

**Release note**:

```release-note
NONE
```

PR #1199, opened 2020-09-11
Bump up chart version to trigger chart release (release-1.19)

brtknr:
- Also remove dysfunctional Github chart install action.
- Only release charts from release branches.

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,7 +3,7 @@ name: Release Charts
 on:
   push:
     branches:
-      - master
+      - "release-*"
 
 jobs:
   release:
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
 appVersion: latest
-description: Cinder CSI Plugin for OpenStack
+description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 1.1.2
+version: 1.2.0
 home: https://github.com/kubernetes/cloud-provider-openstack
 icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
--- a/charts/cinder-csi-plugin/values.yaml
+++ b/charts/cinder-csi-plugin/values.yaml
@@ -30,7 +30,7 @@ csi:
   plugin:
     image:
       repository: docker.io/k8scloudprovider/cinder-csi-plugin
-      tag: latest
+      tag: v1.19.0
       pullPolicy: IfNotPresent
 
 storageClass:
--- a/charts/manila-csi-plugin/Chart.yaml
+++ b/charts/manila-csi-plugin/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
 appVersion: latest
-description: Manila CSI Plugin for OpenStack
+description: Manila CSI Chart for OpenStack
 name: openstack-manila-csi
-version: 0.1.2
+version: 0.2.0
 home: http://github.com/kubernetes/cloud-provider-openstack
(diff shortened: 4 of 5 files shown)


charts/cinder-csi-plugin/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 appVersion: latest
```
ramineni, line 2: can u change the appVersion to v1.19 ? latest can be kept for only master branch

brtknr:
There is another PR to address bumping up appVersion

brtknr:
I can bump it here if you like but we need to merge #1190 first.

ramineni:
> I can bump it here if you like but we need to merge #1190 first which blocks creation of chart releases from master branch.

@brtknr Ok, One clarification, once we bump the release on master to 1.2.0 with #1190, we need to update the same version in release-1.19 branch as well right, does that trigger chart release ?

ramineni:
> There is another PR to address bumping up appVersion

not the chart version , you need to change the appVersion to v1.19 from latest

.github/workflows/release.yaml
```diff
@@ -3,7 +3,7 @@ name: Release Charts
 on:
   push:
     branches:
-      - master
+      - "release-*"
```
lingxiankong, line 6: Should it be `release-1.19`?

charts/cinder-csi-plugin/Chart.yaml
```diff
@@ -1,8 +1,8 @@
...
-description: Cinder CSI Plugin for OpenStack
+description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 1.1.2
+version: 1.2.0
```
lingxiankong, line 5: Maybe I missed the discussion, but may I know where this version comes from? I see it's still 1.1.2 on master branch?

.github/workflows/release.yaml
brtknr: The reason for making this a pattern is so that we don’t need to update on every release.

lingxiankong: ack

lingxiankong:
@brtknr could you answer my second question above?

charts/cinder-csi-plugin/Chart.yaml
brtknr: It should be 1.2.0 on master now with #1190

brtknr:
> @brtknr could you answer my second question above?

Hi @lingxiankong sorry I was on vacation last few days. I believe #1190 bumped up the chart version to 1.2.0 and 0.2.0 respectively. Since the chart on release-1.19 is the same as master branch (apart from the appVersion), it makes sense to have the same chart version as the master branch here (since charts are NOT released from master branch).

lingxiankong:
> > @brtknr could you answer my second question above?
> 
> Hi @lingxiankong sorry I was on vacation last few days. I believe #1190 bumped up the chart version to 1.2.0 and 0.2.0 respectively. Since the chart on release-1.19 is the same as master branch (apart from the appVersion), it makes sense to have the same chart version as the master branch here (since charts are NOT released from master branch).

Got it, thanks! Let's merge this PR once CI passes.

lingxiankong:
Can you check the commits in this PR, it includes a commit already merged on master.

brtknr:
Hold on, there is an extra commit here thats not supposed to be here, lemme fix it.

brtknr:
Ok its fixed.

jichenjc:
I assume the chart failure is acceptable 

/approve

brtknr:
This ultimately failed to successfully publish a release because gh-pages is configured as a protected branch. Please see the error https://github.com/kubernetes/cloud-provider-openstack/runs/1115341006?check_suite_focus=true

brtknr:
@lingxiankong can you please check the branch protection status of gh-pages branch?

PR #1203, opened 2020-09-14
[cinder-csi-plugin] make capacity check consistent

### LINKED ISSUE #1201: [cinder-csi-plugin] avoid default size in ephemeral disk

>
> /kind bug
> /kind feature

**What happened**:

https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/nodeserver.go#L136

will default to 1 if no capacity is given, this is inconsistent to 
https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/nodeserver.go#L63 and doesn't make sense

we should report error instead of use default value if capacity not given

**How to reproduce it**:

**Environment**:
- OpenStack version:
- Others:

jichenjc:

fixes #1201 

**Release note**:

```release-note
NONE
```


ramineni:
/hold 
please check my comment on the issue , #1201

PR #1211, opened 2020-09-15
[cinder-csi-plugin] Update snapshotter to v2.1.1 in charts

ramineni:

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: latest
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 1.2.0
+version: 1.2.1
 home: https://github.com/kubernetes/cloud-provider-openstack
 icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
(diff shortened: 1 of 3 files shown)


brtknr:
lgtm

jichenjc:
do we need additional RBAC changes ?

brtknr:
This has already merged on release-1.19 branch.

ramineni:
> do we need additional RBAC changes ?

yes, I missed it . will update.

ramineni:
> do we need additional RBAC changes ?

Done.

brtknr:
ping, lets merge this, its blocking a few other things.

jichenjc:
/approve 

speed up merge actions as this is helm chart only and relatively small impact

brtknr:
The problem is that now the rbac changes need to be backported to release-1.19 too.

ramineni:
> The problem is that now the rbac changes need to be backported to release-1.19 too.  It would have been better if the rbac change was a separate PR.

hmm, I was thinking about that , do you think we can merge rbac changes with one of your backport fixes , to be consistent with version again, but not sure this is the right approach though? Im ok to split the PRs if that ensures consistency with release branch , could you suggest the best approach?

brtknr:
I have pushed the missing backport to https://github.com/kubernetes/cloud-provider-openstack/pull/1233. If you agree, we can merge that and proceed this way as a one off.

brtknr:
Nothing is going to get released until https://github.com/kubernetes/cloud-provider-openstack/pull/1216 is merged anyway due to the bug I mentioned.

PR #1213, opened 2020-09-15
[occm] lookup flavor name if only id is provided

seanschneeweiss:
This calls the API for a flavor name, if only the id is used to lookup the flavor. This provides a more meaningful return value.

fixes #

**Release note**:

```release-note
[openstack-cloud-controller-manager] Always use flavor name for the Node label `node.kubernetes.io/instance-type`.
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack_instances.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_instances.go
@@ -281,8 +281,9 @@ func srvInstanceType(client *gophercloud.ServiceClient, srv *servers.Server) (st
 			flavor, ok := val.(string)
 			if ok {
 				if key == "id" {
+					mc := metrics.NewMetricContext("flavor", "get")
 					f, err := flavors.Get(client, flavor).Extract()
-					if err == nil {
+					if mc.ObserveRequest(err) == nil {
 						return f.Name, nil
 					}
 				}
--- a/pkg/cloudprovider/providers/openstack/openstack_instances.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_instances.go
@@ -274,7 +274,7 @@ func (i *Instances) InstanceType(ctx context.Context, name types.NodeName) (stri
 }
 
 func srvInstanceType(client *gophercloud.ServiceClient, srv *servers.Server) (string, error) {
-	keys := []string{"name", "id", "original_name"}
+	keys := []string{"original_name", "id"}
 	for _, key := range keys {
 		val, found := srv.Flavor[key]
 		if found {
@@ -282,10 +282,9 @@ func srvInstanceType(client *gophercloud.ServiceClient, srv *servers.Server) (st
 			if ok {
 				if key == "id" {
 					f, err := flavors.Get(client, flavor).Extract()
-					if err != nil {
-						return "", err
+					if err == nil {
+						return f.Name, nil
 					}
-					return f.Name, nil
 				}
 				return flavor, nil
 			}
--- a/pkg/cloudprovider/providers/openstack/openstack_instances.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_instances.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 
 	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	"k8s.io/klog/v2"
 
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/providers/openstack/openstack_instances.go
```diff
@@ -266,16 +267,23 @@ func (i *Instances) InstanceType(ctx context.Context, name types.NodeName) (stri
...
 		val, found := srv.Flavor[key]
 		if found {
 			flavor, ok := val.(string)
 			if ok {
+				if key == "id" {
+					f, err := flavors.Get(client, flavor).Extract()
+					if err != nil {
+						return "", err
+					}
+					return f.Name, nil
```
jichenjc, line 285: I don't understand why need force return f.Name, as we need some value to indicate it's the desired flavor..
then why we need force return name? 

https://github.com/openstack/nova/blob/master/nova/api/validation/parameter_types.py#L273
indicate flavor name min length is 1 ,so I think `name` will be satified and returned anyway

seanschneeweiss: The idea is to provide a more meaningful return, if we only have the ID to lookup. So when using `kubectl get nodes -L "node.kubernetes.io/instance-type"` we don't see the flavor id, but rather the name of it.

seanschneeweiss: Maybe @sbueringer could comment on this.

jichenjc: ok, I think ID is also reasonable output ...
anyway, wait for other folks to comment

ramineni: @seanschneeweiss I guess "id" is available only till microversion 2.46 and "original_name" is the way forward in the latest version , so I'm ok to return name for consistency .

2 comments from my side
1. L282, we can just return the name if no error , else keep the existing behavior of returning 'id'
1. L274, even though , this is not of existing code change , could you check if "name" is valid key in the flavor object returned by server , I dont see it in compute API spec. If not, could you remove the same

lingxiankong: Agree with @ramineni on L282. For L274, please move `original_name` to the first and remove `name`

seanschneeweiss: Thank you. I've committed the proposed changes.

ramineni:
/override cloud-provider-openstack-acceptance-test-csi-manila

ramineni:
/lgtm

DataMinded:
Im having some trouble with the 1.20 release , prior version does not seem to have this problem

`I0108 21:26:51.582167       1 node_controller.go:390] Initializing node blue-master-1 with cloud provider
I0108 21:26:53.014582       1 openstack.go:937] Claiming to support Zones
I0108 21:26:53.352618       1 node_controller.go:492] Adding node label from cloud provider: beta.kubernetes.io/instance-type=Small HD 4GB
I0108 21:26:53.352643       1 node_controller.go:493] Adding node label from cloud provider: node.kubernetes.io/instance-type=Small HD 4GB
I0108 21:26:53.352648       1 node_controller.go:504] Adding node label from cloud provider: failure-domain.beta.kubernetes.io/zone=AMS-EQ1
I0108 21:26:53.352652       1 node_controller.go:505] Adding node label from cloud provider: topology.kubernetes.io/zone=AMS-EQ1
E0108 21:26:53.360535       1 node_controller.go:212] error syncing 'blue-master-1': Node "blue-master-1" is invalid: metadata.labels: Invalid value: "Small HD 4GB": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?'), requeuing`


Seems my cloud provider has a name ('Small HD 4GB' )with spaces for its server flavors , but the cloud-controller doesnt like it and breaks , not adding any labels.   
Some help would be great

seanschneeweiss:
We do have a fix for that. I will try to provide it by the start of next week.

seanschneeweiss:
@DataMinded Please have a look at https://github.com/kubernetes/cloud-provider-openstack/pull/1364.

DataMinded:
> @DataMinded Please have a look at #1364.

ill test it out in a few hours on a clean install and let u know , looks good .
Thanks for the headsup

DataMinded:
> > @DataMinded Please have a look at #1364.
> 
> ill test it out in a few hours on a clean install and let u know , looks good .
> Thanks for the headsup

i think ill be waiting for the next release to try it out , kinda forgot that part

PR #1218, opened 2020-09-16
[cinder-csi-plugin][manila-csi-plugin] Make cloud-config/cacert volume/volumeMounts configurable

brtknr:

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: latest
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 1.2.1
+version: 1.2.2
 home: https://github.com/kubernetes/cloud-provider-openstack
 icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
--- a/charts/cinder-csi-plugin/templates/nodeplugin-daemonset.yaml
+++ b/charts/cinder-csi-plugin/templates/nodeplugin-daemonset.yaml
@@ -47,7 +47,7 @@ spec:
             capabilities:
               add: ["SYS_ADMIN"]
             allowPrivilegeEscalation: true
-          image: "{{ .Values.csi.plugin.image.repository }}:{{ .Values.csi.plugin.image.tag }}"
+          image: "{{ .Values.csi.plugin.image.repository }}:{{ .Values.csi.plugin.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.csi.plugin.image.pullPolicy }}
           args:
             - /bin/cinder-csi-plugin
@@ -73,14 +73,8 @@ spec:
             - name: pods-probe-dir
               mountPath: /dev
               mountPropagation: "HostToContainer"
-            - name: cacert
-              mountPath: /etc/cacert
-              readOnly: true
+            {{- .Values.csi.plugin.volumeMounts | toYaml | trimSuffix "\n" | nindent 12 }}
       volumes:
-        - name: cacert
-          hostPath:
-            path: /etc/cacert
-            type: Directory
         - name: socket-dir
           hostPath:
             path: /var/lib/kubelet/plugins/cinder.csi.openstack.org
@@ -101,3 +95,4 @@ spec:
           hostPath:
             path: /dev
             type: Directory
+        {{ .Values.csi.plugin.volumes | toYaml | trimSuffix "\n" | nindent 8 }}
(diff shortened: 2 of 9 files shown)


charts/cinder-csi-plugin/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: latest
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 1.2.0
+version: 1.2.2
```
jichenjc, line 5: just curious on the bump from .0 to .2 , any reason behind it?

ramineni: @jichenjc this PR , https://github.com/kubernetes/cloud-provider-openstack/pull/1211 , bumps to 1.2.1

brtknr:
I am increasingly coming to the conclusion that the default appVersion on the master branch should be the last stable release but we can define instructions for users to override with latest tag. Backporting every little change to release-1.19 branch seems like unnecessary hassle.

jichenjc:
> I am increasingly coming to the conclusion that the default appVersion on the master branch should be the last stable release but we can define instructions for users to override with latest tag. Backporting every little change to release-1.19 branch seems like unnecessary hassle.

+1 to this suggestion

ramineni:
> I am increasingly coming to the conclusion that the default appVersion on the master branch should be the last stable release but we can define instructions for users to override with latest tag. Backporting every little change to release-1.19 branch seems like unnecessary hassle.

@brtknr I dont agree, because directly updating to latest tag might not be compatible with the 1.19 sidecars everytime as I mentioned before also. And also , generally we dont backport fixes to 1.19 unless it is required / needed . This time we missed the changes which actually need to be on 1.19, as last min changes, otherwise I dont think there will be much updates to release branches

brtknr:
> > I am increasingly coming to the conclusion that the default appVersion on the master branch should be the last stable release but we can define instructions for users to override with latest tag. Backporting every little change to release-1.19 branch seems like unnecessary hassle.
> 
> @brtknr I dont agree, because directly updating to latest tag might not be compatible with the 1.19 sidecars everytime as I mentioned before also. And also , generally we dont backport fixes to 1.19 unless it is required / needed . This time we missed the changes which actually need to be on 1.19, as last min changes, otherwise I dont think there will be much updates to release branches

Hi @ramineni I am not clear on what you mean by 1.19 sidecars and why there would be compatibility issues there. Could you point me to the portion of the code you are speaking about? Thanks

charts/cinder-csi-plugin/templates/nodeplugin-daemonset.yaml
```diff
@@ -74,12 +74,12 @@ spec:
...
               mountPropagation: "HostToContainer"
             - name: cacert
-              mountPath: /etc/cacert
+              mountPath: /etc/kubernetes
               readOnly: true
       volumes:
         - name: cacert
           hostPath:
-            path: /etc/cacert
+            path: /etc/kubernetes
```
ramineni, line 82: @brtknr what is the reason for this change?

ramineni:
@brtknr Still the CI job fails , is this expected? could you let us the reason for the failure

ramineni:
> > > I am increasingly coming to the conclusion that the default appVersion on the master branch should be the last stable release but we can define instructions for users to override with latest tag. Backporting every little change to release-1.19 branch seems like unnecessary hassle.
> > 
> > 
> > @brtknr I dont agree, because directly updating to latest tag might not be compatible with the 1.19 sidecars everytime as I mentioned before also. And also , generally we dont backport fixes to 1.19 unless it is required / needed . This time we missed the changes which actually need to be on 1.19, as last min changes, otherwise I dont think there will be much updates to release branches
> 
> Hi @ramineni I am not clear on what you mean by 1.19 sidecars and why there would be compatibility issues there. Could you point me to the portion of the code you are speaking about? Thanks

https://github.com/kubernetes/cloud-provider-openstack/blob/master/manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml#L36-L62 , these are sidecars

ramineni:
looks good. Thanks.
/lgtm

jichenjc:
@ramineni is the PR https://github.com/kubernetes/cloud-provider-openstack/pull/1211 you talked about?
many PRs related to this fly around. so kind of lost context now

ramineni:
> @ramineni is the PR #1211 you talked about?
> many PRs related to this fly around. so kind of lost context now
> 

ya, I understand, because of version , it has 1.2.1, so better to merge sequentially

ramineni:
Thanks.
/lgtm

PR #1219, opened 2020-09-16
[cinder-csi-plugin][manila-csi-plugin] Make cloud-config/cacert volume/volumeMounts configurable (release-1.19)

brtknr:
see https://github.com/kubernetes/cloud-provider-openstack/pull/1218

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/manila-csi-plugin/Chart.yaml
+++ b/charts/manila-csi-plugin/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: latest
+appVersion: v1.19.0
 description: Manila CSI Chart for OpenStack
 name: openstack-manila-csi
-version: 0.2.0
+version: 0.2.1
 home: http://github.com/kubernetes/cloud-provider-openstack
 icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: "v1.19.0"
+appVersion: v1.19.0
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 1.2.1
+version: 1.2.2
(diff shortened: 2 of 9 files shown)


charts/manila-csi-plugin/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: latest
```
ramineni, line 2: @brtknr appVersion need to be updated here , v1.19.0

ramineni (changes requested):

ramineni:
@brtknr Is it required to backport this, this is not a bug but only an improvement . We only backport critical/required fixes as need basis

ramineni:
@brtknr Is this required for 1.19 , as I mentioned we only backport bug fixes .

brtknr:
Hi @ramineni I agree it is an improvement but since there hasn't been a working chart release yet, it would be safe to incorporate this into the first chart release from v1.19.0 branch.

lingxiankong:
I think it's fine to have this for release-1.19, hopefully we could have some official, stable Charts for this release.

/lgtm

ramineni:
> Hi @ramineni I agree it is an improvement but since there hasn't been a working chart release yet, it would be safe to incorporate this into the first chart release from v1.19.0 branch. I have seen other changes backported too which are not strictly bug fixes, e.g. the snapshotter image version bump.

@brtknr ok, sounds reasonable. by the way , snapshotter version bump is a required fix , otherwise snapshot feature doesnt work due to incompatible of sidecar versions.

charts/manila-csi-plugin/Chart.yaml
ramineni: this is not addressed?

ramineni:
@brtknr As we want stable release of charts as @lingxiankong mentioned, as an exception, we could incorporate these changes .
please address comment , this should be good to go

brtknr:
This needs /approved label to merge. Assuming that @lingxiankong disabled branch protection on gh-pages branch, this should correctly create a new release and publish to charts.yaml.

ramineni:
As we already have lgtm from @lingxiankong , approving this.
/approve

Issue #993, opened 2020-03-27
Move out OpenStack new client logic to a separate package


kayrus:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**The binaries affected**:

- [x] openstack-cloud-controller-manager
- [ ] cinder-csi-plugin
- [ ] k8s-keystone-auth
- [ ] client-keystone-auth
- [ ] octavia-ingress-controller
- [ ] manila-csi-plugin
- [ ] manila-provisioner
- [ ] magnum-auto-healer
- [ ] barbican-kms-plugin

**What happened**:

Currently openstack new client logic is defined in openstack-cloud-controller-manager, but used by all openstack related controllers.

**What you expected to happen**:

A better place for this logic is a separate package.

PR #867, opened 2019-12-05
Core: Enable OpenStack API debugger, when verbosity is set to 6

kayrus:
**What this PR does / why we need it**:

This PR adds verbose OpenStack API request/response debugging, when `--v=6` (or higher) verbosity is set.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*:

fixes #840 

**Special notes for your reviewer**:

**Release note**:

```release-note
Core: Enable OpenStack API debugger, when verbosity is set to 6 or higher
```

PR #1222, opened 2020-09-16
[k8s-keystone-auth] improve keystone connection trace.

### LINKED ISSUE #1207: [k8s-keystone-auth] Implement a tracing http roundtripper client
Affected binary:  k8s-keystone-auth

/kind feature

**What happened**:

Due to the following bug https://bugs.launchpad.net/charm-kubernetes-worker/+bug/1892067, we
identified that the current log tracing levels aren't enough to diagnose  issues in the communication
to the keystone endpoints.

In particular,  this line https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/identity/keystone/keystone.go#L281 doesn't logs any error during the authorization
phase.

Also, would be very helpful to have http connection level tracing in order to see when a connection
is started, finished and or any timeouts ocurred, increasing the logging levels that can be useful when diagnosing issues in the communication to the keystone endpoints at the http level.

* We would like to be able to see why an authentication failed in the kslog , why a token authorization failed,
and be able to enable trace level debugging for the http client.

niedbalski:
[k8s-keystone-auth] improve keystone connection trace.

* Log all errors on failures. (authorizeToken)
* Add httptrace transport if log level == 4.

Fixes #1207

Due to the following bug https://bugs.launchpad.net/charm-kubernetes-worker/+bug/1892067, we
identified that the current log tracing levels aren't enough to diagnose issues in the communication
to the keystone endpoints.

In particular, this line https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/identity/keystone/keystone.go#L281 doesn't logs any error during the authorization
phase.

Also, would be very helpful to have http connection level tracing in order to see when a connection
is started, finished and or any timeouts ocurred, increasing the logging levels that can be useful when diagnosing issues in the communication to the keystone endpoints at the http level.

fixes #1207

### DIFF
--- a/pkg/identity/keystone/keystone.go
+++ b/pkg/identity/keystone/keystone.go
@@ -22,11 +22,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/http/httptrace"
 	"time"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
 	"github.com/gophercloud/gophercloud/openstack/utils"
+
 	"github.com/gorilla/mux"
 	"github.com/spf13/pflag"
 	"gopkg.in/yaml.v2"
@@ -246,6 +248,7 @@ func (k *KeystoneAuth) Handler(w http.ResponseWriter, r *http.Request) {
 	defer r.Body.Close()
 	err := decoder.Decode(&data)
 	if err != nil {
+		klog.Errorf("an error decoding data: %v", err)
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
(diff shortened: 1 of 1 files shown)


kayrus:
@niedbalski I don't think this is a proper concept. See #993 and #867. From my perspective it would be better to detach the client related logic to a separate package, add a trace logic there and reuse it for all cloud provider components. If we start introducing the logic for each component independently again, we will have a lot of duplicate code with a mess we had before.

niedbalski:
@jichenjc I don't think the test failure is related to my change, any chance you can help me to check?

jichenjc:
/retest
let's retest and see what happen @niedbalski 
and you might want to check with @kayrus on his comments above

niedbalski:
Hello @kayrus 

Thank you for reviewing. I wonder if you could guide me on your idea of switching the scope of the createKeystoneClient method into something that's re-usable across components. 

This PR just intends to make the transport more verbose in situations on which communication is broken or
a failure event occurs in the authorization phase.

Thanks

kayrus:
@niedbalski sorry for a late response. have you tried to use `--v=6` ? It should trace all HTTP calls. This feature was introduced in #867
If you want to introduce something new, then I just suggest to move out the client logic into a separate package: #993 
Unfortunately I don't have any specific requirements, since it should be quite straight-forward.

PR #1226, opened 2020-09-18
[cinder-csi-plugin] [doc] update doc about VolumeAttributes is optional

jichenjc:

indicate `VolumeAttributes ` for ephemeral disk is optional and default to 1G

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/using-cinder-csi-plugin.md
+++ b/docs/using-cinder-csi-plugin.md
@@ -405,6 +405,9 @@ $ kubectl create -f manifests/cinder-csi-plugin/
 ```
 $ kubectl create -f examples/cinder-csi-plugin/inline/inline-example.yaml
 ```
+
+> NOTE: specifying volume `capacity` is optional, if it's not provided, it will default to `1Gi`
+
 3. Get the pod description, verify created volume in Volumes section.
 ```
 $ kubectl describe pod


jichenjc:
@ramineni please help to check the doc update

docs/using-cinder-csi-plugin.md
```diff
@@ -423,6 +423,8 @@ Volumes:
 
 ```
 
+> NOTE: `VolumeAttributes` is optional, if it's not provided, it will be default to `1Gi`
```
brtknr, line 426: will default to (delete be)

lingxiankong: Can we move this after `Create a pod with inline volume`

jichenjc: updated both , thanks

docs/using-cinder-csi-plugin.md
```diff
@@ -405,6 +405,9 @@ $ kubectl create -f manifests/cinder-csi-plugin/
 ```
 $ kubectl create -f examples/cinder-csi-plugin/inline/inline-example.yaml
 ```
+
+> NOTE: `VolumeAttributes` is optional, if it's not provided, it will default to `1Gi`
```
ramineni, line 409: @jichenjc As `capacity` value defaults to 1Gi, its better to add like 
specifying volume `capacity` is optional , if it's not provided, it will default to `1Gi`

ramineni (changes requested):
Thanks @jichenjc for updating. One suggestion inline

docs/using-cinder-csi-plugin.md
jichenjc: done, thanks~

ramineni:
Thanks.
/lgtm

Issue #1230, opened 2020-09-20
[octavia-ingress-controller] Deletion of loadbalancer fails due to 403 (Forbidden) against the Barbican API


mikejoh:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind bug

**What happened**:
When testing a bit back and forth with a newer version of the Octavia Ingress Controller (`v1.18.0`) and using a seperate OIC OpenStack user with the `member` and `load-balancer_admin` roles (i'm only using the default Keystone roles) attached the deletion of the loadbalancer fails due to:
```
Failed to delete openstack resources for ingress nginx/nginx: failed to remove Barbican secrets: Request forbidden: [GET https://barbican.example.com/v1/secrets?secret_type=opaque], error message: {"code": 403, "description": "Secret(s) retrieval attempt not allowed - please review your user/project privileges", "title": "Forbidden"}
```
The `Ingress` manifest looks like this:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: nginx
  namespace: nginx
  annotations:
    kubernetes.io/ingress.class: "openstack"
    octavia.ingress.kubernetes.io/internal: "false"
spec:
  rules:
  - host: nginx.example.com
    http:
      paths:
      - path: /
        backend:
          serviceName: nginx-svc
          servicePort: 80
```

This did work before the TLS encryption feature was added and merged. Also note that when adding the `creator` role to the OIC user everything works as expected.

I've traced this particular problem to this part of the `deleteIngress` method:
https://github.com/kubernetes/cloud-provider-openstack/blob/35187fcbadf447a41fb0972bd43093f3bd88e19d/pkg/ingress/controller/controller.go#L517-L525

The `c.osClient.Barbican` object will always be set AFAIT since it's initialized with the rest of the OpenStack clients used by OIC here:

https://github.com/kubernetes/cloud-provider-openstack/blob/35187fcbadf447a41fb0972bd43093f3bd88e19d/pkg/ingress/controller/openstack/client.go#L73-L81

By adding another condition to that if-statement we could ensure that we don't try to contact the Barbican API at every delete event.

**What you expected to happen**:
Since i haven't specified any `TLS` options in the `Ingress` manifest i expect OIC to not try to delete any secrets in Barbican.

**How to reproduce it**:
1. Use a seperate user with when running the OIC and attach the OpenStack default roles of `member` and `load-balancer_admin` to it.
[...]

PR #1231, opened 2020-09-20
[octavia-ingress-controller] Check if TLS field in Ingress spec is set

mikejoh:
At the moment the OIC connects to the Barbican API on every delete event since the if-statement checking if Barbican client is not `nil` will always be true.

This PR adds another condition to that if-statement, in this case to see if the `TLS` field of the `Ingress` spec is not `nil`. 

With that said there's two other ways to approach this when it comes to the if-statement:

1. As in the `ensureIngress` method check the `len()` of `ing.spec.TLS`.
2. Remove the current if-statement in the `deleteIngress` method and replace it with either a `len()` check or `!= nil` check as i've proposed in this PR.

I've tested this locally and it seems to work as i would expect but i need some more eyes on this.

Proposed fix to #1230

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/ingress/controller/controller.go
+++ b/pkg/ingress/controller/controller.go
@@ -515,7 +515,7 @@ func (c *Controller) deleteIngress(ing *nwv1beta1.Ingress) error {
 	lbName := utils.GetResourceName(ing.Namespace, ing.Name, c.config.ClusterName)
 
 	// Delete Barbican secrets
-	if c.osClient.Barbican != nil {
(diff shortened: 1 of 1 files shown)


jichenjc:
seems reasonable but not expert here, @lingxiankong can you take a further look?

/lgtm

lingxiankong:
/lgtm
/approve

Need to backport to release v1.19 and v1.18.

Issue #913, opened 2020-01-24
[cinder-csi-plugin] metrics are not accessible


bgagnon:
/kind bug

**The binaries affected**:

- [x] cinder-csi-plugin

**What happened**:

Tried to access the metrics at `/metrics` for the cinder-csi-plugin controller and node plugins.

**What you expected to happen**:

Some prometheus metrics.

**Anything else we need to know?**:

Recent version of CSI sidecar containers expose Prometheus metrics on a configurable bind address, port and path (defaults to `/metrics`). This should also be provided for `cinder-csi-plugin`, the main container of the CSI Cinder Controller.

jichenjc:
so the question is other side cars already exposed some metrics and cinder-csi-plugin doesn't have the output from /metrics?  if so what's the metrics are recommended? e.g some cinder info? @bgagnon

bgagnon:
I'm not sure if _any_ metrics exist in memory at the moment. It's hard to tell without the `/metrics` handler registered. I think the following metrics would be relevant in the case of the Cinder CSI driver:

* basic Openstack API client metrics (counters for requests, errors)
* any other relevant CSI-related counters
* (maybe) gRPC metrics (eg. [grpc-ecosystem/go-grpc-prometheus](https://github.com/grpc-ecosystem/go-grpc-prometheus))

Since the CSI sidecars have some metrics, I think it's good for consistency if the "main" container has some as well.

jichenjc:
um.. yes, maybe a framework first then check what can be added..

jichenjc:
https://github.com/kubernetes-csi/external-snapshotter
has --metrics-address and --metrics-path param

vendor/github.com/kubernetes-csi/csi-lib-utils/metrics/metrics.go
but that's for CSI side car only, not sure it can be used directly, need further check

HaveFun83:
any news here?
/remove-lifecycle rotten

jichenjc:
I was wrong on above comments, it's not only related to snapshot but also related to others.
so need check CSI driver as well

jichenjc:
/assign 
maybe we can refer to #1178

MPV:
Would this help? https://github.com/kubernetes-csi/docs/issues/339

PR #1236, opened 2020-09-24
[all]refactory metrics code so that additional code can consume it

jichenjc:

refactory metrics code so that https://github.com/kubernetes/cloud-provider-openstack/issues/913 might consume it

fixes #

**Release note**:

```release-note
NONE
```


jichenjc:
@seanschneeweiss any suggestions?

pkg/cloudprovider/providers/openstack/metrics/metrics.go
```diff
@@ -20,89 +20,57 @@ import (
...
-		errors: metrics.NewCounterVec(
+		Errors: metrics.NewCounterVec(
 			&metrics.CounterOpts{
 				Name: "cloudprovider_openstack_reconcile_errors_total",
 				Help: "Total number of OpenStack cloud controller manager reconciliation errors",
 			}, []string{"operation"}),
 	}
-	requestMetrics = &openstackMetrics{
-		duration: metrics.NewHistogramVec(
+	occmRequestMetrics = &openstack_metrics.OpenstackMetrics{
```
seanschneeweiss, line 47: @jichenjc `occmReconcileMetrics` are somehow specific to the CCM.  Whereas `occmRequestMetrics` is something that is reused by other parts of the OpenStack cloud provider, e.g. Cinder CSI Plugin.
Maybe it would be possible to move the `occmRequestMetrics` out of `pkg/cloudprovider/providers/openstack/metrics` to your proposed location `pkg/metrics/metrics.go`. What do you think?

jichenjc: ok, that make sense, let me refactory the PR, thanks

jichenjc: um... @seanschneeweiss another thought here , the definition created context
then it's up to OCCM or CSI cinder or others to create an instance for them
now occm created 2 , 1 for reconcile and the other for request metrics.. so keep occm_xxx
in existing places should be good? later on a new one named csi_cinder_xxx metrics will be created in cinder side ..
```
type MetricContext struct {
       Start            time.Time
       Attributes       []string
       ReconcileMetrics *OpenstackMetrics
       RequestMetrics   *OpenstackMetrics
}
```

seanschneeweiss: @jichenjc Please excuse my late respone. I would agree to define the metrics inside each component, and have the context creation inside each. That is somehow also the strategy other cloud providers use, imho. The OpenStack API metric (RequestMetrics) is slightly different, because it is used in so many places, not sure how to handle that. I'm fine with either way. But what we might have to keep in mind, that using the same metric names on multiple locations is not possible. So how to name the OpenStack API metrics within OCCM, how to name them within cinder csi? I expect to have a single name `openstack_api_request_...` for all API calls (no matter which component).

jichenjc: um... so you mean , the calls either made by OCCM or CSI will be summed up and showed as one call, correct? 
ok, this is something different to my original thought , I thought we are running CSI/OCCM in different Pod and thus we are tracking there from different pod perspective, now I understand your point and need check it..

jichenjc: @seanschneeweiss sorry, still learning here, let's say OCCM and CSI cinder both use openstack API 
is there someway that we can consolidate the API calls ?but they are in different container, can we do that?

jichenjc: ok, let me try to move the reconcile logic back here and try again..

pkg/util/openstack/keymanager.go
```diff
@@ -22,7 +22,7 @@ import (
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/keymanager/v1/secrets"
-	"k8s.io/cloud-provider-openstack/pkg/cloudprovider/providers/openstack/metrics"
+	request_metrics "k8s.io/cloud-provider-openstack/pkg/metrics"
```
seanschneeweiss, line 25: Can `"k8s.io/cloud-provider-openstack/pkg/metrics"` be imported without giving it an alias - similar to `pkg/util/openstack/loadbalancer.go` and `pkg/util/openstack/network.go`?

pkg/cloudprovider/providers/openstack/metrics/metrics.go
seanschneeweiss: @jichenjc looks good to me now, thank you very much.

pkg/util/openstack/keymanager.go
jichenjc: @seanschneeweiss just updated, thanks~

jichenjc:
I did local test and it works

```
# curl -k https://localhost:10258/metrics | grep open
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0# HELP openstack_api_request_duration_seconds [ALPHA] Latency of an OpenStack API call
# TYPE openstack_api_request_duration_seconds histogram
openstack_api_request_duration_seconds_bucket{request="server_get",le="0.005"} 0
openstack_api_request_duration_seconds_bucket{request="server_get",le="0.01"} 0
openstack_api_request_duration_seconds_bucket{request="server_get",le="0.025"} 0
openstack_api_request_duration_seconds_bucket{request="server_get",le="0.05"} 0
openstack_api_request_duration_seconds_bucket{request="server_get",le="0.1"} 0
openstack_api_request_duration_seconds_bucket{request="server_get",le="0.2
....
```

no `cloudprovider` tested due to no LB env in my env,( switch to code before refactory got same result) 

@seanschneeweiss @lingxiankong @ramineni please help check

pkg/cloudprovider/providers/openstack/metrics/metrics.go
```diff
@@ -20,87 +20,61 @@ import (
 	"sync"
 	"time"
 
+	openstack_metrics "k8s.io/cloud-provider-openstack/pkg/metrics"
```
lingxiankong, line 23: Please don't use snake_case in golang, for the package names, by convention we are using lower case, single-word names, there should be no need for underscores or mixedCaps. 

So here I'd suggest `cpometrics`, anywhere else that referencing this module could use `openstackmetrics` if the name conflicts.

jichenjc: yes, thanks for this ,I will update

jichenjc:
@lingxiankong updated just now .. thanks

lingxiankong:
thanks

/lgtm

jichenjc:
@ramineni can you help to check the updated code? Thanks

ramineni:
@jichenjc sorry for the delay . Did a quick glance today, its lil confusing why we have to import two different metric packages to consume metrics, I feel it should be under one package which should register both occm metrics and openstack_api metrics.

May be I missed the context , Ill take a deeper look at it next week again

jichenjc:
@ramineni because one is for reconcile and one for API call , 2 different metics ..

ramineni (changes requested):
@jichenjc IMO it would be better to define metrics in one place under one package , may be we could go like

    metrics/ 
         metrics.go // generic definition
         reconcile_metrics.go // for reconcile
        apirequest_metrics.go //for api
Any new metrics if required could be added further.
That would be less confusing I think. Does it make sense?  wdyt?

pkg/cloudprovider/providers/openstack/metrics/metrics.go
```diff
@@ -20,87 +20,61 @@ import (
...
-	}
 )
 
 // MetricContext indicates the context for OpenStack metrics.
 type MetricContext struct {
-	start      time.Time
-	attributes []string
+	Start            time.Time
+	Attributes       []string
+	ReconcileMetrics *cpometrics.OpenstackMetrics
```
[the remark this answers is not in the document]
jichenjc: ok, I need to give a thought on this in how to handle .. thanks for pointing this out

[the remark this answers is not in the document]
jichenjc: @ramineni just updated

ramineni:
> @jichenjc IMO it would be better to define metrics in one place under one package , may be we could go like
> 
> ```
> metrics/ 
>      metrics.go // generic definition
>      reconcile_metrics.go // for reconcile
>      apirequest_metrics.go //for api
>      newmetrics.go.....
> ```
> 
> That would be less confusing I think. Does it make sense? wdyt?

@jichenjc looks like you missed above comment?

jichenjc:
> 
> 
> > @jichenjc IMO it would be better to define metrics in one place under one package , may be we could go like
> > ```
> > metrics/ 
> >      metrics.go // generic definition
> >      reconcile_metrics.go // for reconcile
> >      apirequest_metrics.go //for api
> >      newmetrics.go.....
> > ```
> > 
> > 
> > That would be less confusing I think. Does it make sense? wdyt?
> 
> @jichenjc looks like you missed above comment?

oops, need check this ,thanks for reminder

jichenjc:
@ramineni can you help to check this again? Thanks

pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
```diff
@@ -49,7 +49,8 @@ import (
 	cloudprovider "k8s.io/cloud-provider"
 	klog "k8s.io/klog/v2"
 
-	"k8s.io/cloud-provider-openstack/pkg/cloudprovider/providers/openstack/metrics"
+	"k8s.io/cloud-provider-openstack/pkg/metrics"
+	cpometrics "k8s.io/cloud-provider-openstack/pkg/metrics"
```
ramineni, line 53: same package imported twice?

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -291,7 +291,8 @@ func (l Logger) Printf(format string, args ...interface{}) {
 }
 
 func init() {
-	metrics.RegisterMetrics()
+	metrics.RegisterAPIMetrics()
```
ramineni, line 294: one suggestion here, instead of having 2 calls here, how about we define RegisterMetrics() inside metrics.go , which calls all relevant metrics, so that if any new metrics added , we need to change only in metrics code
and here we can have a generic call metrics.RegisterMetrics()

ramineni (changes requested):
@jichenjc Thanks. Looks good to me mostly. minor comments inline.
And also , this is significant refactoring , I hope this is well tested? If not , pls ensure nothing existing is broken with this change. We dont have CI to test this, so manual testing is needed.

jichenjc:
> 
> 
> @jichenjc Thanks. Looks good to me mostly. minor comments inline.
> And also , this is significant refactoring , I hope this is well tested? If not , pls ensure nothing existing is broken with this change. We dont have CI to test this, so manual testing is needed.

I do have a test and showing the metrics is work as non-refactoried code .
and I think we should add some test overtime ..

I can provide some test after code change based on your comments

jichenjc:
@ramineni I used my local build image and tested ,seems works fine

```
# k describe daemonset openstack-cloud-controller-manager -n kube-system | grep Image:
    Image:      jichenjc/openstack-cloud-controller-manager-amd64
```

```
# curl -k https://localhost:10258/metrics
# HELP apiserver_audit_event_total [ALPHA] Counter of audit events generated and sent to the audit backend.
# TYPE apiserver_audit_event_total counter
apiserver_audit_event_total 0
# HELP apiserver_audit_requests_rejected_total [ALPHA] Counter of apiserver requests rejected due to an error in audit logging backend.
# TYPE apiserver_audit_requests_rejected_total counter
apiserver_audit_requests_rejected_total 0
# HELP apiserver_client_certificate_expiration_seconds [ALPHA] Distribution of the remaining lifetime on the certificate used to authenticate a request.
# TYPE apiserver_client_certificate_expiration_seconds histogram
apiserver_client_certificate_expiration_seconds_bucket{le="0"} 0
apiserver_client_certificate_expiration_seconds_bucket{le="1800"} 0
apiserver_client_certificate_expiration_seconds_bucket{le="3600"} 0
apiserver_client_certificate_expiration_seconds_bucket{le="7200"} 0
apiserver_client_certificate_expiration_seconds_bucket{le="21600"} 0
apiserver_client_certificate_expiration_seconds_bucket{le="43200"} 0
apiserver_client_certificate_expiration_seconds_bucket{le="86400"} 0
apiserver_client_certificate_expiration_seconds_bucket{le="172800"} 0
apiserver_client_certificate_expiration_seconds_bucket{le="345600"} 0
apiserver_client_certificate_expiration_seconds_bucket{le="604800"} 0
apiserver_client_certificate_expiration_seconds_bucket{le="2.592e+06"} 0
apiserver_client_certificate_expiration_seconds_bucket{le="7.776e+06"} 0
apiserver_client_certificate_expiration_seconds_bucket{le="1.5552e+07"} 0
apiserver_client_certificate_expiration_seconds_bucket{le="3.1104e+07"} 0

```

jichenjc:
@ramineni @lingxiankong please help to review, thanks

ramineni:
Thanks.
/lgtm

pkg/metrics/metrics_api.go
```diff
@@ -0,0 +1,62 @@
...
+			}, []string{"request"}),
+		Errors: metrics.NewCounterVec(
+			&metrics.CounterOpts{
+				Name: "openstack_api_request_errors_total",
+				Help: "Total number of errors for an OpenStack API call",
+			}, []string{"request"}),
+	}
+)
+
+// ObserveReconcile records the request latency and counts the errors.
```
seanschneeweiss, line 46:
```suggestion
// ObserveRequest records the request latency and counts the errors.
```

pkg/metrics/metrics_reconcile.go
```diff
@@ -0,0 +1,63 @@
...
+			}, []string{"operation"}),
+		Errors: metrics.NewCounterVec(
+			&metrics.CounterOpts{
+				Name: "cloudprovider_openstack_reconcile_errors_total",
+				Help: "Total number of OpenStack cloud controller manager reconciliation errors",
+			}, []string{"operation"}),
+	}
+)
+
+// ObserveReconcile records the request latency and counts the errors.
```
seanschneeweiss, line 47:
```suggestion
// ObserveReconcile records reconciliation duration,
// frequency and number of errors.
```

seanschneeweiss (changes requested):

pkg/metrics/metrics_reconcile.go
jichenjc: done

pkg/metrics/metrics_api.go
jichenjc: done

jichenjc:
@lingxiankong @seanschneeweiss @ramineni can you take another look? thanks

lingxiankong:
/approve

Thanks for picking up this refactoring work!

Issue #1232, opened 2020-09-21
[manila-csi-plugin] Update external-provisioner to v2.0.2


gman0:
We'll need to update manila-csi-plugin's external-provisioner. Currently used v1.4.0 is not compatible with v1beta1 snapshots https://github.com/kubernetes/cloud-provider-openstack/pull/1205

/kind feature

gman0:
Done in https://github.com/kubernetes/cloud-provider-openstack/pull/1205. Closing.

PR #1205, opened 2020-09-14
[manila-csi-plugin] Updated external-snapshotter to v2.1.1 and external-provisioner to v2.0.2

gman0:
**What this PR does / why we need it**:
This PR bumps csi-manila's the external-snapshotter to v2.1.1 and external-provisioner to v2.0.2. This is a [backwards incompatible](https://github.com/kubernetes-csi/external-snapshotter#upgrade-from-v1alpha1-to-v1beta1) change.

**Which issue this PR fixes(if applicable)**:
fixes https://github.com/kubernetes/cloud-provider-openstack/issues/1194 , https://github.com/kubernetes/cloud-provider-openstack/issues/1232

**Special notes for reviewers**:

**Release note**:

```release-note
manila-csi-plugin: Updated external-provisioner to v2.0.2 and external-snapshotter to v2.1.1
```

PR #1245, opened 2020-09-25
[manila-csi-plugin] Updated external-provisioner to v2.0.2

gman0:
Currently, manila-csi-plugin uses external-provisioner sidecar v1.4.0, which doesn't support v1beta1 snapshots as a volume data source. This PR updates it to v2.0.2.

fixes https://github.com/kubernetes/cloud-provider-openstack/issues/1232

This PR needs to be merged before #1205.

**Release note**:

```release-note
manila-csi-plugin: Updated external-provisioner v2.0.2
```

### DIFF
--- a/charts/manila-csi-plugin/values.yaml
+++ b/charts/manila-csi-plugin/values.yaml
@@ -67,7 +67,7 @@ controllerplugin:
   provisioner:
     image:
       repository: quay.io/k8scsi/csi-provisioner
-      tag: v1.4.0
+      tag: v2.0.2
       pullPolicy: IfNotPresent
     resources: {}
   # CSI external-snapshotter container spec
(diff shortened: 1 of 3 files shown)


gman0:
PTAL @ramineni @jichenjc

charts/manila-csi-plugin/values.yaml
```diff
@@ -67,7 +67,7 @@ controllerplugin:
   provisioner:
     image:
       repository: quay.io/k8scsi/csi-provisioner
-      tag: v1.4.0
+      tag: v2.0.2
```
jichenjc, line 70: I used to know some RBAC changes are needed for cinder.. not sure for manila ..

ramineni:
@gman0 Isn't it better to merge this PR with the snapshotter to make sure tests pass as both are related ?

charts/manila-csi-plugin/values.yaml
gman0: @jichenjc There are RBAC changes regarding creating volumes from snapshots, those are in the [external-snapshotter v2 PR](https://github.com/kubernetes/cloud-provider-openstack/pull/1205) though.

jichenjc: I messed up .. thanks for the comments

gman0:
@ramineni I can do that

PR #1247, opened 2020-09-28
[cinder-csi-plugin] Update plugin docs [Part 1]

ramineni:
This PR updates the docs of cinder-csi-plugin for better clarity. Also moves the location to separate folder docs/cinder-csi-plugin to have any additional docs related to the plugin under the dir.

As doc has become fairly large with many features added over period of time, its time to split the doc for better readability. For the rest , Ill split the data across multiple files and I'll update the same on followup PRs

P.S If agreed , for consistency , ill move the respective plugins under their folders .

fixes #

**Release note**:

```release-note
NONE
```


gman0:
I support splitting the docs into separate directories/files, the docs as they currently are, are getting too big and unwieldy. Moving the files from `docs/` to separate dirs will break existing hypertext links which can be used who knows where... How about keeping symlinks to the respective documents in `docs/using-<module>.md`?

ramineni:
> I support splitting the docs into separate directories/files, the docs as they currently are, are getting too big and unwieldy. Moving the files from `docs/` to separate dirs will break existing hypertext links which can be used who knows where... How about keeping symlinks to the respective documents in `docs/using-<module>.md`?

Right, Im planning to split existing doc for better readability. Are you suggesting split the files in docs/ folder itself? 
Im thinking if we could update plugin links here https://github.com/kubernetes/cloud-provider-openstack#cloud-provider-openstack to the right location would suffice and any other internal links. If we have all the docs related to plugin same folder it would be better organized. wdyt?

docs/cinder-csi-plugin/using-cinder-csi-plugin.md
```diff
@@ -41,114 +44,119 @@
...
-```
+### Block Storage
+These configuration options pertain to block storage and should appear in the `[BlockStorage]` section of the `$CLOUD_CONFIG` file.
+
+* `node-volume-attach-limit`
+  Optional. To configure maximum volumes that can be attached to the node. Its default value is `256`.
+* `rescan-on-resize`
+  Optional. Set to `true`, to rescan block device and verify its size before expanding the filesystem. Not all hypervizors have a /sys/class/block/XXX/device/rescan location, therefore if you enable this option and your hypervizor doesn't support this, you'll get a warning log on resize event. It is recommended to disable this option in this case. Defaults to `false`
+
+### Metadata
```
jichenjc, line 89: I think this is generic and should apply to OCCM as well..
so maybe better to doc there then point link to OCCM from here
not a blocking one

docs/cinder-csi-plugin/using-cinder-csi-plugin.md
```diff
@@ -41,114 +44,119 @@
...
-### Example local-up-cluster.sh
+## Driver Config
 
-```ALLOW_PRIVILEGED=true RUNTIME_CONFIG="storage.k8s.io/v1=true" LOG_LEVEL=5 hack/local-up-cluster.sh```
+Implementation of `cinder-csi-plugin` relies on following OpenStack services.
 
-### Deploy
+| Service                        | API Version(s) | Deprecated | Required |
+|--------------------------------|----------------|------------|----------|
+| Identity (Keystone)            | v2             | Yes        | No       |
```
jichenjc, line 69: we actually have nova v2 (current version is v2.1 microversion)
do we want to list deprecated one here or just go with v3 keystone?

jichenjc (commented):
/lgtm

only some comment/suggestions

docs/cinder-csi-plugin/using-cinder-csi-plugin.md
ramineni: @jichenjc I have copied from there , its just one parameter , so I thought good to add instead of redirecting

docs/cinder-csi-plugin/using-cinder-csi-plugin.md
ramineni: @jichenjc Its just supported major versions I have added. Authenticating through keystone v2 API is still supported through our code , not removed completely, even though we don't recommend. So added here as deprecated but still can use it.

docs/cinder-csi-plugin/using-cinder-csi-plugin.md
jichenjc: ok, as I am not sure it will be updated thus later on it might be inconsistent..
anyway, either way not a big deal

docs/cinder-csi-plugin/using-cinder-csi-plugin.md
jichenjc: ok, works for me ,thanks~

PR #1249, opened 2020-09-29
[occm] Fix updating listeners

### LINKED ISSUE #1242: [occm] octavia does not update members on multi-port service edge case

>
/kind bug
> /kind feature

**What happened**:

* created a service type `LoadBalancer` having 2 ports defined
* patched the service type `LoadBalancer` and removed one port, removed the nodePort of the leftover port
* occm reconciles and does not update the members, they will still refer the old/outdated nodePort (for details see below)

* The loadbalancer members get updated

**How to reproduce it**:

* Created a service:
  ```
  apiVersion: v1
  kind: Service
  metadata:
    name: kuard
  spec:
    ports:
    - port: 80
      name: http
      protocol: TCP
      targetPort: 8080
    - port: 443
      name: https
      protocol: TCP
      targetPort: 8080
    type: LoadBalancer
    selector:
      app: kuard
  ```
* Patched service: `kubectl patch svc kuard --type json -p '[{"op": "remove","path": "/spec/ports/1"},{"op": "remove","path": "/spec/ports/0/nodePort"}]'`
  * this removes the https port 
  * and unset the nodePort of the http port (and kubernetes fills in a new random port number)
* Take a look at `openstack loadbalancer member list $POOL_ID` 

The members don't get updated, because the following happens:
* in [ensureOctaviaLoadBalancer L:1405](https://github.com/kubernetes/cloud-provider-openstack/blob/release-1.19/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go#L1405) 
  the listener object is getting returned which is just a pointer to the array object of the passed `oldListeners` slice.
* in [L:1412](https://github.com/kubernetes/cloud-provider-openstack/blob/release-1.19/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go#L1412)
  we call popListener, which removes the listener itself from the `oldListeners` slice.
  However this also then changes our reference of the listener object.

  The following snippet visualizes this a bit better: https://play.golang.org/p/C8S7nDlZJEO
  ```
  Printing the listener
  > Pointer: 0xc0000ca780 , Listener ID: foo - Listener: &{foo     0  [] 0 []  false [] []  0 0 0 0 map[] []}
  Removing the listener itself from oldListeners via popListener
  Printing the listener object again - it should keep the ID foo (which it doesn't)
  > Pointer: 0xc0000ca780 , Listener ID: bar - Listener: &{bar     0  [] 0 []  false [] []  0 0 0 0 map[] []}
  ```

The `ensureOctaviaLoadBalancer` function then proceeds and will execute `lbaas.ensureOctaviaPool` using the wrong `loadbalancer.ID` (in this case it applies the new nodePorts to the pool which afterwards is getting deleted instead of the actual pool of the listener which is kept).

A small workaround is the following patch:
```
diff --git a/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go b/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
index 9f0f8f0..1b0eb37 100644
--- a/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
@@ -1402,7 +1402,8 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
        }
 
        for _, port := range service.Spec.Ports {
-               listener, err := lbaas.ensureOctaviaListener(loadbalancer.ID, oldListeners, service, port, svcConf)
+               listenerPtr, err := lbaas.ensureOctaviaListener(loadbalancer.ID, oldListeners, service, port, svcConf)
+               var listener listeners.Listener = *listenerPtr
                if err != nil {
                        return nil, err
                }
@@ -1411,7 +1412,7 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
                // Pop valid listener.
                oldListeners = popListener(oldListeners, listener.ID)
 
-               pool, err := lbaas.ensureOctaviaPool(loadbalancer.ID, listener, service, port, nodes, svcConf)
+               pool, err := lbaas.ensureOctaviaPool(loadbalancer.ID, &listener, service, port, nodes, svcConf)
                if err != nil {
                        return nil, err
                }

```

However: to not have similar issues on other places: the `ensureOctaviaListener` should maybe get adjusted to return the listener without still depending on oldListeners.

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: Built from `release-1.19`
- OpenStack version: Canonical OpenStack Queens LTS
- Others: Kubernetes v1.19.1

---
<sup>
Christian Schlotter <christian.schlotter@daimler.com>, Daimler TSS GmbH, 
<a href="https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md">Imprint</a>
</sup>

lingxiankong:
1. Fixed updating listener for service update.
2. Added a test case for service update.

fixes #1242

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/e2e/cloudprovider/test-lb-service.sh
+++ b/tests/e2e/cloudprovider/test-lb-service.sh
@@ -14,6 +14,19 @@ TIMEOUT=${TIMEOUT:-300}
 FLOATING_IP=${FLOATING_IP:-""}
 NAMESPACE="octavia-lb-test"
 
+delete_resources() {
+  ERROR_CODE="$?"
+
+  printf "\n>>>>>>> Deleting k8s resources\n"
+  for name in "test-basic" "test-x-forwarded-for" "test-update-port"; do
+    kubectl -n $NAMESPACE delete service ${name}
+  done
+  kubectl -n ${NAMESPACE} delete deploy echoserver
+
+  exit ${ERROR_CODE}
+}
+trap "delete_resources" EXIT;
+
 ########################################################################
 ## Name: wait_for_service
 ## Desc: Waits for a k8s service until it gets a valid IP address
@@ -33,7 +46,36 @@ function wait_for_service {
     fi
     sleep 3
     now=$(date +%s)
-    [ $now -gt $end ] && printf "\n>>>>>>> FAIL: Failed to wait for the Service ${service_name} created in time\n" && exit -1
+    [ $now -gt $end ] && printf "\n>>>>>>> FAIL: Timeout when waiting for the Service ${service_name} created\n" && exit -1
+  done
+}
+
+########################################################################
+## Name: wait_for_loadbalancer
+## Desc: Waits for the load balancer to be ACTIVE
+## Params:
+##   - (required) The load balancer ID.
+########################################################################
+function wait_for_loadbalancer {
+  local lbid=$1
+  local i=0
+
+  end=$(($(date +%s) + ${TIMEOUT}))
+  while true; do
+    status=$(kubectl -n $NAMESPACE exec openstackcli -- openstack loadbalancer show $lbid -f value -c provisioning_status)
+    if [[ $status == "ACTIVE" ]]; then
+      if [[ $i == 2 ]]; then
+        printf "\n>>>>>>> Load balancer ${lbid} is ACTIVE\n"
+        break
+      fi
+      let i++
+    else
+      i=0
+    fi
+
+    sleep 3
+    now=$(date +%s)
+    [ $now -gt $end ] && printf "\n>>>>>>> FAIL: Timeout when waiting for the load balancer ${lbid} ACTIVE\n" && exit -1
   done
 }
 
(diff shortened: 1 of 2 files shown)


lingxiankong:

tests/e2e/cloudprovider/test-lb-service.sh
```diff
@@ -109,6 +151,37 @@ spec:
 EOF
 }
 
+########################################################################
+## Name: create_deployment
```
ramineni, line 155: comments here needs to be updated I suppose

tests/e2e/cloudprovider/test-lb-service.sh
```diff
@@ -109,6 +151,37 @@ spec:
...
+## Desc: Makes sure the echoserver service Deployment is running
+## Params: None
+########################################################################
+function create_openstackcli_pod {
+    kubectl -n $NAMESPACE get pod | grep openstackcli | grep Running > /dev/null
+    if [[ $? -eq 1 ]]; then
```
ramineni, line 161: Is creating openstackcli pod required? Running openstack commands on VM where job is running is supported right?

lingxiankong: Currently, the lb-octavia CI job is running by talking to a k8s cluster running on catalyst cloud, for security reasons, the job has no openstack credential configured. The openstack credential is stored as a secret inside the cluster.

tests/e2e/cloudprovider/test-lb-service.sh
lingxiankong: Thanks, done.

tests/e2e/cloudprovider/test-lb-service.sh
ramineni: Ah, got it . Didnt notice k8s cluster is on catalyst cloud.

lingxiankong: Because citynetwork doesn't support octavia.

lingxiankong:

PR #1261, opened 2020-10-05
[cinder-csi-plugin] Do not ignore use-cloud config option

### LINKED ISSUE #1260: [cinder-csi-plugin] use-clouds config option is ignored
/kind bug

**What happened**:

In cinder csi driver config file there is the option "use-clouds" that allows to read credentials from `clouds.yaml` file:
https://github.com/kubernetes/cloud-provider-openstack/blob/94b377e1f4e59711b1b2835fbe6206b91e34657d/pkg/cloudprovider/providers/openstack/openstack.go#L219

When I set the value to true nothing happens and cinder csi driver doesn't read data from the given `clouds.yaml` file. It leads to the error:

> W1005 20:09:45.915538       1 main.go:108] Failed to GetOpenStackProvider: Get "/": unsupported protocol scheme ""

I expect that with UseClouds enabled in the config my credentials are taken from the `clouds.yaml` file

**How to reproduce it**:

1. Place a valid `clouds.yaml` file in /etc folder
2. Start the driver with the following config

    [Global]
    use-clouds = true
    clouds-file = /etc/clouds.yaml
    cloud = openstack

Fedosin:
Now use-cloud option is ignored in the driver's config.

This commit fixes this issue and allows the driver to read credentials from the given clouds.yaml file.

fixes #1260

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/openstack/fixtures/clouds.yaml
+++ b/pkg/csi/cinder/openstack/fixtures/clouds.yaml
@@ -1,0 +1,12 @@
+clouds:
+    openstack:
+      auth:
+        auth_url: "https://169.254.169.254/identity/v3"
+        username: "user"
+        password: "pass"
+        project_id: "c869168a828847f39f7f06edd7305637"
+        domain_id: "2a73b8f597c04551a0fdc8e95544be8a"
+      region_name: "RegionOne"
+      cacert: "fake-ca.crt"
+      interface: "public"
+      identity_api_version: 3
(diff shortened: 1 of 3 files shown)


lingxiankong:
/lgtm

Another proof that it's not a good idea to maintain a copy of openstack related code.

kayrus:
@ramineni @lingxiankong I think that this code must go into a separate package, as I mentioned in #993

lingxiankong:
> @ramineni @lingxiankong I think that this code must go into a separate package, as I mentioned in #993

Agree

PR #1266, opened 2020-10-06
[cinder-csi-plugin] handle a case when a volume was already resized

### LINKED ISSUE #1265: [cinder-csi-plugin]: "context deadline exceeded" on a resize action

/kind bug

**What happened**:

During the resize action the following error occurred: `resize volume pv-test-11894292-abbf-4200-953d-24247c081184 failed: rpc error: code = DeadlineExceeded desc = context deadline exceeded`

However the volume was resized and further attempts failed with an error: `resize volume pv-test-11894292-abbf-4200-953d-24247c081184 failed: rpc error: code = Internal desc = Could not resize volume "1d01da1e-2639-44f1-a3b0-c474391b3011" to size 2: Bad request with: [POST https://cinder/v3/168b54166ddd42c38e8316d85b50d5e8/volumes/1d01da1e-2639-44f1-a3b0-c474391b3011/action], error message: {"badRequest": {"message": "Invalid input received: New size for extend must be greater than current size. (current: 2, extended: 2).", "code": 400}}`

* `context deadline exceeded` should not happen
* further resize attempts should not fail, when the target volume size equals to the current one.

**How to reproduce it**:

n/a

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: 1.19
- OpenStack version:
- Others:

kayrus:

Allows cinder csi plugi to handle a volume, which was already resized.

fixes #1265

**Release note**:

```release-note
[cinder-csi-plugin] Added the size check for volume expansion.
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -547,7 +547,7 @@ func (cs *controllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 		return nil, status.Error(codes.OutOfRange, "After round-up, volume size exceeds the limit specified")
 	}
 
-	_, err := cs.Cloud.GetVolume(volumeID)
+	volume, err := cs.Cloud.GetVolume(volumeID)
 	if err != nil {
 		if cpoerrors.IsNotFound(err) {
 			return nil, status.Error(codes.NotFound, "Volume not found")
@@ -555,7 +555,16 @@ func (cs *controllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 		return nil, status.Error(codes.Internal, fmt.Sprintf("GetVolume failed with error %v", err))
 	}
 
-	err = cs.Cloud.ExpandVolume(volumeID, volSizeGB)
+	if volume.Size >= volSizeGB {
+		// a volume was already resized
+		klog.V(2).Infof("Volume %q has been already expanded to %d, requested %d", volumeID, volume.Size, volSizeGB)
+		return &csi.ControllerExpandVolumeResponse{
+			CapacityBytes:         int64(volume.Size * 1024 * 1024 * 1024),
+			NodeExpansionRequired: true,
+		}, nil
+	}
+
+	err = cs.Cloud.ExpandVolume(volumeID, volume.Status, volSizeGB)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, fmt.Sprintf("Could not resize volume %q to size %v: %v", volumeID, volSizeGB, err))
 	}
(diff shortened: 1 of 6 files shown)


pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -287,7 +287,13 @@ func (os *OpenStack) ExpandVolume(volumeID string, newSize int) error {
 		return err
 	}
 
-	createOpts := volumeexpand.ExtendSizeOpts{
+	if volume.Size == newSize {
```
ramineni, line 290: @kayrus  I think we could move this check https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/controllerserver.go#L558 , so when equal we dont need to call NodeExpandVolume , so we could return false?
2. We can check both equal and greater than , then we dont need to expand.

ramineni (changes requested):
@kayrus Thanks for the PR. Few comments inline.

pkg/csi/cinder/openstack/openstack_volumes.go
kayrus: Good point, but I don't like that volume details are fetched twice per action:
* https://github.com/kubernetes/cloud-provider-openstack/blob/8cb96459c236d114e9c5f7bfb0ba8fdce232f059/pkg/csi/cinder/controllerserver.go#L550
* https://github.com/kubernetes/cloud-provider-openstack/blob/8cb96459c236d114e9c5f7bfb0ba8fdce232f059/pkg/csi/cinder/openstack/openstack_volumes.go#L285

ramineni: +1, I noticed too. would be good if you could fix it as well.

kayrus:
@ramineni is it possible to merge without a test?

kayrus:
@ramineni ready

lingxiankong:
/lgtm

Thanks for the fix.

pkg/csi/cinder/controllerserver.go
```diff
@@ -547,15 +547,24 @@ func (cs *controllerServer) ControllerExpandVolume(ctx context.Context, req *csi
...
 		return nil, status.Error(codes.Internal, fmt.Sprintf("GetVolume failed with error %v", err))
 	}
 
-	err = cs.Cloud.ExpandVolume(volumeID, volSizeGB)
+	if volume.Size >= volSizeGB {
+		// a volume was already resized
+		klog.V(2).Infof("Volume %q has been already expanded to %d, requested %d", volumeID, volume.Size, volSizeGB)
+		return &csi.ControllerExpandVolumeResponse{
+			CapacityBytes:         int64(volume.Size * 1024 * 1024 * 1024),
+			NodeExpansionRequired: true,
```
ramineni, line 563: @kayrus wondering , if we require NodeExpansion here if its already resized ? could we set to false?

kayrus:
> wondering , if we require NodeExpansion here if its already resized ? could we set to false?
@ramineni there is a probability that NodeExpansion wasn't triggered due to a context deadline mentioned in #1265

jichenjc:
/lgtm
looks good

PR #1272, opened 2020-10-08
[cinder-csi-plugin] Add error handling for list fucntions before creating

takmatsu:
current implementation checks existing volumes before creating.
but if somehow list request is failed (due to temporal API fails or something), next mechanism is not handle correctly.
we should just return error when list operation is failed 
no-issue

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -81,6 +81,7 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	volumes, err := cloud.GetVolumesByName(volName)
 	if err != nil {
 		klog.V(3).Infof("Failed to query for existing Volume during CreateVolume: %v", err)
+		return nil, status.Error(codes.Internal, "Failed to check existing Volume")
 	}
 
 	if len(volumes) == 1 {
@@ -324,6 +325,7 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 	snapshots, _, err := cs.Cloud.ListSnapshots(filters)
 	if err != nil {
 		klog.V(3).Infof("Failed to query for existing Snapshot during CreateSnapshot: %v", err)
+		return nil, status.Error(codes.Internal, "Failed to check existing Snapshot")
 	}
 	var snap *ossnapshots.Snapshot
 
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -81,7 +81,7 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	volumes, err := cloud.GetVolumesByName(volName)
 	if err != nil {
 		klog.V(3).Infof("Failed to query for existing Volume during CreateVolume: %v", err)
-		return nil, status.Error(codes.Internal, "Failed to check existing Volume")
+		return nil, status.Error(codes.Internal, "Failed to get volumes")
 	}
 
 	if len(volumes) == 1 {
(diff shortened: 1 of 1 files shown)


ramineni:
>> but if somehow list request is failed (due to temporal API fails or something), next mechanism is not handle correctly.

@takmatsu  could you elaborate here what issue you are facing to understand how it can be handled better.
I think log the error is better and continue to createVolume as that call was just additional check we are doing .

jichenjc:
@takmatsu are you suggesting we should check the error and return based on error? 
so your case is `only` the list API fail and next API call will success? 
seems additoinal info will be helpful on this PR

takmatsu:
* First of all, `CreateVolume()` should be idempotent
https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume
```
This operation MUST be idempotent
```
* we have the possibility `CreateVolume()` for same volume multiple times
e.g) https://github.com/kubernetes-csi/external-provisioner/blob/080d35df20983a57cc0d1da514b1654822998e94/pkg/controller/controller.go#L1330
even if grpc timeout or not-directly-related-to cinder error, csi-provisioner recognize it as failed, and retry will be queued. (it is actually said as `ProvisioningInBackground`)
* cinder can create same name volume any times
id is unique, name is ok to duplicate

with above condition, we observed sometimes csi-provisioner says `Multiple volumes` error: https://github.com/kubernetes/cloud-provider-openstack/blob/403dcd188f20a1ac1a4a477d5b3cb9f7318b285d/pkg/csi/cinder/controllerserver.go#L95

IMHO, we shouldn't continue to create new volume if list is failed also from CSI spec side.

pkg/csi/cinder/controllerserver.go
```diff
@@ -81,6 +81,7 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	volumes, err := cloud.GetVolumesByName(volName)
 	if err != nil {
 		klog.V(3).Infof("Failed to query for existing Volume during CreateVolume: %v", err)
+		return nil, status.Error(codes.Internal, "Failed to check existing Volume")
```
lingxiankong, line 84: Suggestion for the error msg: `Failed to get volumes`

pkg/csi/cinder/controllerserver.go
```diff
@@ -324,6 +325,7 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 	snapshots, _, err := cs.Cloud.ListSnapshots(filters)
 	if err != nil {
 		klog.V(3).Infof("Failed to query for existing Snapshot during CreateSnapshot: %v", err)
+		return nil, status.Error(codes.Internal, "Failed to check existing Snapshot")
```
lingxiankong, line 328: Suggestion for the error msg: `Failed to get snapshots`

lingxiankong:
Well, personally I agree this change is helpful to raise exception earlier if error occurred before actually creating volume.

jichenjc:
@ramineni I think the error is 100% now (at least I tried >10 times now )
it's 413 error (exceed limit) and maybe we need consult openlab again

ramineni:
> * First of all, `CreateVolume()` should be idempotent
>   https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume
> 
> ```
> This operation MUST be idempotent
> ```
> 
> * we have the possibility `CreateVolume()` for same volume multiple times
>   e.g) https://github.com/kubernetes-csi/external-provisioner/blob/080d35df20983a57cc0d1da514b1654822998e94/pkg/controller/controller.go#L1330
>   even if grpc timeout or not-directly-related-to cinder error, csi-provisioner recognize it as failed, and retry will be queued. (it is actually said as `ProvisioningInBackground`)
> * cinder can create same name volume any times
>   id is unique, name is ok to duplicate
> 
> with above condition, we observed sometimes csi-provisioner says `Multiple volumes` error:
> 
> https://github.com/kubernetes/cloud-provider-openstack/blob/403dcd188f20a1ac1a4a477d5b3cb9f7318b285d/pkg/csi/cinder/controllerserver.go#L95
> 
> IMHO, we shouldn't continue to create new volume if list is failed also from CSI spec side.

@takmatsu Thanks for the details. I'm ok with this change.

jichenjc:
/approve

works for me

ramineni:

PR #1277, opened 2020-10-16
[magnum-auto-healer] Uses OpenStack instance ID from Node ProviderID

### LINKED ISSUE #1278: auto-healer: node MachineID is not always Openstack instance UUID

/kind bug

> /kind feature

**What happened**:

Currently, the magnum auto-healer works with the assumption that the node's MachineID == Openstack Instance UUID. However, that is not always the case, not all hypervisors pass the ``instance.uuid`` to the VM's SMBIOS as libvirt does, and not all OSes rely on the MachineID / ``/etc/machine-id`` (e.g.: Windows VMs).

The OpenStack instance UUID can also be found in the Node's ProviderID with the format:

    openstack:///openstack-uuid

We can use that instead. Additionally, OCCM already uses the ProviderID for the same reason.

``magnum-auto-healer`` should also work for Windows nodes and nodes that don't rely on MachineID / ``/etc/machine-id``.

**How to reproduce it**:

Trigger the auto-healing process for a Windows node.

**Environment**:
- OpenStack version:
- Others:

claudiubelu:

Currently, the magnum auto-healer works with the assumption that the node's MachineID == Openstack Instance UUID. However, that is not always the case, not all hypervisors pass the ``instance.uuid`` to the VM's SMBIOS as libvirt does, and not all OSes rely on the MachineID / ``/etc/machine-id`` (e.g.: Windows VMs).

The OpenStack instance UUID can also be found in the Node's ProviderID with the format:

    openstack:///openstack-uuid

We can use that instead. Additionally, OCCM already uses the ProviderID for the same reason, so this will make it more consistent.

fixes #1278

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/autohealing/cloudprovider/openstack/provider.go
+++ b/pkg/autohealing/cloudprovider/openstack/provider.go
@@ -253,15 +253,14 @@ func (provider OpenStackCloudProvider) Repair(nodes []healthcheck.NodeInfo) erro
 	if isWorkerNode {
 		for _, n := range nodes {
 			nodesToReplace := sets.NewString()
-			machineID := uuid.Parse(n.KubeNode.Status.NodeInfo.MachineID)
-			if machineID == nil {
+			serverID := strings.TrimPrefix(n.KubeNode.Spec.ProviderID, "openstack:///")
+			if serverID == "" {
 				log.Warningf("Failed to get the correct server ID for server %s", n.KubeNode.Name)
 				continue
 			}
-			serverID := machineID.String()
 
 			if err := provider.waitForServerDetachVolumes(serverID, 30*time.Second); err != nil {
-				log.Warningf("Failed to detach volumes from server %s, error: %v", serverID, err)
+				log.Warningf("Failed to shutdown the server %s, error: %v, continue handling...", serverID, err)
 			}
 
 			if err := provider.waitForServerPoweredOff(serverID, 30*time.Second); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/autohealing/cloudprovider/openstack/provider.go
```diff
@@ -253,15 +253,14 @@ func (provider OpenStackCloudProvider) Repair(nodes []healthcheck.NodeInfo) erro
...
+			serverID := strings.TrimPrefix(n.KubeNode.Spec.ProviderID, "openstack:///")
+			if serverID == "" {
 				log.Warningf("Failed to get the correct server ID for server %s", n.KubeNode.Name)
 				continue
 			}
-			serverID := machineID.String()
 
 			if err := provider.waitForServerDetachVolumes(serverID, 30*time.Second); err != nil {
-				log.Warningf("Failed to detach volumes from server %s, error: %v", serverID, err)
+				log.Warningf("Failed to shutdown the server %s, error: %v, continue handling...", serverID, err)
```
jichenjc, line 263: why make this change? looks like this is for detach volume..

pkg/autohealing/cloudprovider/openstack/provider.go
```diff
@@ -253,15 +253,14 @@ func (provider OpenStackCloudProvider) Repair(nodes []healthcheck.NodeInfo) erro
 	if isWorkerNode {
 		for _, n := range nodes {
 			nodesToReplace := sets.NewString()
-			machineID := uuid.Parse(n.KubeNode.Status.NodeInfo.MachineID)
-			if machineID == nil {
+			serverID := strings.TrimPrefix(n.KubeNode.Spec.ProviderID, "openstack:///")
```
adisky, line 256: can we make a util function to get the instance id here? should we consider fall back on `NodeInfo.MachineID` in case `ProviderID` not set?

PR #1290, opened 2020-10-26
make batch member update configurable

hamzazafar:
make batch member update configurable

Octavia supports batch member updates, however some provider drivers for Octavia do not support batch update of pool members. In that case, the pool members are added/removed by occm. 

`batch-member-update` can be set in cloud-config file.

Batch member update is enabled by default for Octavia.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/using-openstack-cloud-controller-manager.md
+++ b/docs/using-openstack-cloud-controller-manager.md
@@ -201,7 +201,10 @@ Although the openstack-cloud-controller-manager was initially implemented with N
 
 * `cascade-delete`
   Determines whether or not to perform cascade deletion of load balancers. Default: true.
-  
+
+* `batch-member-update`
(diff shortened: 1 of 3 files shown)


hamzazafar:
/assign @lingxiankong  

@lingxiankong  can you please review. Thanks

lingxiankong:
Hi @hamzazafar , I am not sure which driver you are using, but from Octavia official API document (https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=batch-update-members-detail#batch-update-members) I didn't see anything mentioned that different drivers may have different implementations.

As an official cloud provider that supports official openstack API, I don't think we should support any 'non-standard' openstack API. So, rather than adding an extra config option to OCCM, could you please provide more information about your driver and maybe talk to Octavia team about the API support?

/hold

Issue #1280, opened 2020-10-19
[manila-csi-plugin] Add an option to set share metadata when creating a new volume


gman0:
Add an option to set share metadata when creating a new volume. This was implemented for the (now deprecated) Manila external provisioner, but not for manila-csi.

Related PR: https://github.com/kubernetes/cloud-provider-openstack/pull/782

/assign

PR #1291, opened 2020-10-27
[manila-csi-plugin] Added appendShareMetadata volume parameter

gman0:

This PR adds the ability to set user-defined metadata on newly created shares.

fixes https://github.com/kubernetes/cloud-provider-openstack/issues/1280

**Release note**:

```release-note
[manila-csi-plugin] Added appendShareMetadata volume parameter
```

### DIFF
--- a/pkg/csi/manila/volumesource.go
+++ b/pkg/csi/manila/volumesource.go
@@ -17,6 +17,8 @@ limitations under the License.
 package manila
 
 import (
+	"encoding/json"
+
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/gophercloud/gophercloud/openstack/sharedfilesystems/v2/shares"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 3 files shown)


gman0:
PTAL @gouthampacha @tombarron @ramineni

pkg/csi/manila/volumesource.go
```diff
@@ -124,3 +138,12 @@ func (volumeFromSnapshot) create(req *csi.CreateVolumeRequest, shareName string,
 
 	return share, err
 }
+
+func parseStringMapFromJson(data string) (m map[string]string, err error) {
+	if data == "" {
+		return
+	}
```
jichenjc, line 145: you mentioned it must be valid `Must be a string with a valid JSON object.`
but you allow '' here .. is it on purpose?

gman0: The "Required" column in the docs says it's not required, so it can be either empty or must be a valid JSON object. I'll try to formulate the field description better so that this behavior is reflected there. Thanks.

gman0:
The "append" in `appendShareMetadata` means that if there is any metadata that is set to the share by the driver (there's currently none), `appendShareMetadata` would take precedence, and in case of conflicting metadata keys, keys from this parameter would override those entries set by the driver. If anyone can suggest a better name for this parameter, let me know in the comments please!

jichenjc:
/lgtm

thanks~

tombarron (approved):

gouthampacha (approved):
/lgtm

gouthampacha (approved):
/assign @ramineni 

Hi Anusha, Can you help us with this PR :)

PR #1292, opened 2020-10-28
[k8s-keystone-auth] Fix authorization error

### LINKED ISSUE #1308: [k8s-keystone-auth]Policy rule matching is flawed and The authorization process needs to be optimized
/kind bug

**What happened**:
1. When the namespace in the authorization request is nil, getAllowed(nsDef, ns) will always pass the verification.
2. When the kubernetes cluster turns on anonymous authentication, the authentication will always pass, but the authorization request will not include the extra field, and keystone-auth will still perform authorization verification for such requests.

1. the policy rules like `exampe/daemonsets:*` will pass the request which its namespace is nil like `namespace  , resource: daemonsets , verb: list`, so the client will get all of daemonsets in the cluster.
2. Every authenticated request will go through a complete authorization, which is not necessary

Bevisy:
Signed-off-by: bevisy <binbin36520@gmail.com>

1. Add user authorization information existence judgment
2. Repair request authorization error without namespace

fixes #1308 

**Release note**:

```release-note
[k8s-keystone-auth] Fix authorization error
```

### DIFF
--- a/pkg/identity/keystone/authorizer.go
+++ b/pkg/identity/keystone/authorizer.go
@@ -124,6 +124,11 @@ func resourcePermissionAllowed(permissionSpec map[string][]string, attr authoriz
 		nsDef := strings.ToLower(strings.TrimSpace(keyList[0]))
 		resDef := strings.ToLower(strings.TrimSpace(keyList[1]))
 
+		// When request without namespace, only policy which nsDef is equal to "*" could be executed.
+		if nsDef != "*" && ns == "" {
+			continue
+		}
+
 		allowedNamespaces, err := getAllowed(nsDef, ns)
 		if err != nil {
 			continue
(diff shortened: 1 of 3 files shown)


jichenjc:
@Bevisy please follow the template of PR (you can refer to other PRs) to fix the issue of non-label

lingxiankong:
What issue are you fixing? Could you please submit an issue for that and add that reference in the PR?

/hold

Bevisy:
@lingxiankong 
Issue link has been added, please review.

lingxiankong:
Change looks good, please add some tests to cover the issue you are fixing.

lingxiankong:
We need to cover the test cases below to cover the issue submitted:

1. A user can't access the namespace scoped resources when no namespace is specified in the request.
2. If no `users` defined in a policy, the resource access should be denied.

Bevisy:
@lingxiankong already new test cases to cover the issue.

Issue #1293, opened 2020-10-28
[cinder-csi-plugin] Not possible to use cinder CSI with multiple AZs


zetaab:
**Is this a BUG REPORT or FEATURE REQUEST?**: Not possible to use multiple AZs with cinder CSI. I have computes in zones called helpo, helpa, helpe. Also I do have volume zones called helpo, helpa, helpe. Similar setup that AWS has, volumes can be mounted inside zone only. This means that PV object should have zone labels added. 

I am using following storageclass:

```
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    storageclass.beta.kubernetes.io/is-default-class: "true"
  labels:
    k8s-addon: storage-openstack.addons.k8s.io
  name: default
provisioner: cinder.csi.openstack.org
reclaimPolicy: Delete
volumeBindingMode: WaitForFirstConsumer
```

However, this do have 2 problems: 1) it is creating volume to zone called nova, I do not have zone called nova at all. 2) it does not add labels to PV (which it should, otherwise provisioning does not work correctly between pod and pvc). Second option is to use node affinity in pv, but its not using that either

So instead this should work in similar way than aws-ebs is doing, it should check the zones where we have instances and provision the volume in those. Now it just assumes that we do have az specified in storageclass config (or if not it will use nova). Which is incorrect

/kind bug

**What happened**:

  Warning  FailedAttachVolume  59s (x9 over 3m7s)  attachdetach-controller                 AttachVolume.Attach failed for volume "pvc-83bce05d-b5b5-4dcd-aa71-2794115cbeff" : rpc error: code = Internal desc = [ControllerPublishVolume] Attach Volume failed with error failed to attach 789ea232-90eb-4208-94f6-1f2db95ac2cd volume to ab612a23-9dc5-4e23-a385-6132e8b9a8a4 compute: Bad request with: [POST https://xxxx.com:13774/v2.1/servers/ab612a23-9dc5-4e23-a385-6132e8b9a8a4/os-volume_attachments], error message: {"badRequest": {"code": 400, "message": "Invalid volume: Instance ab612a23-9dc5-4e23-a385-6132e8b9a8a4 and volume 789ea232-90eb-4208-94f6-1f2db95ac2cd are not in the same availability_zone. Instance is in helpa. Volume is in nova"}}

**What you expected to happen**:

I expect that we should have way to use multiple AZs in case of computes and volumes. 

[...]

PR #1296, opened 2020-10-29
[cinder-csi-plugin] update topology feature docs

ramineni:
https://github.com/kubernetes-csi/external-provisioner/blob/v1.6.0/pkg/features/features.go#L38 , Topology feature doesnt seem to be enabled by default. So updating the docs.

xref: #1293
fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/cinder-csi-plugin/features.md
+++ b/docs/cinder-csi-plugin/features.md
@@ -29,7 +29,7 @@ This feature enables driver to consider the topology constraints while creating 
 
 * Supported topology keys:
   `topology.cinder.csi.openstack.org/zone` : Availability by Zone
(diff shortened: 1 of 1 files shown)


lingxiankong:
Hmm...I found github automatically add `approved` label in the PRs submitted by the person who has approved permission.

jichenjc:
@lingxiankong seems it's added recently, we should remove such auto label IMHO

zetaab:
Thank you @ramineni

ramineni:
> @lingxiankong seems it's added recently, we should remove such auto label IMHO

+1 , @lingxiankong any idea we could do that?

PR #1307, opened 2020-11-05
[cinder-csi-plugin] support ignore-volume-az

### LINKED ISSUE #1300: [cinder-csi-plugin] IgnoreVolumeAZ does not work for volumes with Topology
**Is this a BUG REPORT or FEATURE REQUEST?**: bug

 /kind bug

**What happened**:  I am testing this new topology thing in two different openstack. In one openstack I have setup where I have 3 compute azs (zone-1, zone-2, zone-3) and 1 volume az (nova). I should be able to use IgnorevolumeAZ option (https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/cloudprovider/providers/openstack/openstack.go#L152) somehow. However, when topology is enabled using featureflag (in next version it will be enabled for ALL so this will break things for everybody who is using ignorevolumeaz) the nova zone for volumes cannot be used. 

**What you expected to happen**: I expect that I could still use csi provider after topology feature.

**How to reproduce it**: 

1) OpenStack region with zone names that does not match between compute and volume, you have that if you have been using ignorevolumeaz option.
2) enable topology featureflag for provisioner (and use like 2.0.3 version)
3) use storageclass

```
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    storageclass.beta.kubernetes.io/is-default-class: "true"
  name: csi-sc-cinderplugin
parameters:
  availability: nova
  type: solidfire
reclaimPolicy: Delete
volumeBindingMode: Immediate
```
4) try to create volume

I have loglevel 4 in cinder-csi-plugin and the result is this:

```
I1030 15:05:39.988102       1 controllerserver.go:44] CreateVolume: called with args {Name:pvc-6084d54f-c2d9-43af-b7af-2ac62f8bd33d CapacityRange:required_bytes:17179869184  VolumeCapabilities:[mount:<fs_type:"ext4" > access_mode:<mode:SINGLE_NODE_WRITER > ] Parameters:map[availability:nova type:solidfire] Secrets:map[] VolumeContentSource:<nil> AccessibilityRequirements:requisite:<segments:<key:"topology.cinder.csi.openstack.org/zone" value:"zone-2" > > requisite:<segments:<key:"topology.cinder.csi.openstack.org/zone" value:"zone-3" > > requisite:<segments:<key:"topology.cinder.csi.openstack.org/zone" value:"zone-1" > > preferred:<segments:<key:"topology.cinder.csi.openstack.org/zone" value:"zone-1" > > preferred:<segments:<key:"topology.cinder.csi.openstack.org/zone" value:"zone-2" > > preferred:<segments:<key:"topology.cinder.csi.openstack.org/zone" value:"zone-3" > >  XXX_NoUnkeyedLiteral:{} XXX_unrecognized:[] XXX_sizecache:0}
```

The problem is here https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/controllerserver.go#L68-L76 because, now the topologies has even some data. The availability parameter is just ignored. It should not be ignored if we have ignoreaz use-case.

ramineni:

fixes #1300 

**Release note**:

```release-note
[cinder-csi-plugin] support `ignore-volume-az` config parameter.
```

### DIFF
--- a/docs/cinder-csi-plugin/using-cinder-csi-plugin.md
+++ b/docs/cinder-csi-plugin/using-cinder-csi-plugin.md
@@ -68,7 +68,7 @@ These configuration options pertain to block storage and should appear in the `[
 * `rescan-on-resize`
   Optional. Set to `true`, to rescan block device and verify its size before expanding the filesystem. Not all hypervizors have a /sys/class/block/XXX/device/rescan location, therefore if you enable this option and your hypervizor doesn't support this, you'll get a warning log on resize event. It is recommended to disable this option in this case. Defaults to `false`
 * `ignore-volume-az`
-  Optional. When `Topology` feature enabled, by default, PV volume node affinity is populated with volume accessible topology, which is volume AZ. But, some of the openstack users do not have compute zones named exactly the same as volume zones. This might cause pods to go in pending state as no nodes available in volume AZ. Enabling `ignore-volume-az=true`, ignores volumeAZ and schedules on any of the available node AZ.
+  Optional. When `Topology` feature enabled, by default, PV volume node affinity is populated with volume accessible topology, which is volume AZ. But, some of the openstack users do not have compute zones named exactly the same as volume zones. This might cause pods to go in pending state as no nodes available in volume AZ. Enabling `ignore-volume-az=true`, ignores volumeAZ and schedules on any of the available node AZ. Default `false`.
 
 ### Metadata
 These configuration options pertain to metadata and should appear in the `[Metadata]` section of the `$CLOUD_CONFIG` file.
--- a/docs/cinder-csi-plugin/using-cinder-csi-plugin.md
+++ b/docs/cinder-csi-plugin/using-cinder-csi-plugin.md
@@ -67,6 +67,8 @@ These configuration options pertain to block storage and should appear in the `[
   Optional. To configure maximum volumes that can be attached to the node. Its default value is `256`.
 * `rescan-on-resize`
   Optional. Set to `true`, to rescan block device and verify its size before expanding the filesystem. Not all hypervizors have a /sys/class/block/XXX/device/rescan location, therefore if you enable this option and your hypervizor doesn't support this, you'll get a warning log on resize event. It is recommended to disable this option in this case. Defaults to `false`
+* `ignore-volume-az`
+  Optional. When `Topology` feature enabled, by default, PV volume node affinity is populated with volume accessible topology, which is volume AZ. But, some of the openstack users do not have compute zones named exactly the same as volume zones. This might cause pods to go in pending state as no nodes available in volume AZ. Enabling `ignore-volume-az=true`, ignores volumeAZ and schedules on any of the available node AZ.
 
 ### Metadata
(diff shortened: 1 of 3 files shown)


zetaab:
@ramineni @jichenjc there are now two prs that are trying to solve this same issue. However, this PR uses reading the ignore-volume-az value from config, which means that we are backwards compatbile. So if this PR works, I would prefer merging this PR

jichenjc:
@ramineni @zetaab works for me if this is better , I can close #1300 when this is merged

zetaab:
I have verified this PR against openstack which compute zones does not match to volume zone, and it works. However, I will still verify this in openstack which compute zones match to volume zones (and using topology).

zetaab (approved):
This works like should! awesome, thank you for both of you @ramineni and @jichenjc but like I said already: this PR do have backward compatibility using cloud config

/lgtm

zetaab:
how to restart that one failing test?

jichenjc:

@zetaab due to https://github.com/theopenlab/openlab/issues/603 , we are blocked now
I think we can overwrite this error and force merge as it's not reltaed issue.. let me know your opinion

jichenjc:
we can try following to overwrite, but I am not sure it's a good way ..
`override openlab/cloud-provider-openstack-acceptance-test-csi-cinder`

docs/cinder-csi-plugin/using-cinder-csi-plugin.md
```diff
@@ -67,6 +67,8 @@ These configuration options pertain to block storage and should appear in the `[
   Optional. To configure maximum volumes that can be attached to the node. Its default value is `256`.
 * `rescan-on-resize`
   Optional. Set to `true`, to rescan block device and verify its size before expanding the filesystem. Not all hypervizors have a /sys/class/block/XXX/device/rescan location, therefore if you enable this option and your hypervizor doesn't support this, you'll get a warning log on resize event. It is recommended to disable this option in this case. Defaults to `false`
+* `ignore-volume-az`
+  Optional. Set to `true`, incase volume AZ is different from node AZ and to avoid populating accessible topologies to volume AZ.
```
lingxiankong, line 71: Can we have a more detailed description of this option? I'm a little confused about how to configure this option properly, e.g. What's the default value? In which case it should be set to true?

lingxiankong (changes requested):
/approve cancel

zetaab:
@ramineni can you add more detailed description to documentation? You can perhaps take a look legacy cinder if that contains helpful information

docs/cinder-csi-plugin/using-cinder-csi-plugin.md
ramineni: @lingxiankong @zetaab updated. Let me know if its bit more clear now. Thanks.

jichenjc:
the wording looks good to me
/lgtm

lingxiankong:
/lgtm

@zetaab Please approve if it looks good to you as well.

jichenjc:
this has been here for a while and already tested works well
so we can use follow up to fix any remaining issue
/lgtm
/approve

Issue #1322, opened 2020-11-30
[k8s-keystone-auth] doesn't works with psp policies


niedbalski:
**What happened**:

Using the following pod security policy https://pastebin.ubuntu.com/p/xT7ZwsPgPH/ , container with
this deployment  https://github.com/kubernetes/cloud-provider-openstack/blob/master/examples/webhook/keystone-deployment.yaml,won't start as is missing the runAsUser security 

**What you expected to happen**:

Should work with PSP policies

**Environment**:
- OpenStack version: Ussuri

lingxiankong:
Could you please check your paste to correct the format? I'm not very clear what's defined in the PodSecurityPolicy.

jichenjc:
not expert in this area, just curious 
is `https://pastebin.ubuntu.com/p/xT7ZwsPgPH/ ` contains default PSP for all pods? 
otherwise, how do we expect the manifest be modified to adopt for the unique things pasted above? 
@niedbalski

rptaylor:
Yes it would be great if the charts for the CSI drivers included defined PSPs that allow them to run.
On clusters which enable the PSP admission controller for enhanced security, the CSI deployment can only run if it is installed in the kube-system namespace, which has a PSP that allows everything (incl. privileged pods). That is not ideal for security, or for keeping different services isolated in separate namespaces.

Here is an example of the errors seen when trying to deploy openstack-cinder-csi chart v1.2.14:
```
$ kubectl -n openstack-cinder-csi get events
LAST SEEN   TYPE      REASON         OBJECT                                              MESSAGE
3m20s       Warning   FailedCreate   statefulset/openstack-cinder-csi-controllerplugin   create Pod openstack-cinder-csi-controllerplugin-0 in StatefulSet openstack-cinder-csi-controllerplugin failed error: pods "openstack-cinder-csi-controllerplugin-0" is forbidden: PodSecurityPolicy: unable to admit pod: [spec.volumes[2]: Invalid value: "hostPath": hostPath volumes are not allowed to be used]
[...]

PR #1323, opened 2020-11-30
[k8s-keystone-auth] Set the runAsUser: nobody security context.

niedbalski:
In order to allow PSP policies, add security context to the example deployment.

Reference: #1322 

Signed-off-by: Jorge Niedbalski <jorge.niedbalski@canonical.com>

### DIFF
--- a/examples/webhook/keystone-deployment.yaml
+++ b/examples/webhook/keystone-deployment.yaml
@@ -35,6 +35,8 @@ spec:
               readOnly: true
           ports:
             - containerPort: 8443
+          securityContext:
+            runAsUser: 65534
       volumes:
       - name: certs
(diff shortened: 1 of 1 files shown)


niedbalski:
@lingxiankong please review

lingxiankong:
The manifest here https://github.com/kubernetes/cloud-provider-openstack/blob/master/examples/webhook/keystone-deployment.yaml is just an example of how to deploy keystone-auth service, the actual deployment varies between different environments. Please don't rely on that for production without modification.

If you are looking for a production deployment, adding a helm chart [here](https://github.com/kubernetes/cloud-provider-openstack/tree/master/charts) would be a good idea.

PR #1337, opened 2020-12-07
[cinder-csi-plugin] Switch CSI sidecar image repository to gcr

ialidzhikov:
Ref https://github.com/kubernetes-csi/csi-release-tools/issues/86

**Release note**:

```release-note
[cinder-csi-plugin] Use GCR images for CSI sidecar containers.
```

### DIFF
--- a/charts/cinder-csi-plugin/values.yaml
+++ b/charts/cinder-csi-plugin/values.yaml
@@ -5,32 +5,32 @@ timeout: 3m
 csi:
   attacher:
     image:
-      repository: quay.io/k8scsi/csi-attacher
+      repository: k8s.gcr.io/sig-storage/csi-attacher
       tag: v2.2.1
       pullPolicy: IfNotPresent
   provisioner:
     image:
-      repository: quay.io/k8scsi/csi-provisioner
+      repository: k8s.gcr.io/sig-storage/csi-provisioner
       tag: v1.6.1
       pullPolicy: IfNotPresent
   snapshotter:
     image:
-      repository: quay.io/k8scsi/csi-snapshotter
+      repository: k8s.gcr.io/sig-storage/csi-snapshotter
       tag: v2.1.2
       pullPolicy: IfNotPresent
   resizer:
     image:
-      repository: quay.io/k8scsi/csi-resizer
+      repository: k8s.gcr.io/sig-storage/csi-resizer
       tag: v0.5.1
       pullPolicy: IfNotPresent
   livenessprobe:
     image:
-      repository: quay.io/k8scsi/livenessprobe
+      repository: k8s.gcr.io/sig-storage/livenessprobe
       tag: v2.1.0
       pullPolicy: IfNotPresent
   nodeDriverRegistrar:
     image:
-      repository: quay.io/k8scsi/csi-node-driver-registrar
+      repository: k8s.gcr.io/sig-storage/csi-node-driver-registrar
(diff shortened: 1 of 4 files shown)


gouthampacha:
Hi @ialidzhikov - We've two CSI drivers in this repository, manila-csi being the other one [1][2]; so this change could update both of them at the same time. However, just to be sure - will the CSI container images no longer get published to quay.io? Or is this a temporary change?

[1] https://github.com/kubernetes/cloud-provider-openstack/tree/master/charts/manila-csi-plugin
[2] https://github.com/kubernetes/cloud-provider-openstack/tree/master/manifests/manila-csi-plugin

ialidzhikov:
> We've two CSI drivers in this repository, manila-csi being the other one [1][2]; so this change could update both of them at the same time

Currently one of the sidecar images for manilla-csi (I think it was the external-provisioner image) is outdated and this version is only available in quay. If you don't have objections I can file a separate PR for manilla-csi as I will need also to update the version of the sidecar container. 

> However, just to be sure - will the CSI container images no longer get published to quay.io? Or is this a temporary change?

As stated in https://github.com/kubernetes-csi/csi-release-tools/issues/86#issuecomment-736719937, my understanding is that the push to quay will be dropped at some point in the time. For now I see that the recent images are also pushed to quay (probably for backwards compatibility reasons) but I think that soon the sidecar images will be maintained only in gcr. You can also see that the advertised images in the `README.md`s or in the releases are the gcr ones. @msau42 , @pohly feel free to add more details or correct me if I am wrong. : )

gouthampacha:
> > We've two CSI drivers in this repository, manila-csi being the other one [1][2]; so this change could update both of them at the same time
> 
> Currently one of the sidecar images for manilla-csi (I think it was the external-provisioner image) is outdated and this version is only available in quay. If you don't have objections I can file a separate PR for manilla-csi as I will need also to update the version of the sidecar container.

That would be great - I've no objections. Thank you for submitting this change, @ialidzhikov 

/cc @gman0 @tombarron

charts/cinder-csi-plugin/values.yaml
```diff
@@ -5,32 +5,32 @@ timeout: 3m
...
   attacher:
     image:
-      repository: quay.io/k8scsi/csi-attacher
+      repository: k8s.gcr.io/sig-storage/csi-attacher
       tag: v2.2.1
       pullPolicy: IfNotPresent
   provisioner:
     image:
-      repository: quay.io/k8scsi/csi-provisioner
+      repository: k8s.gcr.io/sig-storage/csi-provisioner
```
jichenjc, line 13: this works 
```
# docker pull k8s.gcr.io/sig-storage/csi-provisioner:v1.6.1
v1.6.1: Pulling from sig-storage/csi-provisioner
Digest: sha256:3b9b5ae677fd74de3b1cde5d3606a970f615b598730aa003c0f52e755c78862e
Status: Downloaded newer image for k8s.gcr.io/sig-storage/csi-provisioner:v1.6.1
k8s.gcr.io/sig-storage/csi-provisioner:v1.6.1

```

jichenjc:
just curious , should we update the `release notes` about the image modification ? @ialidzhikov

ialidzhikov:
> just curious , should we update the `release notes` about the image modification ? @ialidzhikov

I added a release note.

ramineni:
Thanks.
/lgtm

PR #1338, opened 2020-12-07
[cinder-csi-plugin] Add support for PSP and ability to set priority class in the helm chart

jlyheden:
Makes it possible to deploy the cinder-csi-plugin helm chart in Kubernetes clusters that has the Pod Security Policy admission controller enabled. Also allows the user to specify priority class for the statefulset and daemonset pods.

N/A

The provided pod security policy was inspired by the full access policy example from K8S docs:  https://kubernetes.io/docs/concepts/policy/pod-security-policy/#example-policies but this might be possible to restrict further?

**Release note**:
```release-note
NONE
```

### DIFF
--- a/charts/cinder-csi-plugin/values.yaml
+++ b/charts/cinder-csi-plugin/values.yaml
@@ -72,3 +72,12 @@ storageClass:
   retain:
     isDefault: false
     allowVolumeExpansion: true
+
+priorityClassName:
+  controllerPlugin: ""
+  nodePlugin: ""
+
+podSecurityPolicy:
+  enabled: false
+  existingPolicyName: ""
+  create: false
--- a/charts/cinder-csi-plugin/values.yaml
+++ b/charts/cinder-csi-plugin/values.yaml
@@ -78,6 +78,12 @@ priorityClassName:
   nodePlugin: ""
 
 podSecurityPolicy:
+  # Setting this to true will create a role and rolebinding granting the service accounts access to use the selected pod security policy
   enabled: false
+
+  # Leave empty if you want the helm chart to create a pod security policy for you (and set create to true)
+  # If not empty this should be the name of a pre-existing pod security policy resource deployed in the cluster (and set create to false)
   existingPolicyName: ""
+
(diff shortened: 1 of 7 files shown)


charts/cinder-csi-plugin/values.yaml
```diff
@@ -75,3 +75,12 @@ storageClass:
   retain:
     isDefault: false
     allowVolumeExpansion: true
+
+priorityClassName:
+  controllerPlugin: ""
+  nodePlugin: ""
+
+podSecurityPolicy:
```
lingxiankong, line 83: Please add some comments to those options (especially for `existingPolicyName` and `create `) here.

charts/cinder-csi-plugin/values.yaml
```diff
@@ -75,3 +75,12 @@ storageClass:
...
+  enabled: false
```
jichenjc, line 84: just curious, is this for Helm only or both for manifest ? or should we update 
https://github.com/kubernetes/cloud-provider-openstack/tree/master/manifests/cinder-csi-plugin
as well?

lingxiankong: I don't think it's appropriate to add this in the yaml manifest as this is totally optional.

charts/cinder-csi-plugin/values.yaml
jlyheden: Added comments in the latest commit

charts/cinder-csi-plugin/values.yaml
lingxiankong: yeah, basically Helm should be more flexible, and manifest is just for testing purpose IMO.

lingxiankong:
Please fix the errror "Chart version not ok. Needs a version bump"

rptaylor:
https://github.com/kubernetes/cloud-provider-openstack/pull/1561  already added priorityClass.  PSP would be nice to have too though.

lingxiankong:
> #1561 already added priorityClass. PSP would be nice to have too though.

PSP is deprecated and will be removed in future release.

PR #1341, opened 2020-12-08
Upgrade k/k dependency to v1.20

### LINKED ISSUE #1080: [occm] Investigate to remove vendored code from occm

ramineni:
This PR updates k/k dependency to 1.20

fixes #1080

**Release note**:

```release-note
NONE
```


cmd/openstack-cloud-controller-manager/main.go
```diff
@@ -101,21 +95,30 @@ the cloud specific control loops shipped with Kubernetes.`,
...
 
-			if err := app.Run(c.Complete(), wait.NeverStop); err != nil {
+			cloudconfig := c.Complete().ComponentConfig.KubeCloudShared.CloudProvider
+			cloud, err := cloudprovider.InitCloudProvider(cloudconfig.Name, cloudconfig.CloudConfigFile)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "%v\n", err)
+			}
+			if cloud == nil {
+				fmt.Fprintf(os.Stderr, "%v\n", err)
+			}
```
jichenjc, line 111: I assume we should error out here and line 107?

pkg/cloudprovider/.import-restrictions
```diff
@@ -7,7 +7,8 @@
 				"k8s.io/cloud-provider",
 				"k8s.io/component-base",
 				"k8s.io/klog",
-				"k8s.io/utils"
+				"k8s.io/utils",
+                                "k8s.io/mount-utils"
```
jichenjc, line 11: intent seems can be updated

cmd/openstack-cloud-controller-manager/main.go
ramineni: right. Ill update. Thanks

ramineni:
@jichenjc There is some issue , we have errors in ccm logs , unable to contact api server , still not exactly sure what is the issue

jichenjc:
> @jichenjc There is some issue , we have errors in ccm logs , unable to contact api server , still not exactly sure what is the issue.
> strange happening only for keystone job. @lingxiankong could you have a look?

I think in general we need have more logs including 
devstack, all pods log etc ,will take a look at what I can do after holiday 
it should be something like 

https://logs.openlabtesting.org/logs/12/712/7456fdaa676ab2c2da8b71dc79d204d8519a0107/check/cluster-api-provider-openstack-current-acceptance-test-v1.18/3784683/logs/_artifacts/logs/

https://logs.openlabtesting.org/logs/12/712/7456fdaa676ab2c2da8b71dc79d204d8519a0107/check/cluster-api-provider-openstack-current-acceptance-test-v1.18/3784683/logs/

so we will have more info

lingxiankong:
I'm looking at this issue

pkg/cloudprovider/.import-restrictions
ramineni: seems indent is correct when u open raw file

ramineni:
@jichenjc could you review? I guess we can cut release branch , once this is merged

jichenjc:
> @jichenjc could you review? I guess we can cut release branch , once this is merged

sorry, just saw this : ) 
glad it's merged already

PR #1344, opened 2020-12-09
[occm] Add handling for supported LBProvider

### LINKED ISSUE #1332: [occm] Octavia OVN Provider create listener error
/kind bug

**What happened**:
Setting `lb-provider=ovn` works, and allows the user to provision a Octavia OVN LoadBalancer, but breaks while creating the Listener. The configuration looks like so:

```
[LoadBalancer]
lb-provider=ovn
use-octavia=true
floating-network-id=180941c5-9e82-41c7-b64d-6a57302ec211
```

However, it fails to create the Listener with the following error:
```
I1205 11:33:24.169540       1 event.go:291] "Event occurred" object="awx/awx-web-lb" kind="Service" apiVersion="v1" type="Normal" reason="EnsuringLoadBalancer" message="Ensuring load balancer"
E1205 11:33:24.517822       1 controller.go:275] error processing service awx/awx-web-lb (will retry): failed to ensure load balancer: failed to create listener for loadbalancer 6c670e04-7c66-4f1a-9a01-9ad2ba7e2a92: Expected HTTP response code [] when accessing [POST https://openstack.bne-home.net:13876/v2.0/lbaas/listeners], but got 501 instead
{"faultcode": "Server", "faultstring": "Provider 'ovn' does not support a requested option: OVN provider does not support allowed_cidrs option", "debuginfo": null}
I1205 11:33:24.518078       1 event.go:291] "Event occurred" object="awx/awx-web-lb" kind="Service" apiVersion="v1" type="Warning" reason="SyncLoadBalancerFailed" message="Error syncing load balancer: failed to ensure load balancer: failed to create listener for loadbalancer 6c670e04-7c66-4f1a-9a01-9ad2ba7e2a92: Expected HTTP response code [] when accessing [POST https://openstack.bne-home.net:13876/v2.0/lbaas/listeners], but got 501 instead\n{\"faultcode\": \"Server\", \"faultstring\": \"Provider 'ovn' does not support a requested option: OVN provider does not support allowed_cidrs option\", \"debuginfo\": null}"
```

When a operator sets OVN as the provider, we will need to adjust the API request sent to Octavia to ensure the request is valid.

**How to reproduce it**:
1. Ensure you have Octavia OVN installed on OpenStack;
2. Add the above mentioned options to the `[LoadBalancer]` section in `cloud.conf`
3. Create a Service with `type: LoadBalancer`
4. Observe the LoadBalancer itself is created, but it will fail while creating the listener, because it's trying to use the `allowed_cidr` option which is not supported by the OVN provider

It looks like there are efforts to support this in the `ovn_octavia_provider`:
https://opendev.org/openstack/ovn-octavia-provider/src/branch/master/ovn_octavia_provider/driver.py#L61-L69

But this breaks for Kubernetes users attempting to use it. So I believe we either need to block `lb-provider=ovn`, or we need to create the listener without `allowed_cidrs`.

We can consider this a feature request if you prefer. But since it breaks for users, I raised this as a bug. Let me know if you need any additional logs

**Environment**:
Kubernetes environment version is: `v1.19.0`
```
Server Version: 4.6.0-0.okd-2020-11-27-200126
Kubernetes Version: v1.19.0-rc.2.1077+43983cda8af930-dirty
```
-  This is the image I'm using for occm:
https://hub.docker.com/layers/k8scloudprovider/openstack-cloud-controller-manager/latest/images/sha256-e9d3de3c0cbc6edca78f0e5ca0cdb7643af10225feed88fa6c1bdd90fc90139d?context=explore

- OpenStack version: Victoria
- Others:
Octavia Version:
```
[root@overcloud-controller-0 ~]# podman exec octavia_api rpm -qa | grep octavia
python3-octavia-lib-2.2.0-0.20201118104745.fe022cd.el8.noarch
openstack-octavia-common-7.1.0-0.20201203065445.9d38db9.el8.noarch
python3-ovn-octavia-provider-0.4.0-0.20201118135846.5b0715d.el8.noarch
python3-octavia-7.1.0-0.20201203065445.9d38db9.el8.noarch
python3-octaviaclient-2.2.0-0.20201112084714.78c08ed.el8.noarch
openstack-octavia-api-7.1.0-0.20201203065445.9d38db9.el8.noarch
```

bshephar:
Currently, when provided with a unsupported lb-provider,
occm will accept providers that we can't support at this stage.
Rather than accept it, we will parse the provided value, compare
with our new supportedLBProvider map and fail for unsupported
providers.

This change introduces supportedLBProvider map, and a small section
during ReadConfig() to ensure the provided value can be supported.

[occm] Check for supported LBProvider type

This PR will add a map of supported LoadBalancer providers, and fail if the user specifies an unsupported Provider.

fixes #1332 1332

**Release note**:
```release-note
[openstack-cloud-controller-manager] Added a check that only "amphora" and "octavia" are supported for the config `lb-provider`.
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack.go
+++ b/pkg/cloudprovider/providers/openstack/openstack.go
@@ -86,6 +86,9 @@ var ErrIPv6SupportDisabled = errors.New("IPv6 support is disabled")
 // userAgentData is used to add extra information to the gophercloud user-agent
 var userAgentData []string
 
+// supportedLBProvider map is used to define LoadBalancer providers that we support
+var supportedLBProvider = []string{"amphora", "octavia"}
+
 // AddExtraFlags is called by the main package to add component specific command line flags
 func AddExtraFlags(fs *pflag.FlagSet) {
 	fs.StringArrayVar(&userAgentData, "user-agent", nil, "Extra data to add to gophercloud user-agent. Use multiple times to add more than one component.")
@@ -410,6 +413,10 @@ func ReadConfig(config io.Reader) (Config, error) {
 	// Set the default values for search order if not set
 	if cfg.Metadata.SearchOrder == "" {
 		cfg.Metadata.SearchOrder = fmt.Sprintf("%s,%s", metadata.ConfigDriveID, metadata.MetadataID)
+	}
+
+	if !util.Contains(supportedLBProvider, cfg.LoadBalancer.LBProvider) {
+		return Config{}, fmt.Errorf("Unsupported LoadBalancer Provider: %s", cfg.LoadBalancer.LBProvider)
 	}
 
 	return cfg, err


pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -86,6 +86,13 @@ var ErrIPv6SupportDisabled = errors.New("IPv6 support is disabled")
 // userAgentData is used to add extra information to the gophercloud user-agent
 var userAgentData []string
 
+// supportedLBProvider map is used to define LoadBalancer providers that we support
+var supportedLBProvider = map[string]bool{
```
lingxiankong, line 90: I would suggest defining a string array instead of a map:

```golang
var supportedLBProvider = [...]string{"amphora", "octavia"}
```

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -412,6 +419,10 @@ func ReadConfig(config io.Reader) (Config, error) {
 		cfg.Metadata.SearchOrder = fmt.Sprintf("%s,%s", metadata.ConfigDriveID, metadata.MetadataID)
 	}
 
+	if !supportedLBProvider[cfg.LoadBalancer.LBProvider] {
```
lingxiankong, line 422: And use:

```golang
if util.Contains(supportedLBProvider, cfg.LoadBalancer.LBProvider) {
    ...
}
```

lingxiankong (changes requested):

pkg/cloudprovider/providers/openstack/openstack.go
bshephar: Hey, tbh I wasn't aware we had util.Contains. I have changed this now. I'll amend my commit and push it

pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -412,6 +415,10 @@ func ReadConfig(config io.Reader) (Config, error) {
 		cfg.Metadata.SearchOrder = fmt.Sprintf("%s,%s", metadata.ConfigDriveID, metadata.MetadataID)
 	}
 
+	if !util.Contains(supportedLBProvider, cfg.LoadBalancer.LBProvider) {
```
jichenjc, line 418: the commit msg said it's must fit and `Contains` are used here, if it's a typo such as `octaviaa` then still pass this test?

lingxiankong: @jichenjc `Contains searches if a string list contains the given string or not.`

jichenjc: my bad ...

lingxiankong: :-) all good

lingxiankong:
/lgtm

In good shape now.

PR #1345, opened 2020-12-10
[occm] store LB address in .Hostname field when using PROXY protocol

### LINKED ISSUE #1287: [occm] PROXY protocol breaks connections from inside the cluster
/kind bug

This is a manifestation of kubernetes/kubernetes#66607. When a service of type `LoadBalancer` is exposed via an IP address (which is the case for OCCM), iptables rules "optimize" the traffic and send the packets directly to the underlying pods.

The problem is that:
* the client isn't aware it needs to send the `PROXY` header (it may not even be capable of doing that)
* the server, per the specification, _must_ refuse the connection if the header is missing

The workaround used by other cloud providers is to expose the LB's external endpoint via the `.Hostname` field instead of the `.IP` field.

bgagnon:

Workaround for kubernetes/kubernetes#66607. In PROXY mode, use Hostname and `.nip.io` suffix to expose IP via the Hostname field.
The `nip.io` suffix is configurable via an optional annotation (other possible values include `.xip.io` or custom domain suffixes).

Fixes #1287 

This doesn't merge nicely to `master` at the moment due to the recent refactors.

I wanted to get the code out there so that my colleagues @davidovich or @rajiteh can pick it up.

**Release note**:
```release-note
LBaaS services in PROXY mode now use the Hostname field to publish the IP address to prevent an iptables bypass (#1287)
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
@@ -44,6 +44,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/wait"
 	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/klog"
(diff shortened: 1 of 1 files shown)


lingxiankong:
I see this PR is targeting release-1.18, why not propose a PR for master?

lingxiankong:
@bgagnon Are you still working on this PR?

bgagnon:
Unfortunately I can no longer pursue this PR -- see my comment in #1287.

lingxiankong:
for master branch: https://github.com/kubernetes/cloud-provider-openstack/pull/1449

PR #1346, opened 2020-12-11
[occm] Refer to the correct error variable

pyama86:

It is to fix cording miss.

I realzed this error.
```
Error creating load balancer (will retry): failed to ensure load balancer for service default/wordpress: failed to add tag xxxxxx to port xxxxx: <nil>
```

but, error detail is `nil` because rever variable is deffrent and I fix it.

**Release note**:

```release-note
NONE
```


pyama86:
@lingxiankong  thanks for merge!

PR #1353, opened 2020-12-18
[cinder-csi-plugin] chart: better optional cloud-config secret handling

remche:
This PR offers better secret handling for `cloud-config` : 

* use an existing secret : 
```yaml
secret:
  use: true
  name: cinder-csi-cloud-config
```

* create a new secret for `cloud-config` :
```yaml
secret:
  use: true
  name: cinder-csi-cloud-config
  data:
    cloud-config: |-
      ...
```

### DIFF
--- a/charts/cinder-csi-plugin/templates/controllerplugin-statefulset.yaml
+++ b/charts/cinder-csi-plugin/templates/controllerplugin-statefulset.yaml
@@ -125,4 +125,12 @@ spec:
       volumes:
         - name: socket-dir
           emptyDir:
+        - name: cloud-config
+        {{- if .Values.secret.use }}
+          secret:
+            secretName: {{ .Values.secret.name }}
+        {{- else }}
+          hostPath:
+            path: /etc/kubernetes
+        {{- end }}
         {{ .Values.csi.plugin.volumes | toYaml | trimSuffix "\n" | nindent 8 }}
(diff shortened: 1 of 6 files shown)


charts/cinder-csi-plugin/values.yaml
```diff
@@ -56,6 +51,8 @@ csi:
         readOnly: true
 
 secret:
+  use: false
```
lingxiankong, line 54: Can we use `enabled` instead of `use`?

lingxiankong (changes requested):

charts/cinder-csi-plugin/values.yaml
remche: Totally agree that's better. Fixed !

lingxiankong:
Haven't tested, but changes look good to me.

/lgtm

charts/cinder-csi-plugin/templates/nodeplugin-daemonset.yaml
```diff
@@ -118,4 +118,12 @@ spec:
           hostPath:
             path: /dev
             type: Directory
+        - name: cloud-config
+        {{- if .Values.secret.use }}
```
ramineni, line 122: it should be enabled?

charts/cinder-csi-plugin/templates/controllerplugin-statefulset.yaml
```diff
@@ -125,4 +125,12 @@ spec:
       volumes:
         - name: socket-dir
           emptyDir:
+        - name: cloud-config
+        {{- if .Values.secret.use }}
```
ramineni, line 129: enabled?

remche: Oops, bad `git add`, fixed.

jichenjc:
/approve

thanks for the change

PR #1355, opened 2021-01-05
[All] Update container image tag to v1.20.0

lingxiankong:
Update the container image tag in the manifest files for release v1.20.0

fixes #

**Release note**:

```release-note
NONE
```


jichenjc:
not sure it's something need to be put into release notes? 
seems we will use 1.20 version by default now (which is good)

lingxiankong:
> not sure it's something need to be put into release notes?

There is nothing significant that should bring user's or operator's notice for this particular commit. This is part of v1.20 release tasks.

jichenjc:
/lgtm

thanks

PR #1357, opened 2021-01-05
[occm] don't error on unsupported octavia provider

### LINKED ISSUE #1356: [occm] octavia stopped working with custom provider driver
**Is this a BUG REPORT or FEATURE REQUEST?**: Bug

/kind bug

**What happened**:
Cloud provider could not be initialized: could not init cloud provider "openstack": Unsupported LoadBalancer Provider: f5

The pull request #1344 introduced an allow-list of supported octavia provider plugins that breaks existing setups using octavia provider plugins that are _not_ amphora based. In our case, we use the F5 provider plugin which is fully compatible with occm:
https://github.com/sapcc/octavia-f5-provider-driver
`lb-provider=f5`

occm is supposed to use what the customer/admin has configured.

Custom provider plugins shouldn't be inferred by occm since customers could use a vendor provider plugin (e.g. Juniper has closed-source provider plugin, and many more:
https://docs.openstack.org/octavia/latest/admin/providers/index.html  

**How to reproduce it**:
cloud-config.conf:
```
[LoadBalancer]
lb-provider=f5
```

I suggest following solutions:

1. Revert #1344 as is doesn't add any benefit.
2. Instead of returning an error, just print a warning that the provider plugin has not been tested/supported.

**Environment**:
- OpenStack version: Queens

notandy:
Emit's a warning instead of initialisation failure if an unsupported octavia provider plugin is used.

fixes #1356

**Release note**:
```release-note
[openstack-cloud-controller-manager] Allow load balancer providers other than 'amphora' and 'octavia'.
```

### DIFF
--- a/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
+++ b/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
@@ -176,7 +176,7 @@ Although the openstack-cloud-controller-manager was initially implemented with N
   The load balancing algorithm used to create the load balancer pool. The value can be `ROUND_ROBIN`, `LEAST_CONNECTIONS`, or `SOURCE_IP`. Default: `ROUND_ROBIN`
 
 * `lb-provider`
-  Optional. Used to specify the provider of the load balancer, e.g. "amphora" or "octavia".
+  Optional. Used to specify the provider of the load balancer, e.g. "amphora" or "octavia". Only "amphora" or "octavia" provider are officially tested, other provider will cause a warning log.
 
 * `lb-version`
   Optional. If specified, only "v2" is supported.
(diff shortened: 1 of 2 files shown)


lingxiankong:
Please also add a description for the option `lb-provider` (https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#load-balancer) that using other custom providers is allowed but only `amphora` provider is officially tested.

notandy:
> Please also add a description for the option `lb-provider` (https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#load-balancer) that using other custom providers is allowed but only `amphora` provider is officially tested.

sure, thanks.

lingxiankong:
/lgtm

Thanks!

PR #1364, opened 2021-01-11
[occm] validate flavor name, fallback on flavor id

seanschneeweiss:
The node.kubernetes.io/instance-type label accepts alphanumeric
characters, '-', '_', and '.' only. If the flavor name contains
invalid characters, the flavor id is used instead.

fixes #

This addresses the issue mentioned in https://github.com/kubernetes/cloud-provider-openstack/pull/1213#issuecomment-757008618.

```
error syncing 'blue-master-1': Node "blue-master-1" is invalid: metadata.labels: Invalid value: "Small HD 4GB": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue', or 'my_value', or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?'), requeuing
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/providers/openstack/openstack_instances.go
+++ b/pkg/cloudprovider/providers/openstack/openstack_instances.go
@@ -29,6 +29,7 @@ import (
 
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation"
 	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/cloud-provider-openstack/pkg/metrics"
 	"k8s.io/cloud-provider-openstack/pkg/util/errors"
@@ -280,21 +281,38 @@ func srvInstanceType(client *gophercloud.ServiceClient, srv *servers.Server) (st
 	keys := []string{"original_name", "id"}
 	for _, key := range keys {
 		val, found := srv.Flavor[key]
-		if found {
-			flavor, ok := val.(string)
-			if ok {
-				if key == "id" {
-					mc := metrics.NewMetricContext("flavor", "get")
-					f, err := flavors.Get(client, flavor).Extract()
-					if mc.ObserveRequest(err) == nil {
-						return f.Name, nil
-					}
-				}
-				return flavor, nil
+		if !found {
+			continue
+		}
+
+		flavor, ok := val.(string)
+		if !ok {
+			continue
+		}
+
+		if key == "original_name" && isValidLabelValue(flavor) {
+			return flavor, nil
+		}
+
+		// get flavor name by id
+		mc := metrics.NewMetricContext("flavor", "get")
+		f, err := flavors.Get(client, flavor).Extract()
+		if mc.ObserveRequest(err) == nil {
+			if isValidLabelValue(f.Name) {
+				return f.Name, nil
 			}
+			// fallback on flavor id
+			return f.ID, nil
 		}
 	}
-	return "", fmt.Errorf("flavor name/id not found")
+	return "", fmt.Errorf("flavor original_name/id not found")
+}
+
+func isValidLabelValue(v string) bool {
+	if errs := validation.IsValidLabelValue(v); len(errs) != 0 {
+		return false
+	}
+	return true
 }
 
 // If Instances.InstanceID or cloudprovider.GetInstanceProviderID is changed, the regexp should be changed too.


lingxiankong:
`-`, `_` and `.` are all allowed for the flavor name in OpenStack Nova, see the following:

```
$ openstack flavor list | grep test
| 75562cd6-03a4-4f41-adfd-b3cccf1b7bcb | my test       |  1024 |    7 |         0 |     1 | True      |
| 76964c13-27bb-4cb1-8456-7aac11413240 | my-test       |  1024 |    7 |         0 |     1 | True      |
| dd998294-a30d-47b0-84ac-e26253b0aea4 | my_test       |  1024 |    7 |         0 |     1 | True      |
| f834cb50-05b2-4729-87d4-caa3fcd583be | my.test       |  1024 |    7 |         0 |     1 | True      |
```

What if `my test` and `my-test ` both exist?

seanschneeweiss:
@lingxiankong `-` and `.` are also not allowed in label names. If someone has `my test` and `my-test` the label would not be unique.
What would you propose if this is not acceptable?
Should we also replace `-` and `.` with something else?

seanschneeweiss:
We could return the ID if the flavor name contains not allowed characters.

lingxiankong:
> We could return the flavor ID, if the flavor name contains non allowed characters. What do you think?

Yeah, that's something I'd prefer, but I'm not sure if some other apps are relying on flavor name. Changing name to ID may probably break the backward compatibility.

@jichenjc @ramineni @zetaab thoughts?

jichenjc:
> 
> 
> > We could return the flavor ID, if the flavor name contains non allowed characters. What do you think?
> 
> Yeah, that's something I'd prefer, but I'm not sure if some other apps are relying on flavor name. Changing name to ID may probably break the backward compatibility.
> 
> @jichenjc @ramineni @zetaab thoughts?

maybe we can provide some option so that we don't break someone who really use the name? 
though I think typically should not rely on the name ..

FYI, our team suffered similar problem previously (our AZ created with blank ,openstack allow AZ with blank but k8s label does not allow blank ..) so I think we should avoid this kind of issue as much as we can

lingxiankong:
According to https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#nodekubernetesioinstance-type, it's not recommended to schedule based on this label, so I guess most of the use cases are just for filtering.

So can we just use flavor ID instead of name and leave a release note about what has been changed?

jichenjc:
I think that's reasonable to update with ID and release notes
but #1379 seems be a problem as it will be used for schedule?

lingxiankong:
> but #1379 seems be a problem as it will be used for schedule?

@jichenjc we can discuss this problem on the issue page or in a separate PR.

ramineni:
> > We could return the flavor ID, if the flavor name contains non allowed characters. What do you think?
> 
> Yeah, that's something I'd prefer, but I'm not sure if some other apps are relying on flavor name. Changing name to ID may probably break the backward compatibility.
> 
> @jichenjc @ramineni @zetaab thoughts?

Agree with @lingxiankong replacing space with '-' internally will be confusing, we should avoid that. I suppose replacing with 'id' is better option as its being done in past releases. But before that , could we raise issue in kubernetes repo on the same , to allow whitespaces and check .

ramineni:
/lgtm cancel 
As under discussion on the best possible approach

lingxiankong:
> could we raise issue in kubernetes repo on the same , to allow whitespaces and check . I think its reasonable to allow whitespaces in instance-type? wdyt?

I don't think it will be acceptable in kubernetes community, as you can see here https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

jichenjc:
> > could we raise issue in kubernetes repo on the same , to allow whitespaces and check . I think its reasonable to allow whitespaces in instance-type? wdyt?
> 
> I don't think it will be acceptable in kubernetes community, as you can see here https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

I commented https://github.com/kubernetes/cloud-provider-openstack/pull/1380#issuecomment-763375737 as well

lingxiankong:
So my suggestion is, we use flavor ID

ramineni:
I'm ok with flavor id .

seanschneeweiss:
Would it be ok to keep flavor name, if no special characters are included in it and use the flavor ID as a fallback? Or would you propose to go back to the previous implementation and return the flavor ID in any case?

ramineni:
@seanschneeweiss fallback to flavor id incase of space , sounds good to me

seanschneeweiss:
@ramineni Maybe not only for space but rather for all non allowed characters.

ramineni:
@seanschneeweiss I'm ok with both. But I'm leaning more towards reverting to flavor id for all ensures consistency, lets wait for others opinion as well.

ramineni:
@lingxiankong could you provide your suggestion on this?

ramineni:
@seanschneeweiss please go ahead with fallback idea

seanschneeweiss:
@ramineni @lingxiankong @jichenjc PTAL, thank you.

lingxiankong:

ramineni:
@seanschneeweiss lets backport this to 1.20 as well

seanschneeweiss:
> @seanschneeweiss lets backport this to 1.20 as well

Please see https://github.com/kubernetes/cloud-provider-openstack/pull/1402

PR #1368, opened 2021-01-12
[cinder-csi-plugin] validate volume capabilities before creation

### LINKED ISSUE #1367: [cinder-csi-plugin] Cinder CSI plugin should not provision ROX or RWX volumes
/kind bug

**What happened**:
When creating a PVC with ROX or RWX access mode, the volume provisioned successfully, but when creating pods assigned to different nodes, it got attach failure.

We should show out the unsupport message as cinder in-tree plugin does:
Warning ProvisioningFailed 7s (x3 over 14s) persistentvolume-controller Failed to provision volume with StorageClass "standard": invalid AccessModes [ReadWriteMany]: only AccessModes [ReadWriteOnce] are supported

**How to reproduce it**:
1. Install OSP cluster and cinder csi driver is installed.
2. Create PVC with ROX/RWX access mode, and create pod to consume it.
3. Check the pod and pvc statuses
4. Create another pod which assigned to another node

1. Volume provisioned successfully
$ oc get pvc -n wduan
NAME STATUS VOLUME CAPACITY ACCESS MODES STORAGECLASS AGE
mypvc02 Bound pvc-ae09e478-2b8e-4350-a83a-b789ff991d7d 1Gi RWX standard-csi 17m

$ oc get pvc -n wduan-01
NAME STATUS VOLUME CAPACITY ACCESS MODES STORAGECLASS AGE
mypvc03 Bound pvc-9404f88c-d437-493b-ba5e-3a2685cdc1c9 1Gi ROX standard-csi 14m

2. Attach failed when creating another pod assigned to another node
Warning FailedAttachVolume 12s attachdetach-controller AttachVolume.Attach failed for volume "pvc-ae09e478-2b8e-4350-a83a-b789ff991d7d" : rpc error: code = Internal desc = [ControllerPublishVolume] Attach Volume failed with error failed to attach ccc11e8b-cfff-40ac-92da-d7baed902b12 volume to fa6d2e8c-eb8f-4a57-a54a-5063f2b8bec7 compute: Bad request with: [POST https://rhos-d.infra.prod.upshift.rdu2.redhat.com:13774/v2.1/servers/fa6d2e8c-eb8f-4a57-a54a-5063f2b8bec7/os-volume_attachments], error message: {"badRequest": {"message": "Invalid input received: Invalid volume: Volume ccc11e8b-cfff-40ac-92da-d7baed902b12 status must be available or downloading (HTTP 400) (Request-ID: req-75c01a31-c877-4582-930e-35c5d8b64eaf)", "code": 400}}

Fedosin:
To comply with in-tree cinder plugin, we need to prevent PVCs with ROX/RWX access modes.

fixes #1367 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -55,6 +55,12 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return nil, status.Error(codes.InvalidArgument, "[CreateVolume] missing Volume capability")
 	}
 
+	for _, cap := range volCapabilities {
+		if cap.GetAccessMode().GetMode() != cs.Driver.vcap[0].Mode {
+			return nil, status.Errorf(codes.InvalidArgument, "[CreateVolume] requested Volume capabilty not supported: %s", cap.GetAccessMode().GetMode())
+		}
+	}
+
 	// Volume Size - Default is 1 GiB
 	volSizeBytes := int64(1 * 1024 * 1024 * 1024)
 	if req.GetCapacityRange() != nil {


Fedosin:
recheck

pkg/csi/cinder/controllerserver.go
```diff
@@ -55,6 +55,12 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return nil, status.Error(codes.InvalidArgument, "[CreateVolume] missing Volume capability")
 	}
 
+	for _, cap := range volCapabilities {
```
mandre, line 58: Feels like some refactoring should be in order, to allow both `CreateVolume()` and `ValidateVolumeCapabilities()` to use the same validate function.

mandre (commented):
/lgtm

pkg/csi/cinder/controllerserver.go
jsafrane: +1

And an unit test would be great.

ramineni:
/hold
until , we understand which version it is being tested etc . https://github.com/kubernetes/cloud-provider-openstack/issues/1367#issuecomment-759208676

tombarron:
A few remarks based on my own understanding of the situatiion.  Please check and correct as appropriate.
#1) K8s now allows one to add  'volumeMode =Block' in the PVC spec [1] in which case ReadWriteMany access mode is also possible.  The default volumeMode remains 'File' even if a block device is used, as with CinderCSI.  With File mode RWX is unsafe *even if* the back end storage supports Cinder multi-attach because a node-local file system is built on the block device before it is first consumed and RWX brings with it the potental that that node local filesystem would be mounted concurrently on separate nodes in the cluster.  With raw block mode, on the other hand, no node-local filesystem is built on the block device by k8s.  The workloads in containers in pods see a raw block device with offsets and no filesystem tree.
#2) Cinder-CSI uses ordinary OpenStack user credentials (member of keystone project) rather than elevated cloud administrator credentials.  This is a good thing -- you don't have to be some kind of power user on OpenStack to be able to run k8s on it and use Cinder-CSI for PVs.  But I think it means that you have no programmatic way to tell if the Cinder back ends behind your storage classes support multi-attach or not.  The cinder folks have talked about building some kind of capabilities API that might expose such information but I don't think it has been done yet.  @eharney knows this stuff better than I do and would be more authoritative.
[1] https://kubernetes.io/blog/2019/03/07/raw-block-volume-support-to-beta/

ramineni:
> A few remarks based on my own understanding of the situatiion. Please check and correct as appropriate.
> #1) K8s now allows one to add 'volumeMode =Block' in the PVC spec [1] in which case ReadWriteMany access mode is also possible. The default volumeMode remains 'File' even if a block device is used, as with CinderCSI. With File mode RWX is unsafe _even if_ the back end storage supports Cinder multi-attach because a node-local file system is built on the block device before it is first consumed and RWX brings with it the potental that that node local filesystem would be mounted concurrently on separate nodes in the cluster. With raw block mode, on the other hand, no node-local filesystem is built on the block device by k8s. The workloads in containers in pods see a raw block device with offsets and no filesystem tree.
> #2) Cinder-CSI uses ordinary OpenStack user credentials (member of keystone project) rather than elevated cloud administrator credentials. This is a good thing -- you don't have to be some kind of power user on OpenStack to be able to run k8s on it and use Cinder-CSI for PVs. But I think it means that you have no programmatic way to tell if the Cinder back ends behind your storage classes support multi-attach or not. The cinder folks have talked about building some kind of capabilities API that might expose such information but I don't think it has been done yet. @eharney knows this stuff better than I do and would be more authoritative.
> [1] https://kubernetes.io/blog/2019/03/07/raw-block-volume-support-to-beta/

@tombarron you are right . Default volume mode is File , for block devices that needs to be explicitly updated. This is default behavior of k8s. And also additionally , there is also a difference when its attached to Pod, it should be added as `volumeDevices` instead of `volumeMounts` . It should be informed decision to be used as file or block device.
For multiattach, I understand the concern. Thats the reason actually, we are asking the admin to specify the volume type in storage class instead of automatically creating it. So, if the volume type if specified as multiattach , and accessmode also RWX then only multiattach works . If correct volume type is not specified and accessmode is RWX , it throws error. To expose the capability of cinder in k8s, the only thing modified in k8s is to allow attaching to multiple nodes if multittach=true. All the backend work , like creating volume type etc should be done by admin beforehand .

But ya without capabilities API , there is always a risk of corruption if not used properly. Let me know your thoughts whats the best way to expose the capability?

pkg/csi/cinder/controllerserver.go
mdbooth: I don't think we can use volCapabilities for this. volCapabilities are the published, *generic* capabilities of the driver. What we need to check here are the *specific* capabilities of the volume type.

We fetch the id of the volume type below. I think we need to pull the extra_specs of the volume type and specifically check for multiattach support.

mdbooth:
I'd like to expand on the above with some hopefully clear and falsifiable statements. Please falsify any that need it :)

Firstly, for volumes with `volumeMode = Block` I believe the cinder CSI driver potentially supports *all* modes. Specifically:

* SINGLE_NODE_WRITER
* SINGLE_NODE_READER_ONLY
* MULTI_NODE_READER_ONLY
* MULTI_NODE_SINGLE_WRITER
* MULTI_NODE_MULTI_WRITER

Therefore I believe we should add all of these to the driver's VolumeCapabilityAccessModes, but return InvalidArgument for specific method calls where the mode can't be supported. As noted, this means the driver capabilities aren't very useful.

We can determine if a volume type is multiattach by fetching its extra_specs and looking for multiattach = true. This isn't a privileged operation.

All volume with a multiattach volume type are multiattach. All other volumes are not multiattach. (Or at least we can make this true by not using the volume.multiattach field, which seems to be deprecated.)

**Therefore for `volumeMode = Block` we always support `SINGLE_NODE_*`, and we support `MULTI_NODE_*` iff the volume type is multiattach as pulled from volume type extra_specs.**

For `volumeMode = File`, we should note that the filesystem is provisioned by us in NodeStageVolume. Therefore we know that the provisioned filesystem is not a cluster filesystem and therefore doesn't support `MULTI_NODE_*`.

**Therefore for `volumeMode = File` we always support `SINGLE_NODE_*`, and we never support `MULTI_NODE_*`.**

pkg/csi/cinder/controllerserver.go
mdbooth: I've written up what I believe should be the correct test here: https://github.com/kubernetes/cloud-provider-openstack/pull/1368#issuecomment-818831513

tombarron:
@mdbooth   Very nice summary, but the catch may be here:

> We can determine if a volume type is multiattach by fetching its extra_specs and looking for multiattach = true. This isn't a privileged operation.

I'd love to be wrong but I don't think a regular user (member of keystone project) can access the relevant extra-specs to determine if multiattach is true.

I've set multiattach to true for the 'tripleo' volume type below.  Note the difference in what is shown to an OpenStack administrator (in this case, with 'overcloudrc' credentials and prompt) versus a regular keystone project member (in this case, with 'demorc' credentials and prompt).   The commands (and underlying APIs) that show the extra spec are not Forbidden to the regular user, but the extra specs information is redacted:

(overcloud) [stack@undercloud-0 ~]$ cinder type-list
+--------------------------------------+-------------+---------------------+-----------+
| ID                                   | Name        | Description         | Is_Public |
+--------------------------------------+-------------+---------------------+-----------+
| 1f1b3653-36d8-4b6d-9573-ca67a7711e77 | tripleo     | -                   | True      |
| ca93e753-c674-47a1-ba3c-d8cf6f3ce1f5 | __DEFAULT__ | Default Volume Type | True      |
+--------------------------------------+-------------+---------------------+-----------+
(overcloud) [stack@undercloud-0 ~]$ cinder type-show tripleo
+---------------------------------+--------------------------------------+
| Property                        | Value                                |
+---------------------------------+--------------------------------------+
| description                     | None                                 |
| extra_specs                     | multi-attach : True                  |
| id                              | 1f1b3653-36d8-4b6d-9573-ca67a7711e77 |
| is_public                       | True                                 |
| name                            | tripleo                              |
| os-volume-type-access:is_public | True                                 |
| qos_specs_id                    | None                                 |
+---------------------------------+--------------------------------------+
(overcloud) [stack@undercloud-0 ~]$ cinder type-key tripleo unset multi-attach
(overcloud) [stack@undercloud-0 ~]$ cinder type-key tripleo set multiattach=True
(overcloud) [stack@undercloud-0 ~]$ cinder type-list
+--------------------------------------+-------------+---------------------+-----------+
| ID                                   | Name        | Description         | Is_Public |
+--------------------------------------+-------------+---------------------+-----------+
| 1f1b3653-36d8-4b6d-9573-ca67a7711e77 | tripleo     | -                   | True      |
| ca93e753-c674-47a1-ba3c-d8cf6f3ce1f5 | __DEFAULT__ | Default Volume Type | True      |
+--------------------------------------+-------------+---------------------+-----------+
(overcloud) [stack@undercloud-0 ~]$ cinder type-show tripleo
+---------------------------------+--------------------------------------+
| Property                        | Value                                |
+---------------------------------+--------------------------------------+
| description                     | None                                 |
| extra_specs                     | multiattach : True                   |
| id                              | 1f1b3653-36d8-4b6d-9573-ca67a7711e77 |
| is_public                       | True                                 |
| name                            | tripleo                              |
| os-volume-type-access:is_public | True                                 |
| qos_specs_id                    | None                                 |
+---------------------------------+--------------------------------------+
(overcloud) [stack@undercloud-0 ~]$ cinder extra-specs-list
+--------------------------------------+-------------+-------------------------+
| ID                                   | Name        | extra_specs             |
+--------------------------------------+-------------+-------------------------+
| 1f1b3653-36d8-4b6d-9573-ca67a7711e77 | tripleo     | {'multiattach': 'True'} |
| ca93e753-c674-47a1-ba3c-d8cf6f3ce1f5 | __DEFAULT__ | {}                      |
+--------------------------------------+-------------+-------------------------+
(overcloud) [stack@undercloud-0 ~]$ source demorc
(demo@overcloud) [stack@undercloud-0 ~]$ cinder type-show tripleo
+---------------------------------+--------------------------------------+
| Property                        | Value                                |
+---------------------------------+--------------------------------------+
| description                     | None                                 |
| id                              | 1f1b3653-36d8-4b6d-9573-ca67a7711e77 |
| is_public                       | True                                 |
| name                            | tripleo                              |
| os-volume-type-access:is_public | True                                 |
+---------------------------------+--------------------------------------+
(demo@overcloud) [stack@undercloud-0 ~]$ cinder extra-specs-list
+--------------------------------------+-------------+-------------+
| ID                                   | Name        | extra_specs |
+--------------------------------------+-------------+-------------+
| 1f1b3653-36d8-4b6d-9573-ca67a7711e77 | tripleo     |             |
| ca93e753-c674-47a1-ba3c-d8cf6f3ce1f5 | __DEFAULT__ |             |
+--------------------------------------+-------------+-------------+

mdbooth:
Indeed 🤔

```
$ curl -H "X-Auth-Token: $token" http://10.46.26.42:8776/v3/a2f172c1f76140a7a47d15f249997ef7/types/4f65d95a-65e6-4d91-99ae-3283bff13444/extra_specs
{"forbidden": {"code": 403, "message": "Policy doesn't allow volume_extension:types_extra_specs:index to be performed."}}
```

mdbooth:
However! If I create a volume using that type I get:

```
$ openstack volume create --type tripleo --size 1 foo
+---------------------+--------------------------------------+
| Field               | Value                                |
+---------------------+--------------------------------------+
| attachments         | []                                   |
| availability_zone   | nova                                 |
| bootable            | false                                |
| consistencygroup_id | None                                 |
| created_at          | 2021-04-14T12:30:19.000000           |
| description         | None                                 |
| encrypted           | False                                |
| id                  | 6a0d07e6-543b-4c50-989d-3415a02e41c1 |
| multiattach         | True                                 |
| name                | foo                                  |
| properties          |                                      |
| replication_status  | None                                 |
| size                | 1                                    |
| snapshot_id         | None                                 |
| source_volid        | None                                 |
| status              | creating                             |
| type                | tripleo                              |
| updated_at          | None                                 |
| user_id             | 41e8a7a892364179a924fbf8309e3b56     |
+---------------------+--------------------------------------+
```

So while it seems we should not use the multiattach parameter to volume create because it is deprecated, we can still use the multiattach property of the resulting volume. i.e. The only way we have of knowing if we're going to get a multiattach volume is to create one and see what we got. That's disappointing, but not completely unworkable.

PR #1375, opened 2021-01-15
[occm] pattern based subnet detection for loadbalancers

### LINKED ISSUE #1328: [occm] Wildcard floating ip subnet selection for LoadBalancer services
**Is this a BUG REPORT or FEATURE REQUEST?**: feature

/kind feature

**What happened**:
As of today the Openstack cloud-controller-manager can be configured to select floating ips for services of type Loadbalancer from a specified floating ip subnet of a floating network.

Floating networks can consist of multiple subnets with different purposes e.g. subnets which are routable via the internet and subnets routable in internal networks.

To ensure that per default all LoadBalancer services (with/without a`LoadBalancerClass`) end up with a floating ip of the correct subnets the cloud-controller-manager need to be configured with a fixed floating ip subnet.
This approach is bad as floating ip subnets can exhaust and users wouldn't be able to create additional LoadBalancer services without annotating them  (`loadbalancer.openstack.org/floating-subnet-id="another-fip-subnet"`) to use a different floating subnet.

It would be handy to give the cloud-controller-manager the ability to select floating ip subnets based on a regular expression matching on their name. This would eliminate the need to configure a specific floating ip subnet. The cloud-controller-manager could consider for the selection of a floating ip, all floating subnets whose name match on the configured regex.
It would be required to loop over all matched floating subnets to find a subnet which has free a floating ip for a new LoadBalancer service.

If all floating subnets are allocated it need to fail as well. But it would help users to create new LoadBalancer services without annotating services or without the need to re-configure the cloud-controller-manager in order to create new LoadBalancer services when an administrator of the respective Openstack environment adds new floating subnets.

My proposal is to add an additional option, e.g. `floating-subnet-name-prefix`, in the `[LoadBalancer]` and `[LoadBalancerClass ...]` sections of the cloud provider config, which is used to the cloud-controller-manager.
An example cloud-provider-config could look like this:
```
[LoadBalancer]
floating-network-id="my-fip-network"
floating-subnet-name-prefix="fip-internet" # All floating subnet which match on "fip-internet*" are considered for selecting a floating ip

[LoadBalancerClass "internal"]
floating-network-id="my-fip-network"
floating-subnet-name-prefix="fip-internal"  # All floating subnet which match on "fip-internal*" are considered for selecting a floating ip
```

The cloud-controller-manager could then look up all matching floating subnets in the floating network as described above to allocate a floating ip for a LoadBalancer service.

mandelsoft:
fixes #1328

A floating IP for a loadbalancer is created on a subnet of a floating pool (specified by a network id of the provider network).
Such a subnet has a dedicated IP range (CIDR). Therefore it might happen, that the subnet runs out of IP addresses. To not disrupt the existing FIP users in such a case a new subnet can be added to the provider network. This one will potentially automatically be selected for new FIP creation if the creator does not explicitly specify the subnet id.

But this automated subnet selection makes only sense if all subnets configured for a provider network provide the same kind
of routing. In Openstack it is possible to configure subnets for the same provider network that feature different kinds of routing (for example public internet or company network). In such a scenario so far a dedicated subnet must be specified for the FIP creation of a loadbalancer. If this setting is not done via service annotation, but as configuration of the cloud controller manager it is not possible to transparently add new matching subnets to the provider network by an administrator which is then used for further FIP creation.

This PR provides support for a new configuration possibility to overcome this problem. Instead of a dedicated subnet id it is possible now to optionally configure a subnet name pattern (supporting the standard glob mechanism) or tags. For this configuration option all matching subnets of a provider network are tried to create a new FIP. The first one where the FIP creation succeeds is used. 
This allows configuring subnets with similar names according to name patterns specific for the routing environment, so that FIPs can transparently be created for loadbalancers just by adding a subnet with a name matching the dedicated name pattern.

**Release note**:

```release-note
[openstack-cloud-controller-manager] Added support to use subnet name pattern and tag filtering for creating floating IP for the load balancer.
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -222,7 +222,7 @@ func (s *floatingSubnetSpec) ListSubnetsForNetwork(lbaas *LbaasV2, networkID str
 	if err != nil {
 		return nil, err
 	}
-	if matcher==nil {
+	if matcher == nil {
 		return list, nil
 	}
 
@@ -263,6 +263,25 @@ func (s *floatingSubnetSpec) MatcherConfigured() bool {
 		return true
 	}
 	return false
+}
+
+func addField(s, name, value string) string {
+	if value == "" {
+		return s
+	}
+	if s == "" {
+		s += ", "
+	}
+	return fmt.Sprintf("%s%s: %q", s, name, value)
+}
+
+func (s *floatingSubnetSpec) String() string {
+	if s == nil || (s.subnetID == "" && s.subnet == "" && s.subnetTags == "") {
+		return "<none>"
+	}
+	pat := addField("", "subnetID", s.subnetID)
+	pat = addField(pat, "pattern", s.subnet)
+	return addField(pat, "tags", s.subnetTags)
 }
 
 func (s *floatingSubnetSpec) Matcher(tag bool) (matcher, error) {
@@ -1187,13 +1206,14 @@ func (lbaas *LbaasV2) getServiceAddress(clusterName string, service *corev1.Serv
 					return "", err
 				}
 				if len(foundSubnets) == 0 {
-					return "", fmt.Errorf("no subnet matching pattern %q found for network %s",
-						svcConf.lbPublicSubnetSpec.subnet, svcConf.lbPublicNetworkID)
+					return "", fmt.Errorf("no subnet matching %s found for network %s",
+						svcConf.lbPublicSubnetSpec, svcConf.lbPublicNetworkID)
 				}
 
 				// try to create floating IP in matching subnets (tags already filtered by list options)
+				klog.V(4).Infof("found %d subnets matching %s for network %s", len(foundSubnets),
+					svcConf.lbPublicSubnetSpec, svcConf.lbPublicNetworkID)
 				for _, subnet := range foundSubnets {
-					klog.V(4).Infof("matching subnet %s(s)[%v]", subnet.Name, subnet.ID, subnet.Tags)
 					floatIPOpts.SubnetID = subnet.ID
 					floatIP, err = lbaas.createFloatingIP(fmt.Sprintf("Trying subnet %s for creating", subnet.Name), floatIPOpts)
 					if err == nil {
@@ -1203,8 +1223,8 @@ func (lbaas *LbaasV2) getServiceAddress(clusterName string, service *corev1.Serv
 					klog.V(2).Infof("cannot use subnet %s: %s", subnet.Name, err)
 				}
 				if err != nil {
-					return "", fmt.Errorf("no free subnet matching pattern %q found for network %s (last error %s)",
-						svcConf.lbPublicSubnetSpec.subnet, svcConf.lbPublicNetworkID, err)
+					return "", fmt.Errorf("no free subnet matching %q found for network %s (last error %s)",
+						svcConf.lbPublicSubnetSpec, svcConf.lbPublicNetworkID, err)
 				} else {
 					klog.V(2).Infof("Successfully created floating IP %s for loadbalancer %s on subnet %s(%s)", floatIP.FloatingIP, lb.ID, foundSubnet.Name, foundSubnet.ID)
 				}
@@ -1541,7 +1561,7 @@ func (lbaas *LbaasV2) checkService(service *corev1.Service, nodes []*corev1.Node
 			floatingSubnet.subnetID = lbClass.FloatingSubnetID
 			if floatingSubnet.subnetID == "" {
 				floatingSubnet.subnet = lbClass.FloatingSubnet
-				floatingSubnet.subnetTags = lbClass.FloatingSubnetTag
+				floatingSubnet.subnetTags = lbClass.FloatingSubnetTags
 			}
 		}
 
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -26,6 +26,8 @@ import (
 	"strings"
 	"time"
 
+	"gopkg.in/godo.v2/glob"
+
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/listeners"
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/loadbalancers"
@@ -82,24 +84,24 @@ const (
 	annotationXForwardedFor = "X-Forwarded-For"
 
 	// ServiceAnnotationLoadBalancerInternal defines whether or not to create an internal loadbalancer. Default: false.
-	ServiceAnnotationLoadBalancerInternal             = "service.beta.kubernetes.io/openstack-internal-load-balancer"
-	ServiceAnnotationLoadBalancerConnLimit            = "loadbalancer.openstack.org/connection-limit"
-	ServiceAnnotationLoadBalancerFloatingNetworkID    = "loadbalancer.openstack.org/floating-network-id"
-	ServiceAnnotationLoadBalancerFloatingSubnet       = "loadbalancer.openstack.org/floating-subnet"
-	ServiceAnnotationLoadBalancerFloatingSubnetID     = "loadbalancer.openstack.org/floating-subnet-id"
-	ServiceAnnotationLoadBalancerClass                = "loadbalancer.openstack.org/class"
-	ServiceAnnotationLoadBalancerKeepFloatingIP       = "loadbalancer.openstack.org/keep-floatingip"
-	ServiceAnnotationLoadBalancerPortID               = "loadbalancer.openstack.org/port-id"
-	ServiceAnnotationLoadBalancerProxyEnabled         = "loadbalancer.openstack.org/proxy-protocol"
-	ServiceAnnotationLoadBalancerSubnetID             = "loadbalancer.openstack.org/subnet-id"
-	ServiceAnnotationLoadBalancerNetworkID            = "loadbalancer.openstack.org/network-id"
-	ServiceAnnotationLoadBalancerTimeoutClientData    = "loadbalancer.openstack.org/timeout-client-data"
-	ServiceAnnotationLoadBalancerTimeoutMemberConnect = "loadbalancer.openstack.org/timeout-member-connect"
-	ServiceAnnotationLoadBalancerTimeoutMemberData    = "loadbalancer.openstack.org/timeout-member-data"
-	ServiceAnnotationLoadBalancerTimeoutTCPInspect    = "loadbalancer.openstack.org/timeout-tcp-inspect"
-	ServiceAnnotationLoadBalancerXForwardedFor        = "loadbalancer.openstack.org/x-forwarded-for"
-	ServiceAnnotationLoadBalancerFlavorID             = "loadbalancer.openstack.org/flavor-id"
-	ServiceAnnotationLoadBalancerAvailabilityZone     = "loadbalancer.openstack.org/availability-zone"
+	ServiceAnnotationLoadBalancerInternal              = "service.beta.kubernetes.io/openstack-internal-load-balancer"
+	ServiceAnnotationLoadBalancerConnLimit             = "loadbalancer.openstack.org/connection-limit"
+	ServiceAnnotationLoadBalancerFloatingNetworkID     = "loadbalancer.openstack.org/floating-network-id"
+	ServiceAnnotationLoadBalancerFloatingSubnet        = "loadbalancer.openstack.org/floating-subnet"
+	ServiceAnnotationLoadBalancerFloatingSubnetID      = "loadbalancer.openstack.org/floating-subnet-id"
+	ServiceAnnotationLoadBalancerClass                 = "loadbalancer.openstack.org/class"
+	ServiceAnnotationLoadBalancerKeepFloatingIP        = "loadbalancer.openstack.org/keep-floatingip"
+	ServiceAnnotationLoadBalancerPortID                = "loadbalancer.openstack.org/port-id"
+	ServiceAnnotationLoadBalancerProxyEnabled          = "loadbalancer.openstack.org/proxy-protocol"
+	ServiceAnnotationLoadBalancerSubnetID              = "loadbalancer.openstack.org/subnet-id"
+	ServiceAnnotationLoadBalancerNetworkID             = "loadbalancer.openstack.org/network-id"
+	ServiceAnnotationLoadBalancerTimeoutClientData     = "loadbalancer.openstack.org/timeout-client-data"
+	ServiceAnnotationLoadBalancerTimeoutMemberConnect  = "loadbalancer.openstack.org/timeout-member-connect"
+	ServiceAnnotationLoadBalancerTimeoutMemberData     = "loadbalancer.openstack.org/timeout-member-data"
+	ServiceAnnotationLoadBalancerTimeoutTCPInspect     = "loadbalancer.openstack.org/timeout-tcp-inspect"
+	ServiceAnnotationLoadBalancerXForwardedFor         = "loadbalancer.openstack.org/x-forwarded-for"
+	ServiceAnnotationLoadBalancerFlavorID              = "loadbalancer.openstack.org/flavor-id"
+	ServiceAnnotationLoadBalancerAvailabilityZone      = "loadbalancer.openstack.org/availability-zone"
 	// ServiceAnnotationLoadBalancerEnableHealthMonitor defines whether or not to create health monitor for the load balancer
 	// pool, if not specified, use 'create-monitor' config. The health monitor can be created or deleted dynamically.
 	ServiceAnnotationLoadBalancerEnableHealthMonitor = "loadbalancer.openstack.org/enable-health-monitor"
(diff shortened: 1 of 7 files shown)


dkistner:
This PR intended to fix https://github.com/kubernetes/cloud-provider-openstack/issues/1328

lingxiankong:
Thanks for proposing this PR.

First, I want to make sure that in your cloud, is normal user able to get subnets of an external network? I'm asking because in devstack (also in our cloud), that's not allowed.

lingxiankong:
Additionally, I still prefer to use subnet tag instead of a name pattern, as suggested in the issue.

lingxiankong:
> This PR intended to fix #1328

Please add that info in the PR description.

mandelsoft:
The usage of a subnet name/patterns seamlessly integrates with the subnet name annotation and solves the meaning of multiple subnets with the same name.
But you are right, supporting a tag would be useful. I will add this.

lingxiankong:
> Thanks for proposing this PR.
> 
> First, I want to make sure that in your cloud, is a normal user able to get subnets of an external network? I'm asking because in devstack (also in our cloud), that's not allowed.

Could you please confirm this question?

dkistner:
@lingxiankong At least for our OpenStack environment this is possible. cc @kayrus

kayrus:
@lingxiankong 

> First, I want to make sure that in your cloud, is normal user able to get subnets of an external network? I'm asking because in devstack (also in our cloud), that's not allowed.

Normal user can allocate FIPs. External FIP network can contain multiple subnets. One subnet can be an internal one, another can contain a public subnet. When public subnets get exhausted, cloud admin can add a new public subnet (with new uuid).
Currently, in case when k8s user wants to allocate public FIPs by default, he must reconfigure OCCM with a new subnet uuid. The aim of the PR is to automate this process by using the subnet name patterns.

> Additionally, I still prefer to use subnet tag instead of a name pattern, as suggested in the issue.

This requires the cloud infrastructure reconfiguration and ask cloud admins to maintain subnet tags in shared external networks. This is quite problematic in public clouds.

lingxiankong:
> @lingxiankong
> 
> > First, I want to make sure that in your cloud, is normal user able to get subnets of an external network? I'm asking because in devstack (also in our cloud), that's not allowed.
> 
> Normal user can allocate FIPs. External FIP network can contain multiple subnets. One subnet can be an internal one, another can contain a public subnet. When public subnets get exhausted, cloud admin can add a new public subnet (with new uuid).
> Currently, in case when k8s user wants to allocate public FIPs by default, he must reconfigure OCCM with a new subnet uuid. The aim of the PR is to automate this process by using the subnet name patterns.

I was talking about the policy actually, can you paste the output of the following commands running as a normal user?

```
$ openstack subnet show <a subnet ID in an external network>
$ openstack floating ip create <external network ID> --subnet <the subnet ID in an external network>
```

> 
> > Additionally, I still prefer to use subnet tag instead of a name pattern, as suggested in the issue.
> 
> This requires the cloud infrastructure reconfiguration and ask cloud admins to maintain subnet tags in shared external networks. This is quite problematic in public clouds.

I don't quite understand why it's problematic in public clouds? I don't see there is much difference between maintaining subnet names and tags. The process of using subnet name pattern proposed in this PR is:

1. The cloud admin create subnet in external network by specifying a name following some pattern.
2. The k8s cluster admin config that name pattern in OCCM configuration
3. OCCM is able to automatically get subnet by the name pattern and allocate floating IPs.

Using tags:

1. The cloud admin create subnet in external network by specifying a tag.
2. The k8s cluster admin config that tag in OCCM configuration
3. OCCM is able to automatically get subnet filtering by tag and allocate floating IPs.

Even if you have already created extra subnets in external network, the only thing that cloud admin should do is to set a tag for the subnet and the cluster admin could config the tag for OCCM.

Could you please let me know what issues you are seeing regarding to the above process?

I don't prefer the name pattern because that's not the recommended way in openstack to query resources, but resource tag is designed for that.

mandelsoft:
I've added the tag support now. For our environment the IT department prefers the selection by name. Therefore I kept the support for name patterns. But all this is optional. You can choose to use tags or a complete subnet name or the given name may be a pattern or you don't specify any subnet kind of filtering (only the floating pool). You can even specify both.
This works identically for the global config of the OCCM, the configured loadbalancer classes or the service annotation based configuration (which basically always allows to configure unexpected by any K8S user, so I would prefer to be able to completely disable which kind of configuration possibility). But only one of those specs is used. There is no merge of the config and annotation based configuration (except for the used floating pool).

kayrus:
> I was talking about the policy actually, can you paste the output of the following commands running as a normal user?

A regular user can do this. And this approach is already coded in OCCM code: https://github.com/kubernetes/cloud-provider-openstack/blob/da85a2c67f283b1b48f118c360fbc00b326e4915/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go#L960-L980

> I don't quite understand why it's problematic in public clouds? I don't see there is much difference between maintaining subnet names and tags.

FIP networks belong to cloud admin scope and shared with users with read only access (with an ability to allocate FIP). External networks are usually created during a cloud creation at the very beginning. Forcing to use only tags will force cloud admins to  maintain tags for a big amount of external network subnets, which were created years ago.

pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
```diff
@@ -110,6 +113,121 @@ type LbaasV2 struct {
...
+		match = regexpNameMatcher(glob.Globexp(pat))
+	}
+	if not {
+		match = negate(match)
+	}
+	return match, nil
+}
+
+// subnetTagMatcher matches a subnet by a given tag spec
+func subnetTagMatcher(tag string) matcher {
```
kayrus, line 176: gophercloud supports tags filtering using the neutron API: https://github.com/gophercloud/gophercloud/blob/bb95efe2e1e7225d263125171bed6f10770605ea/openstack/networking/v2/subnets/requests.go#L37-L40

pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
```diff
@@ -2069,6 +2121,23 @@ func (lbaas *LbaasV2) ensureLoadBalancer(ctx context.Context, clusterName string
...
+	}
+	subs, err := subnets.ExtractSubnets(allPages)
+	if err != nil {
+		return nil, fmt.Errorf("error extracting subnets from pages: %v", err)
+	}
+
+	if len(subs) == 0 {
+		return nil, fmt.Errorf("could not find subnets for network %s", networkID)
+	}
+	return subs, nil
```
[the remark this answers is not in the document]
kayrus: gophercloud supports sorting via neutron API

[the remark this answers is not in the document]
mandelsoft: Sorting is not necessary, because a) any matching subnet can be chosen and b) when deleting LBs again, there will be always any number of subnets wuth free IPs

pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
mandelsoft: I added the tag filtering using the REST API

mandelsoft:
I added the tag filtering support of the REST API. Now is its possible to use tag lists in and/or mode and as for the patterns negated, also

mandelsoft:
@lingxiankong The cinder tests fail, but I haven't done any changes there. Some problems look like errors in shell scripts used for testing.

kayrus:
recheck

kayrus:
retest

pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
```diff
@@ -2886,3 +3152,38 @@ func GetLoadBalancerSourceRanges(service *corev1.Service) (netsets.IPNet, error)
 	}
 	return ipnets, nil
 }
+
+// PreserveGopherError preserves the error details delivered with the response
+// that are explicitly omitted by dedicated error types.
+// error types from provider_client.go
+func PreserveGopherError(rawError error) error {
```
kayrus, line 3159: @mandelsoft in case when gophercloud gets reauthenticated and it receives a 400 or 409 error on FIP allocation, you need to asser an error to `gophercloud.ErrErrorAfterReauthentication`, then extract the original error message:

```go
                if v, ok := err.(gophercloud.ErrErrorAfterReauthentication); ok {
                        err = v.ErrOriginal
                }
                // do further error assertion
```

pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go
```diff
@@ -2886,3 +3152,38 @@ func GetLoadBalancerSourceRanges(service *corev1.Service) (netsets.IPNet, error)
...
+	switch e := rawError.(type) {
+	case gophercloud.ErrDefault400:
+	case gophercloud.ErrDefault401:
+		details = e.Body
+	case gophercloud.ErrDefault403:
+	case gophercloud.ErrDefault404:
+		details = e.Body
+	case gophercloud.ErrDefault405:
+		details = e.Body
+	case gophercloud.ErrDefault408:
```
kayrus, line 3173: Don't you want to add a support for `gophercloud.ErrDefault409`? FIP allocation may return two types of code with a body:

400:
```
{
  "NeutronError": {
    "detail": "",
    "message": "Unable to find any IP address on external network 99e8a528-0b85-419a-9f7f-ee59a92dc7a2.",
    "type": "ExternalIpAddressExhausted"
  }
}
```

409:
```
{
  "NeutronError": {
    "detail": "",
    "message": "No more IP addresses available on network 99e8a528-0b85-419a-9f7f-ee59a92dc7a2.",
    "type": "IpAddressGenerationFailure"
  }
}
```

kayrus: See how it's done in https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1163/files
and https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1171/files

mandelsoft: Error 409 implementation already does not hide the error details. I've added only those errors here, where its implementation overrides the informative base Error implementation with a simple text hiding the details (what even the reason for this might be).

Should I add all errors here, regardless of its standard implementation?

mandelsoft: Yes, I added the others also, with a default behaviour, therefore I added now the missed 409

pkg/openstack/loadbalancer.go
```diff
@@ -1324,37 +1554,50 @@ func (lbaas *LbaasV2) checkService(service *corev1.Service, nodes []*corev1.Node
...
-						klog.Warningf("Failed to find floating-subnet-id for Service %s: %v", serviceName, err)
-					} else {
-						floatingSubnetID = lbSubnet.ID
-					}
-				}
+		// apply defaults from CCM config
+		if floatingNetworkID == "" {
+			floatingNetworkID = lbaas.opts.FloatingNetworkID
+		}
+		if !floatingSubnet.Configured() {
```
lingxiankong, line 1573: This can be combined with Line1557

pkg/openstack/loadbalancer.go
```diff
@@ -973,20 +1168,56 @@ func (lbaas *LbaasV2) getServiceAddress(clusterName string, service *corev1.Serv
...
 			}
-			if svcConf.lbPublicSubnetID != "" {
-				floatIPOpts.SubnetID = svcConf.lbPublicSubnetID
-			}
-			if loadBalancerIP != "" {
-				floatIPOpts.FloatingIP = loadBalancerIP
-			}
 
-			klog.V(4).Infof("Creating floating ip with opts %+v", floatIPOpts)
+			if loadBalancerIP == "" && svcConf.lbPublicSubnetSpec.MatcherConfigured() {
```
lingxiankong, line 1172: The condition `loadBalancerIP == ""` is always true here, please remove.

pkg/openstack/loadbalancer.go
```diff
@@ -973,20 +1168,56 @@ func (lbaas *LbaasV2) getServiceAddress(clusterName string, service *corev1.Serv
...
+				var found_subnet subnets.Subnet
+				// tweak list options for tags
+				subnets, err := lbaas.listSubnetsForNetwork(svcConf.lbPublicNetworkID, svcConf.lbPublicSubnetSpec.TweakListOptsFunction())
```
lingxiankong, line 1175: Variable 'subnets' collides with imported package name

pkg/openstack/loadbalancer.go
```diff
@@ -973,20 +1168,56 @@ func (lbaas *LbaasV2) getServiceAddress(clusterName string, service *corev1.Serv
...
-			if svcConf.lbPublicSubnetID != "" {
-				floatIPOpts.SubnetID = svcConf.lbPublicSubnetID
-			}
-			if loadBalancerIP != "" {
-				floatIPOpts.FloatingIP = loadBalancerIP
-			}
 
-			klog.V(4).Infof("Creating floating ip with opts %+v", floatIPOpts)
+			if loadBalancerIP == "" && svcConf.lbPublicSubnetSpec.MatcherConfigured() {
+				var found_subnet subnets.Subnet
```
lingxiankong, line 1173: Please use camelcase instead of snake case

pkg/openstack/loadbalancer.go
```diff
@@ -2910,3 +3176,42 @@ func GetLoadBalancerSourceRanges(service *corev1.Service) (netsets.IPNet, error)
 	}
 	return ipnets, nil
 }
+
+// PreserveGopherError preserves the error details delivered with the response
+// that are explicitly omitted by dedicated error types.
+// error types from provider_client.go
+func PreserveGopherError(rawError error) error {
```
lingxiankong, line 3183: What's the problem does this function solve? Any examples?

lingxiankong (changes requested):
* I don't think the definition `type TweakSubNetListOpsFunction` is very useful but only makes the code hard to read. Could you please simplify `type floatingSubnetSpec` a little bit?
* Basically, in `type floatingSubnetSpec`, we need a method called something like `GetAvailableSubnets()` which should consider the tags and name pattern and return a list of subnets.

pkg/openstack/loadbalancer.go
mandelsoft: I'm not sure about the control flow outside the function, but here it might be possible for `loadbalancerIP` to be empty. This is because the earlier `if` block `if floatIP == nil && loadBalancerIP != "" {` might be left without a return. Therefore the following `if` block `if floatIP == nil {` might be entered with a non-empty `loadbalancerIP`.

pkg/openstack/loadbalancer.go
mandelsoft: I'll replace this with `foundSubnets`

pkg/openstack/loadbalancer.go
mandelsoft: I added a more detailed comment and an example

pkg/openstack/loadbalancer.go
mandelsoft: I'll do this, but it makes the control flow less clear, because it mashes up the various defaulting steps.

mandelsoft:
@lingxiankong I've reworked the subnet filter spec to provide a method for getting a filtered subnet list. Concerning the `loadBalancerIP` proposal. please see my comment to your topic.

mandelsoft:
The failing check complains about the wrong kubernetes version:
```
  Detected go version: go version go1.15 linux/amd64.
  Kubernetes requires go1.16.0 or greater.
  Please install go1.16.0 or later.
```

mandelsoft:
> * I don't think the definition `type TweakSubNetListOpsFunction` is very useful but only makes the code hard to read. Could you please simplify `type floatingSubnetSpec` a little bit?
> * Basically, in `type floatingSubnetSpec`, we need a method called something like `GetAvailableSubnets()` which should consider the tags and name pattern and return a list of subnets.

I've reworked the subnet filtering as proposed, by using the filter spec as base to get a filtered list of subnets matching the selection criteria.

lingxiankong:
/test cloud-provider-openstack-acceptance-test-lb-octavia

The job cloud-provider-openstack-acceptance-test-e2e-conformance-stable-branch-v1.18 may have some problem.

lingxiankong:
don't worry about `cloud-provider-openstack-acceptance-test-e2e-conformance-stable-branch-v1.18` for now.

lingxiankong:

lingxiankong:
Simplified the release note

lingxiankong:
/lgtm

I will probably do a follow up patch for some minor improvements.

dkistner:
Thank you @lingxiankong for reviewing this PR. 
What's the next step? Can we merge it or is another review required?

lingxiankong:
We need review from another maintainer.

pkg/openstack/loadbalancer.go
ramineni: @mandelsoft  I suppose this func could be moved to https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/util/errors/errors.go

ramineni (commented):
@mandelsoft Thanks for the PR. Have minor suggestion inline. 
Approving this, later could be improved in followup PRs 
/approve

PR #1382, opened 2021-01-21
[cinder-csi-plugin] Fix pagination support for ListVolumes/ListSnapshots

### LINKED ISSUE #1383: [cinder-csi-plugin] ListVolumes pagination doesn't work as expected

>
/kind bug
> /kind feature

**What happened**:

The #969 PR introduced the pagination support, however the pagination didn't make sense, since the `AllPages()` method was used, which automatically fetches all pages.
This PR uses a proper pagination and requests only one page, when the `MaxItems` limit is provided.

When `ListVolumes` is called with a limit, the function should return max the limited amount of volumes.

**How to reproduce it**:

Not possible. Limit was introduced as a spec and not used by any CSI controllers. However this bug can appear, when controllers start to set the limit option.

**Environment**:
- OpenStack version:
- Others:

kayrus:

The #969 PR introduced the pagination support, however the pagination didn't make sense, since the `AllPages()` method was used, which automatically fetches all pages.
This PR uses a proper pagination and requests only one page, when the `MaxItems` limit is provided.

fixes #1383

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/openstack/openstack_volumes.go
+++ b/pkg/csi/cinder/openstack/openstack_volumes.go
@@ -26,6 +26,7 @@ import (
 	volumeexpand "github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/volumeactions"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/volumeattach"
+	"github.com/gophercloud/gophercloud/pagination"
 	"k8s.io/apimachinery/pkg/util/wait"
 	cpoerrors "k8s.io/cloud-provider-openstack/pkg/util/errors"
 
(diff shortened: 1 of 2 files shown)


jichenjc:
@kayrus maybe you want to create another issue to track the bug and link this PR to that new issue?

jichenjc:
and I don't know whether we have any test cases related to this ,as this might need some kind of UT/CI cover? not a mandatory request , just curious

kayrus:
> and I don't know whether we have any test cases related to this ,as this might need some kind of UT/CI cover? not a mandatory request , just curious

I made a research and I didn't find that the `MaxItems` is set to something in CSI attacher. It is just a spec, which is not yet enabled: https://github.com/container-storage-interface/spec/blob/baa71a34651e5ee6cb983b39c03097d7aa384278/lib/go/csi/csi.pb.go#L2310
You can assume that this is a proactive bugfix.

As for the test cover, it is quite problematic, because we're talking about the pagination in gophercloud and pagination in CSI driver. For functional tests we need to have ~10 volumes and have an ability to set the limit, for instance to 1 and ensure that `ListVolumes` returns exactly 1 volume, not more.

dkistner (approved):
/lgtm

jichenjc:
/lgtm

I am ok to just fix it and I think later on we need find some way to improve the test cases...

ramineni:
@kayrus Thanks for updating . could you update ListSnapshots as well.
And also, how did you test this if its working fine?

kayrus:
@ramineni I tested the behavior in a small [go app](https://gist.github.com/kayrus/f03c3bcb5b803f8a5df7d572a859aa23).
It reproduces the issue, when limit is set to 1.

I fixed the `ListSnapshots` as well.

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -83,22 +84,37 @@ func (os *OpenStack) CreateVolume(name string, size int, vtype, availability str
...
+
+		if nextPageURL != "" {
+			queryParams, err := url.ParseQuery(nextPageURL)
+			if err != nil {
+				return false, err
+			}
 			nextPageToken = queryParams.Get("marker")
 		}
+
+		return false, nil
```
ramineni, line 112: @kayrus trying to understand here, dont we need to return true here? false will come out of List Block?

kayrus: correct. if we return true, then the ForEach loop will be continued

kayrus: https://pkg.go.dev/github.com/gophercloud/gophercloud/pagination#Pager.EachPage
```
EachPage iterates over each page returned by a Pager, yielding one at a time to a handler function. Return "false" from the handler to prematurely stop iterating.
```

ramineni: @kayrus Thanks. understood. you might have tested , but just to confirm

1. From the attacher , MaxEntries value can be 0, if not set. So when limit is 0, it by default returns all the items in the list right?
2. If NextPageURL returns error, then also it should be ok to return volumes.
    Like instead of L114-116 , we can directly return L118
          return vols, nextPageToken, err

kayrus: You're right, by default the limit is 0, in this case openstack API will return all items without a pagination.
If `NextPageURL` returns an error (malformed json), then an error should be returned. The same error will be returned even earlier, when `ExtractSnapshots` is called.

ramineni: right, if ExtractVolumes() return error , anyway vols value would be nil/empty. So that statement should suffice. Either way should be fine I suppose.

ramineni:
@kayrus Thanks. Looks good. waiting for tests pass

kayrus:
@ramineni ready for a merge

PR #1390, opened 2021-01-29
[csi-plugin] Bump CSI spec to 1.3

jichenjc:

bump CSI to sepc 1.3 

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/driver_test.go
+++ b/pkg/csi/cinder/driver_test.go
@@ -28,7 +28,7 @@ const (
 )
 
 var (
-	vendorVersion = "1.2.2"
+	vendorVersion = "1.3.0"
 )
 
 func NewFakeDriver() *CinderDriver {
--- a/docs/cinder-csi-plugin/using-cinder-csi-plugin.md
+++ b/docs/cinder-csi-plugin/using-cinder-csi-plugin.md
@@ -31,7 +31,7 @@ The Cinder CSI Driver is a CSI Specification compliant driver used by Container 
 
 ## CSI Compatibility
 
-This plugin is compatible with CSI versions v1.2.0 , v1.1.0, and v1.0.0
+This plugin is compatible with CSI versions v1.3.0, v1.2.0 , v1.1.0, and v1.0.0
 
 ## Downloads
 
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.15
 
 require (
 	github.com/NYTimes/gziphandler v1.1.1 // indirect
-	github.com/container-storage-interface/spec v1.2.0
+	github.com/container-storage-interface/spec v1.3.0
 	github.com/coreos/go-systemd v0.0.0-20190620071333-e64a0ec8b42a // indirect
 	github.com/emicklei/go-restful v2.9.6+incompatible // indirect
 	github.com/golang/protobuf v1.4.3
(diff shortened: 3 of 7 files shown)


bzhaoopenstack:
for the new test image ubuntu-focal, if some specific test jobs need reply on docker, then the docker version need to upgrade also.. see https://download.docker.com/linux/ubuntu/dists/focal/pool/stable/amd64/ , the lowest docker version is 19.03， but the job configured is still 18.09

lingxiankong:
/override openlab/cloud-provider-openstack-acceptance-test-keystone-authentication-authorization

I'm pretty sure all the changes here are not affecting k8s-keystone-auth binary, so override the job for now.

pkg/csi/cinder/driver_test.go
```diff
@@ -28,7 +28,7 @@ const (
 )
 
 var (
-	vendorVersion = "1.2.2"
+	vendorVersion = "1.3.0"
```
lingxiankong, line 31: I'm curious what's the best practice (for developers) to update `specVersion`, `Version` and `vendorVersion`? Can we document that either in this PR or in a separated one?

jichenjc: that's a good suggestion , I will create one for this

jichenjc: please note vendorVersion is only used in test...

lingxiankong: Yeah, I know, but what is it used for? When should a developer bump it?

jichenjc: ok, will document in the DOC PR as well

ramineni: @lingxiankong we bump it for each release without fail and also if any major feature/breaking change added in the plugin, then also it needs a bump (but till now we didnt follow diligently for each major feature addition). But ya , we bump the version for each release

jichenjc: @ramineni so what's your opinion for this PR? we can talk about future stuffs in that doc PR

ramineni: @jichenjc could you let me know what is the advantage of bumping the spec version ?  looks like we are still not supporting the extra functions that are provided by latest spec.

ramineni:
@jichenjc I did glance through the PR, From the PR , I couldn't see any added advantage of bumping the spec version as we are still not supporting the extra functions that are provided by latest spec.  Have to check what highlights that are provided by latest spec and what is changed from existing behavior. If you could highlight in the PR that would be good . I prefer we bump the spec version , when we support its functionalities.

/hold 
need more discussion

ramineni:
@jichenjc I did investigate the highlights of spec 1.3.0 , apart from additional call, there doesnt seem to be much of breaking changes. And also, the added call seems to be alpha feature , so Im not sure if its worth adding the implementation or leaving unimplemented. Thats required only if GET_VOLUME capability is  exposed anyway.

But still Im confused , whether we have to bump vendor version in the driver , eventhough we are not supporting the additional calls or leave it as it is. 

/hold cancel

jichenjc:
@ramineni my thought is we should catch up the latest version if no breaking changes (I think those spec should not bring break changes if it's a y version update e.g 1.2->1.3) 

I think we can bump the version and telling user we now support v1.3 spec though no big changes in our side or we don't even have no implemenation for it , just for sync and fit for our version definition rule..

ramineni:
/approve
Just for sync , no major updates.

PR #1399, opened 2021-02-04
[cinder-csi-plugin] Use values to configure livenessprobe in Helm chart

### LINKED ISSUE #1387: [cinder-csi-plugin] Apparently-completed connections to Keystone and Cinder fail with "i/o timeout"
> /kind bug

**What happened**:
Some of the time, requests to the OpenStack APIs are successful, and with plugin debug logging turned way up I can see that the plugin is sometimes able to connect to Keystone, get a token, and then connect to Cinder to list and create volumes. However, most of the time connections to either Keystone or Cinder will appear to retry several times at 10-second intervals, and eventually fail with a message like:

I0120 17:39:17.616376 1 openstack.go:581] OpenStack connection error, retries exhausted. Aborting
W0120 17:39:17.616511 1 main.go:108] Failed to GetOpenStackProvider: Post "https://<CLOUD_DOMAIN>:5000/v3/auth/tokens": OpenStack connection error, retries exhausted. Aborting. Last error was: dial tcp <CLOUD_IP>:5000: i/o timeout

For connections to the OpenStack API services to succeed as they do with other OpenStack clients, even with cURL from inside the cinder-csi-plugin containers.

**How to reproduce it**:
Just starting the plugin will cause the failure condition, almost immediately. This will result in containers exiting and going into CrashLoopBackoff, although enough connections succeed that if left to restart itself for long enough, volume actions will eventually get performed.

This result occurs with both POSTs and GETs, and requests that often fail still do sometimes succeed.

Since I'm also in control of the OpenStack cloud infrastrucutre, I've been able to confirm in the API server logs that the requests which are being retried have in fact been received and responded to by the API servers, all with 2xx or 3xx response codes and in response times well under 10 seconds (most GETs are under 1 second), which leads me to think that there's not a genuine timeout condition here.

Basic connectivity testing with cURL inside the container where the plugin is running confirms that there's no difficulty or delay in querying the API server endpoints; tcpdump on the Kubernetes worker node hosting the container shows complete TCP conversations between the container and the API servers, with no indication of RSTs or broken connections.

Because I initially thought this must be an issue with the underlying gophercloud implementation that cinder-csi-plugin relies on, I first reported this bug on that project (https://github.com/gophercloud/gophercloud/issues/2100) and had some good conversation with the project team about where this problem might arise, which ultimately ended up in the suggestion to open an issue on this repository.

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: 1.20
- OpenStack version: Stein

jralbert:
This PR makes the CSI plugin livenessprobe configurable via values in the Helm chart that deploys it, making it easier to work with 

fixes #1387 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/cinder-csi-plugin/values.yaml
+++ b/charts/cinder-csi-plugin/values.yaml
@@ -28,6 +28,10 @@ csi:
       repository: k8s.gcr.io/sig-storage/livenessprobe
       tag: v2.1.0
       pullPolicy: IfNotPresent
+    failureThreshold: 5
+    initialDelaySeconds: 10
+    timeoutSeconds: 3
+    periodSeconds: 10
   nodeDriverRegistrar:
     image:
       repository: k8s.gcr.io/sig-storage/csi-node-driver-registrar
--- a/charts/cinder-csi-plugin/values.yaml
+++ b/charts/cinder-csi-plugin/values.yaml
@@ -30,8 +30,8 @@ csi:
       pullPolicy: IfNotPresent
     failureThreshold: 5
     initialDelaySeconds: 10
-    timeoutSeconds: 3
-    periodSeconds: 10
+    timeoutSeconds: 10
+    periodSeconds: 60
   nodeDriverRegistrar:
     image:
       repository: k8s.gcr.io/sig-storage/csi-node-driver-registrar
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: latest
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 1.3.1
+version: 1.3.2
 home: https://github.com/kubernetes/cloud-provider-openstack
 icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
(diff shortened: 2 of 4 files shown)


charts/cinder-csi-plugin/values.yaml
```diff
@@ -28,6 +28,10 @@ csi:
       repository: k8s.gcr.io/sig-storage/livenessprobe
       tag: v2.1.0
       pullPolicy: IfNotPresent
+    failureThreshold: 5
+    initialDelaySeconds: 10
+    timeoutSeconds: 3
+    periodSeconds: 10
```
jichenjc, line 34: I think https://github.com/kubernetes/cloud-provider-openstack/pull/1395/files
is proposing to update those values 
how about make them sync  ? @ramineni @jralbert

ramineni: @jichenjc yes, they should be in sync. 10s perodicity might be overhead as it does API call as well. Appreciate others opinions as well , on what could be optimum default values. @lingxiankong @jralbert ^^

jralbert: It certainly seems right that the defaults should match up between the manifests and the charts. As for optimum default values, I suppose that depends a bit on what common deployment scenarios look like. The OpenStack clouds that I've run have generally been pretty sluggish on API requests, especially those that require referencing multiple resources, so I'd definitely agree with defaults that are a bit more forgiving than the current values. 10 seconds of timeout within a 60 second period seems pretty reasonable; I'll adjust this PR to match @ramineni's.

ramineni:
Thanks.
/lgtm
/approve

PR #1411, opened 2021-02-12
[cinder-csi-plugin] Removed CI Job cloud-provider-openstack-acceptance-test-csi-cinder

chrigl:

We run
* cloud-provider-openstack-e2e-test-csi-cinder
* cloud-provider-openstack-multinode-csi-migration-test

in the same cases, we don't get any benefit out of
cloud-provider-openstack-acceptance-test-csi-cinder, but we must compile
the binaries multiple times, as well as creating another kubernetes for
only a single testcase. This exact testcase is covered by the mentioned
ones.

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -51,27 +51,6 @@
               - ^OWNERS$
               - ^SECURITY_CONTACTS$
               - ^.gitignore$
-    cloud-provider-openstack-acceptance-test-csi-cinder:
-      jobs:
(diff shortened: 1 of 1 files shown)


ramineni:
Thanks. please take care of removing from openlab as well.
/lgtm
/approve

lingxiankong:
> Thanks. please take care of removing from openlab as well.
> /lgtm
> /approve

Please don't remove from openlab, the stable branches are still using it.

PR #1414, opened 2021-02-22
WIP: Add readOnly mode support for attachments.

### LINKED ISSUE #1169: [cinder-csi-plugin] Add Support to readonly volumes
**Is this a BUG REPORT or FEATURE REQUEST?**: /kind feature

>
> /kind bug
> /kind feature

**What happened**:
Currently there is no support of making volume readonly. This feature is to enable the same.

**How to reproduce it**:

**Environment**:
- OpenStack version:
- Others:

ramineni:

fixes #1169

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/openstack/openstack_volumes.go
+++ b/pkg/csi/cinder/openstack/openstack_volumes.go
@@ -24,6 +24,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/apiversions"
 	volumeexpand "github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/volumeactions"
+	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/attachments"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/volumeattach"
 	"github.com/gophercloud/gophercloud/pagination"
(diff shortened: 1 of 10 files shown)


pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -186,14 +187,23 @@ func (os *OpenStack) AttachVolume(instanceID, volumeID string) (string, error) {
...
-	_, err = volumeattach.Create(computeServiceClient, instanceID, &volumeattach.CreateOpts{
+	vAttachRes, err := volumeattach.Create(computeServiceClient, instanceID, &volumeattach.CreateOpts{
 		VolumeID: volume.ID,
 	}).Extract()
 
 	if err != nil {
 		return "", fmt.Errorf("failed to attach %s volume to %s compute: %v", volumeID, instanceID, err)
 	}
 
+	if readOnly {
```
jichenjc, line 198: not sure why we can't add this param when create volume attach at l 190?

ramineni: @jichenjc Its not supported

jichenjc: it's supported in cinder
https://docs.openstack.org/api-ref/block-storage/v3/?expanded=update-an-attachment-detail,create-attachment-detail#create-attachment

so maybe a follow up PR can go to gophercloud to add this param for create?

jichenjc: https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/v3/attachments/requests.go#L31

looks like it support ro/rw already?

ramineni: @jichenjc we are using call compute API call for volume attach . Pls check L190. ITs not supported there .
And also , its not recommended to use blockstorage api directly for volume attach as you mentioned for volume attach (that should be internally called by nova), because that misses the update in nova DB . So we need to add additional call for updating the mode. Hope its clear now.

jichenjc: ok, I need check compute calls and see whether we can add some params
IMHO it should create with param instead of update later on..

jichenjc: ok, read the code again and seems we are using nova api to create the attachment 
and use cinder api to update the attachment..  not sure this is desired way but this looks like what it is
and no API of nova side to specify the ro now

ramineni:
closing this for now as it has dependency on openstack fix.
/close

PR #1431, opened 2021-03-08
[occm] Move CCM rbac to manifests/

ramineni:
This PR moves the required RBAC for OCCM to new location manifests/controller-manager 
And also delete the rbac for pvl-controller, which is not required.

/kind cleanup 

fixes #

**Release note**:

```release-note
NONE
```


manifests/controller-manager/cloud-controller-manager-roles.yaml
```diff
@@ -10,8 +10,10 @@ items:
     resources:
     - leases
     verbs:
-    - get
     - create
+    - get
+    - list
```
jichenjc, line 15: THe commit message said it's move from somewhere ? so where its original place and why not remove them?
or if we add those list and watch, what's the reason behind it?

ramineni:
@lingxiankong @jichenjc PTAL

manifests/controller-manager/cloud-controller-manager-roles.yaml
lingxiankong: @jichenjc Is your concern resolved?

PR #1449, opened 2021-03-18
[OCCM] Set Hostname in load balancer status when using PROXY protocol

### LINKED ISSUE #1287: [occm] PROXY protocol breaks connections from inside the cluster

lingxiankong:
Related issues in other repos:
- https://github.com/kubernetes/kubernetes/issues/66607
- https://github.com/kubernetes/ingress-nginx/issues/3996

In PROXY mode, set Hostname field using IP address and a suffix `.nip.io` to expose the Service of LoadBalancer type.

A new config option `enable-ingress-hostname` is introduced but may be removed in the future, please see the document change.

fixes #1287

**Release note**:

```release-note
[openstack-cloud-controller-manager] Fixed the broken header issue when accessing the load balancer service with PROXY protocol enabled from within the cluster.
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -107,6 +107,9 @@ const (
 	// ServiceAnnotationLoadBalancerEnableHealthMonitor defines whether or not to create health monitor for the load balancer
 	// pool, if not specified, use 'create-monitor' config. The health monitor can be created or deleted dynamically.
 	ServiceAnnotationLoadBalancerEnableHealthMonitor = "loadbalancer.openstack.org/enable-health-monitor"
+
+	// See https://nip.io
+	defaultProxyHostnameSuffix = "nip.io"
 )
 
 // LbaasV2 is a LoadBalancer implementation for Neutron LBaaS v2 API
@@ -1807,6 +1810,16 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 
 	status := &corev1.LoadBalancerStatus{
 		Ingress: []corev1.LoadBalancerIngress{{IP: addr}},
+	}
+
+	// If the load balancer is using the PROXY protocol, expose its IP address via
+	// the Hostname field to prevent kube-proxy from injecting an iptables bypass.
+	// This is a workaround until
+	// https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1860-kube-proxy-IP-node-binding
+	// is implemented (maybe in v1.22).
+	if svcConf.enableProxyProtocol && lbaas.opts.EnableIngressHostname {
+		fakeHostname := fmt.Sprintf("%s.%s", status.Ingress[0].IP, defaultProxyHostnameSuffix)
+		status.Ingress = []corev1.LoadBalancerIngress{{Hostname: fakeHostname}}
 	}
 
 	return status, nil
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -121,7 +121,7 @@
         - cloud-provider-openstack-acceptance-test-e2e-conformance:
             files:
               - cmd/openstack-cloud-controller-manager/.*
-              - pkg/cloudprovider/.*
+              - pkg/openstack/.*
               - pkg/util/.*
               - tests/e2e/cloudprovider/.*
               - go.mod$
@@ -138,7 +138,7 @@
         - cloud-provider-openstack-acceptance-test-e2e-conformance-stable-branch-v1.20:
             files:
               - cmd/openstack-cloud-controller-manager/.*
-              - pkg/cloudprovider/.*
+              - pkg/openstack/.*
               - pkg/util/.*
               - tests/e2e/cloudprovider/.*
               - go.mod$
@@ -155,7 +155,7 @@
         - cloud-provider-openstack-acceptance-test-e2e-conformance-stable-branch-v1.19:
             files:
               - cmd/openstack-cloud-controller-manager/.*
-              - pkg/cloudprovider/.*
+              - pkg/openstack/.*
               - pkg/util/.*
               - tests/e2e/cloudprovider/.*
               - go.mod$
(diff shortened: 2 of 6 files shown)


zetaab:
@lingxiankong if this https://github.com/kubernetes/kubernetes/pull/97681 some day arrives, it will solve this issue also

pkg/openstack/loadbalancer.go
```diff
@@ -1809,6 +1812,16 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 		Ingress: []corev1.LoadBalancerIngress{{IP: addr}},
 	}
 
+	// If the load balancer is using the PROXY protocol, expose its IP address via
+	// the Hostname field to prevent kube-proxy from injecting an iptables bypass.
+	// This is a workaround until
+	// https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1860-kube-proxy-IP-node-binding
+	// is implemented (maybe in v1.22).
+	if svcConf.enableProxyProtocol && lbaas.opts.EnableIngressHostname {
```
zetaab, line 1820: we could maybe have also annotation to override this setting?

lingxiankong: This is something I tried to avoid, as a temporary solution that will be removed in the future, the user should not be involved too much.

lingxiankong:
The `Set hostname` task in the ansible role cloud-instance-hostname is quite unstable. We'd better fix that first to make our jobs reliable.

```
2021-03-19 05:18:52.064876 | TASK [cloud-instance-hostname : Hostname | Set hostname]
2021-03-19 05:18:52.989693 | k8s-node-2 | ERROR
2021-03-19 05:18:52.990325 | k8s-node-2 | {
2021-03-19 05:18:52.990415 | k8s-node-2 |   "msg": "Command failed rc=1, out=, err=Failed to create bus connection: No such file or directory\n"
2021-03-19 05:18:52.990490 | k8s-node-2 | }
```

lingxiankong:
> The `Set hostname` task in the ansible role cloud-instance-hostname is quite unstable. We'd better fix that first to make our jobs reliable.
> 
> ```
> 2021-03-19 05:18:52.064876 | TASK [cloud-instance-hostname : Hostname | Set hostname]
> 2021-03-19 05:18:52.989693 | k8s-node-2 | ERROR
> 2021-03-19 05:18:52.990325 | k8s-node-2 | {
> 2021-03-19 05:18:52.990415 | k8s-node-2 |   "msg": "Command failed rc=1, out=, err=Failed to create bus connection: No such file or directory\n"
> 2021-03-19 05:18:52.990490 | k8s-node-2 | }
> ```

Hopefully [this](https://github.com/theopenlab/openlab-zuul-jobs/pull/1135) could help.

lingxiankong:
@jichenjc @ramineni please help review this PR

ramineni:
Thanks. changes looks ok to me .
/lgtm

jichenjc:
/approve
thanks~

multi-io:
Looks like this breaks external-dns, which sees that `status.loadBalancer.ingress[].hostname` (of the service or of an ingress resource if you use an ingress controller with the service) now contains `<IP>.nip.io` instead of just `<IP>` and as a consequence deletes the A record (removing the service from DNS) and tries to create a CNAME record, which DNS APIs (only tested Openstack Designate for now, but probably true for others as well) reject because apparently RFC 1033 forbids CNAME records with the same name as other records (and external-dns already creates a TXT record with the same name).

```
time="2021-04-12T12:40:47Z" level=info msg="Creating records: ingresstest.scratch.io-multi.de./CNAME: 185.56.132.175.nip.io."
time="2021-04-12T12:40:47Z" level=info msg="Creating records: ingresstest.scratch.io-multi.de./TXT: \"heritage=external-dns,external-dns/owner=default,external-dns/resource=ingress/ingress-test/ingress-test\""
time="2021-04-12T12:40:48Z" level=info msg="Deleting records for ingresstest.scratch.io-multi.de./A"
time="2021-04-12T12:40:48Z" level=error msg="Bad request with: [POST https://designate.cloud.syseleven.net:9001/v2/zones/7b2f1706-e0d3-4bb3-8400-7ecc78eb463c/recordsets], error message: {\"message\": \"CNAME recordsets may not share a name with any other records\", \"code\": 400, \"type\": \"invalid_recordset_location\", \"request_id\": \"req-949d736b-1e8a-4e53-a423-3048c8da55ea\"}"
```

lingxiankong:
Hi @multi-io, I'm not familiar with external-dns, but if using external-dns, I would assume the pod is using dns (rather than external IP) to access the service, so without enabling the config `enable-ingress-hostname`, that would work, right?

multi-io:
Hi @lingxiankong, external-dns just connects to some external DNS service like route53 and sets up DNS records for your ingress hostnames, pointing to the LB IP. The pod looks up those names and they resolve to the LB IP, which the pod then tries to connect to, which will fail without the new `enable-ingress-hostname` config enabled. But with this config enabled, 
the `status.loadBalancer.ingress[].hostname` of a service or ingress resource  now contains `<IP>.nip.io` instead of just `<IP>`, which causes external-DNS to try to create a CNAME record for `<IP>.nip.io` instead of just an A record for `<IP>`, and this breaks because the CNAME record clashes with a TXT record of the same name that external-dns also creates for internal purposes (storing some metadata). I've created a PR to external-DNS that circumvents this, see https://github.com/kubernetes-sigs/external-dns/pull/2049 for more information.

lingxiankong:
Thanks for the detailed clarification, I understand now.

Yes, this workaround doesn't work properly with external-dns, to fully fix this issue, we have to wait for the k8s upstream solution. I will propose a PR to add something in the doc.

PR #1452, opened 2021-03-22
[Helm manila-csi-plugin] Added priorityClassName to DaemonSet

JoaoBraveCoding:
This PR introduces the ability to configure `priorityClassName`  for the manila-csi-plugin DaemonSet
fixes : #1460
None

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/manila-csi-plugin/Chart.yaml
+++ b/charts/manila-csi-plugin/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: latest
 description: Manila CSI Chart for OpenStack
 name: openstack-manila-csi
-version: 1.1.0
+version: 1.1.1
 home: http://github.com/kubernetes/cloud-provider-openstack
 icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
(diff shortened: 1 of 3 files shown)


tombarron:
@JoaoBraveCoding It would be helpful if the commit message or a related issue explained the problem that this change is intended to solve.  IIUC priorityClass is used to influence scheduler premption.  I don't see it used for other pods in cloud provider openstack -- why do we need it for the manila csi nodeplugin?

JoaoBraveCoding:
@tombarron I can surely create an issue with a more detailed explanation on why I proposed adding priorityClass to the manila csi nodeplugin. But to summarize I've proposed this change as in our clusters we started seeing a problem with some pods of important DaemonSets like manila csi being evicted and then staying in a `Pending` state. This happened because of other pods running in the node consuming to much memmory and having a higher priorityClass, so we would like the ability to control that for the manila csi nodeplugin as we would like our scheduler to consider the pods of this Daemonset as system pods.

tombarron:
@JoaoBraveCoding What you say makes lots of sense.  If you'd be so kind as to make an issue and refer to it in the commit message, I think that will help others who might face similar issues and make the commit history a bit clearer.

tombarron (approved):
LGTM.

JoaoBraveCoding:
Done :) Thank you for the feedback!

ramineni:
/lgtm
Linked issue in PR description.

PR #1454, opened 2021-03-24
[cinder-csi-plugin] Start using --nodeid option

### LINKED ISSUE #1453: [cinder-csi-plugin] --nodeid option is ignored by the driver
/kind bug

**What happened**:
Now the binary provides an option called "[--nodeid](https://github.com/kubernetes/cloud-provider-openstack/blob/master/cmd/cinder-csi-plugin/main.go#L77-L78)" that allows to set node ID for the driver. Normally it is set to `spec.nodeName` like in this [chart](https://github.com/kubernetes/cloud-provider-openstack/blob/master/charts/cinder-csi-plugin/templates/nodeplugin-daemonset.yaml#L68-L71).
Unfortunately this value is ignored by the driver, because in reality it fetches node id from [metadata](https://github.com/kubernetes/cloud-provider-openstack/blob/5c287c5f5e4e05ab73a92c853fa5ab34b4ce795d/pkg/csi/cinder/nodeserver.go#L448).
On the other hand Manila CSI driver does it [right](https://github.com/kubernetes/cloud-provider-openstack/blob/27b6fc483451b6df2112a6a4a40a34ffc9093635/pkg/csi/manila/nodeserver.go#L314).

Because of this inconsistency we have next results:

```sh
❯ kubectl get csinode mfedosin-rgrmg-worker-0-bmwx4 -oyaml
...
spec:
  drivers:
  - name: manila.csi.openstack.org
    nodeID: mfedosin-rgrmg-worker-0-bmwx4
    topologyKeys: null
  - allocatable:
      count: 256
    name: cinder.csi.openstack.org
    nodeID: 5934efa7-8169-4490-bf10-922f159c5ad4
    topologyKeys:
    - topology.cinder.csi.openstack.org/zone
```

Either `--nodeid` option is deprecated or node id should be the same as specified in `--nodeid`.

**How to reproduce it**:
Deploy Cinder CSI driver and set `--nodeid` like in the [example](https://github.com/kubernetes/cloud-provider-openstack/blob/master/charts/cinder-csi-plugin/templates/nodeplugin-daemonset.yaml#L68-L71).
Execute `kubectl get csinode <nodename> -oyaml` to make sure that `nodeID` is different from `spec.nodeName`.

Fedosin:
Currently "--nodeid" is not used by the driver and its value is ignored. The real node ID is fetched from metadata service.
With this PR we start using value provided with --nodeid. If it's not set, then the plugin continues fetching node ID from metadata. 

fixes #1453 

**Release note**:
```release-note
--nodeid argument is now optional and used by the plugin. If it's not set, then the plugin fetches node ID from metadata.
```

### DIFF
--- a/cmd/cinder-csi-plugin/main.go
+++ b/cmd/cinder-csi-plugin/main.go
@@ -75,7 +75,6 @@ func main() {
 	cmd.Flags().AddGoFlagSet(flag.CommandLine)
 
 	cmd.PersistentFlags().StringVar(&nodeID, "nodeid", "", "node id")
-	cmd.MarkPersistentFlagRequired("nodeid")
 
 	cmd.PersistentFlags().StringVar(&endpoint, "endpoint", "", "CSI endpoint")
 	cmd.MarkPersistentFlagRequired("endpoint")
@@ -99,8 +98,6 @@ func main() {
 }
 
 func handle() {
-
-	d := cinder.NewDriver(nodeID, endpoint, cluster)
 	// Initiliaze cloud
 	openstack.InitOpenStackProvider(cloudconfig)
 	cloud, err := openstack.GetOpenStackProvider()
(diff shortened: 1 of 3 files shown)


lingxiankong:
/hold

I didn't see we have made a consensus in #1453

ramineni:
@Fedosin please check my comment here , https://github.com/kubernetes/cloud-provider-openstack/issues/1453#issuecomment-844945025

ramineni:
/close

In favor of https://github.com/kubernetes/cloud-provider-openstack/pull/1565

PR #1455, opened 2021-03-25
[cinder-csi-plugin] custom StorageClass in Helm Chart

eumel8:

The pre-defined StorageClasses won't fit for every use case. Give the user the possibility to define their own StorageClasses or VolumeSnapShotClass

none

none

**Release note**:
```release-note
[cinder-csi-plugin] Support custom StorageClass in Helm Chart.
```

### DIFF
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -3,6 +3,7 @@ appVersion: latest
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
 version: 1.3.6
+>>>>>>> bump chart version, add maintainer
 home: https://github.com/kubernetes/cloud-provider-openstack
 icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -8,3 +8,5 @@ icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
   - name: brtknr
     email: brtknr@bath.edu
+  - name: eumel8
+    email: f.kloeker@telekom.de
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: latest
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 1.3.5
+version: 1.3.6
 home: https://github.com/kubernetes/cloud-provider-openstack
 icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -8,5 +8,3 @@ icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
   - name: brtknr
     email: brtknr@bath.edu
-  - name: eumel8
-    email: f.kloeker@telekom.de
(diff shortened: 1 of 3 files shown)


charts/cinder-csi-plugin/values.yaml
```diff
@@ -70,10 +70,49 @@ secret:
 #      ca-file=/etc/cacert/ca-bundle.crt
 
 storageClass:
-  enabled: true
+  enabled: false
   delete:
```
jichenjc, line 74: is this change plans to disable the default storageclass? seems Helm chart is helpful for deploy and tets
remove this from true to false means we lost this bydefault? 

I Think anyway we can manually edit this yaml for complicated case?

jichenjc:
guess this change don't need relesae notes.

charts/cinder-csi-plugin/values.yaml
eumel8: Yes. our cloud provider offers [fixed volume_types](https://docs.otc.t-systems.com/api/evs/evs_04_2065.html) for Cinder volumes where we not able to create our own volume types to pass the default storage class provided by this Helm chart.

ramineni: @eumel8 Could we keep default as enabled: true as they work for most of the cases right.
If custom required , could be changed false and uncomment below staorageclass as per requirement?

charts/cinder-csi-plugin/values.yaml
```diff
@@ -70,10 +70,49 @@ secret:
...
+#    apiVersion: storage.k8s.io/v1
+#    kind: StorageClass
+#    metadata:
+#      name: sata
+#    provisioner: cinder.csi.openstack.org
+#    volumeBindingMode: WaitForFirstConsumer
+#    allowVolumeExpansion: true
+#    parameters:
+#      type: SATA
+#    apiVersion: snapshot.storage.k8s.io/v1beta1
```
ramineni, line 112: need to remove VolumeSnapshotClass def here?

lingxiankong: Seems what you need is to define `type` for the StorageClass, why not just add an option `volumeType`?

charts/cinder-csi-plugin/values.yaml
lingxiankong: Agree with @ramineni

lingxiankong (changes requested):

charts/cinder-csi-plugin/values.yaml
eumel8: sorry, the --- was in the wrong line

ramineni:
@eumel8 you still would need to bump chart version , lint charts seem to fail

lingxiankong:
Maybe you did an incorrect rebase, which caused your commits contain many others from the master.

eumel8:
@lingxiankong Yes, this rebase was completly wrong, sorry. Hopefuly push --force is doing the right thing now

charts/cinder-csi-plugin/Chart.yaml
```diff
@@ -2,9 +2,11 @@ apiVersion: v1
...
 name: openstack-cinder-csi
-version: 1.3.5
+version: 1.3.6
 home: https://github.com/kubernetes/cloud-provider-openstack
 icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
   - name: brtknr
     email: brtknr@bath.edu
+  - name: eumel8
+    email: f.kloeker@telekom.de
```
lingxiankong, line 12: Please don't add yourself as maintainers at this stage.

charts/cinder-csi-plugin/values.yaml
```diff
@@ -82,3 +82,42 @@ storageClass:
...
+#     allowVolumeExpansion: true
+#     parameters:
+#       type: SATA
+#     ---
+#     apiVersion: snapshot.storage.k8s.io/v1beta1
+#     kind: VolumeSnapshotClass
+#     metadata:
+#       name: csi-cinder-snapclass
+#     driver: cinder.csi.openstack.org
+#     deletionPolicy: Delete
```
lingxiankong, line 123: I would prefer to only keep one example.

lingxiankong (changes requested):

charts/cinder-csi-plugin/values.yaml
eumel8: I have now two different, so the user got an idea.

ramineni:
@eumel8 Thanks.
/lgtm

PR #1457, opened 2021-03-27
[cinder-csi-plugin] add snapshot controller to Helm Chart

eumel8:

Extends the current Helm Chart for Cinder CSI Plugin with a deployment of a Snapshot Controller

none

none

**Release note**:
```release-note
[cinder-csi-plugin] Support Snapshot Controller deployment in Helm Chart.
```

### DIFF
--- a/charts/cinder-csi-plugin/templates/snapshot-controller-rbac.yaml
+++ b/charts/cinder-csi-plugin/templates/snapshot-controller-rbac.yaml
@@ -1,0 +1,85 @@
+{{- if .Values.csi.snapshotController.enabled -}}
+# RBAC file for the snapshot controller.
+#
+# The snapshot controller implements the control loop for CSI snapshot functionality.
+# It should be installed as part of the base Kubernetes distribution in an appropriate
+# namespace for components implementing base system functionality. For installing with
+# Vanilla Kubernetes, kube-system makes sense for the namespace.
+
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: snapshot-controller-sa
+  namespace: {{ .Release.Namespace }}
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: snapshot-controller-runner
+rules:
+  - apiGroups: [""]
+    resources: ["persistentvolumes"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims"]
+    verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["storageclasses"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["list", "watch", "create", "update", "patch"]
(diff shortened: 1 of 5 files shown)


charts/cinder-csi-plugin/values.yaml
```diff
@@ -54,6 +54,11 @@ csi:
       - name: cloud-config
         mountPath: /etc/kubernetes
         readOnly: true
+  snapshotController:
+    enabled: false
+    image:
+      repository: k8s.gcr.io/sig-storage/snapshot-controller
+      tag: v3.0.0
```
ramineni, line 61: It should be 2.1.3 , should be compatible with csi-snapshotter version .

charts/cinder-csi-plugin/templates/snapshot-controller-rbac.yaml
```diff
@@ -0,0 +1,85 @@
+{{- if .Values.csi.snapshotController.enabled -}}
+# RBAC file for the snapshot controller.
+#
+# The snapshot controller implements the control loop for CSI snapshot functionality.
+# It should be installed as part of the base Kubernetes distribution in an appropriate
+# namespace for components implementing base system functionality. For installing with
+# Vanilla Kubernetes, kube-system makes sense for the namespace.
+
+apiVersion: v1
```
ramineni, line 9: To avoid any issues, pls make sure this RBAC is compatible with version used. 2.1.3 . For example, https://github.com/kubernetes-csi/external-snapshotter/blob/v2.1.3/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml

charts/cinder-csi-plugin/templates/snapshot-controller-rbac.yaml
```diff
@@ -0,0 +1,85 @@
+{{- if .Values.csi.snapshotController.enabled -}}
+# RBAC file for the snapshot controller.
+#
+# The snapshot controller implements the control loop for CSI snapshot functionality.
+# It should be installed as part of the base Kubernetes distribution in an appropriate
+# namespace for components implementing base system functionality. For installing with
+# Vanilla Kubernetes, kube-system makes sense for the namespace.
```
ramineni, line 7: May be its better to add comment here , enable this only if your Kubernetes distribution does not bundle the snapshot controller. It should be installed only once per cluster , independent of CSI driver.

charts/cinder-csi-plugin/templates/snapshot-controller-statefulset.yaml
```diff
@@ -0,0 +1,35 @@
...
+    metadata:
+      labels:
+        {{- include "cinder-csi.snapshot-controller.labels" . | nindent 8 }}
+    spec:
+      serviceAccount: snapshot-controller-sa
+      containers:
+        - name: snapshot-controller
+          image: "{{ .Values.csi.snapshotController.image.repository }}:{{ .Values.csi.snapshotController.image.tag }}"
+          args:
+            - "--v=5"
```
ramineni, line 32: please remove verbose here. Debug logging not required by default.

ramineni:
@eumel8 Thanks. looks good mostly. one comment inline.

ramineni:
Thanks.
/lgtm

PR #1459, opened 2021-03-29
[manila-csi-plugin] Add --cluster-id option 

Fedosin:

This commit add a new optional string parameter "cluster-id" to manila csi plugin binary. If this parameter value is provided, then the plugin will add "manila.csi.openstack.org/cluster: <clusterID>" to metadata of created shares. It simplifies finding shares created for specific clusters.

Similar solution exists for Cinder: https://github.com/kubernetes/cloud-provider-openstack/blob/master/cmd/cinder-csi-plugin/main.go#L86
https://github.com/kubernetes/cloud-provider-openstack/pull/451

**Release note**:
```release-note
[manila-csi-plugin] support `--cluster-id` argument to get the the CO's identifier.
```

### DIFF
--- a/docs/manila-csi-plugin/using-manila-csi-plugin.md
+++ b/docs/manila-csi-plugin/using-manila-csi-plugin.md
@@ -37,6 +37,7 @@ Option | Default value | Description
 `--with-topology` | _none_ | CSI Manila is topology-aware. See [Topology-aware dynamic provisioning](#topology-aware-dynamic-provisioning) for more info
 `--share-protocol-selector` | _none_ | Specifies which Manila share protocol to use for this instance of the driver. See [supported protocols](#share-protocol-support-matrix) for valid values.
 `--fwdendpoint` | _none_ | [CSI Node Plugin](https://github.com/container-storage-interface/spec/blob/master/spec.md#rpc-interface) endpoint to which all Node Service RPCs are forwarded. Must be able to handle the file-system specified in `share-protocol-selector`. Check out the [Deployment](#deployment) section to see why this is necessary.
+`--cluster-id` | _none_ | The identifier of the cluster that the plugin is running in. If set then the plugin will add "manila.csi.openstack.org/cluster: \<clusterID\>" to metadata of created shares.
 
 ### Controller Service volume parameters
 
--- a/cmd/manila-csi-plugin/main.go
+++ b/cmd/manila-csi-plugin/main.go
@@ -44,6 +44,7 @@ var (
 	fwdEndpoint           string
 	userAgentData         []string
 	compatibilitySettings string
+	clusterID             string
 )
 
 func validateShareProtocolSelector(v string) error {
@@ -140,6 +141,7 @@ func main() {
 					ManilaClientBuilder: manilaClientBuilder,
 					CSIClientBuilder:    csiClientBuilder,
 					CompatOpts:          compatOpts,
+					ClusterID:           clusterID,
 				},
 			)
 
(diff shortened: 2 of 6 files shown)


tombarron:
This looks reasonable, but I think it would be a good idea to get the main author of manila-csi to take a look.

@gman0 PTAL when you can.

gman0:
@Fedosin thanks for the PR! This looks good but I have a couple of notes/questions:

1. `--cluster` is a deployment-wide parameter, whereas [`appendShareMetadata`](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/manila-csi-plugin/using-manila-csi-plugin.md#controller-service-volume-parameters) that we already offer is a parameter localized to a particular `CreateVolume` request (or a StorageClass parameter). Would it make sense to make it possible to override `--cluster` with `appendShareMetadata`? I can't think of a use-case for this from the top of my head, but IMHO it's nicer to let the admins to override the metadata if they wish to do so (but letting them know with a log message like you've done already in this PR), instead of the driver telling them "hey, I just overwritten your settings -- despite what you've told me in the StorageClass". What do you think?
2. If you agree to with (1), then maybe you could add an optional `manila.csi.openstack.org/cluster` parameter to [`options.ControllerVolumeContext`](https://github.com/kubernetes/cloud-provider-openstack/blob/ef0206889cb7c39f04d215e8cbb6c69616e59ea4/pkg/csi/manila/options/shareoptions.go#L23) and, if it's empty, populate it with whatever is inside `--cluster` in [`controllerServer.CreateVolume`](https://github.com/kubernetes/cloud-provider-openstack/blob/ef0206889cb7c39f04d215e8cbb6c69616e59ea4/pkg/csi/manila/controllerserver.go#L80). This would go instead of passing the [`cluster`](https://github.com/kubernetes/cloud-provider-openstack/pull/1459/files#diff-fe927bddf3f07254d963ccb604ace68ec6d1ac08c2376263c2fbc99e61a80a15R34) variable around.
3. Either way `cluster` variable name is not very descriptive of what it contains. Maybe `clusterID / clusterIdentifier` since that's what you mention in the docs?

Fedosin:
@gman0 thanks for your review! I updated my patch as you suggested.

docs/manila-csi-plugin/using-manila-csi-plugin.md
```diff
@@ -37,6 +37,7 @@ Option | Default value | Description
 `--with-topology` | _none_ | CSI Manila is topology-aware. See [Topology-aware dynamic provisioning](#topology-aware-dynamic-provisioning) for more info
 `--share-protocol-selector` | _none_ | Specifies which Manila share protocol to use for this instance of the driver. See [supported protocols](#share-protocol-support-matrix) for valid values.
 `--fwdendpoint` | _none_ | [CSI Node Plugin](https://github.com/container-storage-interface/spec/blob/master/spec.md#rpc-interface) endpoint to which all Node Service RPCs are forwarded. Must be able to handle the file-system specified in `share-protocol-selector`. Check out the [Deployment](#deployment) section to see why this is necessary.
+`--cluster` | _none_ | The identifier of the cluster that the plugin is running in. If set then the plugin will add "manila.csi.openstack.org/cluster: \<cluster\>" to metadata of created shares.
```
gman0, line 40:
```suggestion
`--cluster-id` | _none_ | The identifier of the cluster that the plugin is running in. If set then the plugin will add "manila.csi.openstack.org/cluster: \<cluster\>" to metadata of created shares.
```

pkg/csi/manila/volumesource_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+package manila
+
+import (
+	"fmt"
+	"testing"
+)
+
+// Tests FindExportLocation with AnyExportLocation predicate
+func TestFindExportLocationAny(t *testing.T) {
```
gman0, line 22: ^^

gman0:
@Fedosin Left some comments inline, and also probably a cleaner way to this would be to take out the logic for populating the metadata map out of `volumesource`. Maybe try to populate it beforehand and then just pass the final share metadata map to `volumecreator.create()`?

gman0:
@Fedosin excellent, thanks!

gman0:
@ramineni PTAL and then help us merge this, thank you

Fedosin:
@gman0 thanks a lot for your support!

ramineni:
@Fedosin Thanks
/approve

PR #1461, opened 2021-04-07
[magnum-auto-healer] Fix k8s master node repair if it's volume-based

### LINKED ISSUE #1446: [magnum-auto-healer] Magnum auto healer cannot handle volume based master nodes

/kind bug

**What happened**:
When the k8s nodes are created based on volume, the repair will fail with the error that the root volume is still in in-use status.

The volume-based nodes can be repaired successfully

**How to reproduce it**:
Create a k8s cluster with Magnum based on volume-based nodes, then stop a kubelet process in a master node, then magnum auto healer will try to fix it and fail.

**Environment**:
- magnum-auto-healer version: v1.19.4
- OpenStack version: stable/victoria
- Others:

lingxiankong:
cherry-picked from https://github.com/kubernetes/cloud-provider-openstack/pull/1447

fixes #1446

**Release note**:

```release-note
[magnum-auto-healer] Fixed an issue that k8s master node repair failed when the node is boot from volume.
```

### DIFF
--- a/pkg/autohealing/cloudprovider/openstack/provider.go
+++ b/pkg/autohealing/cloudprovider/openstack/provider.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/blockstorage/v2/volumes"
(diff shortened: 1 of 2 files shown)


jichenjc:
weird UT test failed..
but this is backport so should be ok? 

/lgtm

PR #1473, opened 2021-04-09
[manila-csi-plugin] close GRPC connections

### LINKED ISSUE #1472: [manila-csi-plugin] GRPC connection leaks

/kind bug

**What happened**:

We've noticed increasing memory consumption in manila-csi Node plugin pods. The more RPCs any particular Node plugin pod saw, the more memory it consumed. We've observed the same issue on the partner Node plugin that manila-csi uses to perform node-local operations (in our case this was cephfs-csi).

Since both manila-csi and its partner Node plugin is affected, this suggests that the gRPC communication between them might be the culprit.

Please see the table below:

csi-memory-consumption

Node plugins that were idle kept their memory consumption low. On the other hand, node plugins that saw a lot of RPCs consumed more memory.

Node plugin memory consumption should stay constant over time.

**How to reproduce it**:

Perform a few hundreds of mounts&unmounts and see Node plugin memory consumption (`kubectl top pod ...`).

PR soon

@tombarron @gouthampacha 

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: v1.20.0
- OpenStack version:
- Others:

gman0:

This PR closes opened gRPC connections made to the partner node plugins.

Please see the table below for memory consumption comparison without and with this patch:

manila-csi-mem-consumption-comparison

Without this fix, the memory consumption gradually increases. This PR fixes this and keeps the consumption more or less constant.

fixes #1472

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/manila/driver.go
+++ b/pkg/csi/manila/driver.go
@@ -247,6 +247,7 @@ func (d *Driver) initProxiedDriver() (csiNodeCapabilitySet, error) {
 	if err != nil {
 		return nil, fmt.Errorf("connecting to %s endpoint failed: %v", d.fwdEndpoint, err)
 	}
+	defer conn.Close()
 
 	identityClient := d.csiClientBuilder.NewIdentityServiceClient(conn)
 
--- a/pkg/csi/manila/identityserver.go
+++ b/pkg/csi/manila/identityserver.go
@@ -43,6 +43,7 @@ func (ids *identityServer) Probe(ctx context.Context, req *csi.ProbeRequest) (*c
 	if err != nil {
 		return nil, status.Error(codes.FailedPrecondition, fmtGrpcConnError(ids.d.fwdEndpoint, err))
 	}
+	defer csiConn.Close()
 
 	return ids.d.csiClientBuilder.NewIdentityServiceClient(csiConn).Probe(ctx, req)
 }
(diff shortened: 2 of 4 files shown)


gman0:
@tombarron @gouthampacha @ramineni  PTAL

Fedosin (commented):
/lgtm

Fedosin:
One question: do we have a similar issue with cinder csi plugin?

tombarron (approved):
/LGTM

Thanks for the change and the data!

gman0:
@Fedosin no, cinder-csi is designed differently and doesn't rely on other csi drivers like manila-csi does, so there's no such driver-to-driver communication going on as it is in this case.

gman0:
I believe this is also worth backporting to v1.20.

tombarron:
> I believe this is also worth backporting to v1.20.

+1

gouthampacha (approved):
/lgtm /approve

Thanks for this fix, @gman0

gouthampacha:
/assign @ramineni 

Can you please take a look?

gouthampacha:
/approve


change doesn't touch files outside the cloud-provider-openstack/pkg/csi/manila path. I suppose I can do this?

ramineni:
/lgtm
/approve

Please raise a PR for back-porting the same as well

PR #1476, opened 2021-04-09
[cinder-csi-plugin] Allow multiple config files

### LINKED ISSUE #1481: [cinder-csi-plugin] Allow passing multiple config files
We would like to give the user the ability to pass arbitrary arguments to cinder-csi-plugin beyond those specified by the operator. Currently the operator specifies a 'base' configuration in a configmap, which is essentially necessary for the correct operation of the plugin in the cluster. This is currently:

```
[Global]
use-clouds = true
clouds-file = /etc/kubernetes/secret/clouds.yaml
cloud = openstack
```

The immediate driver is to give the user the ability to specify `ignore-volume-az = true`. There's an open PR adding that to openstack-cinder-csi-driver-operator here: https://github.com/openshift/openstack-cinder-csi-driver-operator/pull/29

The PR adds a second, user-owned configmap. As currently written the operator will simply pass this user-owned config in place of the 'base' config if it exists. However, there are 2 major problems with this approach:

* It requires the user to specify the base configuration as well as any configuration they need to add, which is both more complex and more error-prone for the user
* It requires the user to 'fork' the base configuration, which makes it very difficult to update should it ever change.

We considered several other approaches to the problem:

* Add parameters to the operator's config object, and have the operator generate a complete configuration

This is certainly viable. The issue is, looking at other projects which have taken this approach, e.g. libvirt for qemu, and TripleO for openstack, the result is that all options must be implemented in 2 places before they can be exposed to the user. It can also result in user confusion when the user finds the online documentation for the service they are trying to configure, but they can't directly use they find.

* Have the operator compose the base and custom configmaps into a new configmap which is used by cinder-csi-plugin

We didn't like this due to the potential confusion for the user of having 2 additional configmaps, only one of which they're supposed to edit.

* Compose the base and custom configmaps in an init container before starting cinder-csi-plugin

This is a viable option, and there's a chance we may do this as an interim solution. The issue is that it feels more complicated than necessary. As well as the additional step, and therefore potential for user/maintainer confusion, we need to package a tool for composing ini files in an image somewhere.

The solution of passing multiple configuration files directly to cinder-csi-plugin has a number of advantages:

* It's very simple to understand
* It's a common pattern, e.g. it is used by all OpenStack services
* It's very simple to implement. I have submitted a PR here: https://github.com/kubernetes/cloud-provider-openstack/pull/1476
* There's no potential for incompatibility by having config files parsed by different tools

It would be useful to OpenShift's use case, but the simplicity of it should make it useful for any deployment tooling which may want to compose cinder-csi-plugin configuration. The requirement for a 'base' configuration specific to a cluster is likely universal. Potentially layered configuration may include directly user-specified configuration as in this case, or 'zoned' configuration where certain workers have different configuration based on connectivity.

It may also make sense to make the change consistent across all services provided by CPO.

/kind feature

mdbooth:
This change allows cinder-csi-plugin to be invoked with multiple config
files. These will be parsed sequentially in the order specified. Values
in later config files override values in earlier config files.

This makes 'layered' configurations simpler to manage. For example a
base configuration containing authentication information and a per-node
override configuration which applies only to a single node. Keeping
these in separate files simplifies the task of changing the base
configuration for all nodes.

fixes #1481 

**Release note**:

```release-note
[cinder-csi-plugin] The --cloud-config option may now be given multiple times. Multiple configuration files will be merged, with later configuration files taking precedence.
```

### DIFF
--- a/cmd/cinder-csi-plugin/main.go
+++ b/cmd/cinder-csi-plugin/main.go
@@ -34,7 +34,7 @@ import (
 var (
 	endpoint    string
 	nodeID      string
-	cloudconfig string
+	cloudconfig []string
 	cluster     string
 )
 
@@ -80,7 +80,7 @@ func main() {
 	cmd.PersistentFlags().StringVar(&endpoint, "endpoint", "", "CSI endpoint")
 	cmd.MarkPersistentFlagRequired("endpoint")
 
-	cmd.PersistentFlags().StringVar(&cloudconfig, "cloud-config", "", "CSI driver cloud config")
+	cmd.PersistentFlags().StringSliceVar(&cloudconfig, "cloud-config", nil, "CSI driver cloud config. This option can be given multiple times")
 	cmd.MarkPersistentFlagRequired("cloud-config")
 
 	cmd.PersistentFlags().StringVar(&cluster, "cluster", "", "The identifier of the cluster that the plugin is running in.")
--- a/docs/cinder-csi-plugin/using-cinder-csi-plugin.md
+++ b/docs/cinder-csi-plugin/using-cinder-csi-plugin.md
@@ -6,11 +6,12 @@
   - [CSI Compatibility](#csi-compatibility)
   - [Downloads](#downloads)
   - [Kubernetes Compatibility](#kubernetes-compatibility)
+  - [Driver Deployment](#driver-deployment)
+    - [Command-line arguments](#command-line-arguments)
   - [Driver Config](#driver-config)
     - [Global](#global)
     - [Block Storage](#block-storage)
     - [Metadata](#metadata)
-  - [Driver Deployment](#driver-deployment)
     - [Using the manifests](#using-the-manifests)
     - [Using the Helm chart](#using-the-helm-chart)
   - [Supported Features](#supported-features)
@@ -42,6 +43,52 @@ Stable released version images of the plugin can be found at [Docker Hub](https:
 For each kubernetes official release, there is a corresponding release of Cinder CSI driver which is compatible with k8s release. It is recommended to use the corresponding version w.r.t kubernetes.
 
 For sidecar version compatibility with kubernetes, please refer [Compatibility Matrix](https://kubernetes-csi.github.io/docs/sidecar-containers.html) of each sidecar.
+
+## Driver Deployment
+
+You can either use the manifests under `manifests/cinder-csi-plugin` or the Helm chart `charts/cinder-csi-plugin`.
+
+### Command-line arguments
+
+In addition to the standard set of klog flags, `cinder-csi-plugin` accepts the following command-line arguments:
+
+<dl>
+  <dt>--nodeid &lt;node id&gt;</dt>
+  <dd>
+  This argument is required.
+
+  An identifier for the current node which will be used in OpenStack API calls.  This can be either the UUID or name of the OpenStack server, but note that if using name it must be unique.
+  </dd>
+
+  <dt>--endpoint &lt;endpoint&gt;</dt>
+  <dd>
+  This argument is required.
+
+  The endpoint of the gRPC server agents will use to connect to this CSI plugin, typically a local unix socket.
+
+  The manifests default this to `unix://csi/csi.sock`, which is supplied via the `CSI_ENDPOINT` environment variable.
+  </dd>
+
+  <dt>--cloud-config &lt;config file&gt; [--cloud-config &lt;config file&gt; ...]</dt>
+  <dd>
+  This argument must be given at least once.
+
+  The path to a driver config file. The format of this file is specified in [Driver Config](#driver-config).
+
+  If multiple configuration files are supplied they will be merged. In the case of a conflict, the last supplied config file will take precedence.
+
+  The manifests default this to `/etc/config/cloud.conf`, which is supplied via the `CLOUD_CONFIG` environment variable.
+  </dd>
+
+  <dt>--cluster &lt;cluster name&gt;</dt>
+  <dd>
+  This argument is optional.
+
+  The identifier of the cluster that the plugin is running in.
+
+  This will be added as metadata to every Cinder volume created by this plugin.
+  </dd>
+</dl>
 
 ## Driver Config
 
@@ -83,10 +130,6 @@ These configuration options pertain to metadata and should appear in the `[Metad
     service first if available, then the configuration drive.
 
   Influencing this behavior may be desirable as the metadata on the configuration drive may grow stale over time, whereas the metadata service always provides the most up to date view. Not all OpenStack clouds provide both configuration drive and metadata service though and only one or the other may be available which is why the default is to check both.
-
-## Driver Deployment
-
-You can either use the manifests under `manifests/cinder-csi-plugin` or the Helm chart `charts/cinder-csi-plugin`.
 
 ### Using the manifests
 
--- a/pkg/csi/cinder/openstack/openstack.go
+++ b/pkg/csi/cinder/openstack/openstack.go
@@ -90,20 +90,25 @@ func logcfg(cfg Config) {
 	klog.Infof("Block storage opts: %v", cfg.BlockStorage)
 }
 
-// GetConfigFromFile retrieves config options from file
-func GetConfigFromFile(configFilePath string) (Config, error) {
+// GetConfigFromFiles retrieves config options from file
+func GetConfigFromFiles(configFilePaths []string) (Config, error) {
 	var cfg Config
-	config, err := os.Open(configFilePath)
-	if err != nil {
-		klog.Errorf("Failed to open OpenStack configuration file: %v", err)
-		return cfg, err
-	}
-	defer config.Close()
 
-	err = gcfg.FatalOnly(gcfg.ReadInto(&cfg, config))
-	if err != nil {
-		klog.Errorf("Failed to read OpenStack configuration file: %v", err)
-		return cfg, err
+	// Read all specified config files in order. Values from later config files
+	// will overwrite values from earlier ones.
+	for _, configFilePath := range configFilePaths {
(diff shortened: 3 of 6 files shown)


jichenjc:
@mdbooth can you help to open an issue and put the reason/purpose of this PR there?
I think it's reasonable change and this might need further doc as well (not in this PR is ok)
but some info in the issue itself will be easier to track

mdbooth:
Thanks, @jichenjc. I opened https://github.com/kubernetes/cloud-provider-openstack/issues/1481 which has a much more complete description of where this is coming from and hopefully other ways it could potentially be used.

Incidentally this is my first time submitting any k8s feature, so please let me know if there's anything I can do to help you and myself.

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -124,20 +129,20 @@ func GetConfigFromFile(configFilePath string) (Config, error) {
...
 }
 
 // CreateOpenStackProvider creates Openstack Instance
 func CreateOpenStackProvider() (IOpenStack, error) {
 	// Get config from file
-	cfg, err := GetConfigFromFile(configFile)
+	cfg, err := GetConfigFromFiles(configFiles)
 	if err != nil {
-		klog.Errorf("GetConfigFromFile %s failed with error: %v", configFile, err)
+		klog.Errorf("GetConfigFromFile %s failed with error: %v", configFiles, err)
```
mdbooth, line 145: typo: GetConfigFromFile

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -124,20 +129,20 @@ func GetConfigFromFile(configFilePath string) (Config, error) {
...
-var configFile = "/etc/cloud.conf"
+var configFiles = []string{"/etc/cloud.conf"}
 var cfg Config
 
-func InitOpenStackProvider(cfgFile string) {
-	configFile = cfgFile
-	klog.V(2).Infof("InitOpenStackProvider configFile: %s", configFile)
+func InitOpenStackProvider(cfgFiles []string) {
+	configFiles = cfgFiles
+	klog.V(2).Infof("InitOpenStackProvider configFile: %s", configFiles)
```
mdbooth, line 137: typo: configFiles

cmd/cinder-csi-plugin/main.go
```diff
@@ -80,7 +80,7 @@ func main() {
 	cmd.PersistentFlags().StringVar(&endpoint, "endpoint", "", "CSI endpoint")
 	cmd.MarkPersistentFlagRequired("endpoint")
 
-	cmd.PersistentFlags().StringVar(&cloudconfig, "cloud-config", "", "CSI driver cloud config")
+	cmd.PersistentFlags().StringSliceVar(&cloudconfig, "cloud-config", nil, "CSI driver cloud config")
```
jichenjc, line 83: based on this , I think we might need upgrade CSI version defined at 
https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/cinder-csi-plugin/development.md

jichenjc (commented):
better to give an example on how to use this in the doc as well

cmd/cinder-csi-plugin/main.go
mdbooth: Yes. Now that you mention it this is also critical to the operator, as the change is backwards compatible but not forwards compatible. I've bumped the version. As it's the first version which no longer follows the CSI Spec version I've also reformatted the comment to include a version history. Having previously worked with nova I have frequently found this to be useful in practise: https://opendev.org/openstack/nova/src/commit/fa59cde51ffd761ac552b93af6c67e2432288548/nova/compute/rpcapi.py#L76

I also updated the help text of the cloudconfig param to note that it can be given multiple times.

pkg/csi/cinder/openstack/openstack_test.go
```diff
@@ -76,10 +77,41 @@ rescan-on-resize=true`
...
+	}
+
+	_, err = f.WriteString(fakeOverrideFileContent)
+	f.Close()
+	if err != nil {
+		t.Errorf("failed to write file: %v", err)
+	}
+	defer os.Remove(fakeOverrideFileName)
+
+	// expectedOpts should reflect the overriden value of rescan-on-resize. All
```
tombarron, line 106: cloud-provider-openstack-format job wants s/overriden/overridden/

mdbooth: Thanks, @tombarron ! I have not yet learned how to read that output and had missed the actual error.

tombarron: I wouldn't say I have learned either -- best I know on this log is to look for ERROR then check whether the error is ignored since it may be an expected condition.  The ansible RECAP messages seem to reveal less than with OpenStack logs.

mandre:
/retest
Only `[Fail] [sig-storage] [cinder-csi-e2e] CSI Volumes [Driver: cinder.csi.openstack.org] [Testpattern: Dynamic PV (default fs)] provisioning [It] should provision storage with pvc data source in parallel [Slow] ` failed  for [cloud-provider-openstack-e2e-test-csi-cinder](https://logs.openlabtesting.org/logs/76/1476/c14880b038d930af574a7e3388daafcb4c3f7c7b/cloud-provider-openstack-e2e-test-csi-cinder/cloud-provider-openstack-e2e-test-csi-cinder/6c1df27/) and this doesn't seem related to this patch.

lingxiankong:
Please also make some changes/amendments in https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/cinder-csi-plugin/using-cinder-csi-plugin.md#driver-config

mdbooth:
> 
> 
> Please also make some changes/amendments in https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/cinder-csi-plugin/using-cinder-csi-plugin.md#driver-config
> 
> You can do this in a separate PR if there is no comment from other reviewers.

I need to rebase anyway, so I'll do it in this PR. It may be a few days before I get back to this, though.

raildo (approved):
LGTM

docs/cinder-csi-plugin/using-cinder-csi-plugin.md
```diff
@@ -43,6 +44,52 @@ For each kubernetes official release, there is a corresponding release of Cinder
...
+  <dt>--cluster &lt;cluster name&gt;</dt>
+  <dd>
+  This argument is optional.
+
+  The identifier of the cluster that the plugin is running in.
+
+  This will be added as metadata to every Cinder volume created by this plugin.
+  </dd>
+</dl>
+
```
lingxiankong, line 92: Could you convert those using markdown? The pure HTML would make it harder to maintain.

lingxiankong:
I can confirm the failed jobs have nothing to do with this PR, from the log, the cinder-csi components are all up and running. The failure is due to the CI cloud provider issue.

@jichenjc @ramineni @chrigl we can also consider using the similar CI job I did for occm, so we could have more control of the cloud resources (in devstack) created, also benefit from the latest version Cinder features.

For this PR, please leave your comments, we shouldn't let the contributors wait for so long for the reasons they couldn't control.

docs/cinder-csi-plugin/using-cinder-csi-plugin.md
mdbooth: I can't, because AFAICT this is the 'supported' way to do description lists. I don't like it either, but I couldn't find anything better.

Do you know a better way to do this?

lingxiankong: https://raw.githubusercontent.com/kubernetes/cloud-provider-openstack/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md

Search `enable-ingress-hostname`

ramineni:
updated description to link correct issue.
@mdbooth could you add alternate manifests  as discussed in issue

mdbooth:
> updated description to link correct issue.
> @mdbooth could you add alternate manifests as discussed in issue

Will do, although it will be a few days before I get back to this.

ramineni:
@mdbooth code looks good to me. Please add alternate manifests with examples in seperate PR.
Lets just wait for one functional job run before merge.

ramineni:
/approve

@mdbooth Please update docs in seperate PR

artpej:
Yesterday csi cinder plugin was able to retrieve instance id using metadata service or config drive, today it seems that nodeid is mandatory. did i miss something?

- thank for your work -

lingxiankong:
What do you mean by "nodeid is mandatory"? From https://github.com/kubernetes/cloud-provider-openstack/pull/1565 I can see it's deprecated and shouldn't be needed.

ramineni:
@artpej Yes, Its not a mandatory/required argument and the value is not used anywhere. Please share the logs

artpej:
@lingxiankong @ramineni , thanks for your response, I just ran another try and everything works as expected (without the need to specify the `nodeid`). I should had encounter a problem with the image cache. Sorry for the inconvenience.

However the documentation in this PR specify that the `node-id` is required ( still present in the master branch https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/cinder-csi-plugin/using-cinder-csi-plugin.md#command-line-arguments ). It confused me a bit.

Finally thanks for your opensource work i really appreciate, and sorry for my poorly english speaking :wink: .

ramineni:
>> However the documentation in this PR specify that the node-id is required ( still present in the master branch https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/cinder-csi-plugin/using-cinder-csi-plugin.md#command-line-arguments ). It confused me a bit.

@artpej Thanks for pointing. Raised a PR to fix the doc. https://github.com/kubernetes/cloud-provider-openstack/pull/1573

PR #1479, opened 2021-04-12
[cinder-csi-plugin] Check volume status in createVolume

ramineni:
This commit makes sure volume status to be avaiable before returning from createVolume .
It avoids failure at later stage of controllerPublishVolume

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -136,6 +136,16 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 
 	}
 
+	// Wait for volume status to be Available
+	if vol.Status != openstack.VolumeAvailableStatus {
+		targetStatus := []string{openstack.VolumeAvailableStatus}
+		err := cloud.WaitVolumeTargetStatus(vol.ID, targetStatus)
+		if err != nil {
+			klog.Errorf("[CreateVolume] Target volume status is not 'Available': %v", err.Error())
+			return nil, status.Errorf(codes.Internal, err.Error())
+		}
+	}
+
 	klog.V(4).Infof("CreateVolume: Successfully created volume %s in Availability Zone: %s of size %d GiB", vol.ID, vol.AvailabilityZone, vol.Size)
 
 	return getCreateVolumeResponse(vol, ignoreVolumeAZ, req.GetAccessibilityRequirements()), nil
--- a/pkg/csi/cinder/controllerserver_test.go
+++ b/pkg/csi/cinder/controllerserver_test.go
@@ -47,8 +47,8 @@ func TestCreateVolume(t *testing.T) {
 	properties := map[string]string{"cinder.csi.openstack.org/cluster": FakeCluster}
 	// CreateVolume(name string, size int, vtype, availability string, snapshotID string, tags *map[string]string) (string, string, int, error)
 	osmock.On("CreateVolume", FakeVolName, mock.AnythingOfType("int"), FakeVolType, FakeAvailability, "", "", &properties).Return(&FakeVol, nil)
-
 	osmock.On("GetVolumesByName", FakeVolName).Return(FakeVolListEmpty, nil)
+	osmock.On("WaitVolumeTargetStatus", FakeVolID, []string{"available"}).Return(nil)
 	// Init assert
 	assert := assert.New(t)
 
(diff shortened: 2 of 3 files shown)


pkg/csi/cinder/controllerserver.go
```diff
@@ -136,6 +136,16 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 
 	}
 
+	// Wait for volume status to be Available
+	if vol.Status != openstack.VolumeAvailableStatus {
+		targetStatus := []string{openstack.VolumeAvailableStatus}
+		err := cloud.WaitVolumeTargetStatus(vol.ID, targetStatus)
+		if err != nil {
+			klog.Errorf("[CreateVolume] Target volume status is not 'Available': %v", err.Error())
+			return nil, status.Errorf(codes.Internal, err.Error())
```
chrigl, line 145: Any chance of leaking volumes here? Like... when it results in `error_something` we don't pass back the volume_id and thus kubernetes doesn't know about the volume and deletion (from k8s) won't be possible.

jichenjc: create should only results in `error` considering ("error", "error_extending", "error_deleting") as error status
per https://docs.openstack.org/api-ref/block-storage/v3/index.html

so if k8s has revert mechanism it should be reasonable to return that info about the volume ID

chrigl: When I am thinking about it... This might leak volumes that end up in `error_X`... But when a volume is in error state, there is nothing an OpenStack User (not admin) can do about it. So there is nothing we can do about at this point.

ramineni: @chrigl right, there is definitely, volume leak. And also I suppose it retries when it fails I guess (need to confirm this again). I was wondering before , delete volumes is the right thing to do upon error here or not as someone needs to check backend what is the issue , when volume goes to error state . So I couldnt decide upon delete here or leave up to admin.

>> But when a volume is in error state, there is nothing an OpenStack User (not admin) can do about it.
Oh, from the API ref , I couldnt make out its admin only operation. https://docs.openstack.org/api-ref/block-storage/v3/?expanded=delete-a-volume-detail#delete-a-volume
then rules out we could delete in all cases. Then just report back error is what we can do I suppose.

mdbooth: I don't see any special handling in cinder for delete of volumes in an error state, and the only policy check is volume:delete, which defaults to admin_or_owner. Not only that, 'error' is explicitly in the list of permitted volume statuses for delete.

IOW I expect volume delete on an error volume to succeed for the volume owner if the backend is able to delete it. If the backend can't delete it, there's nothing additional an admin can do through the api. An admin can force delete, but that just permits deleting volumes in intermediate states, so it has no effect if the volume is already in error.

mdbooth: Rest API only checks policy for force:
https://opendev.org/openstack/cinder/src/commit/c12375fec6c005ba8d7e14d6f94612e7f5e66dd2/cinder/api/v3/volumes.py#L54-L81

Backend only checks DELETE_POLICY:

https://opendev.org/openstack/cinder/src/commit/c12375fec6c005ba8d7e14d6f94612e7f5e66dd2/cinder/volume/api.py#L392

Backend permits deleting volume with status 'error' in all cases, and doesn't check status at all if force was given:

https://opendev.org/openstack/cinder/src/commit/c12375fec6c005ba8d7e14d6f94612e7f5e66dd2/cinder/volume/api.py#L451-L454

lingxiankong:
@ramineni what particular problem are you trying to fix? Personally I think rather than waiting for volume available after creation, we could check volume status in other operations.

ramineni:
> @ramineni what particular problem are you trying to fix? Personally I think rather than waiting for volume available after creation, we could check volume status in other operations.

@lingxiankong we have many failures recently in e2e , where attach operation fails , when looked into root cause , the problem is volume itself not created properly , its in error state.(which from logs is not evident properly) So I believe erroring out early in createVolume stage would be better , so that admin needs to check what is the issue and also gives more apt error message where the issue is.

lingxiankong:
(continue the discussion in the code)

pkg/csi/cinder/controllerserver.go
lingxiankong: > @lingxiankong we have many failures recently in e2e , where attach operation fails , when looked into root cause , the problem is volume itself not created properly , its in error state.(which from logs is not evident properly) So I believe erroring out early in createVolume stage would be better , so that admin needs to check what is the issue and also gives more apt error message where the issue is.

@ramineni but like @chrigl said, erroring out early will lead to orphan volumes in openstack cinder. What I suggest is to check volume status in other functions so that from a Kubernetes users' point of view, they have the chance to delete the PVC (and so the error status volume could be deleted as well). What do you think?

PR #1486, opened 2021-04-13
[cinder-csi-plugin] add snapshot crds to Helm Chart

eumel8:

Adding required CRDs if snapshot controller is enabled.

reference: https://github.com/kubernetes-csi/external-snapshotter/tree/master/client/config/crd

fixes # none

none

**Release note**:
```release-note
NONE
```

### DIFF
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: latest
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 1.3.5
+version: 1.3.6
 home: https://github.com/kubernetes/cloud-provider-openstack
 icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
--- a/charts/cinder-csi-plugin/templates/snapshot.storage.k8s.io_volumesnapshotclasses.yaml
+++ b/charts/cinder-csi-plugin/templates/snapshot.storage.k8s.io_volumesnapshotclasses.yaml
@@ -1,0 +1,112 @@
+{{- if .Values.csi.snapshotController.enabled -}}
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.4.0
+    api-approved.kubernetes.io: "https://github.com/kubernetes-csi/external-snapshotter/pull/419"
+  creationTimestamp: null
+  name: volumesnapshotclasses.snapshot.storage.k8s.io
+spec:
+  group: snapshot.storage.k8s.io
+  names:
+    kind: VolumeSnapshotClass
+    listKind: VolumeSnapshotClassList
+    plural: volumesnapshotclasses
+    singular: volumesnapshotclass
+  scope: Cluster
+  versions:
+  - additionalPrinterColumns:
+    - jsonPath: .driver
+      name: Driver
+      type: string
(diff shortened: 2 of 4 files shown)


lingxiankong:
/lgtm

Good to have.

ramineni:
/hold
Just to make sure to have a look before merging

ramineni:
@eumel8 we have thought about adding CRDs in the repo before . Two things that crossed my mind-
1. Snapshot controller is not per csi driver , its for cluster . so if its already deployed as part of distribution/cluster we dont require it 
2. Adding here a copy might become outdated later, its hard to track any change in CRDs to be updated here as well for every release. So instead , we updated the docs for doing the same https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/cinder-csi-plugin/features.md#volume-snapshots and ask them to take CRDs corresponding to the version 
3.  Its recommended to use same version CRDs as per the sidecar version compatibility.

So, Im wondering if we could add comment here https://github.com/kubernetes/cloud-provider-openstack/blob/master/charts/cinder-csi-plugin/templates/snapshot-controller-statefulset.yaml#L4 on where to install CRDs before deploying snapshot controller ?

lingxiankong:
What @ramineni said makes sense to me. cinder-csi-plugin provides capability for snapshot CRD functions, but registering the CRDs should be done separately.

/lgtm cancel

eumel8:
@ramineni @lingxiankong thx for reviewing and thoughts. 
Just to explain my idea behind the PR: Yes, the CRDs are developed on another place, and yes, this can be outdated after a while. But we want to give the best experience to our customer, user, and colleagues and for that reason we prefer the "one-click-installation" of a product or a tool. It's not very comfortable to install a Helm chart here, then separately a CRD file there or take few manual steps. In best case you have to deal with one tool and one technology.  I'll create another chart for hosting the CRDs and put this together with a depedeny.

spielkind:
@ramineni actually I don't get your approach, the csi-snapshotter is already deployed by the helm chart, also there is a default version in the values.yaml for the csi-snapshotter (2.1.3). So every time this default get an update we should merge the matching CRDs, also may some stuff on the RBACs or Statefulset may changed. From my point of view it makes no sense to not merge the CRDs but have the statefulset/rbacs in the helm chart.

Either remove all of it, and may submit it chart to csi-snapshotter repo - or just merge the copy of the CRDs to this repo?

Or may I just got the overall picture wrong? :)

ramineni:
@spielkind csi-snapshotter sidecar is deployed alongside of csi driver. 
But the volume snapshot controller and the CRDs are independent of any CSI driver. Regardless of the number CSI drivers deployed on the cluster, there must be only one instance of the volume snapshot controller running and one set of volume snapshot CRDs installed per cluster.  it is strongly recommended that Kubernetes distributors bundle and deploy the controller and CRDs as part of their Kubernetes cluster management process (independent of any CSI Driver)

So thats the reason , Im bit hesitant to maintain the same in this repo.

spielkind:
@ramineni Ok, but the snapshot controller is already deployed (default is false) by the helm chart in this repo:
https://github.com/kubernetes/cloud-provider-openstack/blob/master/charts/cinder-csi-plugin/templates/snapshot-controller-statefulset.yaml
https://github.com/kubernetes/cloud-provider-openstack/blob/master/charts/cinder-csi-plugin/templates/snapshot-controller-rbac.yaml

So my point was either remove the snapshot controller or add the CRDs?
Then the cleanest option would be to remove this part as well and submit a helm chart to:
https://github.com/kubernetes-csi/external-snapshotter
?

ramineni:
@spielkind +1 , adding to external-snapshotter would be the clean option .

PR #1488, opened 2021-04-13
[cinder-csi-plugin] Add microversion to GetVolumeByName

### LINKED ISSUE #1487: [cinder-csi-plugin] volumes list/get needs microversion passed
Since the cinder Train release getting a list of volumes and getting volumes from the API removed a deprecated version of filtering.  This can cause clients that don't pass in a microversion to not be able to filter the list of volumes by name.

Here is the relevant cinder patch that removed the deprecated filtering:
https://github.com/openstack/cinder/commit/ee1ac541253c42e3d94c907cb5b49a1172d68817

The solution is to support passing microversions.  The needed version to support the working filtering is
3.34.   If no microversion is passed in to the API call, then cinder reverts back to the lowest supported
version which is 3.0 and hence the filtering doesn't work.

Need to set the
blockstorageClient.Microversion = "3.34" at a minimum to support filtering by X for volumes list.
https://github.com/kubernetes/cloud-provider-openstack/blob/b20ce967f4b9f259c4148cf809c5111245a53f12/pkg/csi/cinder/openstack/openstack_volumes.go#L89

and here

https://github.com/kubernetes/cloud-provider-openstack/blob/b20ce967f4b9f259c4148cf809c5111245a53f12/pkg/csi/cinder/openstack/openstack_volumes.go#L126

hemna:
This patch adds the microversion needed to be able to properly
filter volumes by name.

fixes #1487

### DIFF
--- a/pkg/csi/cinder/openstack/openstack_volumes.go
+++ b/pkg/csi/cinder/openstack/openstack_volumes.go
@@ -124,8 +124,18 @@ func (os *OpenStack) ListVolumes(limit int, startingToken string) ([]volumes.Vol
 // GetVolumesByName is a wrapper around ListVolumes that creates a Name filter to act as a GetByName
 // Returns a list of Volume references with the specified name
 func (os *OpenStack) GetVolumesByName(n string) ([]volumes.Volume, error) {
+	// Init a local thread safe copy of the Cinder ServiceClient
+	blockstorageClient, err := openstack.NewBlockStorageV3(os.blockstorage.ProviderClient, os.epOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	// cinder filtering in volumes list is available since 3.34 microversion
+	// https://docs.openstack.org/cinder/latest/contributor/api_microversion_history.html#id32
+	blockstorageClient.Microversion = "3.34"
+
 	opts := volumes.ListOpts{Name: n}
-	pages, err := volumes.List(os.blockstorage, opts).AllPages()
+	pages, err := volumes.List(blockstorageClient, opts).AllPages()
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -86,7 +86,7 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	volumes, err := cloud.GetVolumesByName(volName)
 	if err != nil {
 		klog.Errorf("Failed to query for existing Volume during CreateVolume: %v", err)
-		return nil, status.Error(codes.Internal, "Failed to get volumes")
+		return nil, status.Error(codes.Internal, fmt.Sprintf("Failed to get volumes: %s", err))
 	}
 
 	if len(volumes) == 1 {


hemna:
> Thanks for your pull request. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please follow instructions at https://git.k8s.io/community/CLA.md#the-contributor-license-agreement to sign the CLA.**
> 
> It may take a couple minutes for the CLA signature to be fully registered; after that, please reply here with a new comment and we'll verify. Thanks.
> 
> * If you've already signed a CLA, it's possible we don't have your GitHub username or you're using a different email address.  Check your existing CLA data and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * If you signed the CLA as a corporation, please sign in with your organization's credentials at https://identity.linuxfoundation.org/projects/cncf to be authorized.
> * If you have done the above and are still having issues with the CLA being reported as unsigned, please log a ticket with the Linux Foundation Helpdesk: https://support.linuxfoundation.org/
> * Should you encounter any issues with the Linux Foundation Helpdesk, send a message to the backup e-mail support address at: [login-issues@jira.linuxfoundation.org](mailto:login-issues@jira.linuxfoundation.org)

I just signed it

jichenjc:
thanks for the PR , can you help to update `fixes # 1487` to `fixes #1487` ?

jichenjc:
you mentioned 2 APIs are affected but this PR only updated 1, on purpose?

databus23:
> you mentioned 2 APIs are affected but this PR only updated 1, on purpose?

@jichenjc Only one of the two api calls mentioned in the #1487 is actually using a filter (by name) and therefore requires the microversion.

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -124,8 +124,18 @@ func (os *OpenStack) ListVolumes(limit int, startingToken string) ([]volumes.Vol
 // GetVolumesByName is a wrapper around ListVolumes that creates a Name filter to act as a GetByName
 // Returns a list of Volume references with the specified name
 func (os *OpenStack) GetVolumesByName(n string) ([]volumes.Volume, error) {
+	// Init a local thread safe copy of the Cinder ServiceClient
+	blockstorageClient, err := openstack.NewBlockStorageV3(os.blockstorage.ProviderClient, os.epOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	// cinder online resize is available since 3.34 microversion
```
databus23, line 133: we have a copy and paste error here. the microversion is not passed because of online resize but because filtering list volumes require the microversion

hemna:
> you mentioned 2 APIs are affected but this PR only updated 1, on purpose?

Yah, I read the volumes list code a bit closer and realized that there is no filtering by field in that function, so it's not needed.
The microversion is only needed when filtering volumes/snapshots by field.

pkg/csi/cinder/openstack/openstack_volumes.go
hemna: This has been fixed.

databus23:
The e2e test is failing and it could be related. Seems like volumes can't be created anymore. I can't find the log output of the cinder-csi-plugin in the tests logs. I suspect the openstack environment is not accepting the microversion specification somehow.

jichenjc:
The gate e2e test is not stable recently

let's give another try

databus23:
I'm pretty sure the test failure is related to this PR, the error I linked is exactly a return from the function that was modified in this PR. Now it would be good to get the actual error logged with klog, but It seems the log output of the cinder-csi-plugin is not part of the test output.
I confirmed that this code works in our openstack environment and volumes can be created. So it would be interesting to get the logs to see why it is failing in the test environment.

jichenjc:
>>> klog.Errorf("Failed to query for existing Volume during CreateVolume: %v", err) 
 return nil, status.Error(codes.Internal, "Failed to get volumes") 

the gate issue and your issue both exist in the logs 
for now, I also can't figure out what's wrong there as the code is simliar to other micorversion usage 
I suggest you update 
 return nil, status.Error(codes.Internal, "Failed to get volumes") 
to 
 return nil, status.Error(codes.Internal, "Failed to get volumes %s", err) 
so that we can know what's wrong in the gate log (I admit we should put all container logs into test result)

databus23:
@hemna Could you update the PR to change this line: https://github.com/hemna/cloud-provider-openstack/blob/5e62d0c1f076193acd0ed0890c87be4bcfa869e6/pkg/csi/cinder/controllerserver.go#L89
to
```
		return nil, status.Error(codes.Internal, fmt.Sprintf("Failed to get volumes: %s", err))
```

databus23:
Ok the error now says:

```
Failed to get volumes: Bad request with: [GET https://lon1.citycloud.com:8776/v3/06ce5499f716488d985845cddcabdc16/volumes/detail?name=pvc-6d2e0605-d9ef-4f7f-b0fe-2056070337d5], error message: {"badRequest": {"message": "Invalid filters name are found in query options.", "code": 400}}
```

databus23:
What version of cinder is running in the e2e test environment. It seems like it behaves differently then out version.

jichenjc:
I think we are using master (latest code) from 
https://github.com/theopenlab/openlab-zuul-jobs/blob/master/roles/clone-devstack-gate-to-workspace/tasks/main.yml ? @lingxiankong do you know about this?

lingxiankong:
For the job `cloud-provider-openstack-e2e-test-csi-cinder`, it is talking with citynetwork block storage service (Cinder based) but not sure about the version.

ramineni:
can be tested once we have new CI up and running

ramineni:
/override openstack-cloud-keystone-authentication-authorization-test

lingxiankong:
/lgtm

Give  @jichenjc a chance to have a look.

jichenjc:
/approve

thanks~

PR #1492, opened 2021-04-22
[cinder-csi-plugin] Tag volume with optional pvc/pv metadata

databus23:
The external CSI provisioner optionally injects metadata in the CreateVolume that we can pass onto the volume as additional tags.
See https://github.com/kubernetes-csi/external-provisioner/pull/399

** Release notes **
```release-note
Tag volumes with additional metadata about pv and pvc if provided
```

### DIFF
--- a/manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml
+++ b/manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml
@@ -54,6 +54,7 @@ spec:
             - "--timeout=3m"
             - "--default-fstype=ext4"
             - "--feature-gates=Topology=true"
+            - "--extra-create-metadata"
           env:
             - name: ADDRESS
               value: /var/lib/csi/sockets/pluginproxy/csi.sock
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -103,6 +103,12 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 
 	// Volume Create
 	properties := map[string]string{"cinder.csi.openstack.org/cluster": cs.Driver.cluster}
+	//Tag volume with metadata if present: https://github.com/kubernetes-csi/external-provisioner/pull/399
+	for _, mKey := range []string{"csi.storage.k8s.io/pvc/name", "csi.storage.k8s.io/pvc/namespace", "csi.storage.k8s.io/pv/name"} {
+		if v, ok := req.Parameters[mKey]; ok {
+			properties[mKey] = v
+		}
+	}
 	content := req.GetVolumeContentSource()
 	var snapshotID string
 	var sourcevolID string
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: latest
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 1.3.6
+version: 1.3.7
 home: https://github.com/kubernetes/cloud-provider-openstack
 icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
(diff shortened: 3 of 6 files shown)


pkg/csi/cinder/controllerserver.go
```diff
@@ -103,6 +103,14 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
 	// Volume Create
 	properties := map[string]string{"cinder.csi.openstack.org/cluster": cs.Driver.cluster}
+	//Tag volume with metadata if present: https://github.com/kubernetes-csi/external-provisioner/pull/399
+	for k, v := range req.Parameters {
+		for _, mKey := range []string{"csi.storage.k8s.io/pvc/name", "csi.storage.k8s.io/pvc/namespace", "csi.storage.k8s.io/pv/name"} {
+			if k == mKey {
+				properties[k] = v
+			}
+		}
+	}
```
Fedosin, line 113:
```suggestion
	for _, mKey := range []string{"csi.storage.k8s.io/pvc/name", "csi.storage.k8s.io/pvc/namespace", "csi.storage.k8s.io/pv/name"} {
		if v, ok := req.GetParameters()[mKey]; ok {
			properties[mKey] = v
		}
	}
```

Fedosin: lgtm, but I would simplify this code a little.

Fedosin (changes requested):

pkg/csi/cinder/controllerserver.go
databus23: Thanks, I simplified as suggested.

pkg/csi/cinder/controllerserver.go
```diff
@@ -103,6 +103,12 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 
 	// Volume Create
 	properties := map[string]string{"cinder.csi.openstack.org/cluster": cs.Driver.cluster}
+	//Tag volume with metadata if present: https://github.com/kubernetes-csi/external-provisioner/pull/399
```
jichenjc, line 106: looks like `--extraCreateMetadata` is needed for such behavior to take effect ,so do we want to enable that in the mnaifest?

https://github.com/kubernetes-csi/external-provisioner/pull/399/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R68

databus23: the flag is called `--extra-create-metadata` by now. I added it to the chart and the manifests.

Fedosin (approved):
/lgtm

manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml
```diff
@@ -54,6 +54,7 @@ spec:
             - "--timeout=3m"
             - "--default-fstype=ext4"
             - "--feature-gates=Topology=true"
+            - "--extra-create-metadata"
```
lingxiankong, line 57: Shouldn't we also bump the tag of image k8s.gcr.io/sig-storage/csi-provisioner?

pkg/csi/cinder/controllerserver.go
```diff
@@ -103,6 +103,12 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+	for _, mKey := range []string{"csi.storage.k8s.io/pvc/name", "csi.storage.k8s.io/pvc/namespace", "csi.storage.k8s.io/pv/name"} {
+		if v, ok := req.Parameters[mKey]; ok {
+			properties[mKey] = v
+		}
+	}
```
lingxiankong, line 111: Unit test?

lingxiankong (changes requested):

manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml
databus23: Its not required for this PR.  The flag `--extra-create-metadata` is present since 1.6.0, currently the manifests contain 2.1.1

pkg/csi/cinder/controllerserver.go
databus23: I added a unit test for the extra metadata case.

pkg/csi/cinder/controllerserver.go
lingxiankong: awesome, thanks!

lingxiankong:
/lgtm

cc @ramineni @jichenjc  please take a look

jichenjc:
/approve 

I think it's a good thing to have , thanks~

PR #1495, opened 2021-04-26
[cinder-csi-plugin] Add log for POST and DELETE actions

jichenjc:
add logs so it's easier to track what we do on create/delete actions

fixes #

**Release note**:

```release-note
NONE
```


pkg/csi/cinder/controllerserver.go
```diff
@@ -312,6 +314,8 @@ func (cs *controllerServer) ListVolumes(ctx context.Context, req *csi.ListVolume
 }
 
 func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequest) (*csi.CreateSnapshotResponse, error) {
+	klog.V(4).Infof("Createsnapshot: called with args %+v", protosanitizer.StripSecrets(*req))
```
Fedosin, line 317: "CreateSnapshot"

Fedosin (changes requested):
One nit with "Createsnapshot" and it's ready to go

Issue #1498, opened 2021-04-26
[cinder-csi-plugin] if node-expand is called after node-publish, the test failed


jichenjc:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
>
/kind bug
> /kind feature

**What happened**:

https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/node.go#L731

failed, 

```
Unexpected error:
2021-04-26 09:51:19.538596 | ubuntu-xenial-citynetwork |         <*status.statusError | 0xc000089e00>: {
2021-04-26 09:51:19.538609 | ubuntu-xenial-citynetwork |             state: {
2021-04-26 09:51:19.538622 | ubuntu-xenial-citynetwork |                 NoUnkeyedLiterals: {},
2021-04-26 09:51:19.538635 | ubuntu-xenial-citynetwork |                 DoNotCompare: [],
2021-04-26 09:51:19.538648 | ubuntu-xenial-citynetwork |                 DoNotCopy: [],
2021-04-26 09:51:19.538661 | ubuntu-xenial-citynetwork |                 atomicMessageInfo: nil,
2021-04-26 09:51:19.538674 | ubuntu-xenial-citynetwork |             },
2021-04-26 09:51:19.538686 | ubuntu-xenial-citynetwork |             sizeCache: 0,
2021-04-26 09:51:19.538700 | ubuntu-xenial-citynetwork |             unknownFields: nil,
2021-04-26 09:51:19.538712 | ubuntu-xenial-citynetwork |             Code: 13,
2021-04-26 09:51:19.538725 | ubuntu-xenial-citynetwork |             Message: "Unable to find Device path for volume",
2021-04-26 09:51:19.538738 | ubuntu-xenial-citynetwork |             Details: nil,
2021-04-26 09:51:19.538751 | ubuntu-xenial-citynetwork |         }
2021-04-26 09:51:19.538764 | ubuntu-xenial-citynetwork |         rpc error: code = Internal desc = Unable to find Device path for volume
2021-04-26 09:51:19.538777 | ubuntu-xenial-citynetwork |     occurred[0m
```

**What you expected to happen**:

**How to reproduce it**:

**Anything else we need to know?**:

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version:
- OpenStack version:
- Others:

jichenjc:
[...]

PR #1496, opened 2021-04-26
[cinder-csi-plugin] Update csi-test from 3.1.0 to 4.2.0

jichenjc:
**What this PR does / why we need it**:

upgrade csi-test and fixed a set of issues after the test cases upgraded

**Which issue this PR fixes(if applicable)**:
fixes #1509 
fixes #1497 
fixes #1498

**Special notes for reviewers**:

**Release note**:

```release-note
NONE
```

PR #1499, opened 2021-04-26
[cinder-csi-plugin] Check volume path and volume ID before Expand

### LINKED ISSUE #1497: [cinder-csi-plugin] volume length is not checked and volumeID not verified during expand volume

>
/kind bug
> /kind feature

**What happened**:

https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/node.go#L697
https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/node.go#L716
failed (we are not using master branch now) 

**How to reproduce it**:

**Environment**:
- OpenStack version:
- Others:

jichenjc:

fixes #1497 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/nodeserver.go
+++ b/pkg/csi/cinder/nodeserver.go
@@ -526,6 +526,17 @@ func (ns *nodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 		return nil, status.Error(codes.InvalidArgument, "Volume ID not provided")
 	}
 	volumePath := req.GetVolumePath()
+	if len(volumePath) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Volume path not provided")
+	}
+
+	_, err := ns.Cloud.GetVolume(volumeID)
+	if err != nil {
+		if cpoerrors.IsNotFound(err) {
+			return nil, status.Error(codes.NotFound, fmt.Sprintf("Volume with ID %s not found", volumeID))
+		}
+		return nil, status.Error(codes.Internal, fmt.Sprintf("NodeExpandVolume failed with error %v", err))
+	}
 
 	args := []string{"-o", "source", "--first-only", "--noheadings", "--target", volumePath}
 	output, err := ns.Mount.Mounter().Exec.Command("findmnt", args...).CombinedOutput()


pkg/csi/cinder/nodeserver.go
```diff
@@ -526,6 +526,17 @@ func (ns *nodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 		return nil, status.Error(codes.InvalidArgument, "Volume ID not provided")
 	}
 	volumePath := req.GetVolumePath()
+	if len(volumePath) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Volume ID not provided")
```
Fedosin, line 530: "Volume path not provided"

pkg/csi/cinder/nodeserver.go
```diff
@@ -526,6 +526,17 @@ func (ns *nodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
...
+	}
+
+	_, err := ns.Cloud.GetVolume(volumeID)
+	if err != nil {
+		if cpoerrors.IsNotFound(err) {
+			return nil, status.Error(codes.NotFound, "Volume not found")
```
Fedosin, line 536: Maybe `"Volume with ID %s not found", volumeID`?

Fedosin (changes requested):

pkg/csi/cinder/nodeserver.go
```diff
@@ -526,6 +526,17 @@ func (ns *nodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 		return nil, status.Error(codes.InvalidArgument, "Volume ID not provided")
 	}
 	volumePath := req.GetVolumePath()
+	if len(volumePath) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Volume path not provided")
+	}
+
+	_, err := ns.Cloud.GetVolume(volumeID)
```
ramineni, line 533: @jichenjc From the issue details this is not required. 
https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/node.go#L682 , this checks for volumeID , which we are already doing in L525. So ideally shouldnt fail. Could you check again why that test case is failing.

jichenjc: my bad, actually the test case is https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/node.go#L716
I updated issue just now

ramineni: @jichenjc got it. I think this is not caught by current test as we are using old version of csi-test . Could you update the csi-test to new version as well to make sure this part of code is tested in CI as well. Thanks.

jichenjc:
> 
> 
> @jichenjc Thanks for the PR. code changes looks good. But please update the csi-test module to latest version , so that this code change can tested in CI as well.

https://github.com/kubernetes/cloud-provider-openstack/pull/1496
this is the PR that handling the update, I am not sure combine several things together will make thing way more complated
so this is a seperated PR

ramineni:
> > @jichenjc Thanks for the PR. code changes looks good. But please update the csi-test module to latest version , so that this code change can tested in CI as well.
> 
> #1496
> this is the PR that handling the update, I am not sure combine several things together will make thing way more complated
> so this is a seperated PR

@jichenjc I think combining should be fine , as we can actually see if test is fixed or any more changes required.

jichenjc:
> 
> 
> > > @jichenjc Thanks for the PR. code changes looks good. But please update the csi-test module to latest version , so that this code change can tested in CI as well.
> > 
> > 
> > #1496
> > this is the PR that handling the update, I am not sure combine several things together will make thing way more complated
> > so this is a seperated PR
> 
> @jichenjc I think combining should be fine , as we can actually see if test is fixed or any more changes required. As I was wondering before why it wasnt caught on CI as sanity test is there

I indicated why it not failed at beginning in the issue :)

as mentioned, there might be more issues (actually I am working on) and it might cause more time 
https://github.com/kubernetes/cloud-provider-openstack/issues/1498

so if we can fix this now ,I think about how about let's merge it (anyway it's a bug...)

jichenjc:
I find a way to fix #1498
so guess then we can combine those whole patch set

PR #1500, opened 2021-04-26
[cinder-csi-plugin] made ControllerUnpublishVolume call idempotent

axelgobletbdr:

The [CSI spec](https://github.com/container-storage-interface/spec/blob/master/spec.md#controllerunpublishvolume) lists that the `ControllerUnpublishVolume` function should be idempotent. If the volume corresponding to the `volume_id` is not attached to the node corresponding to the `node_id`, the Plugin MUST reply `0 OK`.

This is not the case for the cinder-csi-plugin. If the disk is not attached to the compute, the `DetachVolume` function returns an [error](https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/openstack/openstack_volumes.go#L285). This in turn causes the `ControllerUnpublishVolume` function to return an [error](https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/controllerserver.go#L256).

The non-idempotency of the `ControllerUnpublishVolume` function causes Kubernetes to keep retrying detachment of volumes that have somehow been detached via a different process than the cinder-csi-plugin. It is not able to recover from this state by itself. Generally, the cinder-csi-plugin should manage all volume detachments, but it is always possible that this happens externally. We should prevent such state inconsistencies from occuring.

This PR ensures that `DetachVolume` does not return an error if the volume is not attached to the compute, making the `ControllerUnpublishVolume` function idempotent to adhere to the CSI spec. This ensures that Kubernetes is able to put `VolumeAttachments` in the expected state, even if some external processes detached a volume from a node.

fixes #

I have tested this fix in our Kubernetes cluster, where the above described `VolumeAttachment` inconsistencies were present. The problem was solved by this fix, and the plugin is still functioning as it should.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/openstack/openstack_volumes.go
+++ b/pkg/csi/cinder/openstack/openstack_volumes.go
@@ -282,7 +282,8 @@ func (os *OpenStack) DetachVolume(instanceID, volumeID string) error {
 		}
 	}
 
-	return fmt.Errorf("disk: %s has no attachments or not attached to compute %s", volume.ID, instanceID)
+	// Disk has no attachments or not attached to provided compute
+	return nil
 }
 
 // WaitDiskDetached waits for detached


axelgobletbdr:
Some server provisioning steps fail during the [openlab/cloud-provider-openstack-multinode-csi-migration-test](https://logs.openlabtesting.org/logs/00/1500/d301da3c223cbc2f37740eb51104923d19d5abf6/cloud-provider-openstack-multinode-csi-migration-test/cloud-provider-openstack-multinode-csi-migration-test/11a3197/job-output.txt.gz). I doubt that this has something to do with my change in application logic. Can we restart that test?

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -282,7 +282,8 @@ func (os *OpenStack) DetachVolume(instanceID, volumeID string) error {
 		}
 	}
 
-	return fmt.Errorf("disk: %s has no attachments or not attached to compute %s", volume.ID, instanceID)
+	// Disk has no attachments or not attached to provided compute
+	return nil
```
ramineni, line 286: @axelgobletbdr  Thanks for the PR. Looks good. 
Just trying to understand better, in your case , If volume has no attachments , doesn't it mean that volume is in `Available` state  , then this function shouldn't throw error as per L264?

axelgobletbdr: @ramineni if the status is `Available` that is indeed correct. This does not necessarily need to be the case though. The volume can also already be attached to another instance. in that case, the state will be `in-use`, and the code will return an error at L285.

ramineni: @axelgobletbdr  right . I want to confirm that it fails only in case of multiattach volume type. Thanks

axelgobletbdr: It can happen for single attach volumes as well, through the following process:
1. Volume is attached to node X, and Kubernetes `VolumeAttachment` exists
2. Volume is somehow externally detached from node X. Kubernetes `VolumeAttachment` still exists
3. Pod using the volume is scheduled to a different node Y
4. Kubernetes will try to delete the current `VolumeAttachment` and create a new one. 
5. The deletion will not succeed due to the finalizers that need to be removed by the CSI plugin. The CSI plugin fails when the `ControllerUnpublishVolume` endpoint is called, since the volume is already scheduled to a different node.

It is quite an edge case. This also appeared for us because we are having some volume issues with our openstack deployment. Nevertheless, Kubernetes should be able to reconcile and end up in the expected state. It will be able to if the function call is idempotent as defined in the CSI spec.

PR #1501, opened 2021-04-26
Update snapshots examples to start using v1 API

Fedosin:

Snapshots v1 API is GA from 1.20, so we can address it in our examples:
https://kubernetes.io/blog/2020/12/10/kubernetes-1.20-volume-snapshot-moves-to-ga/

**Release note**:
```release-note
NONE
```


Fedosin:
We need to bump external-snapshotter first: https://github.com/kubernetes-csi/external-snapshotter/pull/493

jichenjc:
the manila CI is not related 
can you overwrite @lingxiankong ?

/lgtm

PR #1504, opened 2021-04-28
[occm] recreate load balancer pool on protocol mismatch

### LINKED ISSUE #1167: [occm] LB Pool not updated when proxy-protocol service annotation changed

/kind bug

**What happened**:
A change to the service annotation
* `loadbalancer.openstack.org/proxy-protocol`

is not picked up by the openstack-cloud-controller-manager. Changes are only propagated to the OpenStack loadbalancer pool when it is initially created.

Changing the above service annotation should be supported and the pool updated/recreated accordingly.

**How to reproduce it**:
Add the service annotation to an existing service of type `LoadBalancer`.
```yaml
apiVersion: v1
kind: Service
metadata:
   name: test-service
   annotations:
     loadbalancer.openstack.org/proxy-protocol: "true"
spec:
   ports:
   - port: 22000
     protocol: TCP
     targetPort: 8080
   type: LoadBalancer
```

Updating the loadbalancer pool protocol is not supported by Octavia.
See the Octavia API pool documentation [Update a pool detail](https://docs.openstack.org/api-ref/load-balancer/v2/?expanded=update-a-pool-detail#update-a-pool).
For updating the protocol, the pool has to be deleted and recreated keeping the following in mind:
* keep the floating ip
* keep the loadbalancer and listener (if possible)
* re-create the pool (and all others we may need to recreate)

This issue is a follow up from #1149.

**Environment**:
- Openstack version: Octavia

chrischdi:

Previously adding the proxy-protocol annotation after the pool got
created did not change anything. With this fix changing the annotation
also recreates the load balancer pool because the protocol of a pool
can't get changed via an update call.

fixes #1167 

**Release note**:

```release-note
[openstack-cloud-controller-manager] Recreate load balancer pool if the protocol changed based on the service definition.
```

---
<sup>
Christian Schlotter <christian.schlotter@daimler.com>, Daimler TSS GmbH,
<a href="https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md">Imprint</a>
</sup>

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1346,7 +1346,7 @@ func (lbaas *LbaasV2) ensureOctaviaPool(lbID string, listener *listeners.Listene
 	if svcConf.enableProxyProtocol {
 		poolProto = v2pools.ProtocolPROXY
 	} else if svcConf.keepClientIP && poolProto != v2pools.ProtocolHTTP {
-		// klog.V(4).Infof("Forcing to use %q protocol for pool because annotation %q is set", v2pools.ProtocolHTTP, ServiceAnnotationLoadBalancerXForwardedFor)
+		klog.V(4).Infof("Forcing to use %q protocol for pool because annotation %q is set", v2pools.ProtocolHTTP, ServiceAnnotationLoadBalancerXForwardedFor)
 		poolProto = v2pools.ProtocolHTTP
 	}
 
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1345,8 +1345,7 @@ func (lbaas *LbaasV2) ensureOctaviaPool(lbID string, listener *listeners.Listene
 	poolProto := v2pools.Protocol(listener.Protocol)
 	if svcConf.enableProxyProtocol {
 		poolProto = v2pools.ProtocolPROXY
-	} else if svcConf.keepClientIP && poolProto != v2pools.ProtocolHTTP {
-		klog.V(4).Infof("Forcing to use %q protocol for pool because annotation %q is set", v2pools.ProtocolHTTP, ServiceAnnotationLoadBalancerXForwardedFor)
+	} else if (svcConf.keepClientIP || svcConf.tlsContainerRef != "") && poolProto != v2pools.ProtocolHTTP {
 		poolProto = v2pools.ProtocolHTTP
 	}
 
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1340,6 +1340,38 @@ func (lbaas *LbaasV2) ensureOctaviaPool(lbID string, listener *listeners.Listene
 	if err != nil && err != openstackutil.ErrNotFound {
 		return nil, fmt.Errorf("error getting pool for listener %s: %v", listener.ID, err)
 	}
+
+	// By default, use the protocol of the listener
+	poolProto := v2pools.Protocol(listener.Protocol)
+	if svcConf.enableProxyProtocol {
+		poolProto = v2pools.ProtocolPROXY
+	} else if svcConf.keepClientIP && poolProto != v2pools.ProtocolHTTP {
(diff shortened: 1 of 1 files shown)


pkg/openstack/loadbalancer.go
```diff
@@ -1340,6 +1340,38 @@ func (lbaas *LbaasV2) ensureOctaviaPool(lbID string, listener *listeners.Listene
 	if err != nil && err != openstackutil.ErrNotFound {
 		return nil, fmt.Errorf("error getting pool for listener %s: %v", listener.ID, err)
 	}
+
+	// By default, use the protocol of the listener
+	poolProto := v2pools.Protocol(listener.Protocol)
+	if svcConf.enableProxyProtocol {
+		poolProto = v2pools.ProtocolPROXY
+	} else if svcConf.keepClientIP && poolProto != v2pools.ProtocolHTTP {
+		// klog.V(4).Infof("Forcing to use %q protocol for pool because annotation %q is set", v2pools.ProtocolHTTP, ServiceAnnotationLoadBalancerXForwardedFor)
```
chrischdi, line 1349:
```suggestion
		klog.V(4).Infof("Forcing to use %q protocol for pool because annotation %q is set", v2pools.ProtocolHTTP, ServiceAnnotationLoadBalancerXForwardedFor)
```

pkg/openstack/loadbalancer.go
```diff
@@ -1340,6 +1340,38 @@ func (lbaas *LbaasV2) ensureOctaviaPool(lbID string, listener *listeners.Listene
 	if err != nil && err != openstackutil.ErrNotFound {
 		return nil, fmt.Errorf("error getting pool for listener %s: %v", listener.ID, err)
 	}
+
+	// By default, use the protocol of the listener
+	poolProto := v2pools.Protocol(listener.Protocol)
+	if svcConf.enableProxyProtocol {
+		poolProto = v2pools.ProtocolPROXY
+	} else if svcConf.keepClientIP && poolProto != v2pools.ProtocolHTTP {
```
zetaab, line 1348: hmm, we should also remember that there can be udp and tcp loadbalancers.

zetaab (changes requested):

pkg/openstack/loadbalancer.go
chrischdi: You are right, I updated the if clause to match the case inside `lbaas.buildPoolCreateOpt` now.

I also removed the log line because it would be printed later on too when the pool is getting created again.

However: I think udp and tcp is ok because we will get this information from the `listener.Protocol`.

What do you think?

chrischdi:
@zetaab do you maybe have some time to take another look at it? :-)

pkg/openstack/loadbalancer.go
```diff
@@ -1340,6 +1340,37 @@ func (lbaas *LbaasV2) ensureOctaviaPool(lbID string, listener *listeners.Listene
...
+		poolProto = v2pools.ProtocolHTTP
+	}
+
+	// Delete the pool and its members if it already exists and has the wrong protocol
+	if pool != nil && v2pools.Protocol(pool.Protocol) != poolProto {
+		klog.V(2).Infof("Deleting wrong pool %s for listener %s because the protocol does not match", pool.ID, listener.ID)
+
+		// Delete pool automatically deletes all its members.
+		mc := metrics.NewMetricContext("loadbalancer_pool", "delete")
+		if err = v2pools.Delete(lbaas.lb, pool.ID).ExtractErr(); err != nil {
```
lingxiankong, line 1358: `if err = v2pools.Delete(lbaas.lb, pool.ID).ExtractErr(); mc.ObserveRequest(err) != nil {` so you could remove Line 1366

chrischdi: But this would also record `IsNotFound` as an error in the metrics, would this be correct? We are doing the same in line 2383.

lingxiankong: Hmm, yes, you are right. I didn't notice that.

lingxiankong:
/lgtm
/override openlab/cloud-provider-openstack-acceptance-test-lb-octavia

lingxiankong:
BTW, I've changed the `Release note` to make it more clear and readable.

lingxiankong:
@jichenjc @raminen Could any of you please review this PR?

PR #1515, opened 2021-05-03
[manila-csi-plugin] Add shareAccessTo volume parameter

gman0:

This PR adds `shareAccessTo` volume / storage class parameter. Its value is sourced into `access_to` field when creating a share access rule when provisioning a share.

One of the use cases is when creating CephFS cephx clients. Ceph kernel clients using the same cephx ID may re-use MDS cache and OSD connections, improving the performance and reducing the load on the Ceph cluster.

If no `shareAccessTo` volume parameter is provided, the behavior remains as it was: `access_to` is set to the volume name (PV name).

**Release note**:

```release-note
[manila-csi-plugin] Added shareAccessTo volume/StorageClass parameter
```


gman0:
@tombarron @gouthampacha @ramineni PTAL

gman0:
There's a similar volume parameter for NFS shares, `nfs-shareClient`, which also sets `access_to` but only for NFS shares. Similarly `shareAccessTo` from this patch is relevant only for CephFS shares - which is way less general than what a parameter name `shareAccessTo` would suggest.

I don't like adding even more volume parameters (which carry essentially the same information: the value of `access_to`), but for the sake of consistency I'll rework this PR. Closing.

PR #1519, opened 2021-05-04
[manila-csi-plugin] Add cephfs-clientID to volume parameters

gman0:

This PR adds `cephfs-clientID` volume / storage class parameter. Its value is sourced into `access_to` field when creating a share access rule for the provisioned share.

Ceph kernel clients using the same cephx ID may re-use MDS cache and OSD connections, improving the performance and reducing the load on the Ceph cluster. Using `cephfs-clientID`, users are able to make use of this optimization.

If no `cephfs-clientID` volume parameter is provided, the behavior remains as it was: `access_to` is set to the volume name (PV name).

**Release note**:

```release-note
[manila-csi-plugin] Added cephfs-clientID volume/StorageClass parameter
```


gman0:
@tombarron @gouthampacha @ramineni PTAL

tombarron (approved):
Looks good to me.

tombarron:
/lgtm
/approve

but touches docs outside pkg/csi/manila, so really just for the record :D

gman0:
Thanks @tombarron !

PR #1522, opened 2021-05-05
[cinder-csi-plugin] Cleanup: handle errors from GetDevicePath

Fedosin:
Now GetDevicePath from metadata doesn't return an error if something bad happens. This makes it difficult to debug Cinder CSI driver.

This commit modifies the function to return an error value, and also adds additional logging that should help with debugging.

```release-note
NONE
```


ricolin (commented):
/lgtm

jichenjc:
/approve

thanks for the PR

PR #1524, opened 2021-05-06
Less hasty backoff for loadbalancers

ottosulin:
Affects:
* octavia-ingress-controller

Current backoff factor when creating or deleting Octavia loadbalancers is only 1.2. If a large number of loadbalancers is created at the same time, the load can be too much for OpenStack API. The proposed change helps OpenStack API cope with the number of requests, while still maintaining a good total backoff time.


lingxiankong:
@ottosulin Thanks for contributing, please read https://raw.githubusercontent.com/kubernetes/cloud-provider-openstack/master/.github/PULL_REQUEST_TEMPLATE.md and make changes to the PR commit message accordingly.

PR #1528, opened 2021-05-11
[manila-csi-plugin] Fix error connecting to nfs plugin when using cluster proxy

### LINKED ISSUE #1527: [manila-csi-plugin] Cannot connect to csi-nfsplugin if the cluster uses a proxy

/kind bug

**What happened**:
Configure a cluster-wide proxy. On OpenShift this can be done as described in [Configuring the cluster-wide proxy](https://docs.openshift.com/container-platform/4.7/networking/enable-cluster-wide-proxy.html). The net result is that `openstack-manila-csi-nodeplugin` is launched with, e.g.:

```
    - name: HTTPS_PROXY
      value: https://proxy.example.com:3128
    - name: HTTP_PROXY
      value: http://proxy.example.com:3128
    - name: NO_PROXY
      value: .cluster.local,.svc,10.0.128.0/17,10.128.0.0/14,127.0.0.1,169.254.169.254,172.30.0.0/16,api-int.cluster-dsal.shiftstack.test,localhost
```

The problem is that gRPC uses these proxy settings and doesn't have default behaviour to ignore the proxy for unix:// connections. The result is logs full of:

```
W0510 14:00:06.094110       1 builder.go:82] still connecting to unix:///var/lib/kubelet/plugins/csi-nfsplugin/csi.sock
```

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: master (eb0eebee)

mdbooth:
fixes #1527

This is very simple to reproduce on OpenShift by setting a cluster-wide proxy. See #1527 for reproducer steps.

**Release note**:

```release-note
NONE
```

**NOTE**

A better fix would be to update gRPC to at least v1.34.0 which contains https://github.com/grpc/grpc-go/pull/3890. This supports the unix protocol explicitly and never proxies it. However, due to our transitive dep on etcd clientv3 and https://github.com/etcd-io/etcd/issues/11563 we are pinned to gRPC <= v1.30.0. The etcd 3.5 branch has finally bumped the gRPC version, but that is not released yet. None of the etcd stable branch tags contain a gRPC bump.

/cc @Fedosin

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,7 +32,7 @@ require (
 	golang.org/x/crypto v0.0.0-20210220033148-5ea612d1eb83
 	golang.org/x/net v0.0.0-20210224082022-3d97a244fca7
 	golang.org/x/sys v0.0.0-20210225134936-a50acf3fe073
-	google.golang.org/grpc v1.27.1
+	google.golang.org/grpc v1.29.0
 	gopkg.in/gcfg.v1 v1.2.3
 	gopkg.in/godo.v2 v2.0.9
 	gopkg.in/warnings.v0 v0.1.2 // indirect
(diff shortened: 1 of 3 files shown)


tombarron (approved):

tombarron:
@gman0 ptal

Fedosin:
/override openlab/cloud-provider-openstack-acceptance-test-lb-octavia

mdbooth:
> /retest

Ah, looks like the replies from @k8s-ci-robot to @Fedosin saying that the command was incorrect are just noise, and the test are in fact being executed by a less noisy bot responding to the same command?

Oh well, looks like we're re-running all the tests now 🤷‍♂️

mdbooth:
Update commit message and added a comment to document the results of my git spelunking in etc and gRPC.

ramineni:
lb-octaia job is disabled now.
/override openlab/cloud-provider-openstack-acceptance-test-lb-octavia

tombarron (approved):
Let's see if we can get this in.

ramineni:
overriding 1.18-e2e-conformance , as no need for this PR and also would be removed in future on master branch

tombarron:
@ramineni Thanks much!

PR #1530, opened 2021-05-12
[magnum-auto-healer] Fix duplicated repair action

### LINKED ISSUE #1529: [magnum-auto-healer] Fix duplicated healing action

/kind bug

**What happened**:
When there is a node down, the repair action might be triggered more than one time.

Before the cluster back to healthy status, only one repair action is executed.

**How to reproduce it**:
1) Create a new cluster
2) Login one of the master node, stop the kubelet
3) Monitor the auto healer log

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: v1.20.3
- OpenStack version: master
- Others:

openstacker:
This is kind of a regression issue. After introducing the feature that restarts the broken node if it's found the first time, then the stack update action is quick because there is no real action involved. But the controller is still polling the status periodically, so another round check coming and will trigger the repair again.

fixes #1529 

**Release note**:
```release-note
[magnum-auto-healer] Fix duplicated auto-healing actions and the default health check period is changed from 30s to 60s.
```

### DIFF
--- a/pkg/autohealing/cloudprovider/openstack/provider.go
+++ b/pkg/autohealing/cloudprovider/openstack/provider.go
@@ -250,6 +250,67 @@ func (provider OpenStackCloudProvider) waitForServerDetachVolumes(serverID strin
 	return rootVolumeID, err
 }
 
+// FirstTimeRepair Handle the first time repair for a node
+// 1) If the node is the first time in error, reboot and uncordon it
+// 2) If the node is not the first time in error, check if the last reboot time is in provider.Config.RebuildDelayAfterReboot
+//    That said, if the node has been found in broken status before but has been long time since then, the processed variable
+//    will be kept as False, which means the node need to be rebuilt to fix it, otherwise it means the has been processed.
+// The bool type return value means that if the node has been processed from a first time repair PoV
+func (provider OpenStackCloudProvider) firstTimeRepair(n healthcheck.NodeInfo, serverID string, firstTimeRebootNodes map[string]healthcheck.NodeInfo) (bool, error) {
+	var firstTimeUnhealthy = true
+	for id := range unHealthyNodes {
+		log.V(5).Infof("comparing server ID %s with known broken ID %s", serverID, id)
+		if id == serverID {
+			log.Infof("it is NOT the first time that node %s is being repaired", serverID)
+			firstTimeUnhealthy = false
+			break
+		}
+	}
+
+	var processed = false
+	if firstTimeUnhealthy == true {
+		log.Infof("rebooting node %s to repair it", serverID)
+
+		if res := servers.Reboot(provider.Nova, serverID, servers.RebootOpts{Type: servers.SoftReboot}); res.Err != nil {
+			// Usually it means the node is being rebooted
+			log.Warningf("failed to reboot node %s, error: %v", serverID, res.Err)
+			if strings.Contains(res.Err.Error(), "reboot_started") {
+				// The node has been restarted
+				firstTimeRebootNodes[serverID] = n
+				processed = true
+			}
+		} else {
+			// Uncordon the node
+			if n.IsWorker == true {
+				nodeName := n.KubeNode.Name
+				newNode := n.KubeNode.DeepCopy()
+				newNode.Spec.Unschedulable = false
+				if _, err := provider.KubeClient.CoreV1().Nodes().Update(context.TODO(), newNode, metav1.UpdateOptions{}); err != nil {
+					log.Errorf("Failed to cordon node %s, error: %v", nodeName, err)
+				} else {
+					log.Infof("Node %s is cordoned", nodeName)
+				}
+			}
+
+			n.RebootAt = time.Now()
+			firstTimeRebootNodes[serverID] = n
+			unHealthyNodes[serverID] = n
+			log.Infof("Node %s has been rebooted at %s", serverID, n.RebootAt)
+		}
+		processed = true
+	} else {
+		// If the node was rebooted within mins (defined by RepairDelayAfterReboot), ignore it
+		log.Infof("Node %s is found in unhealthy again which was rebooted at %s", serverID, unHealthyNodes[serverID].RebootAt)
+		if unHealthyNodes[serverID].RebootAt.After(time.Now().Add(-1 * provider.Config.RebuildDelayAfterReboot)) {
+			log.Infof("Node %s is found in unhealthy again, but we're going to defer the repair because it maybe in reboot process", serverID)
+			firstTimeRebootNodes[serverID] = n
+			processed = true
+		}
+	}
+
+	return processed, nil
+}
+
 // Repair  For master nodes: detach etcd and docker volumes, find the root
 //         volume, then shutdown the VM, marks the both the VM and the root
 //         volume (heat resource) as "unhealthy" then trigger Heat stack update
@@ -275,6 +336,8 @@ func (provider OpenStackCloudProvider) Repair(nodes []healthcheck.NodeInfo) erro
 		masters = nodes
 	}
 
+	firstTimeRebootNodes := make(map[string]healthcheck.NodeInfo)
+
 	err := provider.UpdateHealthStatus(masters, workers)
 	if err != nil {
 		return fmt.Errorf("failed to update the helath status of cluster %s, error: %v", clusterName, err)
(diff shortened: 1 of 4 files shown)


pkg/autohealing/cloudprovider/openstack/provider.go
```diff
@@ -250,6 +250,61 @@ func (provider OpenStackCloudProvider) waitForServerDetachVolumes(serverID strin
...
+		processed = true
+	} else {
+		n.RebootAt = unHealthyNodes[serverID].RebootAt
+		// If the node was rebooted within mins (defined by RepairDelayAfterReboot), ignore it
+		log.Infof("Node %s is found in unhealthy again which was rebooted at %s", serverID, n.RebootAt)
+		if n.RebootAt.After(time.Now().Add(-1 * provider.Config.RebuildDelayAfterReboot)) {
+			log.Infof("Node %s is found in unhealthy again, but we're going to defer the repair because it maybe in reboot process", serverID)
+			firstTimeRebootNodes[serverID] = n
+			processed = true
+		}
```
ricolin, line 302: If it's rebooted for a while ago but still presented as un healthy, we just delete it

pkg/autohealing/config/config.go
```diff
@@ -83,12 +86,13 @@ type kubeConfig struct {
...
-		DryRun:               false,
-		CloudProvider:        "openstack",
-		MonitorInterval:      30 * time.Second,
-		MasterMonitorEnabled: true,
-		WorkerMonitorEnabled: true,
-		LeaderElect:          true,
-		CheckDelayAfterAdd:   10 * time.Minute,
+		DryRun:                  false,
+		CloudProvider:           "openstack",
+		MonitorInterval:         60 * time.Second,
```
ricolin, line 91: Is there reason why we change this to 60?

pkg/autohealing/cloudprovider/openstack/provider.go
openstacker: @ricolin  What do you mean "we just delete it"? We cannot just delete the node, since it's being rebooted.

pkg/autohealing/config/config.go
openstacker: Good question. The default report interview from kubelet to kube-apiserver is 40s, so it's possible that the node status hasn't been updated within 30s. And based on our experience on prod, 30s is a bit too often. So I'm proposing to change it to a number just more than 40s so 60s is a good balance. I will update the release to cover this.

ricolin: I'm fine with 60 secs, but I wonder is there any proper way to notify users about the default config change. releasenote?:)

openstacker: I have already updated the relesse note.

pkg/autohealing/cloudprovider/openstack/provider.go
ricolin: Just noted for the behavior, so correct me if I misinterpret it

I mean we will not found node in firstTimeRebootNodes [1] if it's not firsttimeUnhealthy and already passed delay mins since last reboot. And that node will be deleted at line 471

[1] https://github.com/kubernetes/cloud-provider-openstack/pull/1530/commits/3d06649e21aa6a19c59477640b970efe6ec96736#diff-9846ce0766626342e9df737eff34e6aba6b203aa80e0ba8fa021ac26ddbad785R467

ricolin (commented):
/lgtm

openstacker:
> /lgtm

pkg/autohealing/cloudprovider/openstack/provider.go
openstacker: Yes, if the node is not firsttimeUnhealthy and already passed delayed mins, then it will be deleted from the k8s PoV.

pkg/autohealing/cloudprovider/openstack/provider.go
```diff
@@ -275,6 +330,8 @@ func (provider OpenStackCloudProvider) Repair(nodes []healthcheck.NodeInfo) erro
 		masters = nodes
 	}
 
+	firstTimeRebootNodes := make(map[string]healthcheck.NodeInfo)
```
lingxiankong, line 333: `firstTimeRebootNodes` is initialized as an empty map every time when `Repair()` is called, that's not what we expected.

pkg/autohealing/cloudprovider/openstack/provider.go
```diff
@@ -250,6 +250,61 @@ func (provider OpenStackCloudProvider) waitForServerDetachVolumes(serverID strin
 	return rootVolumeID, err
 }
 
+// FirstTimeRepair Handle the first time repair for a node
+// 1) If the node is the first time in error, reboot it
```
lingxiankong, line 254: reboot and uncordon it.

Also, please describe explicitly what's the meaning of the returned value(`processed`) and what the called should do.

pkg/autohealing/cloudprovider/openstack/provider.go
```diff
@@ -250,6 +250,61 @@ func (provider OpenStackCloudProvider) waitForServerDetachVolumes(serverID strin
...
+				if _, err := provider.KubeClient.CoreV1().Nodes().Update(context.TODO(), newNode, metav1.UpdateOptions{}); err != nil {
+					log.Errorf("Failed to cordon node %s, error: %v", nodeName, err)
+				} else {
+					log.Infof("Node %s is cordoned", nodeName)
+				}
+			}
+
+			n.RebootAt = time.Now()
+			firstTimeRebootNodes[serverID] = n
+			unHealthyNodes[serverID] = n
```
lingxiankong, line 290: duplicate with Line 269

pkg/autohealing/cloudprovider/openstack/provider.go
```diff
@@ -250,6 +250,61 @@ func (provider OpenStackCloudProvider) waitForServerDetachVolumes(serverID strin
...
+				newNode := n.KubeNode.DeepCopy()
+				newNode.Spec.Unschedulable = false
+				if _, err := provider.KubeClient.CoreV1().Nodes().Update(context.TODO(), newNode, metav1.UpdateOptions{}); err != nil {
+					log.Errorf("Failed to cordon node %s, error: %v", nodeName, err)
+				} else {
+					log.Infof("Node %s is cordoned", nodeName)
+				}
+			}
+
+			n.RebootAt = time.Now()
```
lingxiankong, line 288: The `nodes []healthcheck.NodeInfo` outside of this function is not going to be changed because `n` here is just a copy.

lingxiankong (changes requested):

pkg/autohealing/controller/controller.go
```diff
@@ -316,6 +316,10 @@ func (c *Controller) repairNodes(unhealthyNodes []healthcheck.NodeInfo) {
 					newNode := node.KubeNode.DeepCopy()
 					newNode.Spec.Unschedulable = true
 
+					// Skip cordon for master node
```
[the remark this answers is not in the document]
openstacker: Because by default we have already set the UnScheduable for all master nodes. It's not necessary to set it again here.

pkg/autohealing/cloudprovider/openstack/provider.go
```diff
@@ -250,6 +250,61 @@ func (provider OpenStackCloudProvider) waitForServerDetachVolumes(serverID strin
 	return rootVolumeID, err
 }
 
+// FirstTimeRepair Handle the first time repair for a node
+// 1) If the node is the first time in error, reboot it
+// 2) If the node is not the first time in error, check if the last reboot time is in provider.Config.RebuildDelayAfterReboot
+func (provider OpenStackCloudProvider) FirstTimeRepair(n healthcheck.NodeInfo, serverID string, firstTimeRebootNodes map[string]healthcheck.NodeInfo) (bool, error) {
```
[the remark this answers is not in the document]
openstacker: Good point. Will fix in next patch set.

pkg/autohealing/controller/controller.go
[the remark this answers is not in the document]
lingxiankong: If this is specific to magnum, I would recommend we add a config option to decide if master node should be uncordoned or not. The `repairNodes` is supposed to be agnostic to providers.

pkg/autohealing/cloudprovider/openstack/provider.go
openstacker: It's expected that the variable `firstTimeRebootNodes` is only available in the Repair function because all the broken nodes information are stored in variable `unHealthyNodes`. Maybe we should get a better name, the variable is used to temporarily storing broken nodes that have been processed by the function `firstTimeRepair`.

pkg/autohealing/cloudprovider/openstack/provider.go
openstacker: I have removed the duplicated line at 269 and the variable `n` will be saved at line 290. So I think the new value of RebootAt will be saved. Please correct me if I missed something. Cheers.

openstacker:
Doing a ninja approval based on the lgtm from Lingxian :)

PR #1535, opened 2021-05-17
[occm] Add support of trunk ports

jingczhang:
Currenlty, getAttachedInterfacesByID() only returns neutron ports directly attached to VM.
For vlan-aware VMs that have trunk ports, subports attached to the trunk ports are not returned.
This pull request checks if VM has trunk ports, if so, adds their subports in the return.

Signed-off-by: Jing Zhang <jing.c.zhang@nokia.com>

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/instances.go
+++ b/pkg/openstack/instances.go
@@ -29,6 +29,9 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/attachinterfaces"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/trunks"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
 	"github.com/gophercloud/gophercloud/pagination"
 	"github.com/mitchellh/mapstructure"
 	v1 "k8s.io/api/core/v1"
@@ -47,6 +50,7 @@ import (
 // Instances encapsulates an implementation of Instances for OpenStack.
 type Instances struct {
 	compute        *gophercloud.ServiceClient
+	network        *gophercloud.ServiceClient
 	opts           metadata.Opts
 	networkingOpts NetworkingOpts
 }
@@ -77,8 +81,15 @@ func (os *OpenStack) instances() (*Instances, bool) {
 		return nil, false
 	}
 
+	network, err := client.NewNetworkV2(os.provider, os.epOpts)
+	if err != nil {
+		klog.Errorf("Failed to create an OpenStack Network client: %v", err)
+		return nil, false
+	}
+
 	return &Instances{
 		compute:        compute,
+		network:        network,
 		opts:           os.metadataOpts,
 		networkingOpts: os.networkingOpts,
 	}, true
@@ -115,7 +126,7 @@ func (i *Instances) AddSSHKeyToAllInstances(ctx context.Context, user string, ke
 func (i *Instances) NodeAddresses(ctx context.Context, name types.NodeName) ([]v1.NodeAddress, error) {
 	klog.V(4).Infof("NodeAddresses(%v) called", name)
 
-	addrs, err := getAddressesByName(i.compute, name, i.networkingOpts)
+	addrs, err := getAddressesByName(i.compute, name, i.networkingOpts, i.network)
 	if err != nil {
 		return nil, err
 	}
@@ -143,7 +154,7 @@ func (i *Instances) NodeAddressesByProviderID(ctx context.Context, providerID st
 		return []v1.NodeAddress{}, err
 	}
 
-	interfaces, err := getAttachedInterfacesByID(i.compute, server.ID)
+	interfaces, err := getAttachedInterfacesByID(i.compute, server.ID, i.network)
 	if err != nil {
 		return []v1.NodeAddress{}, err
 	}
@@ -235,7 +246,7 @@ func (i *Instances) InstanceMetadata(ctx context.Context, node *v1.Node) (*cloud
 		return nil, err
 	}
 
-	interfaces, err := getAttachedInterfacesByID(i.compute, srv.ID)
+	interfaces, err := getAttachedInterfacesByID(i.compute, srv.ID, i.network)
 	if err != nil {
 		return nil, err
 	}
@@ -465,9 +476,15 @@ func getServerByName(client *gophercloud.ServiceClient, name types.NodeName) (*S
 // * metadata hostname
 // * server object Addresses (floating type)
 func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface, networkingOpts NetworkingOpts) ([]v1.NodeAddress, error) {
+	type Address struct {
+		IPType string `mapstructure:"OS-EXT-IPS:type"`
+		Addr   string
+	}
+
 	addrs := []v1.NodeAddress{}
 
 	// parse private IP addresses first in an ordered manner
+	allPrivates := make(map[string][]Address)
 	for _, iface := range interfaces {
 		for _, fixedIP := range iface.FixedIPs {
 			if iface.PortState == "ACTIVE" {
@@ -479,6 +496,10 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
 							Address: fixedIP.IPAddress,
 						},
 					)
+					if len(iface.NetID) > 0 {
+						addr := Address{IPType: "fixed", Addr: fixedIP.IPAddress}
+						allPrivates[iface.NetID] = append(allPrivates[iface.NetID], addr)
+					}
 				}
 			}
 		}
@@ -513,15 +534,33 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
 	}
 
 	// process the rest
-	type Address struct {
-		IPType string `mapstructure:"OS-EXT-IPS:type"`
-		Addr   string
-	}
-
 	var addresses map[string][]Address
 	err := mapstructure.Decode(srv.Addresses, &addresses)
 	if err != nil {
 		return nil, err
+	}
+
+	// add subports if exist to the server
+	extraPrivates := make(map[string][]Address)
+	for k, v := range allPrivates {
+		ok := false
+		for _, a := range v {
+			for _, v1 := range addresses {
+				for _, a1 := range v1 {
+					if a.Addr == a1.Addr {
+						ok = true
+						break
+					}
+				}
+			}
+		}
+		if !ok {
+			extraPrivates[k] = v
+		}
+	}
+	klog.V(5).Infof("Node '%s' extraPrivates '%s'", srv.Name, extraPrivates)
+	for k, v := range extraPrivates {
+		addresses[k] = v
 	}
 
 	var networks []string
@@ -573,13 +612,13 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
 	return addrs, nil
 }
 
-func getAddressesByName(client *gophercloud.ServiceClient, name types.NodeName, networkingOpts NetworkingOpts) ([]v1.NodeAddress, error) {
-	srv, err := getServerByName(client, name)
+func getAddressesByName(compute *gophercloud.ServiceClient, name types.NodeName, networkingOpts NetworkingOpts, network *gophercloud.ServiceClient) ([]v1.NodeAddress, error) {
+	srv, err := getServerByName(compute, name)
 	if err != nil {
 		return nil, err
 	}
 
-	interfaces, err := getAttachedInterfacesByID(client, srv.ID)
+	interfaces, err := getAttachedInterfacesByID(compute, srv.ID, network)
 	if err != nil {
 		return nil, err
 	}
@@ -587,12 +626,12 @@ func getAddressesByName(client *gophercloud.ServiceClient, name types.NodeName, 
 	return nodeAddresses(&srv.Server, interfaces, networkingOpts)
 }
 
-func getAddressByName(client *gophercloud.ServiceClient, name types.NodeName, needIPv6 bool, networkingOpts NetworkingOpts) (string, error) {
+func getAddressByName(compute *gophercloud.ServiceClient, name types.NodeName, needIPv6 bool, networkingOpts NetworkingOpts, network *gophercloud.ServiceClient) (string, error) {
 	if needIPv6 && networkingOpts.IPv6SupportDisabled {
 		return "", errors.ErrIPv6SupportDisabled
 	}
 
-	addrs, err := getAddressesByName(client, name, networkingOpts)
+	addrs, err := getAddressesByName(compute, name, networkingOpts, network)
 	if err != nil {
 		return "", err
 	} else if len(addrs) == 0 {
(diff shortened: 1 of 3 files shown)


lingxiankong:
Thanks for the contribution!

But first, I want to know more about the problems this PR is trying to fix, what issues have you seen? Why the k8s nodes are using truck port which should be supposed to use in NFV use case? Is there any particular reason?

jingczhang:
Hi,

I had tried to put more information for the PR in the commit message but messed up the format, so they did not show up.

Major reasons for this PR are:

(1) To support CNFs that tag their own vlans i.e. vlan-aware CNFs

There is work on-going at sriov-cni side to enable this:
https://github.com/k8snetworkplumbingwg/sriov-cni/issues/133
Implement interface for vlan trunking

(2) To align Kubernetes deployment over VM to that over BM (bare-metal)

So that, from CNF point of view, same "look-and-feel". The same CNFs can be deployed in exactly the same way regardless the underlaying infra-structure i.e. virtualized or not.

By extension, enable the same SDN solution for Kubernetes deployments, either over VM or over BM.

Please check and review.

Jing

lingxiankong:
> Hi,
> 
> I had tried to put more information for the PR in the commit message but messed up the format, so they did not show up.
> 
> Major reasons for this PR are:
> 
> (1) To support CNFs that tag their own vlans i.e. vlan-aware CNFs
> 
> There is work on-going at sriov-cni side to enable this:
> [k8snetworkplumbingwg/sriov-cni#133](https://github.com/k8snetworkplumbingwg/sriov-cni/issues/133)
> Implement interface for vlan trunking
> 
> (2) To align Kubernetes deployment over VM to that over BM (bare-metal)
> 
> So that, from CNF point of view, same "look-and-feel". The same CNFs can be deployed in exactly the same way regardless the underlaying infra-structure i.e. virtualized or not.
> 
> By extension, enable the same SDN solution for Kubernetes deployments, either over VM or over BM.
> 
> Please check and review.
> 
> Jing

Thanks for that info, I totally understand how vlan-aware ports work and k8s deployment over VM and BM. What I was asking is about the real use case. Are you working with an environment that already has used vlan-aware port in k8s VM deployment and seen issues with openstack-cloud-controller-manager? Or you just want to contribute something that may be useful in the future?

Another thing I'm not quite sure about but is very important to openstack-cloud-controller-manager is, have you tested if the subport in neutron trunk can work with octavia?

jingczhang:
Hi,

This PR is to solve the real issue in deployment scenario where VMs are on flat networks. VMs are assigned with trunk ports, IPs are assigned on the subports. Without this PR, those VMs can not be initialized as Kubernetes nodes since cloud-manager-openstack does not see VMs' IP on subports.

This PR has no impact to VMs not assigned with trunk ports. Can you expand the concern with Octavia?

Jing

lingxiankong:
> Hi,
> 
> This PR is to solve the real issue in deployment scenario where VMs are on flat networks. VMs are assigned with trunk ports, IPs are assigned on the subports. Without this PR, those VMs can not be initialized as Kubernetes nodes since cloud-manager-openstack does not see VMs' IP on subports.
> 
> This PR has no impact to VMs not assigned with trunk ports. Can you expand the concern with Octavia?
> 
> Jing

Could you please show me the port information of k8s worker nodes in your deployment? Because I don't have such environment neither the CI, could you also test with this PR the service can be created successfully? I was mentioning Octavia because openstack-cloud-controller-manager is create Octavia load balancer with subnet info of the worker nodes, I'm wondering which subnet the load balancer is going to use and which subnet is used for adding load balancer members.

Sorry for so many questions but I just want to fully understand the use case as this is the first time I've heard that creating k8s cluster on top of vlan aware VMs.

jingczhang:
Hi,

(1)
Below are the excerpts from logs without and with the PR:

=== without the PR ===
I0525 15:39:22.462829       1 node_controller.go:325] Initializing node mycluster-01-control-01 with cloud provider
E0525 15:39:24.132876       1 node_controller.go:370] failed to initialize node mycluster-01-control-01 at cloudprovider: failed to find kubelet node IP from cloud provider


=== with the PR ===
I0525 18:29:06.924591       1 node_controller.go:390] Initializing node mycluster-01-control-01 with cloud provider
I0525 18:29:09.859509       1 openstack.go:698] Claiming to support Zones
I0525 18:29:10.517786       1 node_controller.go:454] Successfully initialized node mycluster-01-control-01 with cloud provider

(2)
Below are the excerpts of nova instance attached ports (trunk) and sub-ports.

[stack@undercloud (overcloudrc) ~]$ nova interface-list mycluster-01-control-01
+------------+--------------------------------------+--------------------------------------+----------------------------------+-------------------+-----+
| Port State | Port ID                              | Net ID                               | IP addresses                     | MAC Addr          | Tag |
+------------+--------------------------------------+--------------------------------------+----------------------------------+-------------------+-----+
| ACTIVE     | 46c7638a-bb37-4ee2-a9cb-be3f4b29f6a5 | 13dd900a-4cfa-4cff-8ad4-ead0e1bfe9ef | 13.13.13.206                     | fa:16:3e:a2:1f:6b | -   |
......

[stack@undercloud (overcloudrc) ~]$ openstack network trunk list 
+--------------------------------------+---------------------------------------------+--------------------------------------+-------------+
| ID                                   | Name                                        | Parent Port                          | Description |
+--------------------------------------+---------------------------------------------+--------------------------------------+-------------+
| b1d0b4fa-dc26-4729-b1eb-c6f63304d911 | mycluster-01-control-01-infra_bond_trunk_1  | 46c7638a-bb37-4ee2-a9cb-be3f4b29f6a5 |             |
......

[stack@undercloud (overcloudrc) ~]$ openstack network subport list --trunk mycluster-01-control-01-infra_bond_trunk_1
+--------------------------------------+-------------------+-----------------+
| Port                                 | Segmentation Type | Segmentation ID |
+--------------------------------------+-------------------+-----------------+
| e86489cb-f37e-424d-8e10-2c1f448e69f9 | vlan              |              50 |
| 2f60a3d1-ad2b-43ae-b9e7-cb64f582449a | vlan              |             701 |
+--------------------------------------+-------------------+-----------------+

[stack@undercloud (overcloudrc) ~]$ openstack port show 2f60a3d1-ad2b-43ae-b9e7-cb64f582449a | grep fixed_ip
...
| fixed_ips               | ip_address='10.10.10.50', subnet_id='b2f87457-af78-4dcd-b03d-aa0b49423535'
...

(3) With regard to your question "could you also test with this PR the service can be created successfully", can you pelase recommend a specific test case? I mean a specific one with steps, so as to make sure there is no confusion.

Jing

lingxiankong:
Thanks for those info, please see my comments inline.

> Below are the excerpts from logs without and with the PR:
> 
> === without the PR ===
> I0525 15:39:22.462829 1 node_controller.go:325] Initializing node mycluster-01-control-01 with cloud provider
> E0525 15:39:24.132876 1 node_controller.go:370] failed to initialize node mycluster-01-control-01 at cloudprovider: failed to find kubelet node IP from cloud provider

From the following console output, I see the trunk port has an IP 13.13.13.206, which should be returned in the function `getAttachedInterfacesByID()`, why the openstack-cloud-controller-manager failed to get node IP?
 
> (3) With regard to your question "could you also test with this PR the service can be created successfully", can you pelase recommend a specific test case? I mean a specific one with steps, so as to make sure there is no confusion.

The main function of openstack-cloud-controller-manager is to provide Service of type LoadBalancer implementation, please make sure the Service can be created successfully with the change.

jingczhang:
Hi,

(1) It is because the nodeIp 10.10.10.50 is on the vlan interface (subport), not on the trunk port

(2) Just to be sure I follow you...I have assumed this PR is transparent to deployments not using trunk ports (by the nature of this PR itself)...while looking at the verification test runs now, I see it failed two tests (test-e2e-comformance-stable-brach-v1.19 and test-lb-octavia). You suggest we to verify Octavia...we will verify that...suppose that is done, this test-lb-octavia test would most likely pass?

Is there a way to check for what reason the above two tests fail?

Jing

lingxiankong:
> Hi,
> 
> (1) It is because the nodeIp 10.10.10.50 is on the vlan interface (subport), not on the trunk port

I know the subport also has an IP, I was asking why 13.13.13.206 was not returned by the existing `getAttachedInterfacesByID()` function [here](https://github.com/kubernetes/cloud-provider-openstack/blob/991c792973620717a789bbcd644683ba276c8888/pkg/openstack/openstack.go#L574)? Maybe I missed something, but even with your change, 13.13.13.206 is still visible to cloud controller manager.

> 
> (2) Just to be sure I follow you...I have assumed this PR is transparent to deployments not using trunk ports (by the nature of this PR itself)...while looking at the verification test runs now, I see it failed two tests (test-e2e-comformance-stable-brach-v1.19 and test-lb-octavia). You suggest we to verify Octavia...we will verify that...suppose that is done, this test-lb-octavia test would most likely pass?

Please ignore those 2 tests, `test-e2e-comformance-stable-brach-v1.19` doesn't test Service functions, and `test-lb-octavia` has been disabled as the backend k8s cluster is still running an old version that can't work with master openstack-cloud-controller-manager. That's the reason we need to manually verify all the PRs related to openstack-cloud-controller-manager before the job is improved and added back.

jingczhang:
Hi,

(1) This PR is NOT to remove the IPs assigned on the ports (trunk port) directly attached to the instance, this is why I said this PR is transparent to the instances not using trunk ports.
This PR is to additionally get the IPs assigned on the sub ports (vlan port). The nodeIP (10.10.10.50) is assigned on the sub ports, VM is vlan-aware, this nodeIP is just one of the IPs assigned on VM's vlan interfaces.

(2) Thanks for the clarification for the failed tests are irrelevant. I will update on Octavia test, we do use Octavia.

Jing

lingxiankong:
> Hi,
> 
> (1) This PR is NOT to remove the IPs assigned on the ports (trunk port) directly attached to the instance, this is why I said this PR is transparent to the instances not using trunk ports.
> This PR is to additionally get the IPs assigned on the sub ports (vlan port). The nodeIP (10.10.10.50) is assigned on the sub ports, VM is vlan-aware, this nodeIP is just one of the IPs assigned on VM's vlan interfaces.

Sorry, you didn't answer my question. In your log:

```
E0525 15:39:24.132876 1 node_controller.go:370] failed to initialize node mycluster-01-control-01 at cloudprovider: failed to find kubelet node IP from cloud provider
```

It said node_controller can't find node IP, but the trunk port does have an IP in your console output. I am trying to better understand the bug you want to fix here. Unfortunately, I don't think this PR could fix that bug.

jingczhang:
Hi,

Not sure what exactly is not clear to you. Let me try again by breaking down the details step-by-step:
(1) VMs are vlan-aware.
(2) VMs are attached with trunk ports, trunk ports are created on flat networks. IPs assigned on trunk ports are useless.
(3) IPs on the vlan interfaces are assigned on the sub-ports (on VLAN networks). One of those IPs is the Kubernetes node IP.
(4) Without this PR, cloud controller only returns IPs on the neutron ports directly attached to the VMs, trunk ports in this case.
(5) With this PR, cloud controller additionally returns IPs on the sub-ports, one of those IPs is the Kubernetes node IP.
(6) This PR HAS resolved the real life issue.

I hope it is clear this time.

Jing

lingxiankong:
> (4) Without this PR, cloud controller only returns IPs on the neutron ports directly attached to the VMs, trunk ports in this case.

This step is not clear to me. You said the cloud controller could find an IP (whether it's useless or not), but how do you explain the log `failed to find kubelet node IP from cloud provider`?

> (5) With this PR, cloud controller additionally returns IPs on the sub-ports, one of those IPs is the Kubernetes node IP.

How cloud controller could choose the right subport's IP rather than trunk port's IP?

jingczhang:
Hi,

(1) Node IP is on on the sub-port, not on the trunk ports.
(2) IPs on the trunk ports are filtered by internal-network-name.

We expect to verify Octivia w/ this networking scheme sometime in June, and will update.

Jing

jingczhang:
Hi @lingxiankong,
We managed to check out the integration with Octavia ingress controller.
No regression issue with VMs what are not attached with trunk ports.
But we find octavia ingress controller does not support direct port i.e. SRIOV.
Is this a known issue? Is someone already working on a fix for enhancement?
I am working on a local fix by adding a configuration parameter "port-type" [normal, direct]. Would you be in favor of this fix or enhancement?

Thank you

lingxiankong:
> We managed to check out the integration with Octavia ingress controller.

@jingczhang, thanks for working on this. Maybe I didn't make myself understood, I'm not saying octavia-ingress-controller, I am suggesting doing some test with this PR in Octavia enabled cloud.

> But we find octavia ingress controller does not support direct port i.e. SRIOV.

No, it doesn't support, again, I'd like to hear more details for the use case.

BTW, seems you submitted some extra commits, please fix.

jingczhang:
Hi @lingxiankong,

Yes, I indeed thought you want the interaction with octavia-ingress-controller be verified. As part of this exercise, Octavia LB is checked out, and it is found not support SRIOV attachment.

Let me try to describe the use case for the pull request again, it is to deploy CNFs in the same way across BM and VM in terms of both deployment scheme and runtime performance.
(1) Kubernetes nodes (Openstack VMs) are attached with SRIOV flat networks.
(2) Inside the Kubernetes nodes, some of the VFs are bonded to support IP vlan CNFs, the rest VFs are pooled to support SRIOV CNFs

I will checkout the failed CI builds.

jichenjc:
>>> test-lb-octavia test 

looks like the test is gone now (should we rewrite this from openlab to praw or it's not there due to OCCM test already cover ?) @lingxiankong @jingczhang seems this PR need such test anyway?

pkg/openstack/instances.go
```diff
@@ -625,5 +664,64 @@ func getAttachedInterfacesByID(client *gophercloud.ServiceClient, serviceID stri
 		return interfaces, err
 	}
 
+	// Check if trunk ports are attached
+	var subports []trunks.Subport
+
+	listOpts := trunks.ListOpts{}
+	allPages, err := trunks.List(network, listOpts).AllPages()
```
jichenjc, line 671: will this impact env does not have Trunk support?

jichenjc (commented):
>>>getAttachedInterfacesByID() only returns neutron ports directly attached to VM.

This is an internal function, what's the impact from end user perspective ? will it be a 
good thing to include into issue or the PR commit msg

jichenjc:
@jingczhang don't worry, those test are obsolete and will be updated later on (overwrite when the code review is done) ,so let's focus on function first

pkg/openstack/instances.go
lingxiankong: Same question for me.

jingczhang: Hi, 

There is no impact on env that does not have trunk support, ourselves do deployment with two flavors using trunks (SRIOV) and not using trunks (legacy VLAN/VxLAN), that is also the reason why those "acceptance" tests passed in CI (they all don't support trunk?).

This pull request was created before the nodeAddresses() function is moved from openstack.go to instances.go, at that time all acceptance tests are passed. I rebased the pull request to the master, after that, openlab/cloud-provider-openstack-acceptance-test-e2e-conformance is able to pass, but those 3 openlab/cloud-provider-openstack-acceptance-test-e2e-conformance-stable-branch-vXX failed, not sure why.

Please help scrutinize the pull request, comment and approve.

Thanks

Jing

pkg/openstack/instances.go
```diff
@@ -625,5 +664,64 @@ func getAttachedInterfacesByID(client *gophercloud.ServiceClient, serviceID stri
...
+	if err != nil {
+		klog.Errorf("Failed to list trunks: %v", err)
+		return interfaces, nil
```
lingxiankong, line 674: I'm wondering why the errors below are all ignored?

Additionally, I'd rather put the logic here into a separate function.

pkg/openstack/instances.go
```diff
@@ -625,5 +664,64 @@ func getAttachedInterfacesByID(client *gophercloud.ServiceClient, serviceID stri
...
+		}
+		n, err := networks.Get(network, p.NetworkID).Extract()
+		if err != nil {
+			return interfaces, nil
+		}
+		var iface = attachinterfaces.Interface{
+			PortState: "ACTIVE",
+			FixedIPs:  []attachinterfaces.FixedIP{},
+			PortID:    p.ID,
+			NetID:     n.Name,
```
lingxiankong, line 711: Why it's setting network name as NetID?

lingxiankong (changes requested):

pkg/openstack/instances.go
jingczhang: (1) The errors were not ignored in the initial pull request. After rebasing to pick up the code re-factor in openstack.go and instances.go, the openlab/cloud-provider-openstack-acceptance-test-e2e-conformance test starts to fail. By ignoring the errors, the CI test passes. I am unable to make a sense of it.
(2) Creating a separate function sounds good, I will do.

pkg/openstack/instances.go
jingczhang: This is because [Networking] config takes network name not network id. IPs on subports need be associated with network names to allow the filtering.
Do you see a better way?

pkg/openstack/instances.go
lingxiankong: > By ignoring the errors, the CI test passes

Ignoring the errors doesn't sound a good approach, if the CI failed just because we catch the error here, there must be bugs. So please add the errors back and let's see what the failure will be.

jingczhang: Reverted back to return error in case of trunk API failures.

PR #1538, opened 2021-05-22
update a url

zhangshj-inspur:
update a url


Fedosin (commented):
/ok-to-test

Technically the old link works, it just redirects you to the new one. I don't mind merging this change, but please add a release note with `None` in the body.

dims:
https://kubernetes.io/security/ redirects are working fine. let's leave this alone please

Issue #1539, opened 2021-05-23
[cinder-csi-plugin] filesystem of a larger volume created from a snapshot/volume is not expanded


alibo:
**Is this a BUG REPORT or FEATURE REQUEST?**: 

/kind bug

**What happened**:

If a new PVC is created from a Snapshot or another volume and its requested size is greater than the source volume/snapshot, its filesystem won't be resized.

**What you expected to happen**:

The filesystem of the new/dest volume should be expanded up to block device size.

**How to reproduce it**:

Source PVC:

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: source-pvc
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
  storageClassName: my-storage-class
```

Cloned PVC + increased its requested storage:

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: cloned-pvc
spec:
  accessModes:
  - ReadWriteOnce
  dataSource:
    kind: PersistentVolumeClaim
    name: source-pvc
  resources:
    requests:
      storage: 2Gi # <-- increased from 1G to 2G
  storageClassName: my-storage-class
```

**Anything else we need to know?**:

It can be expanded by increasing the requested storage of the cloned PVC after it's created, but I think the easier option for a normal k8s user is expanding it in the `NodePublishVolume` or `NodeStageVolume` method implicitly if it's required.

**Environment**:
- cinder-csi-plugin: version: 1.20
- OpenStack version: Rocky
- K8s/Openshift: 1.20/4.7

ramineni:
@alibo please share the logs after enabling verbose to 5 . Add --v=5 here https://github.com/kubernetes/cloud-provider-openstack/blob/master/manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml#L106 .

We do pass the requested larger size in pvc to create volume call, so the same should be passed to cinder backend . Ideally should create of larger size if supported . Do you see any errors ?

alibo:
> We do pass the requested larger size in pvc to create volume call, so the same should be passed to cinder backend . Ideally should create of larger size if supported . Do you see any errors ?

It's true, the cinder creates a larger volume and when it's attached to the node it has the right size when I check it with `lsblk`. 

[...]

PR #1440, opened 2021-03-12
[all] Update to k/k v1.21.0

ramineni:
**What this PR does / why we need it**:
This PR updates k/k to v1.21.0. Some of the test packages have been moved different location in latest k/k repo.
This PR is to sync with the same before adding any new tests.

**Which issue this PR fixes(if applicable)**:
fixes #

**Special notes for reviewers**:

**Release note**:

```release-note
NONE
```

PR #1543, opened 2021-05-26
[cinder-csi-plugin] Verify filesystem size after resize

zuzzas:

Here is an issue that triggered a waterfall of "successful" Expands, which led to improper filesystem size: https://bugs.launchpad.net/openstack-ansible/+bug/1902914. Rescan does not work.

This means that in some cases, we can't trust a positive response from an API, and we should check an actual filesystem size after Expanding it.

**Release note**:

```release-note
[cinder-csi-plugin] Verify filesystem size after resize
```

### DIFF
--- a/pkg/csi/cinder/nodeserver.go
+++ b/pkg/csi/cinder/nodeserver.go
@@ -28,6 +28,7 @@ import (
 	"golang.org/x/net/context"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"k8s.io/cloud-provider-openstack/pkg/util/filesystem"
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/util/resizefs"
 	utilpath "k8s.io/utils/path"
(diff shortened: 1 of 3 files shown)


zuzzas:
Nope, that won't work. Statfs() calculates available size without filesystem-specific.

There are some insights [here](https://github.com/kubernetes/kubernetes/pull/99223/files#r582162509).

pkg/csi/cinder/nodeserver.go
```diff
@@ -540,6 +541,17 @@ func (ns *nodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	if _, err := r.Resize(devicePath, volumePath); err != nil {
 		return nil, status.Errorf(codes.Internal, "Could not resize volume %q:  %v", volumeID, err)
 	}
+
+	newFsSize, err := filesystem.GetFilesystemSize(volumePath)
```
alibo, line 545: You may want to use `r.NeedResize()` instead:

https://github.com/kubernetes/mount-utils/blob/master/resizefs_linux.go#L105

mount-utils is added in this PR: https://github.com/kubernetes/cloud-provider-openstack/pull/1440, but it's interesting it doesn't exist in this file 🤔

It can also be used for https://github.com/kubernetes/cloud-provider-openstack/issues/1539 as well :)

alibo: I thought I've sent it as a comment, not a review! :D

zuzzas: Aye, but the problem is. Due to the bug that I linked, the block device won't be resized as well. I think I'll add a check for the block device size; let's see how it works out. And if I can persuade maintainers to merge this.

zuzzas:
```
[BlockStorage]
rescan-on-resize = true
```

This config triggers a code path on NodeVolumeExpand() that works around the aforementioned bug. This PR has no purpose.

PR #1544, opened 2021-05-27
[cinder-csi-plugin] add cluster id to snapshot metadata

Fedosin:

Now cluster id is set by --cluster argument at the driver's start. When it's set, cinder CSI controller appends the cluster id to volumes metadata during creation. Unfortunately it doesn't happen for volume snapshots, which makes it harder to find them after cluster deletion.

This commit adds cluster id to snapshots metadata to unify the behavior with volumes.

```release-note
[cinder-csi-plugin] Append cluster ID (value of `--cluster`) to metadata of snapshots in the cloud when creating VolumeSnapshot.
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -41,6 +41,10 @@ type controllerServer struct {
 	Cloud  openstack.IOpenStack
 }
 
+const (
+	cinderCSIClusterIDKey = "cinder.csi.openstack.org/cluster"
+)
+
 func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
 	klog.V(4).Infof("CreateVolume: called with args %+v", protosanitizer.StripSecrets(*req))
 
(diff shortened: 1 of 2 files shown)


pkg/csi/cinder/controllerserver.go
```diff
@@ -357,8 +361,14 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
 	} else {
+		// Add cluster ID to the snapshot metadata
+		properties := map[string]string{cinderCSIClusterIDKey: cs.Driver.cluster}
+		for mVal, mKey := range req.Parameters {
+			properties[mKey] = mVal
+		}
+
 		// TODO: Delegate the check to openstack itself and ignore the conflict
-		snap, err = cs.Cloud.CreateSnapshot(name, volumeId, &req.Parameters)
+		snap, err = cs.Cloud.CreateSnapshot(name, volumeId, &properties)
```
jichenjc, line 371: this is not related to this PR
but seems CreateSnapshot and CreateVolume take req.Parameters in different manner

CreateVolume doesn't contain req.parameter into tags and in turn put into metadata...

says param 
```
// Plugin specific parameters passed in as opaque key-value pairs.
  // This field is OPTIONAL. The Plugin is responsible for parsing and
  // validating these parameters. COs will treat these as opaque.
  // Use cases for opaque parameters:
  // - Specify a policy to automatically clean up the snapshot.
  // - Specify an expiration date for the snapshot.
  // - Specify whether the snapshot is readonly or read/write.
  // - Specify if the snapshot should be replicated to some place.
  // - Specify primary or secondary for replication systems that
  //   support snapshotting only on primary.
```

so we should re-consider this to avoid put all those param into metadata or vice versa (line 107 above) , make all param to volume req ..

Fedosin: Yes, that's what I was going to ask... As you said, for volumes we [sanitize](https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/controllerserver.go#L107) req parameters, but for snapshots we put them all in metadata.
I didn't modify this behavior in my PR to support backward compatibility, but I agree that we need to reconsider what we actually put there.

jichenjc: https://github.com/kubernetes/cloud-provider-openstack/issues/1545 opened

PR #1550, opened 2021-05-30
[OCCM] Add new job for e2e conformance test

lingxiankong:
Add new job `cloud-provider-openstack-test-occm-e2e-conformance` to replace the existing one. Use k3s instead of kubeadm to make CI job simple, fast and stable.

We will keep this running for some time before totally replacing `cloud-provider-openstack-acceptance-test-e2e-conformance`

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -183,3 +183,18 @@
               - ^OWNERS$
               - ^SECURITY_CONTACTS$
               - ^.gitignore$
+    cloud-provider-openstack-test-occm-e2e-conformance:
+      jobs:
+        - cloud-provider-openstack-test-occm-e2e-conformance:
+            files:
+              - cmd/openstack-cloud-controller-manager/.*
+              - pkg/openstack/.*
+              - go.mod$
+              - go.sum$
+              - Makefile$
+            irrelevant-files:
+              - ^docs/.*$
+              - ^.*\.md$
+              - ^OWNERS$
+              - ^SECURITY_CONTACTS$
+              - ^.gitignore$


ramineni:
@lingxiankong Thanks for the PR. One suggestion, for this new job , we should consider using kubetest2 instead as kubetest is deprecated https://github.com/kubernetes/test-infra/tree/master/kubetest#deprecation-notice

lingxiankong:
> @lingxiankong Thanks for the PR. One suggestion, for this new job , we should consider using kubetest2 instead as kubetest is deprecated https://github.com/kubernetes/test-infra/tree/master/kubetest#deprecation-notice

Yeah, thanks for pointing that out, I've noticed that announce as well when doing the job refactoring, but I would prefer to only change the way how k8s is installed and keep it running for some time (for stability purpose), then we replace kubetest with kubetest2 in a separate PR.

I created an issue for tracking this: https://github.com/kubernetes/cloud-provider-openstack/issues/1562

ramineni:
> > @lingxiankong Thanks for the PR. One suggestion, for this new job , we should consider using kubetest2 instead as kubetest is deprecated https://github.com/kubernetes/test-infra/tree/master/kubetest#deprecation-notice
> 
> Yeah, thanks for pointing that out, I've noticed that announce as well when doing the job refactoring, but I would prefer to only change the way how k8s is installed and keep it running for some time (for stability purpose), then we replace kubetest with kubetest2 in a separate PR.
> 
> I created an issue for tracking this: #1562

sounds good to me.

/lgtm

PR #1567, opened 2021-06-08
[occm] update url - moving CCM to manifests/

curx:
update documentation, since #1431, the url has changed

```release-note
NONE
```

Signed-off-by: Thorsten Schifferdecker <ts@systs.org>


ramineni:
@curx Good catch ! Thanks for the PR

/lgtm

PR #1570, opened 2021-06-09
[cinder-csi-plugin] Enable ephemeral e2e tests 

ramineni:
This PR enables e2e tests for ephemeral volumes.

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/manifests/cinder-csi-plugin/cinder-csi-nodeplugin.yaml
+++ b/manifests/cinder-csi-plugin/cinder-csi-nodeplugin.yaml
@@ -63,6 +63,7 @@ spec:
             - "--nodeid=$(NODE_ID)"
             - "--endpoint=$(CSI_ENDPOINT)"
             - "--cloud-config=$(CLOUD_CONFIG)"
+            - "--v=5"
           env:
             - name: NODE_ID
               valueFrom:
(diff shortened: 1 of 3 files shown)


ramineni:
@lingxiankong @jichenjc Ready for review. PTAL

manifests/cinder-csi-plugin/cinder-csi-nodeplugin.yaml
```diff
@@ -63,6 +63,7 @@ spec:
             - "--nodeid=$(NODE_ID)"
             - "--endpoint=$(CSI_ENDPOINT)"
             - "--cloud-config=$(CLOUD_CONFIG)"
+            - "--v=5"
```
jichenjc, line 66: previously , I was told in different PR this should not be set to verbose by default as user always has capability to do so by default we should not set log level..

ramineni: @jichenjc Thanks for pointing out. I just added to debug the failures before. Missed to remove it. Ill update

ramineni:
@jichenjc updated. PTAL

jichenjc:
/lgtm
/approve

thanks for the update, we have lgtm before and there is only minor change

PR #1572, opened 2021-06-09
[cinder-csi-plugin][doc] Update CSIMigration status

ramineni:
From release1.21 , CSIMigration is ON by default. This PR
is to reflect the same in doc.

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/cinder-csi-plugin/using-cinder-csi-plugin.md
+++ b/docs/cinder-csi-plugin/using-cinder-csi-plugin.md
@@ -242,9 +242,7 @@ Optionally, to test the driver csc tool could be used. please refer, [usage guid
 
 ## In-tree Cinder provisioner to cinder CSI Migration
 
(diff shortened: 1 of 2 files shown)


lingxiankong:
Please also update https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/migrate-to-ccm-with-csimigration.md#enable-csimigration

ramineni:
> Please also update https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/migrate-to-ccm-with-csimigration.md#enable-csimigration

@lingxiankong Thanks . Updated .I see that all the feature gates are now enabled by default. So removed the same.
PTAL

ramineni:
@jichenjc PTAL

PR #1577, opened 2021-06-19
[OCCM] Restore the previous name for e2e-conformance job

lingxiankong:
The job name is still used in https://github.com/kubernetes/test-infra/blob/master/config/testgrids/kubernetes/sig-cloud-provider/openstack/config.yaml, this PR changes the job name back (but using the new job definition, see [here](https://github.com/theopenlab/openlab-zuul-jobs/blob/master/zuul.d/jobs.yaml#L616)) to avoid test-infra change.

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -115,9 +115,9 @@
               - ^OWNERS$
               - ^SECURITY_CONTACTS$
               - ^.gitignore$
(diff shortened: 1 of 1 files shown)


ramineni:
/lgtm
/approve

Just the name change

PR #1578, opened 2021-06-21
[cinder-csi-plugin] Bump snapshotter version to 4.0.0

ramineni:
This PR bumps snapshotter to 4.0.0 and also added support --extra-create-metadata tag.
Also updates other sidecars to latest version . 
fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/cinder-csi-plugin/values.yaml
+++ b/charts/cinder-csi-plugin/values.yaml
@@ -13,13 +13,13 @@ csi:
     topology: "true"
     image:
       repository: k8s.gcr.io/sig-storage/csi-provisioner
-      tag: v2.1.1
+      tag: v2.2.0
       pullPolicy: IfNotPresent
     resources: {}
   snapshotter:
     image:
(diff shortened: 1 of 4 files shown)


ramineni:
@jichenjc PTAL

charts/cinder-csi-plugin/values.yaml
```diff
@@ -31,7 +31,7 @@ csi:
   livenessprobe:
     image:
       repository: k8s.gcr.io/sig-storage/livenessprobe
-      tag: v2.1.0
+      tag: v2.2.0
```
MPV, line 34: You might have known this already, either way — this change had great impact, in resolving this memory leak: https://github.com/kubernetes-csi/livenessprobe/pull/94

jichenjc: thanks for the info, seems this is not memory leak but memory usage high 
but anyway, glad it fixed the problem and we adopted it

PR #1579, opened 2021-06-25
[manila-csi-plugin] Add support for online volume expansion

### LINKED ISSUE #1475: [manila-csi-plugin] allow share volume expansion
/kind feature

**What happened**:
I'm trying to extend a pvc with the manila storageclass but i'm unable to do so, allowVolumeExpansion is set to true in the SC.
Extending a Manila share seems not possible in current situation. While is is possible in Openstack:
https://docs.openstack.org/manila/latest/admin/shared-file-systems-share-resize.html

I couldn't find any reference in the code to extend a pvc for manila, is it not yet implemented?

Regards,
Paul

**Environment**:
- OpenStack version: train

gman0:

This PR implements online `EXPAND_VOLUME` controller capability.

fixes #1475

**Release note**:

```release-note
[manila-csi-plugin] Add support for online volume expansion
```


gman0:
PTAL @gouthampacha @tombarron @ramineni

gouthampacha (commented):
/ok-to-test

pkg/csi/manila/controllerserver.go
```diff
@@ -424,7 +424,60 @@ func (cs *controllerServer) ListSnapshots(context.Context, *csi.ListSnapshotsReq
...
+		if clouderrors.IsNotFound(err) {
+			return nil, status.Errorf(codes.NotFound, "share %s not found: %v", req.GetVolumeId(), err)
+		}
+
+		return nil, status.Errorf(codes.Internal, "failed to retrieve share %s: %v", req.GetVolumeId(), err)
+	}
+
+	// Check for pending operations on this volume
+	if _, isPending := pendingVolumes.LoadOrStore(share.Name, true); isPending {
+		return nil, status.Errorf(codes.Aborted, "volume named %s is already being processed", share.Name)
```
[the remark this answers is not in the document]
gman0: You're right, and this is all over the codebase. We can consolidate all log messages that refer to Manila storage resources in the next PR just after this one. Would that be ok?

gman0:
@lingxiankong is the CI disabled? No acceptance tests were run.

lingxiankong:
> @lingxiankong is the CI disabled? No acceptance tests were run.

No idea, I checked the http://status.openlabtesting.org/status, seems fine

lingxiankong:
Maybe it was restarted, now I see the job is running.

gman0:
Weird... Thank you @lingxiankong !

pkg/csi/manila/controllerserver.go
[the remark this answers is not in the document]
gouthampacha: Yep, WFM. Thanks, @gman0

gouthampacha (approved):
/lgtm

charts/manila-csi-plugin/templates/controllerplugin-statefulset.yaml
```diff
@@ -59,6 +59,21 @@ spec:
               mountPath: /var/lib/kubelet/plugins/{{ printf "%s.%s" .protocolSelector $.Values.driverName | lower }}
           resources:
 {{ toYaml $.Values.controllerplugin.snapshotter.resources | indent 12 }}
+        - name: {{ .protocolSelector | lower }}-resizer
+          image: "{{ $.Values.controllerplugin.resizer.image.repository }}:{{ $.Values.controllerplugin.resizer.image.tag }}"
+          args:
+            - "--v=5"
```
ramineni, line 65: I think we should avoid using verbose logging by default . Its applicable for all other sidecars as well

gman0: I checked cinder-csi's manifests and indeed there's no `--v` flag in any of the sidecars. I can fix this in bulk in a separate PR. WDYT?

ramineni: sounds good

pkg/csi/manila/controllerserver.go
```diff
@@ -424,7 +424,60 @@ func (cs *controllerServer) ListSnapshots(context.Context, *csi.ListSnapshotsReq
...
+
+	// Try to expand the share
+
+	currentSizeInBytes := int64(share.Size * bytesInGiB)
+
+	if currentSizeInBytes >= req.GetCapacityRange().GetRequiredBytes() {
+		// Share is already larger than requested size
+
+		return &csi.ControllerExpandVolumeResponse{
+			CapacityBytes: currentSizeInBytes,
```
ramineni, line 468: Isnt it better to add NodeExpansionRequired: false, if node expansion not required? It says required field in csi spec.
              // Whether node expansion is required for the volume. When true
              // the CO MUST make NodeExpandVolume RPC call on the node. This field
              // is REQUIRED.
              bool node_expansion_required = 2;

pkg/csi/manila/controllerserver.go
```diff
@@ -424,7 +424,60 @@ func (cs *controllerServer) ListSnapshots(context.Context, *csi.ListSnapshotsReq
...
+	}
+
+	desiredSizeInGiB := int(req.GetCapacityRange().GetRequiredBytes() / bytesInGiB)
+
+	if err = extendShare(share, desiredSizeInGiB, manilaClient); err != nil {
+		return nil, err
+	}
+
+	return &csi.ControllerExpandVolumeResponse{
+		CapacityBytes: int64(desiredSizeInGiB * bytesInGiB),
```
ramineni, line 479: ditto

pkg/csi/manila/controllerserver.go
```diff
@@ -424,7 +424,60 @@ func (cs *controllerServer) ListSnapshots(context.Context, *csi.ListSnapshotsReq
...
+		}
+
+		return nil, status.Errorf(codes.Internal, "failed to retrieve share %s: %v", req.GetVolumeId(), err)
+	}
+
+	// Check for pending operations on this volume
+	if _, isPending := pendingVolumes.LoadOrStore(share.Name, true); isPending {
+		return nil, status.Errorf(codes.Aborted, "volume named %s is already being processed", share.Name)
+	}
+	defer pendingVolumes.Delete(share.Name)
```
ramineni, line 458: @gman0 Just wondering here, why we need to delete volumes here if some operation is in progress ?

pkg/csi/manila/identityserver.go
```diff
@@ -57,6 +57,13 @@ func (ids *identityServer) GetPluginCapabilities(ctx context.Context, req *csi.G
 				},
 			},
 		},
+		{
+			Type: &csi.PluginCapability_VolumeExpansion_{
+				VolumeExpansion: &csi.PluginCapability_VolumeExpansion{
+					Type: csi.PluginCapability_VolumeExpansion_ONLINE,
```
ramineni, line 63: @gman0 Is OFFLINE capability not applicable for manila csi?

pkg/csi/manila/controllerserver.go
gman0: `NodeExpansionRequired bool` is initialized to `false` by default

pkg/csi/manila/identityserver.go
gman0: I think `VolumeExpansion.OFFLINE` and `VolumeExpansion.ONLINE` pertain to what happens when the volume is in use while its being expanded, rather than if it's allowed to be online/offline.

> This call MAY be made by the CO during any time in the lifecycle of the volume after creation if plugin has `VolumeExpansion.ONLINE` capability.
https://github.com/container-storage-interface/spec/blob/master/spec.md#controllerexpandvolume

Manila doesn't care, so I think advertising online expansion only is enough. I asked at the CSI slack channel to make sure though.

gman0: I'll send a comment here once somebody replies.

pkg/csi/manila/controllerserver.go
gman0: `pendingVolumes` is a [`sync.Map`](https://golang.org/pkg/sync/#Map). It acts like a try-lock: it tries to lock a resource (with share name as the key), protecting it against concurrent use. The `Delete()` just deletes a key from the map, freeing the lock for that volume.

pkg/csi/manila/identityserver.go
gman0: Got a response: external-resizer doesn't check offline/online caps anyway, and they will be even deprecated from the CSI spec. See release notes at https://github.com/container-storage-interface/spec/releases/tag/v1.5.0

tombarron: Agree that manila doesn't care if share is in use or not.

tombarron (approved):
This change seems good to me if ramineni's questions have been answered.

pkg/csi/manila/controllerserver.go
ramineni: Ah, got it. Thanks

pkg/csi/manila/controllerserver.go
ramineni: ok fine then

pkg/csi/manila/identityserver.go
ramineni: Ok

ramineni:
Looks good to me

/approve

gman0:
Thank you all for reviews! I'll follow up with pull requests like we agreed: consolidating log messages and lowering verbosity levels.

PR #1580, opened 2021-06-30
[cinder-csi-plugin] Sync RBAC as per latest sidecar versions

ramineni:
This PR is to sync the RBAC rules as per latest sidecars
versions.

fixes #

**Release note**:

```release-note
NONE
```


lingxiankong:
The RBAC manifests are tested in the job cloud-provider-openstack-e2e-test-csi-cinder which is great

/approve

PR #1583, opened 2021-07-02
[manila-csi-plugin] Use low verbosity logging by default

gman0:

This PR addresses a review comment here https://github.com/kubernetes/cloud-provider-openstack/pull/1579#discussion_r660297101 .

Deployment manifests and manila-csi helm chart have been modified so that they don't use `--v=LEVEL` log verbosity level flags by default.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/manila-csi-plugin/Chart.yaml
+++ b/charts/manila-csi-plugin/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: latest
 description: Manila CSI Chart for OpenStack
 name: openstack-manila-csi
-version: 1.3.0
+version: 1.3.1
 home: http://github.com/kubernetes/cloud-provider-openstack
 icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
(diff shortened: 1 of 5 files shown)


ramineni:
Thanks

/lgtm
/approve

gman0:
@ramineni @jichenjc  I just noticed:

https://github.com/kubernetes/cloud-provider-openstack/blob/ab36705fb740299c57cfeeefec6f4185e5f68579/pkg/csi/cinder/utils.go#L80-L90

and

https://github.com/kubernetes/cloud-provider-openstack/blob/ab36705fb740299c57cfeeefec6f4185e5f68579/pkg/csi/manila/driver.go#L311-L323

We're using levels 3 and 5 for logging GRPC requests and responses. Bugs like https://github.com/kubernetes/cloud-provider-openstack/issues/1587 would be very hard to debug without having access to logs, especially when it's not obvious how to reproduce the bug -- so even after users turn up the verbosity level they wouldn't be able to deliver logs for us to troubleshoot.

Do you think it's ok to always log at least the GRPC communication?

jichenjc:
>>>Do you think it's ok to always log at least the GRPC communication?

not sure how much such logs will be logged, do we have any data about that?  tradeoff need to be considered for issue tracking and tons of logs

PR #1585, opened 2021-07-06
[cinder-csi-plugin] Add volume condition for controller and node plugin

### LINKED ISSUE #1584: [cinder-csi-plugin] Support VOLUME_CONDITION of CSI

>
> /kind bug
 /kind feature

**What happened**:
has `// VOLUME_CONDITION controller capability is supported.`
we need enable this so that we can support `https://github.com/kubernetes-csi/external-health-monitor/`

if we don't support this option, we will get 
`CSI driver does not support Controller ListVolumes and GetVolume service or does not implement VolumeCondition, exiting` at the monitor container start stage

**How to reproduce it**:

**Environment**:
- OpenStack version:
- Others:

jichenjc:

fixes #1584

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/openstack/openstack_volumes.go
+++ b/pkg/csi/cinder/openstack/openstack_volumes.go
@@ -398,3 +398,14 @@ func (os *OpenStack) diskIsUsed(volumeID string) (bool, error) {
 func (os *OpenStack) GetBlockStorageOpts() BlockStorageOpts {
 	return os.bsOpts
 }
+
+// IsAbnormalVolume returns whether the volume is abnormal
+func IsAbnormalVolume(status string) bool {
+	for _, eState := range volumeErrorStates {
+		if status == eState {
+			return true
+		}
+	}
+
+	return false
+}
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -313,6 +313,12 @@ func (cs *controllerServer) ListVolumes(ctx context.Context, req *csi.ListVolume
 		for _, attachment := range v.Attachments {
 			status.PublishedNodeIds = append(status.PublishedNodeIds, attachment.ServerID)
 		}
+
+		condition := &csi.VolumeCondition{
+			Abnormal: openstack.IsAbnormalVolume(v.Status),
+			Message:  v.Status,
+		}
+		status.VolumeCondition = condition
 		ventry.Status = status
 
 		ventries = append(ventries, &ventry)
@@ -586,6 +592,11 @@ func (cs *controllerServer) ControllerGetVolume(ctx context.Context, req *csi.Co
 	for _, attachment := range volume.Attachments {
 		status.PublishedNodeIds = append(status.PublishedNodeIds, attachment.ServerID)
 	}
+	condition := &csi.VolumeCondition{
+		Abnormal: openstack.IsAbnormalVolume(volume.Status),
+		Message:  volume.Status,
+	}
+	status.VolumeCondition = condition
 	ventry.Status = status
 
 	return &ventry, nil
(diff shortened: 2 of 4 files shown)


jichenjc:
@lingxiankong @ramineni can you help take a look? The error is not related

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -398,3 +398,14 @@ func (os *OpenStack) diskIsUsed(volumeID string) (bool, error) {
 func (os *OpenStack) GetBlockStorageOpts() BlockStorageOpts {
 	return os.bsOpts
 }
+
+// IsAbnormalVolume returns whether the volume is abnormal
+func IsAbnormalVolume(status string) bool {
+	for _, eState := range volumeErrorStates {
+		if status == eState {
+			return false
```
lingxiankong, line 406: I think you mean `return true` here.

pkg/csi/cinder/nodeserver_test.go
```diff
@@ -440,6 +444,10 @@ func TestNodeGetVolumeStatsFs(t *testing.T) {
 			{Total: FakeFsStats.TotalBytes, Available: FakeFsStats.AvailableBytes, Used: FakeFsStats.UsedBytes, Unit: csi.VolumeUsage_BYTES},
 			{Total: FakeFsStats.TotalInodes, Available: FakeFsStats.AvailableInodes, Used: FakeFsStats.UsedInodes, Unit: csi.VolumeUsage_INODES},
 		},
+		VolumeCondition: &csi.VolumeCondition{
+			Abnormal: true,
+			Message:  "available",
```
lingxiankong, line 449: "available" means abnormal?

pkg/csi/cinder/nodeserver_test.go
```diff
@@ -401,6 +401,10 @@ func TestNodeGetVolumeStatsBlock(t *testing.T) {
 		Usage: []*csi.VolumeUsage{
 			{Total: FakeBlockDeviceStats.TotalBytes, Unit: csi.VolumeUsage_BYTES},
 		},
+		VolumeCondition: &csi.VolumeCondition{
+			Abnormal: true,
+			Message:  "available",
```
lingxiankong, line 406: "available" means abnormal?

lingxiankong (changes requested):

pkg/csi/cinder/nodeserver_test.go
jichenjc: yes, I need update this logic

pkg/csi/cinder/nodeserver_test.go
jichenjc: ditto

jichenjc:
@ramineni @lingxiankong can you help review this? 
I Think we can overwrite old openlab CI ? thanks

lingxiankong:
/override openstack-cloud-keystone-authentication-authorization-test

@jichenjc May I know what is the current status of the job openstack-cloud-keystone-authentication-authorization-test? It's here but I see it's never passed.

ramineni:
@jichenjc Thanks for the PR. Trying to understand better, is this volume conditions are used only if external health monitor is enabled is supported/enabled? not used otherwise? because, I see that external health monitor is a separate sidecar and alpha feature only, so if we dont have plans to add support now , I dont think we should add this

jichenjc:
@ramineni I thought we should catch up with CSI spec and adopt new features time to time?
or if it's alpha, maybe we can put it as optional feature of CPO? let me know your thoughts :)

ramineni:
> @ramineni I thought we should catch up with CSI spec and adopt new features time to time? 

@jichenjc +1 , yes I agree.

> or if it's alpha, maybe we can put it as optional feature of CPO? let me know your thoughts :)

Actually 3 things crossed my mind - 
1. I see external health monitor is a big feature and needs proper testing /configuration before adding it here. 
2. As VOLUME_CONDITION is only part of the feature, may be its better to add it as a whole feature with tests enabled.
3. And as its only alpha , Im wondering is it worth the effort , may be we should wait until beta . 

wdyt?

PR #1589, opened 2021-07-08
Fix `make import-boss` gate failure 

ramineni:
**What this PR does / why we need it**:
This PR enables installing import-boss using go install command

**Which issue this PR fixes(if applicable)**:
fixes #1588

**Special notes for reviewers**:

**Release note**:

```release-note
NONE
```

PR #1590, opened 2021-07-08
[manila-csi-plugin] Do not populate nodeStageCache when an error occurred previously

### LINKED ISSUE #1587: [manila-csi-plugin] Possible cache corruption when an error occurs during NodeStageVolume gRPC call?

/kind bug

**What happened**:

After a network connectivity issue (presumably), some of our nodes weren't able to mount any Manila PVCs any more. The affected Pods showed events saying "stage secrets cannot be nil or empty", and similar messages showed up in the logs fo the Manila CSI node plugin:

```
I0707 11:21:43.071234       1 driver.go:309] [ID:164772] GRPC call: /csi.v1.Node/NodeStageVolume
I0707 11:21:43.071251       1 driver.go:310] [ID:164772] GRPC request: {"secrets":"***stripped***","staging_target_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-9e5dea96-d06b-4ed4-8676-ce10baedd75a/globalmount","volume_capability":{"AccessType":{"Mount":{}},"access_mode":{"mode":5}},"volume_context":{"cephfs-mounter":"kernel","shareAccessID":"cf32d375-4f2b-4e8d-a7f4-fcb3117de78e","shareID":"3b3a8bf1-3c86-4966-b674-423aace78b11","storage.kubernetes.io/csiProvisionerIdentity":"1622370232102-8081-cephfs.manila.csi.openstack.org"},"volume_id":"3b3a8bf1-3c86-4966-b674-423aace78b11"}
I0707 11:21:43.071766       1 builder.go:39] [ID:1740] FWD GRPC call: /csi.v1.Node/NodeStageVolume
I0707 11:21:43.071774       1 builder.go:40] [ID:1740] FWD GRPC request: {"staging_target_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-9e5dea96-d06b-4ed4-8676-ce10baedd75a/globalmount","volume_capability":{"AccessType":{"Mount":{}},"access_mode":{"mode":5}},"volume_id":"3b3a8bf1-3c86-4966-b674-423aace78b11"}
E0707 11:21:43.074678       1 driver.go:313] [ID:164772] GRPC error: rpc error: code = InvalidArgument desc = stage secrets cannot be nil or empty
```

From reading the source, I inferred (please feel free to correct me) that the messages logged in both `driver.go:310` and `builder.go:40` should both be instances of the `csi.NodeStageVolumeRequest` message (first received by the Manila CSI plugin, and then forwarded to another CSI Node Plugin, in our case the CephFS CSI plugin listening at `unix:///var/lib/kubelet/plugins/cephfs.csi.ceph.com/csi.sock`). The CephFS CSI plugin [validates these messages](https://github.com/ceph/ceph-csi/blob/dd0884310f07c06d04c23d72ee386507f2be208f/internal/util/validate.go#L23-L25) to contain a `secrets` property -- which the message logged in `builder.go:40` apparently does not contain, resulting in the observed `stage secrets cannot be nil or empty` error:

The issue did not resolve itself automatically after the initial network issue was resolved, but persisted until the Manila CSI plugin pod on that node was (manually!) deleted and recreated. With a freshly created Pod, the volumes were able to be mounted again.

The issue should have resolved itself without having to manually delete the CSI node plugin pod.

**How to reproduce it**:

Honestly, no idea. I analysed the source code and might have an idea of the underlying issue -- see next section for more details.

I believe this issue might occur due to a cache corruption while handling the `csi.v1.Node/NodeStageVolume` gRPC call.

The `manila.nodeServer` struct maintains an internal cache of volume contexts and stage secrets in its `nodeStageCache` property:

https://github.com/kubernetes/cloud-provider-openstack/blob/ab36705fb740299c57cfeeefec6f4185e5f68579/pkg/csi/manila/nodeserver.go#L35-L42

Whenever a `NodeStageVolume` gRPC call is processed, the volume ID will be used for a cache lookup. On a cache miss, the `ns.buildVolumeContext` and `buildNodeStageSecret` functions will be invoked to determine the volume context and stage secrets:

https://github.com/kubernetes/cloud-provider-openstack/blob/ab36705fb740299c57cfeeefec6f4185e5f68579/pkg/csi/manila/nodeserver.go#L262-L282

Later, the volume context and stage secrets (either retrieved from cache or computed on-demand) will replace the original values in the `csi.NodeStageVolumeRequest` struct, before the request is forwarded:

https://github.com/kubernetes/cloud-provider-openstack/blob/ab36705fb740299c57cfeeefec6f4185e5f68579/pkg/csi/manila/nodeserver.go#L292-L293

HOWEVER:

I believe the cache logic described above might contain an implementation error: When either `ns.buildVolumeContext` or `buildNodeStageSecret` returns an error, the cache will be populated anyway (in which case any one of `volumeCtx` or `stageSecret` might be `nil`) BEFORE the error is checked.

The initial gRPC call will fail (since the error _is_ checked eventually), but because the cache is already populated at that point, subsequent gRPC calls will trigger a cache hit and the nil values will be used further down the line. This will cause the forwarded gRPC messages to have an empty volume context and stage secret and be rejected by the validation layer in the upstream CephFS CSI node plugin (or its gRPC service implementation of the CSI node server, to be more precise).

Since `ns.buildVolumeContext` instantiates a Manila client (and presumably communicates with the Manila API), it seems plausible that this function might fail when the Manila client cannot connect. This is in turn a plausible explanation for the behaviour described initially.

DISCLAIMER:

I concluded this from reading the source code, but did not verify my assumptions with a debugger. However, it seems obvious to me that the error return value is being checked too late in this case, and my assumptions provide a plausible explanation for the originally observed behaviour.

**Environment**:
- manila-csi-plugin version: In our cluster, we're using `docker.io/k8scloudprovider/manila-csi-plugin:v1.19.0`. The code samples in my analysis above are the current master branch (latest commit to touch that file was b20ce96, which is also the case for v1.12.1)
- cephcsi version: `quay.io/cephcsi/cephcsi:v3.1.2`
- OpenStack version: Train release, exact version unknown

martin-helmich:

This change prevents the node server's `nodeStageCache` from being populated with invalid (nil) values, when an error occurred previously.

fixes #1587

n/a

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/manila/nodeserver.go
+++ b/pkg/csi/manila/nodeserver.go
@@ -273,7 +273,9 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 			publishSecret, err = buildNodePublishSecret(accessRight, getShareAdapter(ns.d.shareProto), volID)
 		}
 
-		ns.nodeStageCache[volID] = stageCacheEntry{volumeContext: volumeCtx, stageSecret: stageSecret, publishSecret: publishSecret}
+		if err == nil {
+			ns.nodeStageCache[volID] = stageCacheEntry{volumeContext: volumeCtx, stageSecret: stageSecret, publishSecret: publishSecret}
+		}
 	}
 	ns.nodeStageCacheMtx.Unlock()
 


gman0:
/lgtm

@martin-helmich thank you!

gman0:
Don't mind `pull-cloud-provider-openstack-check`, it's being fixed.

jichenjc:
/approve

let's wait for the `check` job pass

martin-helmich:
Am I correct in assuming the failing build is unrelated to my change? Is there anything left to do for me at this point? 😳

gman0:
@martin-helmich correct, no action needed from your side. I believe the failing CI job is being fixed here https://github.com/kubernetes/cloud-provider-openstack/pull/1589

gouthampacha (approved):
/lgtm

gouthampacha:
https://github.com/kubernetes/cloud-provider-openstack/pull/1589 just merged

PR #1591, opened 2021-07-09
Update cinder csi rpc log with count number

jichenjc:
steal from manila
https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/manila/driver.go#L312

fixes #

**Release note**:

```release-note
NONE
```


lingxiankong:
/override openstack-cloud-csi-cinder-e2e-test

Issue #720, opened 2019-08-12
[Manila CSI] add support for CephFS snapshots


gman0:
**Current situation:**

Currently, CephFS backends in Manila are limited to only creating snapshots - Manila does not offer `create_share_from_snapshot_support` capability for CephFS just yet. From CSI side of things, the spec doesn't allow advertising `CREATE_DELETE_SNAPSHOT` without supporting snapshot volume source in `CreateVolumeRequest` - which means that csi-manila must forbid CephFS snapshots altogether.

**Proposed solution:**

Given this is one of the more sought-after features, csi-manila tries to fill out this gap with in-driver functionality (i.e. a compatibility layer) until it's supported natively by Manila. The driver checks for Manila capabilities by reading extra specs from the share type of the source share. Based on those the driver either chooses to use Manila's native support or a compatibility layer, if one exists.

In order to create a share from a CephFS snapshot, the data needs to be copied over from the snapshot into a new share. CSI Manila will therefore create a blank (destination) share, mount it, mount the source share containing the snapshot and copy the data to destination using rsync.

**Implementation:**

- https://github.com/gman0/cloud-provider-openstack/tree/csimanila-compat-caps : basis for recognizing Manila capabilities and using compatibility layers when needed
- https://github.com/gman0/cloud-provider-openstack/tree/csimanila-cephfs-snap-compat : `create_share_from_snapshot_support` compatibility layer for native CephFS

The code is finished and ready for PRs. I can post those right after https://github.com/kubernetes/cloud-provider-openstack/issues/714 is done.

gman0:
/remove-lifecycle stale

This is still ongoing...

PR #1592, opened 2021-07-09
[manila-csi-plugin] snapshot caps check

gman0:
This PR removes hard-coded checks for CephFS shares, preventing them from being snapshotted. Instead, share type capabilities are checked and creating snapshots is accepted or denied based on that.

fixes https://github.com/kubernetes/cloud-provider-openstack/issues/720

We've abandoned the original idea described in #720. Since Manila supports snapshot restoration with CephFS backends since the Wallaby release, we're going with that instead.

**Release note**:

```release-note
[manila-csi-plugin] Added support for snapshot and snapshot restore of CephFS shares where backends support it
```


pkg/csi/manila/controllerserver.go
```diff
@@ -248,6 +228,33 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+	}
+
+	requiredShareTypeCaps := []capabilities.ManilaCapability{
+		capabilities.ManilaCapabilitySnapshot,
+		capabilities.ManilaCapabilityShareFromSnapshot,
+	}
+
+	for _, c := range requiredShareTypeCaps {
+		if !shareTypeCaps[c] {
+			return nil, status.Errorf(codes.InvalidArgument,
```
gman0, line 252: The main objective here is to ensure that it will be possible to create a new volume with this snapshot as its source and I'm not sure I'm going the right way about this.

Is it possible to create a share from a snapshot, but using a share type that's different than what was used to create the original share? If so, then checking share type caps here doesn't make much sense -- because maybe one type had `create_share_from_snapshot_support` disabled, but the other may have it enabled and share creation would succeed.

From my tests, it seems the caps on the share (not on the share type) seemed to have the final say. Even if I tried to create a share from snapshot with a type that had `create_share_from_snapshot_support=True`, Manila would refuse to create it if the original share reported `create_share_from_snapshot_support=False`. Unfortunately, there's no way for manila-csi to know at the moment. The gophercloud lib is currently missing `create_share_from_snapshot_support` field in [`shares.Share` struct](https://github.com/gophercloud/gophercloud/blob/932f0916e113975e4d34bcd1d2dce9c01aa96570/openstack/sharedfilesystems/v2/shares/results.go#L18-L76). I'll push a patch there. I'm not entirely sure we should merge this PR without it.

gman0:
PTAL @tombarron @gouthampacha

pkg/csi/manila/controllerserver.go
gman0: https://github.com/gophercloud/gophercloud/pull/2191

gouthampacha: @gman0: no you can't specify both snapshot and share type when creating new shares; new shares from snapshots are expected to retain the share type of the parent; 

and yes, the capability of the share is the best place to look since share type extra specs are mutable. While share type extra-specs are mutable, they're assumed to not change often and manila administrators know that changes aren't going to apply to resources that are already provisioned. However, i totally agree with not relying on the share type extra-specs. 

your proposed way of just checking the share's "snapshot_support" and "create_share_from_snapshot_support" seems correct.

pkg/csi/manila/share.go
```diff
@@ -75,7 +90,7 @@ func getOrCreateShare(shareName string, createOpts *shares.CreateOpts, manilaCli
 		return share, 0, nil
 	}
 
-	return waitForShareStatus(share.ID, shareCreating, shareAvailable, false, manilaClient)
+	return waitForShareStatus(share.ID, []string{shareCreating, shareCreatingFromSnapshot}, shareAvailable, false, manilaClient)
```
gouthampacha, line 93: Awesome!

gouthampacha (commented):
Thanks for the changes @gman0

pkg/csi/manila/controllerserver.go
gman0: @gouthampacha thank you for explanation and suggestions! I'll fix this like we've agreed after the `create_share_from_snapshot_support` patch in gophercloud is in.

gman0:
Do we wait for new release of gophercloud or is it ok if I pin the gophercloud dependency to a particular commit @lingxiankong @ramineni ? I'd like this feature to still make it for v1.22 if possible

pkg/csi/manila/share.go
tombarron: Yeah, this looks right.  (@gouthampacha reminded me yesterday)

ramineni:
> Do we wait for new release of gophercloud or is it ok if I pin the gophercloud dependency to a particular commit @lingxiankong @ramineni (see [#1592 (comment)](https://github.com/kubernetes/cloud-provider-openstack/pull/1592#discussion_r667056421) )? I'd like this feature to still make it for v1.22 if possible

@gman0 we can pin to particular commit , I dont see any issue in that. 
But when is the new release planned?

gman0:
Thanks @ramineni . I just asked Gophercloud's Joe Topjian to find out whether v0.19.0 is planned still this month. It seems releases are 1 month apart, and the last one was 11th of June -- so maybe the next one is coming very soon? If it's not soon enough I'll pin the dependency to the commit from the PR linked earlier.

pkg/csi/manila/share.go
gman0: Thank you both! I'll update this PR once I find out what do we do with updating gophercloud lib dependency.

gman0:
Jobs `cloud-provider-openstack-e2e-test-csi-cinder` and `cloud-provider-openstack-multinode-csi-migration-test` are stuck. I don't think it's caused by this PR.

gman0:
I've updated the caps checks like we've agreed. @tombarron @gouthampacha could you please take a look? @ramineni I'm not sure about the failing jobs. I think I've seen these ones failing in other PRs too... Is that a blocker for this PR?

tombarron:
/lgtm
/approve
But there are files here outside pkg/csi/manila so these are just for the record.

gouthampacha (approved):
/lgtm
/assign @ramineni 

Thank you @gman0

gman0:
Thank you all for feedback! @ramineni your approval didn't trigger a merge so I guess that will happen once CI passes?

ramineni:
@gman0 looks like gate resolution might take some time. If you have setup, could you quickly run e2e on cinder csi manually , so that we  can override and go ahead with the merge

gman0:
@ramineni 

I manually deployed `examples/cinder-csi-plugin/nginx.yaml` that's used by cinder-csi's acceptance test. I can supply more info if you need.

1. Checked-out branch from this PR
2. `make image-cinder-csi-plugin`
3. Customized `cloud-provider-openstack/charts/cinder-csi-plugin/values.yaml` accordingly, along with changes to `cloud-config` secret.

```
$ kubectl get all | grep cinder
pod/openstack-cinder-csi-controllerplugin-0        6/6     Running   0          4m26s
pod/openstack-cinder-csi-nodeplugin-8pq6s          3/3     Running   0          4m26s
service/openstack-cinder-csi-controllerplugin      ClusterIP   None         <none>        <none>      4m26s
daemonset.apps/openstack-cinder-csi-nodeplugin      1         1         1       1            1           <none>          4m26s
statefulset.apps/openstack-cinder-csi-controllerplugin      1/1     4m27s
```

```
$ kubectl create -f examples/cinder-csi-plugin/nginx.yaml 
storageclass.storage.k8s.io/csi-sc-cinderplugin created
persistentvolumeclaim/csi-pvc-cinderplugin created
pod/nginx created
```

```
$ openstack volume list
+--------------------------------------+------------------------------------------+--------+------+-----------------------------------------+
| ID                                   | Name                                     | Status | Size | Attached to                             |
+--------------------------------------+------------------------------------------+--------+------+-----------------------------------------+
| 3594c5db-a6f6-4142-a022-767279cfe944 | pvc-6045e704-08b5-4b12-8b66-b01ad43b3339 | in-use |    1 | Attached to rvasek-k8s-dev on /dev/vdb  |
+--------------------------------------+------------------------------------------+--------+------+-----------------------------------------+
$ openstack volume show 3594c5db-a6f6-4142-a022-767279cfe944
+------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Field                        | Value                                                                                                                                                                                                                                                                                                     |
+------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| attachments                  | [{'id': '3594c5db-a6f6-4142-a022-767279cfe944', 'attachment_id': 'e02b7f79-271d-436c-8afc-a8742283ad2c', 'volume_id': '3594c5db-a6f6-4142-a022-767279cfe944', 'server_id': 'bb3dd693-6af5-459b-b69e-d1d0a9104125', 'host_name': None, 'device': '/dev/vdb', 'attached_at': '2021-08-11T14:57:43.000000'}] |
| availability_zone            | nova                                                                                                                                                                                                                                                                                                      |
| bootable                     | false                                                                                                                                                                                                                                                                                                     |
| consistencygroup_id          | None                                                                                                                                                                                                                                                                                                      |
| created_at                   | 2021-08-11T14:57:00.000000                                                                                                                                                                                                                                                                                |
| description                  | Created by OpenStack Cinder CSI driver                                                                                                                                                                                                                                                                    |
| encrypted                    | False                                                                                                                                                                                                                                                                                                     |
| id                           | 3594c5db-a6f6-4142-a022-767279cfe944                                                                                                                                                                                                                                                                      |
| multiattach                  | False                                                                                                                                                                                                                                                                                                     |
| name                         | pvc-6045e704-08b5-4b12-8b66-b01ad43b3339                                                                                                                                                                                                                                                                  |
| os-vol-tenant-attr:tenant_id | ce8c897f-6252-4454-b1cc-bfee7d447572                                                                                                                                                                                                                                                                      |
| properties                   | cinder.csi.openstack.org/cluster='kubernetes', csi.storage.k8s.io/pv/name='pvc-6045e704-08b5-4b12-8b66-b01ad43b3339', csi.storage.k8s.io/pvc/name='csi-pvc-cinderplugin', csi.storage.k8s.io/pvc/namespace='default'                                                                                      |
| replication_status           | None                                                                                                                                                                                                                                                                                                      |
| size                         | 1                                                                                                                                                                                                                                                                                                         |
| snapshot_id                  | None                                                                                                                                                                                                                                                                                                      |
| source_volid                 | None                                                                                                                                                                                                                                                                                                      |
| status                       | in-use                                                                                                                                                                                                                                                                                                    |
| type                         | standard                                                                                                                                                                                                                                                                                                  |
| updated_at                   | 2021-08-11T14:57:44.000000                                                                                                                                                                                                                                                                                |
| user_id                      | rvasek                                                                                                                                                                                                                                                                                                    |
+------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

```
$  kubectl get pod nginx
NAME    READY   STATUS    RESTARTS   AGE
nginx   1/1     Running   0          3m22s
```

Clean up was successful too.

ramineni:
@gman0 Thanks

ramineni:

ramineni:

PR #1600, opened 2021-07-19
[cinder-csi-plugin (chart)] Make a cinder chart compatible with the occm one

yerzhant:
Make cinder plugin installation more seamless, i.e. make it easier to install the plugin right after installation of occm.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/charts/cinder-csi-plugin/templates/controllerplugin-statefulset.yaml
+++ b/charts/cinder-csi-plugin/templates/controllerplugin-statefulset.yaml
@@ -136,7 +136,7 @@ spec:
             secretName: {{ .Values.secret.name }}
         {{- else }}
           hostPath:
-            path: /etc/kubernetes
+            path: /etc/config
         {{- end }}
         {{ .Values.csi.plugin.volumes | toYaml | trimSuffix "\n" | nindent 8 }}
       affinity: {{ toYaml .Values.csi.plugin.controllerPlugin.affinity | nindent 8 }}
--- a/charts/cinder-csi-plugin/templates/controllerplugin-statefulset.yaml
+++ b/charts/cinder-csi-plugin/templates/controllerplugin-statefulset.yaml
@@ -106,7 +106,7 @@ spec:
             - name: CSI_ENDPOINT
               value: unix://csi/csi.sock
             - name: CLOUD_CONFIG
-              value: /etc/kubernetes/cloud-config
+              value: /etc/config/cloud.conf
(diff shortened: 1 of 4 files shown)


yerzhant:
I signed it.

jichenjc:
I assume when you say `occm one`, you are saying manifests? such as https://github.com/kubernetes/cloud-provider-openstack/blob/master/manifests/cinder-csi-plugin/cinder-csi-controllerplugin.yaml ?

yerzhant:
@jichenjc Sorry for disambiguation. I meant a helm chart of an occm. The idea behind the PR is to streamline process of installation of cinder. I.e. one can right after this:

`helm install openstack-ccm cpo/openstack-cloud-controller-manager -f occm.yaml`

run this:

`helm install cinder-csi cpo/openstack-cinder-csi -f cinder.yaml`,

as the occm's chart creates a secret that is incompatible with cinder's chart and expects a `ca-file` to be placed in another dir. As for the `cacert` location - it is solvable/doable, but as for the secret - one need to create a second secret.

jichenjc:
ok, well understand, thanks for the detailed info

charts/cinder-csi-plugin/templates/controllerplugin-statefulset.yaml
```diff
@@ -106,7 +106,7 @@ spec:
             - name: CSI_ENDPOINT
               value: unix://csi/csi.sock
             - name: CLOUD_CONFIG
-              value: /etc/kubernetes/cloud-config
+              value: /etc/config/cloud.conf
```
ramineni, line 109: This doesnt work , if secret is added via hostpath, L138. May be we have to remove the option to be added by hostpath

yerzhant: Fixed in https://github.com/kubernetes/cloud-provider-openstack/pull/1600/commits/676bed90f314c5823a91ee7dd45ecf60c16851ff

lingxiankong:
openstack-cloud-controller-manager chart is using secret for cloud config, cinder-csi also supports to use secret as volume but the volume is not used in `volumeMounts` which should be fixed there.

brtknr:
OCCM chart needs to be normalised with Cinder CSI chart, not the other way round. The OCCM mounts from non standard locations like /etc/config and /etc/ssl/cert

yerzhant:
Thank for info.

PR #1605, opened 2021-07-21
[manila-csi-plugin] Add options for passing mount opts to cephfs-csi

gman0:

manila-csi already passes mount options to its partner Node Plugins. However, cephfs-csi consumes mount options for CephFS kernel and FUSE client via separate volume context parameters: `kernelMountOptions` and `fuseMountOptions`. In order for user to be able to control these, manila-csi needs to expose these params as well.

Please see https://github.com/ceph/ceph-csi/blob/devel/docs/deploy-cephfs.md#configuration for reference (section "Available volume parameters").

Exposing kernel opts are especially important for environments with SELinux enabled, where users may want to pass in SELinux context options.

**Release note**:

```release-note
[manila-csi-plugin] Add cephfs-kernelMountOptions and cephfs-fuseMountOptions volume parameters
```


tombarron (approved):
Looks right, thanks!

tombarron:
The docs file is outside pkg/csi/manila so my attempts probabably won't be effective, but they are there for the record.

gman0:
Thank you, @tombarron ! @ramineni could you please take a look?

gouthampacha (approved):
/lgtm

PR #1614, opened 2021-07-30
cinder-csi: Disable uuid checks on XFS

jsafrane:
Add 'nouuid' mount option to all XFS mounts to be able to mount a volume and its restored snapshot on the same node. Without the option, such a mount fails, because XFS detects that two different volumes with the same filesystem UUID are being mounted.

See https://github.com/container-storage-interface/spec/issues/482

**Release note**:

```release-note
[cinder-csi-plugin] Added support to mount a volume with XFS filesystem and its restored snapshot on the same node.
```

### DIFF
--- a/pkg/csi/cinder/nodeserver.go
+++ b/pkg/csi/cinder/nodeserver.go
@@ -213,7 +213,7 @@ func nodePublishEphemeral(req *csi.NodePublishVolumeRequest, ns *nodeServer) (*c
 				fsType = mnt.FsType
 			}
 			mountFlags := mnt.GetMountFlags()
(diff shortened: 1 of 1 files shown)


jichenjc:
I am curious why the e2e and other test not started... I knew we have trouble in gate but not sure why it's not testing at all?

jichenjc:
the change looks good

jsafrane:
@jichenjc @ramineni can you please /approve this?

jichenjc:
given the gate already have 1.22 released
I think we can merge this now 

/approve

PR #1616, opened 2021-08-02
[manila-csi-plugin] pass CreateVolumeRequest.parameters as Volume.volume_context

gman0:

Currently, `Volume.volume_context` is populated manually:

https://github.com/kubernetes/cloud-provider-openstack/blob/61bdbe5bf4b75453b67e77b543891e1a5bb3eea2/pkg/csi/manila/controllerserver.go#L167-L179

If we wanted to add more entries to the map, we'd need to do it explicitly which is tedious and error prone. https://github.com/kubernetes/cloud-provider-openstack/pull/1605 added two new parameters, and I forgot to add them into the map. As a result, they are missing in provisioned volumes, despite having them set in my storage class.

This PR makes it so that *all* recognized volume parameters(*) in `CreateVolumeRequest.parameters` map are automatically passed as `Volume.volume_context`. Doing so makes sure no similar bug should happen in the future.

(*) Recognized volume parameters: `validator.Validator` now offers `Fields []string` that's populated with all recognized field names of a struct. `CreateVolumeRequest.parameters` is then filtered against [`options.NodeVolumeContext`](https://github.com/kubernetes/cloud-provider-openstack/blob/61bdbe5bf4b75453b67e77b543891e1a5bb3eea2/pkg/csi/manila/options/shareoptions.go#L39-L49)'s fields, so that only recognized fields will be passed to the resulting `Volume.volume_context`.

No backwards-compatibility breakage is expected.

This is what the resulting PV `volumeAttributes` looks like with this patch:

```json
$ kubectl get pv pvc-19e0358e-12b2-469c-8d13-0d5d61c8668d -o json | jq .spec.csi.volumeAttributes
{
  "cephfs-kernelMountOptions": "context=system_u:system_r:container_t:s0",
  "cephfs-mounter": "kernel",
  "shareAccessID": "025c0070-89cd-4706-b542-90c7c4d8f1d7",
  "shareID": "5ae16a16-727a-4493-a4e3-1c634f9e67a0",
  "storage.kubernetes.io/csiProvisionerIdentity": "1627906097137-8081-cephfs.manila.csi.openstack.org"
}
```

for this storage class:
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: devstack-cephfs
provisioner: cephfs.manila.csi.openstack.org
parameters:
  type: cephfstype

  cephfs-mounter: kernel
  cephfs-kernelMountOptions: context=system_u:system_r:container_t:s0
  
  csi.storage.k8s.io/provisioner-secret-name: devstack-manila
  csi.storage.k8s.io/provisioner-secret-namespace: default
  csi.storage.k8s.io/node-stage-secret-name: devstack-manila
  csi.storage.k8s.io/node-stage-secret-namespace: default
  csi.storage.k8s.io/node-publish-secret-name: devstack-manila
  csi.storage.k8s.io/node-publish-secret-namespace: default
```

You can see that PV volume attributes contain only `cephfs-mounter`, `cephfs-kernelMountOptions`, `shareID` and `shareAccessID`. Other options from the storage class (`type`, as well as the implied `protocol` field) have been filtered away as they are not recognized fields in `NodeVolumeContext`.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/manila/controllerserver.go
+++ b/pkg/csi/manila/controllerserver.go
@@ -66,6 +66,18 @@ func getVolumeCreator(source *csi.VolumeContentSource, shareOpts *options.Contro
 	return nil, status.Error(codes.InvalidArgument, "invalid volume content source")
 }
 
+func filterParametersForVolumeContext(params map[string]string, recognizedFields []string) map[string]string {
+	volCtx := make(map[string]string)
+
+	for _, fieldName := range recognizedFields {
+		if val, ok := params[fieldName]; ok {
+			volCtx[fieldName] = val
+		}
+	}
+
+	return volCtx
+}
+
 func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
 	if err := validateCreateVolumeRequest(req); err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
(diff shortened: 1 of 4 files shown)


gman0:
PTAL @tombarron @gouthampacha

tombarron (approved):
Looks good, Robert.

ramineni:
> @ramineni this is a bugfix for #1605 . The relevant results of tests are in the description of this PR. Can this make it still to v1.22 release?

Is this urgent ? Else, we plan to do minor release after the CI has been fixed , where we can target all blocked PRs.

gman0:
Well without this fix, https://github.com/kubernetes/cloud-provider-openstack/pull/1605 doesn't work at all for dynamically provisioned volumes. Not being able to pass the cephfs kernel client mount opts then causes big performance problems in environments with selinux enabled due to relabeling. We have included this patch in our locally-built image for the moment, but it would be nice to have it upstreamed.

ramineni:
/lgtm 

PR #1638, opened 2021-09-02
[cinder-csi-plugin] add leader election

### LINKED ISSUE #702: cinder-csi-plugin does not have options for leader election
Of all the containers in the pod template of the  `csi-cinder-controllerplugin` StatefulSet, `docker.io/k8scloudprovider/cinder-csi-plugin` is the  only one that does not appear to have options for leader election.

Isn't this a requirement to run in HA with a multi-replica Deployment?

m-yosefpor:

Adding support for running HA CSI controller.

fixes #702

Increase replicas to 3, all operations should work as expected (attach, detach, provision, deprovision, resize, snapshot).

I tested it on k8s 1.20, all operations are completing without errors. Also I can see successful leader election change events:

```
29m         Normal    LeaderElection      lease/csi-cinderplugin                               csi-cinder-controllerplugin-0 became leader
3m30s       Normal    LeaderElection      lease/csi-cinderplugin                               csi-cinder-controllerplugin-0 became leader
29m         Normal    LeaderElection      lease/external-attacher-leader-csi-cinderplugin      csi-cinder-controllerplugin-0 became leader
3m30s       Normal    LeaderElection      lease/external-attacher-leader-csi-cinderplugin      csi-cinder-controllerplugin-0 became leader
29m         Normal    LeaderElection      lease/external-resizer-csi-cinderplugin              csi-cinder-controllerplugin-0 became leader
3m30s       Normal    LeaderElection      lease/external-resizer-csi-cinderplugin              csi-cinder-controllerplugin-0 became leader
29m         Normal    LeaderElection      lease/external-snapshotter-leader-csi-cinderplugin   csi-cinder-controllerplugin-0 became leader
3m30s       Normal    LeaderElection      lease/external-snapshotter-leader-csi-cinderplugin   csi-cinder-controllerplugin-0 became leader
```

Cordon all nodes, deletetd `csi-cinder-controllerplugin-0` (so it won't be scheduled again),  leader changed to `csi-cinder-controllerplugin-2` successfully and all CSI operations still working.

**Release note**:

```release-note
[cinder-csi-plugin] Add leader election flag (for HA CSI controller) in the manifests and switch from StatefulSet to Deployment. The existing cinder-csi-plugin helm chart should be uninstalled before installing the new version. Action required
```

### DIFF
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: latest
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 1.4.8
+version: 1.5.0
 home: https://github.com/kubernetes/cloud-provider-openstack
 icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: latest
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 1.5.0
+version: 2.0.0
 home: https://github.com/kubernetes/cloud-provider-openstack
 icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
--- a/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
+++ b/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
@@ -11,6 +11,7 @@ spec:
 {{- if eq .Values.csi.controllerPlugin.strategy.type "RollingUpdate" }}
     rollingUpdate:
       maxUnavailable: {{ .Values.csi.controllerPlugin.strategy.rollingUpdate.maxUnavailable }}
+      maxSurge: {{ .Values.csi.controllerPlugin.strategy.rollingUpdate.maxSurge }}
 {{- end }}
   selector:
     matchLabels:
--- a/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
+++ b/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
@@ -1,21 +1,17 @@
-apiVersion: v1
-kind: Service
-metadata:
-  name: {{ include "cinder-csi.name" . }}-controllerplugin
-  labels:
-    {{- include "cinder-csi.controllerplugin.labels" . | nindent 4 }}
-spec:
-  clusterIP: None
----
-kind: StatefulSet
+kind: Deployment
 apiVersion: apps/v1
 metadata:
   name: {{ include "cinder-csi.name" . }}-controllerplugin
   labels:
     {{- include "cinder-csi.controllerplugin.labels" . | nindent 4 }}
 spec:
-  serviceName: {{ include "cinder-csi.name" . }}-controllerplugin
   replicas: {{ .Values.csi.plugin.controllerPlugin.replicas }}
+  strategy:
+    type: {{ .Values.csi.controllerPlugin.strategy.type }}
+{{- if eq .Values.csi.controllerPlugin.strategy.type "RollingUpdate" }}
+    rollingUpdate:
+      maxUnavailable: {{ .Values.csi.controllerPlugin.strategy.rollingUpdate.maxUnavailable }}
+{{- end }}
   selector:
     matchLabels:
       {{- include "cinder-csi.controllerplugin.matchLabels" . | nindent 6 }}
--- a/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
+++ b/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
@@ -7,11 +7,11 @@ metadata:
 spec:
   replicas: {{ .Values.csi.plugin.controllerPlugin.replicas }}
   strategy:
-    type: {{ .Values.csi.controllerPlugin.strategy.type }}
-{{- if eq .Values.csi.controllerPlugin.strategy.type "RollingUpdate" }}
+    type: {{ .Values.csi.plugin.controllerPlugin.strategy.type }}
+{{- if eq .Values.csi.plugin.controllerPlugin.strategy.type "RollingUpdate" }}
     rollingUpdate:
-      maxUnavailable: {{ .Values.csi.controllerPlugin.strategy.rollingUpdate.maxUnavailable }}
-      maxSurge: {{ .Values.csi.controllerPlugin.strategy.rollingUpdate.maxSurge }}
+      maxUnavailable: {{ .Values.csi.plugin.controllerPlugin.strategy.rollingUpdate.maxUnavailable }}
+      maxSurge: {{ .Values.csi.plugin.controllerPlugin.strategy.rollingUpdate.maxSurge }}
 {{- end }}
   selector:
     matchLabels:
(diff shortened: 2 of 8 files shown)


m-yosefpor:
@ramineni do we need to also have e2e tests for an HA setup?

bgagnon:
I believe the use of `StatefulSet` with `spec.replicas: 1` was to ensure a single replica at all times. With leader election, it can become a `Deployment` instead.

m-yosefpor:
> I believe the use of `StatefulSet` with `spec.replicas: 1` was to ensure a single replica at all times. With leader election, it can become a `Deployment` instead.

Agree. now with leader election, even with `replicas: 1` users should be able to have a rolling update strategy with no down time rollouts. Also I switched to HA 3 replicas by default, which is also a sane default value used in other csi helm charts by default (such as ceph-csi)

charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
```diff
@@ -1,21 +1,17 @@
-apiVersion: v1
-kind: Service
-metadata:
-  name: {{ include "cinder-csi.name" . }}-controllerplugin
-  labels:
-    {{- include "cinder-csi.controllerplugin.labels" . | nindent 4 }}
-spec:
-  clusterIP: None
----
```
jichenjc, line 9: um... what's the reason for removing this service?

charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
```diff
@@ -1,21 +1,17 @@
...
-kind: Service
-metadata:
-  name: {{ include "cinder-csi.name" . }}-controllerplugin
-  labels:
-    {{- include "cinder-csi.controllerplugin.labels" . | nindent 4 }}
-spec:
-  clusterIP: None
----    
-kind: StatefulSet
+kind: Deployment
```
jichenjc, line 1: change this will make manifests inconsistent to helm chart .. in case we really need fit for HA 
then a open question is whether we need update manifests as well

charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
m-yosefpor: This headless service is not used at all, however it's a best practice to have it for statefulSets and some linters validate that each sts must have a headless service. However if we want to change it to deployment, it would be unnecessary and unused.

charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
m-yosefpor: No it's not a hard requirement and we can keep it as is for better backward compatibility, however the only downside is if a user runs a single replica, he can't benefit from rolling deployment strategy, to have another controller pod ready before removing the old one. The sts-controller first terminate the csi controller pod and then recreates it which causes a short period of service unavailability for single replica ones. For `replicas: 3`, it's not much different. In general it's a better practice to have such leader elected controllers as deployments instead of statefulset. [openshift/openstack-cinder-csi-driver-operator](https://github.com/openshift/openstack-cinder-csi-driver-operator) is using a deployment for cinder-csi, also other csi drivers (such as ceph-csi, etc) provide deployment instead of a statefulSet.

I can update the manifests as well, or if you think it might cause some issues for users during the upgrade, I can revert this to StatefulSet.

charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
jichenjc: ok, understood, thanks for the info

charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
jichenjc: thanks for the detailed info this is really helpful , I think unless we have special reason
we probably can go deployment as others did so and consider HA reason

I think we don't need to be 100% same for manifest / chart but for this kind of difference (Deployment/Statefulset) , we need some strong reason if we want to make them different
so I guess we can update to deployment and put some upgrade notes in case needed?

@ramineni @lingxiankong any comments?

m-yosefpor: @jichenjc thanks. I updated the manifests as well.

charts/cinder-csi-plugin/values.yaml
```diff
@@ -67,6 +67,18 @@ csi:
         - operator: Exists
```
ramineni, line 67: needs version bump if any changes to chart.https://github.com/kubernetes/cloud-provider-openstack/blob/master/charts/cinder-csi-plugin/Chart.yaml

charts/cinder-csi-plugin/values.yaml
```diff
@@ -67,6 +67,18 @@ csi:
...
       kubeletDir: /var/lib/kubelet
     controllerPlugin:
+      replicas: 3
```
ramineni, line 70: Lets keep default value to 1

charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
```diff
@@ -45,6 +43,7 @@ spec:
           args:
             - "--csi-address=$(ADDRESS)"
             - "--timeout={{ .Values.timeout }}"
+            - "--leader-election=true"
```
ramineni, line 46: I see that enabling leader election might require additional rbac , are they inplace already?

ramineni:
Its not the update of only charts now, so please remove the same from title and release notes

charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
m-yosefpor: Yes actually there is a `external-resizer-cfg` Role with correspoding rbac, however as it was not only related to resizer. As all the roles refrence one `sa`, even that was enough. However for better consistency, I refactored it and added the role to each of the 4 components.

charts/cinder-csi-plugin/values.yaml
m-yosefpor: done

charts/cinder-csi-plugin/values.yaml
m-yosefpor: done

m-yosefpor:
@ramineni all your suggestions applied.

charts/cinder-csi-plugin/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: latest
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 1.4.8
+version: 1.5.0
```
lingxiankong, line 5: As this change is not backward compatible, I think we should bump the major version, i.e. 2.0.0. If the users have already installed the previous cinder-csi-plugin chart, they should first `helm uninstall` before installing 2.0.0, right?

lingxiankong (changes requested):

charts/cinder-csi-plugin/Chart.yaml
m-yosefpor: @lingxiankong done

lingxiankong:
/lgtm

@m-yosefpor I edited the release note, please double check.

m-yosefpor:
> /lgtm
> 
> @m-yosefpor I edited the release note, please double check.

SGTM. Thanks.

ramineni:
@m-yosefpor Thanks. Looks good.

/lgtm

m-yosefpor:
@lingxiankong any ETA on when the next release for openstack-cinder-csi helm chart will be?

lingxiankong:
> @lingxiankong any ETA on when the next release for openstack-cinder-csi helm chart will be?

Hi @m-yosefpor, the current chart release policy is we only do the official release together with the project, so for this specific one, we have to wait until v1.23 release.

I know that may sound weird, I will talk with the team and see what we could do.

PR #1642, opened 2021-09-07
[OCCM][CI] make k3s install flag (external cloud ) optional so that other CI won't use it

jichenjc:

make cloud provider flag external flag optional ,so OCCM use it as external but others like CSI 
use in tree one

fixes #

**Release note**:

```release-note
NONE
```


lingxiankong:
> split the install of OCCM and test of OCCM into 2 roles
thus others such as cinder is able to reuse the OCCM install (otherwise the pods will be pending)

@jichenjc May I know why Cinder test should install OCCM?

jichenjc:
>>>@jichenjc May I know why Cinder test should install OCCM?

@lingxiankong thanks
I am using your k3s workflow (and I believe all other test will follow?) if cloud provider is external , should we create OCCM to make it's eligible to schedule pods? in addition, if some code changes some common part, should we also consider run OCCM as well to make sure not side effect?

lingxiankong:
> I am using your k3s workflow (and I believe all other test will follow?) if cloud provider is external , should we create OCCM to make it's eligible to schedule pods? in addition, if some code changes some common part, should we also consider run OCCM as well to make sure not side effect?

Is it possible to change k3s ansible role to not rely on the cloud controller manager?

jichenjc:
I can take a quick look, but I Think build OCCM still have value..

lingxiankong:
> I can take a quick look, but I Think build OCCM still have value..

I'd prefer to not add another unnecessary component as a hard dependency if we could.

jichenjc:
ok, let me see what I can do

jichenjc:
@lingxiankong updated~

tests/playbooks/roles/install-k3s/defaults/main.yaml
```diff
@@ -8,4 +8,6 @@ sg_name: "k3s_sg"
 keypair_name: "k3s_keypair"
 image_url: "https://cloud-images.ubuntu.com/releases/focal/release/ubuntu-20.04-server-cloudimg-amd64.img"
 image_name: "ubuntu-focal"
-master_port_name: "k3s_master"
\ No newline at end of file
+master_port_name: "k3s_master"
+master_external_flag: "--disable-cloud-controller --kubelet-arg='cloud-provider=external'"
```
lingxiankong, line 12: I suggest using `k3s_master_cloud_provider_config` and `k3s_worker_cloud_provider_config`

lingxiankong (changes requested):

tests/playbooks/roles/install-k3s/defaults/main.yaml
jichenjc: updated

tests/playbooks/roles/install-k3s/defaults/main.yaml
```diff
@@ -8,4 +8,6 @@ sg_name: "k3s_sg"
 keypair_name: "k3s_keypair"
 image_url: "https://cloud-images.ubuntu.com/releases/focal/release/ubuntu-20.04-server-cloudimg-amd64.img"
 image_name: "ubuntu-focal"
-master_port_name: "k3s_master"
\ No newline at end of file
+master_port_name: "k3s_master"
+k3s_master_cloud_provider_config: "--disable-cloud-controller --kubelet-arg='cloud-provider=external'"
```
ramineni, line 12: > make k3s install flag (external cloud ) optional

@jichenjc This means we use intree cloud provider?  intree cloud providers will be removed in next release or so , so the job fails when the provider is removed from the repo right? Or I misundersood the intent? Please clarify

jichenjc: @ramineni please see above discussion with @lingxiankong 
I started up with OCCM actually..

ramineni: @jichenjc Ya, I checked , Thats what I would like to confirm, not to rely on CCM means intree? Jobs would fail in that case if removed?
@lingxiankong ^^

jichenjc: it takes a few more steps than current way
I guess another method is to keep this running until in-tree really removed
should be not that hard :)

ramineni: @jichenjc I dont think thats good idea to use deprecated ones in jobs

jichenjc: but actually we run it for long time, don't we? 
at least it makes the first run simple (no OCCM involvement from my test so far ..)

ramineni: The prerequisite for cinder csi driver is we need Kubernetes Cluster on Openstack, the only supported/recommended way is using external provider right. We dont need to build the OCCM, but we can directly pull from docker repo the latest image. wdyt?
@lingxiankong wdyt?

jichenjc: that's another option but I don't see too much difference 
after all, we have docker registry, code repo down etc ,all set for CSI already
build OCCM or pull is not a big deal

lingxiankong: @ramineni @jichenjc Very good discussion,  my original thought is to remove the dependency of occm (especially the image build) from cinder-csi related tests, I'm not sure if the cloud controller is now a required component for a k8s cluster?

jichenjc: I knew it's gonna remove the in-tree provider soon (1.23 or 1.24? ) 
I can go with a w/ OCCM but no build image (just pull latest) approach if that's an acceptable way

ramineni: @jichenjc sounds good to me

ramineni:
@jichenjc Is this still required ? Now we plan to use external for all jobs right

jichenjc:
this doesn't need anymore

PR #1648, opened 2021-09-15
[occm] Support re-use load balancer between Services

### LINKED ISSUE #933: [occm] Allow ip address sharing for services of type loadbalancers

lingxiankong:
By default, different Services of LoadBalancer type should have
different corresponding cloud load balancers, however,
openstack-cloud-controller-manager allows multiple Services to share a
single load balancer if the Octavia service supports the tag feature
(since version 2.5).

Added a new annotation `loadbalancer.openstack.org/load-balancer-id`
that can be specified for creating Services.

Added a new config option `[LoadBalancer] max-shared-lb` to limit the
number of Services that share a load balancer.

Check the doc for more details.

fixes #933

You can use the image `docker.io/lingxiankong/openstack-cloud-controller-manager-amd64:test` for testing.

**Release note**:

```release-note
[openstack-cloud-controller-manager] Support to share a cloud load balancer between different Services. The load balancer can be either created by other Services or outside the Kubernetes cluster. As the load balancer doesn't relate to a dedicated Service anymore, any scripts relying on that relationship previously need to change to use the load balancer tags instead. Action required.
```


ramineni:
@lingxiankong This job seems to fail . Could you check
pull-cloud-provider-openstack-check — Job failed.

lingxiankong:
Oh, yes, I paid more attention to `openstack-cloud-controller-manager-e2e-test`, will fix.

lingxiankong:
I've removed the usage of import-boss, I didn't find there is any other repo using it except k/k, so I'd suggest to remove it to avoid too much overhead.

ramineni:
> I've removed the usage of import-boss, I didn't find there is any other repo using it except k/k, so I'd suggest to remove it to avoid too much overhead.

@lingxiankong import-boss works if we have .import-restrictions in place , it prohibits to import some of the k/k packages which are not meant to be used by external repo. Is it failed because of added dependency or any other failure?

lingxiankong:
> @lingxiankong import-boss works if we have .import-restrictions in place , it prohibits to import some of the k/k packages which are not meant to be used by external repo. Is it failed because of added dependency or any other failure?

@ramineni Yeah, it failed once before, there was a new module needed to add to the `.import-restrictions` file, which made me feel unconfortable to maintain such a file that is only used by kubernetes code repo. I was suggesting removing that first because we don't have a policy describing what packages should be allowed and what shouldn't, the second reason is, given there are no other cloud providers using that, why we should spend extra effort maintaining that?

ramineni:
> > @lingxiankong import-boss works if we have .import-restrictions in place , it prohibits to import some of the k/k packages which are not meant to be used by external repo. Is it failed because of added dependency or any other failure?
> 
> @ramineni Yeah, it failed once before, there was a new module needed to add to the `.import-restrictions` file, which made me feel unconfortable to maintain such a file that is only used by kubernetes code repo. I was suggesting removing that first because we don't have a policy describing what packages should be allowed and what shouldn't, the second reason is, given there are no other cloud providers using that, why we should spend extra effort maintaining that?

@lingxiankong Ya, agree its a overhead , ok to be removed I suppose. But we should keep a lookout in general not to use k/k repo packages if alternate external packages/libs available, else its fine.

pkg/openstack/loadbalancer.go
```diff
@@ -3150,45 +3125,64 @@ func (lbaas *LbaasV2) ensureLoadBalancerDeleted(ctx context.Context, clusterName
...
+	// For neutron-lbaas
+	if !lbaas.opts.UseOctavia {
+		return lbaas.ensureLoadBalancerDeletedLegacy(loadbalancer)
+	}
 
-		err = waitLoadbalancerDeleted(lbaas.lb, loadbalancer.ID)
-		if err != nil {
-			return fmt.Errorf("failed to wait for deleted loadbalancer %s: %v", loadbalancer.ID, err)
+	if needDeleteLB && lbaas.opts.CascadeDelete {
+		klog.InfoS("Deleting load balancer", "lbID", loadbalancer.ID, "service", klog.KObj(service))
```
ramineni, line 3153: @lingxiankong everywhere Im seeing, you kept 2 logs L 3153, L 3157  logged by default (not debug logs). Is it required ? We can add Deleted confirmation logs to default and other as debug logs?

pkg/util/openstack/loadbalancer.go
```diff
@@ -43,9 +44,10 @@ const (
 	OctaviaFeatureTimeout           = 3
 	OctaviaFeatureAvailabilityZones = 4
 
-	loadbalancerActiveInitDelay = 1 * time.Second
-	loadbalancerActiveFactor    = 1.2
-	loadbalancerActiveSteps     = 19
+	waitLoadbalancerInitDelay   = 5 * time.Second
+	waitLoadbalancerFactor      = 1
+	waitLoadbalancerActiveSteps = 36
```
ramineni, line 49: this has been increases significantly. any reason?

pkg/util/openstack/loadbalancer.go
```diff
@@ -168,9 +172,39 @@ func waitLoadbalancerActive(client *gophercloud.ServiceClient, loadbalancerID st
 
 	})
 
+	if err == wait.ErrWaitTimeout {
+		err = fmt.Errorf("timeout waiting for the loadbalancer %s", activeStatus)
```
ramineni, line 176: good to loadbalancer ID here?

pkg/util/openstack/loadbalancer.go
```diff
@@ -196,16 +230,72 @@ func GetLoadbalancerByName(client *gophercloud.ServiceClient, name string) (*loa
...
+	}
+
+	mc := metrics.NewMetricContext("loadbalancer", "update")
+	_, err := loadbalancers.Update(client, lbID, updateOpts).Extract()
+	if mc.ObserveRequest(err) != nil {
+		return err
+	}
+
+	if err := WaitLoadbalancerActive(client, lbID); err != nil {
+		return fmt.Errorf("failed to wait for load balancer ACTIVE after updating: %v", err)
```
ramineni, line 246: we can add ID here as well?

pkg/util/openstack/loadbalancer.go
```diff
@@ -231,13 +321,32 @@ func CreateListener(client *gophercloud.ServiceClient, lbID string, opts listene
...
+		if cpoerrors.IsNotFound(err) {
+			klog.V(2).Infof("Listener %s for load balancer %s was already deleted: %v", listenerID, lbID, err)
+		} else {
+			_ = mc.ObserveRequest(err)
+			return fmt.Errorf("error deleting listener %s for load balancer %s: %v", listenerID, lbID, err)
+		}
+	}
+
+	if err := WaitLoadbalancerActive(client, lbID); err != nil {
+		return fmt.Errorf("failed to wait for load balancer ACTIVE after deleting listener: %v", err)
```
ramineni, line 344: can add lbID here

ramineni:
@lingxiankong Can we add functional test as well for this functionality? or is this already tested as part of CI?

pkg/openstack/loadbalancer.go
lingxiankong: Yeah, I did that on purpose. First, the default log level for klog is V(2), we don't need to explicitly use `klog.V(2)`. Second, log the starting and ending point of operations is considered the best practice (at least in openstack) in production and is useful for troubleshooting purpose, e.g. an operation is stuck, without the starting point logging, it's hard to know what the system is doing at that moment.

pkg/util/openstack/loadbalancer.go
lingxiankong: The comment for the previous code is incorrect, we were not using exponential backoff actually, we were using the fixed interval backoff mechanism, so `loadbalancerActiveFactor    = 1.75` doesn't make any sense. I changed this to: Start with 5s delay, sleep 5s (5*1) between each iteration, try 36 times in max, so the total duration is around 5*36=180s, that should be covering most prod and testing environments.

pkg/util/openstack/loadbalancer.go
lingxiankong: Agree, will add.

pkg/util/openstack/loadbalancer.go
lingxiankong: Sure

lingxiankong:
> @lingxiankong Can we add functional test as well for this functionality? or is this already tested as part of CI?

Yes, please see the change in tests/e2e/cloudprovider/test-lb-service.sh

pkg/util/openstack/loadbalancer.go
ramineni: > we were not using exponential backoff actually, we were using the fixed interval backoff mechanism,

@lingxiankong Didnt get that, I see we are using "wait.ExponentialBackoff" at L158 , https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#ExponentialBackoff , as per my understanding it says it is using longer delays between each retry . Isnt it? Are you suggesting the function is not doing as expected?

ramineni:
> > @lingxiankong Can we add functional test as well for this functionality? or is this already tested as part of CI?
> 
> Yes, please see the change in tests/e2e/cloudprovider/test-lb-service.sh

Got it . Thanks. I might have missed it as diff is minimized

pkg/util/openstack/loadbalancer.go
lingxiankong: If you take a look at the implementation of `ExponentialBackoff`, especially `time.Sleep(backoff.Step())`, see how the sleep time is calculated.

ramineni:
code looks fine to me.Thanks
/lgtm

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
```diff
@@ -450,3 +459,110 @@ To enable PROXY protocol support, the openstack-cloud-controller-manager config
...
+Now, create another Service `service-2` but re-use the load balancer created for `service-1` by specifying the annotation `loadbalancer.openstack.org/load-balancer-id`:
+
+```yaml
+kind: Service
+apiVersion: v1
+metadata:
+  name: service-2
+  namespace: default
+  annotations:
+    loadbalancer.openstack.org/load-balancer-id: "2b224530-9414-4302-8163-5abebdcdc84f"
```
jichenjc, line 519: just curious, if the ID is given but the LB not found or in some race condition like
is deleting, what will happen?

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
```diff
@@ -193,6 +194,14 @@ Request Body:
 
   Not supported when `lb-provider=ovn` is configured in openstack-cloud-controller-manager.
 
+- `loadbalancer.openstack.org/load-balancer-id`
+
+  This annotation is automatically added to the Service if it's not specified when creating. After the Service is created successfully it shouldn't be changed, otherwise the Service won't behave as expected.  
+
+  If this annotation is specified with a valid cloud load balancer ID when creating Service, the Service is reusing this load balancer rather than creating another one. Again, it shouldn't be changed after the Service is created.
+
+  If this annotation is specified, the other annotations which define the load balancer features will be ignored.
```
jichenjc, line 203: I think we should error out on such case, not ignore it ..

jichenjc:
mostly good enough

just some suggestions/questions inline

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
lingxiankong: If the load balancer is not found or in deleting status, service creation will fail.

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
lingxiankong: We can't, with this PR, all the services (either existing or new) will have this annotation created automatically, in controller code it's hard to tell if the request is creating new service or updating an existing one.

lingxiankong:
The job is failing because the devstack has some issues being fixed in the openstack community right now.

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
jichenjc: sorry then I am confused.. if we will create this annottation anyway
and you are saying `If this annotation is specified, the other annotations which define the load balancer features will be ignored` , does this mean we will ignore the other annotation in any case? 

anyway, this is not a blocker so we can wait for CI ready then merge and help me in understand the logic, thanks

lingxiankong: As demonstrated in the doc, if you are creating a new service without this annotation, the workflow is the same as previous, but in the end occm adds this annotation to the service. So next time when you updating this service, occm can directly look for the load balancer by ID rather than searching by name.

However, if you create a service with a valid lb id specified, that means you are just creating listener/pool/members in an existing load balancer, which means you can't config something like fixed/floating IP for the lb, the flavor and az, etc.

jichenjc: ok, I understand, maybe my thought is overkill ,so , let it be, thanks for the detailed info~

lingxiankong:
The job passed.

lingxiankong:
/override openlab/cloud-provider-openstack-acceptance-test-e2e-conformance-stable-branch-v1.19
/override openlab/cloud-provider-openstack-acceptance-test-e2e-conformance-stable-branch-v1.20
/override openlab/cloud-provider-openstack-acceptance-test-e2e-conformance-stable-branch-v1.21
/override openlab/cloud-provider-openstack-acceptance-test-keystone-authentication-authorization

wouterh-dev:
Is there a reason that max-shared-lb was set to only 2 by default?

PR #1656, opened 2021-10-04
[manila-csi-plugin] Add e2e tests

gman0:

This PR adds e2e test for manila-csi-plugin.

fixes #

I've had some problems running tests that dealt with snapshots. I suggest we take what works right now at least, and then enable more tests once we figure out how to fix the rest.

**Release note**:

```release-note
NONE
```


tests/ci-csi-manila-e2e.sh
```diff
@@ -0,0 +1,120 @@
+#!/bin/bash
```
gman0, line 1: This script is copy-paste from csi-cinder's e2e job. I'd prefer to factor this out into a separate file and keep the scripts for the individual e2e entry points minimal. Maybe in a separate PR?

gman0:
PTAL @jichenjc @ramineni

tests/ci-csi-manila-e2e.sh
ramineni: +1

gman0:
```
FailedScheduling  4m43s  default-scheduler  0/1 nodes are available: 1 node(s) had taint {node.cloudprovider.kubernetes.io/uninitialized: true}, that the pod didn't tolerate.",
```

ramineni:
@gman0 only 2 tests have run in this right?

ramineni:
Just triggered cinder tests as well , as this PR have changes.

PR #1665, opened 2021-10-11
[CI] Add install-k8s role

gman0:

This PR adds `install-k8s` role for the CI Ansible playbooks.

fixes #

`install-k3s` brings in many OpenStack dependencies. manila-csi e2e test (https://github.com/kubernetes/cloud-provider-openstack/pull/1656) doesn't install these, so it fails to run.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/playbooks/roles/install-k8s/defaults/main.yaml
+++ b/tests/playbooks/roles/install-k8s/defaults/main.yaml
@@ -1,0 +1,7 @@
+---
+k8s_version: 'master'
+etcd_version: 'v3.5.0'
+
+k8s_src_dir: '{{ ansible_user_dir }}/src/k8s.io/kubernetes'
+k8s_log_dir: '{{ ansible_user_dir }}/workspace/logs/kubernetes'
+kubectl: '{{ ansible_user_dir }}/src/k8s.io/kubernetes/cluster/kubectl.sh'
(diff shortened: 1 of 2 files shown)


jichenjc:
just curious , I think k3s is a k8s compatible suite that we used for our test
if k3s role has dependency, why not split it and we introduce a k8s role so we will have
k8s and k3s both in our test?

gman0:
@lingxiankong thanks for the input! If it's ok to run with stable k8s releases (I'm not sure why this wasn't the case already in OpenLab), there are easier ways to get the cluster going. Would kubeadm be acceptable? A role for this already exists in https://github.com/theopenlab/openlab-zuul-jobs/tree/master/roles/create-single-k8s-cluster-with-kubeadm . As far as manila-csi-plugin goes, this is completely sufficient as it doesn't need any of the cloud provider stuff.

ramineni:
@gman0 I dont quite understand this PR.

Are you planning to use local-up-cluster.sh for bring cluster up or any other? I don't see a reason to build the k8s binaries every time the job is run.

If you want to use local-up-cluster.sh , that's not recommended way forward actually . That script changes a lot and not reliable. So we planned to move everything to kubeadm , now k3s is lightweight option, so using that.

gman0:
Closing. https://github.com/kubernetes/cloud-provider-openstack/pull/1669 got merged instead.

PR #1667, opened 2021-10-12
[manila-csi-plugin] ControllerExpandVolume: fix bad key for deleting from pendingVolumes

### LINKED ISSUE #1666: [manila-csi-plugin] Cannot extend the same volume twice

/kind bug

**What happened**:

When I extend the volume a second time I get this error message: 

```
volume pvc-8f8e61c9-18b5-4706-9161-8346ddcc55bc is already being processed
```

`Resize volume succeeded`

**How to reproduce it**:

1. Create a persistentVolumeClaim: 

```bash
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: client-pvc
  namespace: test-csi-nfs
spec:
  storageClassName: csi-manila-nfs
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: 1Gi
EOF
```

2. Extend the volume for the first time by switching it to 2Gi, the expansion will work correctly 

```bash
kubectl -n test-csi-nfs patch pvc client-pvc -p '{"spec":{"resources":{"requests":{"storage":"2Gi"}}}}'
```

3. Extend the volume again by switching it to 3Gi: 

```bash
kubectl -n test-csi-nfs patch pvc client-pvc -p '{"spec":{"resources":{"requests":{"storage":"3Gi"}}}}'
```

There is an error : 

```bash
kubectl -n test-csi-nfs describe pvc client-pvc
...
Events
...
Warning  VolumeResizeFailed      0s (x2 over 2s)   external-resizer nfs.manila.csi.openstack.org    resize volume "pvc-c7a34569-7a3e-4b49-acf3-520f7e0d8651" by resizer "nfs.manila.csi.openstack.org" failed: rpc error: code = Aborted desc = volume pvc-c7a34569-7a3e-4b49-acf3-520f7e0d8651 is already being processed
```

At line 451 of the controllerserver.go file, I think we should use the value of `req.GetName()` as the key to remove the volume from the pendingVolumes list instead of the value of `share.Name` 

So when we extended the volume the first time, this volume always stayed in the pendingVolumes list while the volume is ready 

https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/manila/controllerserver.go#L448-L451

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: 1.22.0
- OpenStack version: train

m-k8s:

Correctly removes the volume from pendingVolumes when the expansion is complete.

fixes #1666 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/manila/controllerserver.go
+++ b/pkg/csi/manila/controllerserver.go
@@ -445,10 +445,10 @@ func (cs *controllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 	}
 
 	// Check for pending operations on this volume
-	if _, isPending := pendingVolumes.LoadOrStore(req.GetVolumeId(), true); isPending {
+	if _, isPending := pendingVolumes.LoadOrStore(share.Name, true); isPending {
 		return nil, status.Errorf(codes.Aborted, "volume %s is already being processed", share.Name)
 	}
-	defer pendingVolumes.Delete(req.GetVolumeId())
+	defer pendingVolumes.Delete(share.Name)
 
 	// Try to expand the share
 


gman0:
@alexisries thank you for your contribution! The CI job is failing because it's WIP. I believe we can merge this since this is just reverting the change we've done earlier, as @gouthampacha pointed out.

/lgtm
/approve

PR #1678, opened 2021-10-29
[occm] Make volume mounts configurable in Helm Chart

jacksgt:
Depending on which controllers of the OCCM are being used, not all
volume mounts specified in the Helm chart are necessary.
Furthermore, some systems use different paths for these components [1] or
they don't have them at all [2], which will result in failure to attach
the volume to the pod and subsequently run the pod.

This patch makes the volume mounts configurable for the Helm chart
user. By default, the same volumes as before (flexvolume-dir,
ca-certs, k8s-certs) are mounted into the pods.
This makes this change transparent to current users of the Helm
chart (no breaking change).

This patch also updates the deprecated --address parameter to use
--bind-address instead.

**Which issue this PR fixes(if applicable)**: N/A

I performed manual testing with these changes, i.e. using `helm template occm charts/openstack-cloud-controller-manager/` to verify that 1) the output is as expected and 2) the generated YAML manifests are the same as before (no breaking changes).

**Release note**:
```release-note
[openstack-cloud-controller-manager] Make volume mounts of openstack-cloud-controller-manager Helm Chart configurable
```

[1] Depending on the Kubernetes distribution (e.g. OCP, k3s etc.)
[2] For example, our Fedora CoreOS nodes simply don't have /usr/libexec/kubernetes/plugins/volume/exec and since the entire root filesystems of these nodes is read-only, the directory also cannot be created. Consequently, the entire mount fails.

### DIFF
--- a/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
@@ -38,9 +38,9 @@ spec:
             - --cloud-provider=openstack
             - --use-service-account-credentials=true
             {{- if .Values.serviceMonitor.enabled }}
-            - --address=0.0.0.0
+            - --bind-address=0.0.0.0
             {{- else }}
-            - --address=127.0.0.1
+            - --bind-address=127.0.0.1
             {{- end }}
             {{- if .Values.controllerExtraArgs }}
             {{- with .Values.controllerExtraArgs }}
@@ -55,17 +55,12 @@ spec:
             protocol: TCP
           {{- end }}
           volumeMounts:
-            - mountPath: /etc/kubernetes/pki
-              name: k8s-certs
-              readOnly: true
-            - mountPath: /etc/ssl/certs
-              name: ca-certs
-              readOnly: true
             - mountPath: /etc/config
               name: cloud-config-volume
               readOnly: true
-            - mountPath: /usr/libexec/kubernetes/kubelet-plugins/volume/exec
-              name: flexvolume-dir
+          {{- if .Values.extraVolumeMounts }}
+            {{- toYaml .Values.extraVolumeMounts | nindent 12 }}
+          {{- end }}
           {{- if .Values.livenessProbe }}
           livenessProbe:
             {{- toYaml .Values.livenessProbe | nindent 12 }}
@@ -83,18 +78,9 @@ spec:
               value: /etc/config/cloud.conf
       hostNetwork: true
       volumes:
-      - hostPath:
-          path: /usr/libexec/kubernetes/kubelet-plugins/volume/exec
-          type: DirectoryOrCreate
-        name: flexvolume-dir
-      - hostPath:
-          path: /etc/kubernetes/pki
-          type: DirectoryOrCreate
-        name: k8s-certs
-      - hostPath:
-          path: /etc/ssl/certs
-          type: DirectoryOrCreate
-        name: ca-certs
       - name: cloud-config-volume
         secret:
-          secretName: {{ .Values.secret.name | default "cloud-config" }}
+          secretName: {{ .Values.secret.name }}
+      {{- if .Values.extraVolumes }}
+        {{ toYaml .Values.extraVolumes | nindent 6 }}
+      {{- end }}
--- a/charts/openstack-cloud-controller-manager/values.yaml
+++ b/charts/openstack-cloud-controller-manager/values.yaml
@@ -52,9 +52,11 @@ serviceMonitor: {}
 #   enabled: true
 
 # Create a secret resource cloud-config (or other name) to store credentials and settings from cloudConfig
+# You can also provide your own secret (not created by the Helm chart), in this case set create to false
+# and adjust the name of the secret as necessary
 secret:
   create: true
-#  name:
+  name: cloud-config
 
 # Specify settings with the same key as the CCM config: https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#config-openstack-cloud-controller-manager
 cloudConfig:
@@ -63,3 +65,28 @@ cloudConfig:
   loadBalancer:
   blockStorage:
   metadata:
+
+# The following three volumes are required to use all OCCM controllers,
+# but might not be needed if you just use a specific controller
+# Additional volumes that should be available to the pods:
+extraVolumes:
+  - name: flexvolume-dir
+    hostPath:
+      path: /usr/libexec/kubernetes/kubelet-plugins/volume/exec
+  - name: ca-certs
+    hostPath:
+      path: /etc/ssl/certs
+  - name: k8s-certs
+    hostPath:
+      path: /etc/kubernetes/pki
+# Where the additional volumes should be mounted into the pods:
+extraVolumeMounts:
+  - name: flexvolume-dir
+    mountPath: /usr/libexec/kubernetes/kubelet-plugins/volume/exec
+    readOnly: true
+  - name: ca-certs
+    mountPath: /etc/ssl/certs
+    readOnly: true
+  - name: k8s-certs
+    mountPath: /etc/kubernetes/pki
+    readOnly: true


charts/openstack-cloud-controller-manager/Chart.yaml
```diff
@@ -4,7 +4,7 @@ description: Openstack Cloud Controller Manager Helm Chart
 icon: https://object-storage-ca-ymq-1.vexxhost.net/swift/v1/6e4619c416ff4bd19e1c087f27a43eea/www-images-prod/openstack-logo/OpenStack-Logo-Vertical.png
 home: https://github.com/kubernetes/cloud-provider-openstack
 name: openstack-cloud-controller-manager
-version: 1.1.1
+version: 1.2.0
```
jichenjc, line 7: why bump to 1.2? maybe need 1.1.2?

jichenjc:
just one minor question, others seems fine

charts/openstack-cloud-controller-manager/Chart.yaml
jacksgt: Hi, thanks for getting back to me so quickly.

I was following [Semantic versioning](https://semver.org/), which [should be used for Helm charts](https://helm.sh/docs/chart_best_practices/conventions/#version-numbers):

> Given a version number MAJOR.MINOR.PATCH, increment the:
>   MAJOR version when you make incompatible API changes,
>   MINOR version when you add functionality in a backwards compatible manner, and
>   PATCH version when you make backwards compatible bug fixes.

jacksgt:
@jichenjc Hi, the CI test failed, but this does not seem to be cause by this PR, rather a test script seems to be missing:
```
 + ./tests/ci-keystone-e2e.sh
/usr/local/bin/runner.sh: line 114: ./tests/ci-keystone-e2e.sh: No such file or directory 
```

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/cloud-provider-openstack/1678/openstack-cloud-keystone-authentication-authorization-test/1456100003477983232#1:build-log.txt%3A17

jichenjc:
right, the test scripts is supposed to be added later on 
recently a test-infra PR merged for e2e keystone, for now you we can ignore that failure
and overwrite it later on

I will submit a PR to make test pass first (maybe a fake stub first)
thanks for point that out

charts/openstack-cloud-controller-manager/Chart.yaml
lingxiankong: Hi both, appreciate the discussion here. However, it has been decided (within the team) to only bump the chart version when we do release, so please still keep `1.1.1`.

@jichenjc remember? Please see the discussion in Slack. I just need to add some doc for contributors.

Otherwise, this PR looks good to me.

lingxiankong (changes requested):

jacksgt:
Hi @lingxiankong , as requested I have reset the version number to 1.1.1

lingxiankong:
> Hi @lingxiankong , as requested I have reset the version number to 1.1.1

Thanks. Please ignore the chart lint job, we are fixing the failure.

lingxiankong:
We have a fix the helm chart job: https://github.com/kubernetes/cloud-provider-openstack/pull/1691, please rebase the PR once that is merged.

jacksgt:
> We have a fix the helm chart job: #1691, please rebase the PR once that is merged.

Done.

charts/openstack-cloud-controller-manager/templates/daemonset.yaml
```diff
@@ -38,9 +38,9 @@ spec:
             - --cloud-provider=openstack
             - --use-service-account-credentials=true
             {{- if .Values.serviceMonitor.enabled }}
-            - --address=0.0.0.0
+            - --bind-address=0.0.0.0
```
strigazi, line 41: Is this needed for the configurable volume mounts?

charts/openstack-cloud-controller-manager/values.yaml
```diff
@@ -63,3 +65,24 @@ cloudConfig:
   loadBalancer:
   blockStorage:
   metadata:
+
+# Additional volumes that should be mounted into the pods
+# The following three volumes are required to use all OCCM controllers,
+# but might not be needed if you just use a specific controller
+extraVolumeMounts:
```
strigazi, line 72: What about leaving the extra mounts empty by default?

Like this, `/usr/libexec/kubernetes/kubelet-plugins/volume/exec` can not be removed.


Have you considered the approach below? To leave it empty and let users pick what they need?
https://github.com/kubernetes/ingress-nginx/blob/ab4fa4246797a33bc7b45b45097a91086fe9ad48/charts/ingress-nginx/values.yaml#L537-L545

jacksgt: Leaving it empty would be a backwards-incompatible change.
You can override the entire array by specifying `extraVolumeMounts: []` (or adding the desired entries only).

charts/openstack-cloud-controller-manager/templates/daemonset.yaml
jacksgt: This is not required for this, but the `address` flag has been deprecated. When using `--address`, occm prints the following:
```
Flag --address has been deprecated, see --bind-address instead.
```

strigazi: I leave it to the approvers if they want to take this in on commit.

charts/openstack-cloud-controller-manager/values.yaml
strigazi: It seems that it doesn't merge the default extraVolumeMounts and the user-defined ones. We can override the defaults, thanks!

Is this syntax used elsewhere? I still think it would be friendlier to use standard kubernetes syntax for volumes and volumeMounts.
```
  - name: flexvolume-dir
    hostPath: /usr/libexec/kubernetes/kubelet-plugins/volume/exec
    mountPath: /usr/libexec/kubernetes/kubelet-plugins/volume/exec
```

jacksgt: What do you mean by "standard kubernetes syntax"?

strigazi: In the examples below volumes and volumeMounts are defined as they would in kubernetes manifest. Not sure how else to explain "standard kubernetes syntax".

* https://github.com/kubernetes/cloud-provider-openstack/blob/a9da92cddac704db6d940abbc8f1b2dc20b673cc/charts/cinder-csi-plugin/values.yaml#L52-L62
* mentioned above https://github.com/kubernetes/ingress-nginx/blob/ab4fa4246797a33bc7b45b45097a91086fe9ad48/charts/ingress-nginx/values.yaml#L537-L545

jacksgt: I see, thanks for the example. Well, I tried to adapt the data structure to this use-case because I don't like exposing all the implementation details (otherwise the user eventually needs to pass in the entire `spec` structure).

gman0: Hi @jacksgt, thanks for your contribution! This really seems a bit over-complicated IMO. We'd need to keep around the comment documenting various volume types and options. We'd also lose input validation (imagine making a typo `subpath` instead of `subPath`).

I think it's best to stick with what's already in Kubernetes if we can. kubernetes/ingress-nginx seems like a good example to follow, having `extraVolumes` and `extraVolumeMounts` fields. Not only is this approach more flexible for users, we keep all the input validation. Maybe adding a comment referring users to the official Kubernetes docs on what's the syntax for volumes and volume mounts would help too.

jacksgt: Alrighty, thanks for the feedback. I have updated the PR accordingly.
Would you mind taking a look? @gman0 @strigazi

gman0: @jacksgt great, thank you!

gman0:
Some changes were made. @lingxiankong @jichenjc can you PTAL?

jichenjc:
/lgtm

leave to @lingxiankong for a look as he had some questions at first

lingxiankong:
/approve

Thanks all the people involved.

PR #1710, opened 2021-12-01
[All] Modify install-cpo-oocm role to support use of latest image

ramineni:
This PR adds new role to deploy occm which can be reused in multiple tests in cinder csi, manila csi etc.

fixes #

**Release note**:

```release-note
NONE
```


lingxiankong:
Any particular reason for creating a new role rather than modifying the existing one (`install-cpo-occm`)?

jichenjc:
>>>occm which can be reused in multiple tests in cinder csi, manila csi etc.

>>>Any particular reason for creating a new role rather than modifying the existing one (install-cpo-occm)?

is this the discussion that we had before (run OCCM as dependency so that all CSI related test run with cloud provider) in the slack? Not remember the exact conclusion but we are ok to postpone this to later releases

lingxiankong:
> is this the discussion that we had before (run OCCM as dependency so that all CSI related test run with cloud provider) in the slack? Not remember the exact conclusion but we are ok to postpone this to later releases

Maybe we did have a chat before, but I totally forgot how the consensus was made. However, from the changes, maybe we could re-use the `install-cpo-occm` role by introducing several options.

ramineni:
> > is this the discussion that we had before (run OCCM as dependency so that all CSI related test run with cloud provider) in the slack? Not remember the exact conclusion but we are ok to postpone this to later releases
> 
> Maybe we did have a chat before, but I totally forgot how the consensus was made. However, from the changes, maybe we could re-use the `install-cpo-occm` role by introducing several options.

@lingxiankong the consensus was to reuse the docker latest image directly instead of building the occm for csi related plugins. So I have introduced this. I'm not sure if I could reuse existing one , so added new one. Ill explore if it can be done by adding options.

ramineni:
> > is this the discussion that we had before (run OCCM as dependency so that all CSI related test run with cloud provider) in the slack? Not remember the exact conclusion but we are ok to postpone this to later releases
> 
> Maybe we did have a chat before, but I totally forgot how the consensus was made. However, from the changes, maybe we could re-use the `install-cpo-occm` role by introducing several options.

@lingxiankong I have reused existing job itself to support this. PTAL^^

tests/playbooks/roles/install-cpo-occm/tasks/main.yaml
```diff
@@ -51,13 +52,24 @@
...
+    cmd: |
+      check={{ build_image }}
+      if ["$check" = true] ; then
+          image_url="{{ remote_registry_host }}/openstack-cloud-controller-manager-amd64:{{ github_pr }}"
+      else
+          image_url="docker.io/k8scloudprovider/openstack-cloud-controller-manager:latest"
+      fi
+
+      echo $image_url
+  register: image_url
```
lingxiankong, line 71: Why not just define some extra variables? e.g. (may need test)

```
generated_image_url: "{{ remote_registry_host }}/openstack-cloud-controller-manager-amd64:{{ github_pr }}"
image_url: "{{ generated_image_url if build_image else 'docker.io/k8scloudprovider/openstack-cloud-controller-manager:latest' }}"

ramineni:
@lingxiankong PTAL^^

ramineni:
@jichenjc @lingxiankong Ill update the rest of the jobs to reuse this role in a separate PR, if this approach looks ok to you.

lingxiankong:
/lgtm

thanks!

ramineni:
@jichenjc PTAL

jichenjc:
/lgtm
/approve 

thanks for the update

PR #1711, opened 2021-12-02
[cinder-csi-plugin] Enable logs for e2e tests

ramineni:
This PR enables logs  of e2e tests for debugging purpose
fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/ci-csi-cinder-e2e.sh
+++ b/tests/ci-csi-cinder-e2e.sh
@@ -109,10 +109,10 @@ ansible-playbook -v \
   -e github_pr_branch=${PR_BRANCH}
 exit_code=$?
 
-# Fetch devstack logs for debugging purpose
-#scp -i ~/.ssh/google_compute_engine \
-#  -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no \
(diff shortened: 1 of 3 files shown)


ramineni:
@jichenjc @lingxiankong Ready for review. PTAL

jichenjc:
/lgtm

@ramineni I assume this is log for e2e test, seems the `build log` contains those info as well? any additional info we got from those logs? maybe we need consider include the pod logs of manifests such as provisioner etc? thoughts?

ramineni:
> /lgtm
> 
> @ramineni I assume this is log for e2e test, seems the `build log` contains those info as well? any additional info we got from those logs? 

@jichenjc yes, right. build log also contains the same info . e2e log just captures only e2e tests results which makes it much easier to read/debug the test failures . 

> maybe we need consider include the pod logs of manifests such as provisioner etc? thoughts?

@jichenjc sounds good , but we dont need always , we can keep it optional and enable as required if possible.

jichenjc:
>>>sounds good , but we dont need always , we can keep it optional and enable as required if possible.

agree, that's helpful to us , approve as it's passed the test and just enhancement in the test logs 

/approve

PR #1715, opened 2021-12-08
[cinder-csi-plugin] Updated CLOUD_CONFIG in Helm Chart

powellchristoph:

Currently the openstack-ccm helm chart [hardcodes its key name](https://github.com/kubernetes/cloud-provider-openstack/blob/master/charts/openstack-cloud-controller-manager/templates/secret.yaml#L8) when creating its secret as `cloud.conf`. Due to the different secret keys, you must create two separate secrets if running both the CCM and CSI plugin. This change would allow the `cinder-csi-plugin` reuse the existing secret created by the CCM.

```release-note
* Changed CLOUD_CONFIG env var in Helm Chart.
```


powellchristoph:
Completed CLA Authorization.

jichenjc:
I don't remember clearly whether cloud.conf is same in CSI and OCCM
if they do, yes, we need think about make them identical

powellchristoph:
Just a quick look at the source, it appears that each plugin references their own section in the config file so they should be fine.

jichenjc:
could you please help to fix the CLA and other issues?

I think it's ok and reasonable to make the secret reusable if we can

charts/cinder-csi-plugin/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: latest
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 2.0.0
+version: 2.0.1
```
lingxiankong, line 5: Version bump will be done during the release, so please keep the version here.

lingxiankong (changes requested):
and please fix the CLA issue.

ramineni:
@powellchristoph I think the issue 2 commits are done with diff usernames. fixing should solve the cla issue

powellchristoph:
I borked this all up. I'll submit a new one thats proper.

PR #1717, opened 2021-12-09
[openstack-cloud-controller-manager] Log warning when internal-lb=true and requesting external LB

kopiczko:

This logs a warning when `internal-lb=false` is set and external service is requested. The documentation [here](https://github.com/kubernetes/cloud-provider-openstack/tree/master/examples/loadbalancers) says `service.beta.kubernetes.io/openstack-internal-load-balancer` is `false` by default. But `internal-lb` is also `false` by default. We spent some time debugging it without any indication in the logs.

This is my first PR here, so sorry if I messed up with something. Also for me it could be an error, but I'm not sure how important is backward-compatibility here.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1514,6 +1514,9 @@ func (lbaas *LbaasV2) checkService(service *corev1.Service, nodes []*corev1.Node
 
 	// If in the config file internal-lb=true, user is not allowed to create external service.
 	if lbaas.opts.InternalLB {
+		if !getBoolFromServiceAnnotation(service, ServiceAnnotationLoadBalancerInternal, false) {
+			klog.V(1).Infof("Annotation %q is set to false (default) but internal-lb=true, enforcing internal LB", ServiceAnnotationLoadBalancerInternal)
+		}
 		svcConf.internal = true
 	} else {
 		svcConf.internal = getBoolFromServiceAnnotation(service, ServiceAnnotationLoadBalancerInternal, lbaas.opts.InternalLB)
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1515,7 +1515,7 @@ func (lbaas *LbaasV2) checkService(service *corev1.Service, nodes []*corev1.Node
 	// If in the config file internal-lb=true, user is not allowed to create external service.
 	if lbaas.opts.InternalLB {
 		if !getBoolFromServiceAnnotation(service, ServiceAnnotationLoadBalancerInternal, false) {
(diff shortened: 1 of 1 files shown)


kopiczko:
Any hints on the CI here? I ran `go test` for the changed package and it seems to be fine.

pkg/openstack/loadbalancer.go
```diff
@@ -1514,6 +1514,9 @@ func (lbaas *LbaasV2) checkService(service *corev1.Service, nodes []*corev1.Node
 
 	// If in the config file internal-lb=true, user is not allowed to create external service.
 	if lbaas.opts.InternalLB {
+		if !getBoolFromServiceAnnotation(service, ServiceAnnotationLoadBalancerInternal, false) {
+			klog.V(1).Infof("Annotation %q is set to false (default) but internal-lb=true, enforcing internal LB", ServiceAnnotationLoadBalancerInternal)
```
jichenjc, line 1518: I 'd think we should log this unconditionally if it's helpful for operator?

kopiczko: I don't mind. Please provide a suggestion if you think it's better.

jichenjc: yes, please if you can ,thank you~

kopiczko: Would this work?

```
klog.V(1).Infof("Detected internal-lb=true setting, enforcing internal LB")
```

Is the v-level ok?

lingxiankong: Please use the Structured Logging and use `V(3)`:

```
klog.V(3).InfoS("Enforcing internal LB", "annotation", true, "config", false)
```

lingxiankong (changes requested):

pkg/openstack/loadbalancer.go
jichenjc: @lingxiankong any reason for v3? I think it's a mutating and only log once, should we always log it?

lingxiankong: Because the occm is enforcing the load balancer configuration which is different with the user's expectation, the log here makes sense and it's just a warning, the operator still has a chance to see that when debugging. Using v3 because it doesn't fall into the default log level based on https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md

jichenjc: ok, well point, agree , thanks~

ramineni:
@kopiczko could you push the PR again with review comments addressed. CI should be automatically fixed on new patch set.

ramineni:

PR #1723, opened 2021-12-28
[OCCM][doc]: Add tip for single master node LB creation

jichenjc:

took me a while to figure out why the LB failed to create in one node env even I untaint the master
it's because `node.kubernetes.io/exclude-from-external-load-balancers` labeled by default and this will prevent the node to be used 
https://github.com/kubernetes/cloud-provider/blob/25867882d509131a6fdeaf812ceebfd0f19015dd/controllers/service/controller.go#L673

fixes #

**Release note**:

```release-note
NONE
```


docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
```diff
@@ -34,6 +34,8 @@ For more information about cloud-controller-manager, please see:
 
 **NOTE: Now, the openstack-cloud-controller-manager implementation is based on OpenStack Octavia, Neutron-LBaaS has been deprecated in OpenStack since Queens release and no longer maintained in openstack-cloud-controller-manager. So make sure to use Octavia if upgrade to the latest openstack-cloud-controller-manager docker image.**
 
+**NOTE: for test/PoC with only 1 master node environment, you need remove the label `node.kubernetes.io/exclude-from-external-load-balancers` of the master node otherwise the loadbalancer will not be created. search the label [here](https://pkg.go.dev/k8s.io/api/core/v1) for further information.**
```
lingxiankong, line 37: How about we add this to `Limitation` section or in the doc `Exposing applications using services of LoadBalancer type`?

lingxiankong (changes requested):

PR #1728, opened 2021-12-31
[octavia-ingress-controller]: update label from master to excludeLB

### LINKED ISSUE #1724: [octavia-ingress-controller] Master node label usage is inconsistent with cloud provider

>
> /kind bug
> /kind feature

**What happened**:

https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/ingress/controller/controller.go#L210
we are using `LabelNodeRoleMaster = "node-role.kubernetes.io/master"`

but cloud-provider for LB is using 
`node.kubernetes.io/exclude-from-external-load-balancers`

https://github.com/kubernetes/cloud-provider/blob/25867882d509131a6fdeaf812ceebfd0f19015dd/controllers/service/controller.go#L673

why we use different label for LB and ingress ?

**How to reproduce it**:

**Environment**:
- OpenStack version:
- Others:

jichenjc:

fixes #1724 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/ingress/controller/controller.go
+++ b/pkg/ingress/controller/controller.go
@@ -85,9 +85,14 @@ const (
 	// IngressClass specifies which Ingress class we accept
 	IngressClass = "openstack"
 
-	// LabelNodeRoleMaster specifies that a node is a master
+	// LabelNodeExcludeLB specifies that a node should not be used to create a Loadbalancer on
+	// https://github.com/kubernetes/cloud-provider/blob/25867882d509131a6fdeaf812ceebfd0f19015dd/controllers/service/controller.go#L673
+	LabelNodeExcludeLB = "node.kubernetes.io/exclude-from-external-load-balancers"
+
+	// DepcreatedLabelNodeRoleMaster specifies that a node is a master
 	// It's copied over to kubeadm until it's merged in core: https://github.com/kubernetes/kubernetes/pull/39112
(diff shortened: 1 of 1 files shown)


pkg/ingress/controller/controller.go
```diff
@@ -85,9 +85,9 @@ const (
 	// IngressClass specifies which Ingress class we accept
 	IngressClass = "openstack"
 
-	// LabelNodeRoleMaster specifies that a node is a master
-	// It's copied over to kubeadm until it's merged in core: https://github.com/kubernetes/kubernetes/pull/39112
-	LabelNodeRoleMaster = "node-role.kubernetes.io/master"
+	// LabelNodeExcludeLB specifies that a node should not be used to create a Loadbalancer on
+	// https://github.com/kubernetes/cloud-provider/blob/25867882d509131a6fdeaf812ceebfd0f19015dd/controllers/service/controller.go#L673
+	LabelNodeExcludeLB = "node.kubernetes.io/exclude-from-external-load-balancers"
```
lingxiankong, line 90: To not break the existing deployment, can we use a list for the excluded labels including both?

lingxiankong (changes requested):

pkg/ingress/controller/controller.go
jichenjc: updated, thanks~

PR #1739, opened 2022-01-11
[cinder-csi-plugin]not use compute az when ignore-volume-az=true is set

### LINKED ISSUE #1735: [cinder-csi-plugin] topology-aware provisioning broken when compute and volume AZs are distinct

/kind bug

**What happened**:

* If `availability` is not set in storage class parameters, cinder-csi will use a zone supplied by the `CreateVolumeRequest.accessibility_requirements` as a volume AZ, if present:
   https://github.com/kubernetes/cloud-provider-openstack/blob/d46aa87534042ad1e26b812d1ef1aa140317a25e/pkg/csi/cinder/controllerserver.go#L73-L84
* If storage class doesn't specify any `allowedTopologies`, external-provisioner populates `accessibility_requirements` with aggregated cluster (compute) topology by default (see [--immediate-topology](https://github.com/kubernetes-csi/external-provisioner#other-recognized-arguments), and the [effect it has](https://github.com/kubernetes-csi/external-provisioner#topology-support)). That makes the ` != nil` condition above succeed, setting `volAvailability` to some compute AZ.
* If volume and compute AZs are distinct, cinder-csi will always make wrong choice when selection an AZ from `accessibility_requirements`, resulting in a volume provisioning error.

* Helm chart could provide a switch to disable topology-aware provisioning altogether, so that `CreateVolumeRequest.accessibility_requirements` would never be populated, or
* Helm chart could provide a switch to run external-provisioner with `--immediate-topology=false`, or
* cinder-csi could check that if no `availability` volume parameter is supplied, and `ignore-volume-az=true` is set in cloud config, it should not consider zones from `accessibility_requirements` as volume AZs.

**How to reproduce it**:

Provision a volume in a cloud with distinct volume and compute AZs.

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: v1.22.0, external-provisioner v2.2.0
- OpenStack version:
- Others:

jichenjc:

fixes #1735

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -76,15 +76,16 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	// Required, incase vol AZ is different from node AZ
 	volAvailability = req.GetParameters()["availability"]
 
+	cloud := cs.Cloud
+	ignoreVolumeAZ := cloud.GetBlockStorageOpts().IgnoreVolumeAZ
+
 	if len(volAvailability) == 0 {
-		// Check from Topology
-		if req.GetAccessibilityRequirements() != nil {
+		// Check from Topology, if ignore-volume-az=true is set in cloud config
+		// then the AZ of compute will not be used, see #1735 for further info
+		if req.GetAccessibilityRequirements() != nil && ignoreVolumeAZ {
 			volAvailability = getAZFromTopology(req.GetAccessibilityRequirements())
 		}
 	}
-
-	cloud := cs.Cloud
-	ignoreVolumeAZ := cloud.GetBlockStorageOpts().IgnoreVolumeAZ
 
 	// Verify a volume with the provided name doesn't already exist for this tenant
 	volumes, err := cloud.GetVolumesByName(volName)
--- a/pkg/csi/cinder/controllerserver_test.go
+++ b/pkg/csi/cinder/controllerserver_test.go
@@ -46,9 +46,10 @@ func TestCreateVolume(t *testing.T) {
 	// mock OpenStack
 	properties := map[string]string{"cinder.csi.openstack.org/cluster": FakeCluster}
 	// CreateVolume(name string, size int, vtype, availability string, snapshotID string, tags *map[string]string) (string, string, int, error)
-	osmock.On("CreateVolume", FakeVolName, mock.AnythingOfType("int"), FakeVolType, FakeAvailability, "", "", &properties).Return(&FakeVol, nil)
+	osmock.On("CreateVolume", FakeVolName, mock.AnythingOfType("int"), FakeVolType, "", "", "", &properties).Return(&FakeVol, nil)
 
 	osmock.On("GetVolumesByName", FakeVolName).Return(FakeVolListEmpty, nil)
+
 	// Init assert
 	assert := assert.New(t)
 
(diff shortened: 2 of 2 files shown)


pkg/csi/cinder/controllerserver.go
```diff
@@ -76,16 +76,17 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
 
+	cloud := cs.Cloud
+	ignoreVolumeAZ := cloud.GetBlockStorageOpts().IgnoreVolumeAZ
+
 	if len(volAvailability) == 0 {
-		// Check from Topology
-		if req.GetAccessibilityRequirements() != nil {
+		// Check from Topology, if ignore-volume-az=true is set in cloud config
+		// then the AZ of compute will not be used, see #1735 for further info
+		if req.GetAccessibilityRequirements() != nil && ignoreVolumeAZ {
```
gman0, line 85: The problem is that `req.GetAccessibilityRequirements()` is never `nil`. This is because external-provisioner is run with `--immediate-topology=true` by default. See docs for it in https://github.com/kubernetes-csi/external-provisioner#other-recognized-arguments . So if compute and volume AZs are distinct, setting `ignore-volume-az=true` still won't help. It would still use compute AZ as a volume AZ -- which is the original problem described in https://github.com/kubernetes/cloud-provider-openstack/issues/1735.

We might need an option to set `--immediate-topology=false` via Helm values, or a new cloud config option that disables  setting compute topology as a volume AZ.

jichenjc: um... I need more time to read all this and maybe do some test on

for now ,looks like a `--immediate-topology=false` seems more reasonable as it's just configuration change 
will post more analysis later on

gman0: I can share my notes I made before filing the bug report, if it helps any...

external-provisioner configuration:
* cinder-csi runs with `--immediate-topology=true`, and `--strict-topology=false`
* `--strict-topology=false` makes `CreateVolumeRequest.accessibility_requirements` = selected node
* `--immediate-topology=true` makes `CreateVolumeRequest.accessibility_requirements` = aggregated cluster topology, if `allowedTopologies` is empty
* [This table](https://github.com/kubernetes-csi/external-provisioner#topology-support) shows the different behaviors for different combinations of switches and storage class parameters
* With these options, the table linked above reduces to:
  | Delayed binding | Allowed topologies | Resulting accessibility requirements |
  |:-:|:-:|-|
  |Yes|No|`Requisite` = Aggregated cluster topology<br>`Preferred` = `Requisite` with selected node topology as first element|
  |Yes|Yes|`Requisite` = Allowed topologies<br>`Preferred` = `Requisite` with selected node topology as first element|
  |No|No|`Requisite` = Aggregated cluster topology<br>`Preferred` = `Requisite` with randomly selected node topology as first element|
  |No|Yes|`Requisite` = Allowed topologies<br>`Preferred` = `Requisite` with randomly selected node topology as first element|

cinder-csi configuration:
* `availability: <volume AZ name>` storage class parameter:
  * Not empty: `volume.Availability = parameters["availability"]`
  * Empty: `volume.Availability = head(req.AccessibilityRequirements.Preferred)`
* `ignore-volume-az=<bool>` cloud config parameter:
  * True:
  ```go
  resp.AccessibleTopologies =
    req.AccessibilityRequirements.Preferred
  ```
  * False:
  ```go
  resp.AccessibleTopologies = map[string]string{
    "topology.cinder.csi.openstack.org/zone": volume.Availability
  }
  ```

gman0: Maybe @lingxiankong @ramineni have other ideas?

gman0: Essentially, the issue we've run into is that `volume.Availability` is never empty: it's either what you explicitly specify the zone to be in the storage class, or external-provisioner chooses one for you -- from the compute zones.

That means Cinder can never round-robin across AZs allowed in Cinder's volume type. For clouds where compute and volume AZs don't have any relation this is a problem, because we would have to have a separate storage class for each volume AZ (where we'd have to set the `availability` param explicitly).

gman0: And secondly, going with the assumption that "compute and volume AZs don't have any relation" is true for certain clouds, tagging the PV with anything but `nil` in `resp.AccessibleTopologies` may also bring problems (e.g. if a new compute AZ was added in the future).

IMHO the least intrusive way to handle this would be to expose `--immediate-topology` in Helm values, leaving it to `=true` by default.

jichenjc: >>>IMHO the least intrusive way to handle this would be to expose --immediate-topology in Helm values, leaving it to =true by default.

I agree with above, maybe some doc (consolidate your great analysis above) should also be helpful 

@ramineni @lingxiankong thoughts?

gman0: @jichenjc I failed to notice that cinder-csi's Helm chart already provides a way to disable topology altogether using `csi.provisioner.topology`, which is then fed into

https://github.com/kubernetes/cloud-provider-openstack/blob/ee4ecca326219153805b09ccc95911efd93f58f0/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml#L41-L51

big apologies for all the noise and for wasting your time with this! I assumed there is only `ignore-volume-az` cloud config option, and that there is no way to disable topology.

With `csi.provisioner.topology=false` and `ignore-volume-az=true`, it seems to work just fine:
* `req.AccessibilityRequirements` is now always `nil`
* `volume.Availability` is now always `parameters["availability"]` or empty

gman0:
Perhaps the most general way to handle this is to add a Helm chart value `extraArgs` that would be appended to external-provisioner's args. But with the ability to disable topology awareness that already exists, there is no need for it for this particular case described in https://github.com/kubernetes/cloud-provider-openstack/issues/1735. Thanks for the efforts @jichenjc and I'm sorry for the noise again.

If you don't see any other improvements that could be done in this area, feel free to close this PR.

pkg/csi/cinder/controllerserver.go
jichenjc: no worry, learn a lot from this :) thanks for reporting, I am going to close this PR

PR #1741, opened 2022-01-11
[cinder-csi-plugin] Add clusterID Helm value

gman0:

cinder-csi doesn't currently allow to set the cluster ID via its helm chart. It's hardcoded to `kubernetes`. This PR adds a Helm value `clusterID` so that it's configurable.

fixes #

**Release note**:

```release-note
[cinder-csi-plugin] Add clusterID Helm chart value
```


jichenjc:
I don't remember correctly, we should update the version in Charts.yaml for such change?

gman0:
@jichenjc no, chart version is incremented only on new releases now (you can see that the helm linter in CI doesn't report an error either).

jichenjc:
/lgtm
/approve

ok , got it :)

PR #1747, opened 2022-01-19
[cinder-csi-plugin] Updated CLOUD_CONFIG in helm chart.

powellchristoph:
My apologies for messing up the previous one.

Currently the openstack-ccm helm chart [hardcodes its key name](https://github.com/kubernetes/cloud-provider-openstack/blob/master/charts/openstack-cloud-controller-manager/templates/secret.yaml#L8) when creating its secret as `cloud.conf`. Due to the different secret keys, you must create two separate secrets if running both the CCM and CSI plugin. This change would allow the `cinder-csi-plugin` reuse the existing secret created by the CCM.

```release-note
 [cinder-csi-plugin] Make CLOUD_CONFIG secret keys consistent in helm chart for better re-usability 
```

### DIFF
--- a/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
+++ b/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
@@ -113,7 +113,7 @@ spec:
             - name: CSI_ENDPOINT
               value: unix://csi/csi.sock
             - name: CLOUD_CONFIG
-              value: /etc/kubernetes/cloud-config
+              value: /etc/kubernetes/cloud.conf
             - name: CLUSTER_NAME
               value: "{{ .Values.clusterID }}"
           ports:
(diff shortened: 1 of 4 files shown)


ramineni:
updated release note

ramineni:
Thanks for the PR. looks good.
/lgtm
/approve

MPV:
Is this change missing a bump of the helm chart version?

MPV:
Does this change introduce a backwards-incompatible change (by not making the filename configurable, and thus switching it from `cloud-config` to `cloud.conf`)?

Would we want to make it configurable to ease upgrading (so people don't need to create new secrets as part of upgrading)?

lingxiankong:
> s this change missing a bump of the helm chart version?

We don't bump chart version during the dev cycle.

lingxiankong:
> Does this change introduce a backwards-incompatible change (by not making the filename configurable, and thus switching it from `cloud-config` to `cloud.conf`)?
> 
> Would we want to make it configurable to ease upgrading (so people don't need to create new secrets as part of upgrading)?

Could using `checksum` annotation in the Deployment solve this? If yes, maybe we could consider backporting this to stable release.

PR #1762, opened 2022-01-21
[manila-csi-plugin] more e2e tests

gman0:

This PR adds rest of the tests from e2e storage test suite.

Problems I've experienced:
* FsGroupChangePolicyTestSuite:
  * It seems like it's trying to delete a PVC that's still in use by a Pod, and then fails due to timeout when attempting to delete that PVC.
  * It also deleted storage class of a PVC before deleting that PVC, which then causes deletion of that PVC to fail (can't access the provisioner secrets anymore).
* EphemeralTestSuite:
  * Pod: `Warning  FailedMount  16s (x10 over 4m25s)  kubelet            MountVolume.NewMounter initialization failed for volume "my-volume-0" : volume mode "Ephemeral" not supported by driver cephfs.manila.csi.openstack.org (only supports ["Persistent"])`
  * The test doesn't check for this, and is stuck on `STEP: checking the requested inline volume exists in the pod running on node {Name: Selector:map[] Affinity:nil}` forever.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/e2e/csi/manila/manilavolume.go
+++ b/tests/e2e/csi/manila/manilavolume.go
@@ -1,0 +1,113 @@
+package test
+
+import (
+	"bytes"
+	"os/exec"
+	"strconv"
+	"strings"
+
+	"github.com/onsi/ginkgo"
+	"k8s.io/kubernetes/test/e2e/framework"
+	storageframework "k8s.io/kubernetes/test/e2e/storage/framework"
+)
+
+func runCmd(name string, args ...string) ([]byte, error) {
+	var stdout, stderr bytes.Buffer
+	cmd := exec.Command(name, args...)
+
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	framework.Logf("Running %s %v", cmd.Path, cmd.Args)
+	err := cmd.Run()
+
+	framework.Logf("Stdout output %s %v: `%s`", cmd.Path, cmd.Args, stdout.String())
+	framework.Logf("Stderr output %s %v: `%s`", cmd.Path, cmd.Args, stderr.String())
+
+	return stdout.Bytes(), err
+}
+
+// It is assummed that the `openstack` and other related programs
+// are accessible from $PATH on the node.
+
+type manilaVolume struct {
+	proto    string
+	shareID  string
+	accessID string
+}
+
+func manilaCreateVolume(
+	shareProto,
+	accessType,
+	accessTo string,
+	sizeInGiB int,
+	config *storageframework.PerTestConfig,
+) storageframework.TestVolume {
+	ginkgo.By("Creating a test Manila volume externally")
+
+	// Create share.
+
+	out, err := runCmd(
+		"openstack",
+		"share",
+		"create",
+		shareProto,
+		strconv.Itoa(sizeInGiB),
+		"--property=provisioned-by=manila.csi.openstack.org/e2e.test",
+		"--format=value",
+		"--column=id",
+		"--wait",
+	)
+
+	shareID := strings.TrimSpace(string(out))
+
+	framework.ExpectNoError(err)
+	framework.ExpectNotEqual(shareID, "")
+
+	framework.Logf("Created test Manila volume %s", shareID)
+
+	// Grant access to the share.
+
+	out, err = runCmd(
+		"openstack",
+		"share",
+		"access",
+		"create",
+		shareID,
+		accessType,
+		accessTo,
+		"--format=value",
+		"--column=id",
+	)
+
+	// XXX: In case of cephx access rights, the access_key field might
+	//      not be generated in time for when the volume is mounted.
+	//      Tests may fail. A workaround would be to wait until it is ready.
+
+	accessID := strings.TrimSpace(string(out))
+
+	framework.ExpectNoError(err)
+	framework.ExpectNotEqual(accessID, "")
+
+	framework.Logf("Created access right %s for Manila volume %s", accessID, shareID)
+
+	return &manilaVolume{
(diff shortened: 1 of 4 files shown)


gman0:
@tombarron @gouthampacha @ramineni @jichenjc PTAL

Passing and enabled test suites (please consider the times below only as very rough estimates of how long each test *could* run in the CI -- those are single data point values, not average times ; it greatly depends on the backend used (locally I'm using CephFS, but our CI uses NFS LVM) and other conditions of the CI environment):
* [Volumes](https://github.com/kubernetes/kubernetes/blob/release-1.23/test/e2e/storage/testsuites/volumes.go) (took 116.764 seconds)
* [SubPath](https://github.com/kubernetes/kubernetes/blob/release-1.23/test/e2e/storage/testsuites/subpath.go) (took 541.871 seconds)
* [Provisioning](https://github.com/kubernetes/kubernetes/blob/release-1.23/test/e2e/storage/testsuites/provisioning.go) (took 58.566 seconds)
* [VolumeMode](https://github.com/kubernetes/kubernetes/blob/release-1.23/test/e2e/storage/testsuites/volumemode.go) (took 229.025 seconds)
* [Snapshottable](https://github.com/kubernetes/kubernetes/blob/release-1.23/test/e2e/storage/testsuites/snapshottable.go) (took 809.996 seconds)
* [VolumeExpand](https://github.com/kubernetes/kubernetes/blob/release-1.23/test/e2e/storage/testsuites/volume_expand.go) (took 101.469 seconds)
* [VolumeIO](https://github.com/kubernetes/kubernetes/blob/release-1.23/test/e2e/storage/testsuites/volume_io.go) (took 49.459 seconds)
* Total time 1907.150 seconds (~32 minutes)

Test suites that should be enabled but are failing at the moment and therefore disabled:
* [FsGroupChangePolicyTestSuite](https://github.com/kubernetes/kubernetes/blob/release-1.23/test/e2e/storage/testsuites/fsgroupchangepolicy.go): see notes in the original post of this PR

Test suites deliberately not enabled:
* [Capacity](https://github.com/kubernetes/kubernetes/blob/release-1.23/test/e2e/storage/testsuites/capacity.go): manila-csi doesn't support [capacity-aware](https://kubernetes-csi.github.io/docs/storage-capacity-tracking.html) provisioning
* [Topology](https://github.com/kubernetes/kubernetes/blob/release-1.23/test/e2e/storage/testsuites/topology.go): manila-csi needs [topology-aware](https://kubernetes-csi.github.io/docs/topology.html) provisioning only in some specific cases that are not currently present in our CI setup
* [VolumeLimits](https://github.com/kubernetes/kubernetes/blob/release-1.23/test/e2e/storage/testsuites/volumelimits.go): manila-csi doesn't set any [`NodeGetInfoResponse.max_volumes_per_node`](https://github.com/container-storage-interface/spec/blob/master/spec.md#nodegetinfo) that this test suite tries to verify
* [Disruptive](https://github.com/kubernetes/kubernetes/blob/release-1.23/test/e2e/storage/testsuites/disruptive.go):
  * Disruptive tests are skipped in CI:
    
    https://github.com/kubernetes/cloud-provider-openstack/blob/f1499fbb466aebd667e2367ffdee1be89971dfc5/tests/playbooks/roles/install-csi-manila/tasks/main.yaml#L285
    
    In cinder-csi too:
    
    https://github.com/kubernetes/cloud-provider-openstack/blob/f1499fbb466aebd667e2367ffdee1be89971dfc5/tests/playbooks/roles/install-csi-cinder/tasks/main.yaml#L210
    
    @ramineni @jichenjc dou you know why?
  * When I try to run this test suite manually, it fails:
  
    ```
    [Fail] [sig-storage] [manila-csi-e2e] CSI Volumes [Driver: cephfs.manila.csi.openstack.org] [Testpattern: Dynamic PV (filesystem volmode)] disruptive[Disruptive][LinuxOnly] [It] Should test that pv written before kubelet restart is readable after restart. 
    ```
    
    IMHO it's a good thing to test for this case and I'm not sure why it fails. I've tried with both FUSE and kernel CephFS clients, by the way, with the same result. I'll need to check this some more later.
* [VolumePerf](https://github.com/kubernetes/kubernetes/blob/release-1.23/test/e2e/storage/testsuites/volumeperf.go): verifies QPS and burst -- manila-csi itself doesn't offer such QoS guarantees

Other test suites available in sig-storage e2e test suite framework that I haven't worked on integrating yet:
* [VolumeStress](https://github.com/kubernetes/kubernetes/blob/release-1.23/test/e2e/storage/testsuites/volume_stress.go)
* [SnapshottableStress](https://github.com/kubernetes/kubernetes/blob/release-1.23/test/e2e/storage/testsuites/snapshottable_stress.go)

gman0:
If you agree with the current set of enabled test suites, this PR is ready -- from my side at least.

jichenjc:
>>>@ramineni @jichenjc dou you know why?

I don't know the reason, I can dig more here and I agree with you, the e2e need enhance
the PR itself is reasonable to me , thanks

/approve

gouthampacha:
> @tombarron @gouthampacha @ramineni @jichenjc PTAL
> 
> Passing and enabled test suites (please consider the times below only as very rough estimates of how long each test _could_ run in the CI -- those are single data point values, not average times ; it greatly depends on the backend used (locally I'm using CephFS, but our CI uses NFS LVM) and other conditions of the CI environment):
> 

Thank you for working on these and enabling these @gman0! The explanation was really useful to grok the situation and learn about these tests; will you be filing a bug about the FsGroupChangePolicyTestSuite issue?

/lgtm

gman0:
@gouthampacha yes I'll be following this up

PR #1777, opened 2022-02-01
[occm] Add ability to disable Service controller

dulek:
One of the CNI plugins that can be used when Kubernetes runs on OpenStack is Kuryr-Kubernetes. If it's configured, K8s pods will get wired using OpenStack Neutron and trunk ports. Services, on the other hand, will get represented by internal-facing Octavia load balancers. In case of LoadBalancer Services, Kuryr will use the same approach, only adding a FIP to such an LB. This means that when running with Kuryr the Services controller in cloud-provider-openstack should be disabled, as Kuryr handles that on his own.

This commit adds `enabled` option to the `LoadBalancerOpts`, that will default to `true`. If explicitly disabled, the `EnsureLoadBalancer()` and `UpdateLoadBalancer()` functions will only raise `ImplementedElsewhere` error, effectively telling the controller that integration is disabled.

Please note that `GetLoadBalancer()` and `EnsureLoadBalancerDeleted()` are unchanged, as controller still needs to be able to delete the LB in case integration was enabled earlier on the cluster.

**Release note**:
```release-note
[openstack-cloud-controller-manager] Added `enabled` option in the LoadBalancer section of the configuration, when set to false, the Service of LoadBalancer type is managed by other controllers.
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1910,6 +1910,10 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 
 // EnsureLoadBalancer creates a new load balancer or updates the existing one.
 func (lbaas *LbaasV2) EnsureLoadBalancer(ctx context.Context, clusterName string, apiService *corev1.Service, nodes []*corev1.Node) (*corev1.LoadBalancerStatus, error) {
+	if !lbaas.opts.Enabled {
+		return nil, cloudprovider.ImplementedElsewhere
+	}
+
 	mc := metrics.NewMetricContext("loadbalancer", "ensure")
 	status, err := lbaas.ensureLoadBalancer(ctx, clusterName, apiService, nodes)
 	return status, mc.ObserveReconcile(err)
@@ -2733,6 +2737,9 @@ func (lbaas *LbaasV2) updateOctaviaLoadBalancer(ctx context.Context, clusterName
 
 // UpdateLoadBalancer updates hosts under the specified load balancer.
 func (lbaas *LbaasV2) UpdateLoadBalancer(ctx context.Context, clusterName string, service *corev1.Service, nodes []*corev1.Node) error {
+	if !lbaas.opts.Enabled {
+		return cloudprovider.ImplementedElsewhere
+	}
 	mc := metrics.NewMetricContext("loadbalancer", "update")
 	err := lbaas.updateLoadBalancer(ctx, clusterName, service, nodes)
 	return mc.ObserveReconcile(err)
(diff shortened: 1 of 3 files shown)


pkg/openstack/loadbalancer.go
```diff
@@ -1910,6 +1910,10 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 
 // EnsureLoadBalancer creates a new load balancer or updates the existing one.
 func (lbaas *LbaasV2) EnsureLoadBalancer(ctx context.Context, clusterName string, apiService *corev1.Service, nodes []*corev1.Node) (*corev1.LoadBalancerStatus, error) {
+	if !lbaas.opts.Enabled {
+		return nil, cloudprovider.ImplementedElsewhere
+	}
+
```
mdbooth, line 1916: Did I understand correctly from the PR description that we can't either:
* Stop watching Service
* Entirely disable the Service controller

Because it may need to Reconcile on delete? Is that correct? Is there anything else we still need to reconcile in CPO?

dulek: We don't have any control over Service controller here as it's running in kubernetes/cloud-provider and here we only implement an interface that's called there. There is an explanation on how it's supposed to work here: https://github.com/kubernetes/cloud-provider/blob/master/cloud.go#L117-L132

This seems to makes sense - imagine someone running with `enabled: true`, creating a few LBs and then switching to `enabled: false`. It's still expected that when these Services will get deleted, the Octavia LB will get removed, even though the integration is disabled now.

dulek:
I just noticed that an alternative approach could be to return error from [1]. The problem with that is that it doesn't really support the case I described above - when the `enabled: false` got set when a few LBs are already existing.

[1] https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/openstack/openstack.go#L309

lingxiankong:
> when the enabled: false got set when a few LBs are already existing.

Yeah, that depends on if you want to totally disable the controller or just disable it for creating new load balancers.

pkg/openstack/loadbalancer.go
jichenjc: @mdbooth the other resource creation is in other controllers, here we only create LB for LB typed services , so should be ok to return here directly

dulek:
> > when the enabled: false got set when a few LBs are already existing.
> 
> Yeah, that depends on if you want to totally disable the controller or just disable it for creating new load balancers.

Yup, so I think I prefer to just disable creation of new LBs due to arguments I mentioned above - after disabled, it's still expected LBs already created will not get orphaned when the Service gets deleted. It might even be a security risk to leave such LBs running in the cluster as after a while pod IPs might get reused and orphaned LB can be used to direct the outside traffic to pods that shouldn't be exposed.

@lingxiankong, @mdbooth, @jichenjc, what are the next steps required to merge this patch?

jichenjc:
/approve

thanks for the PR

PR #1782, opened 2022-02-08
[all] Remove deprecated option `bs-version`

ramineni:
`bs-version` config option is not supported anymore and by default
v3 api is used. This PR is to remove the same from the code.

This PR also removes the reference to identity v2 in docs as its support is  removed.

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/cinder-csi-plugin/using-cinder-csi-plugin.md
+++ b/docs/cinder-csi-plugin/using-cinder-csi-plugin.md
@@ -96,7 +96,6 @@ Implementation of `cinder-csi-plugin` relies on following OpenStack services.
 
 | Service                        | API Version(s) | Deprecated | Required |
 |--------------------------------|----------------|------------|----------|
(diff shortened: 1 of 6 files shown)


docs/cinder-csi-plugin/using-cinder-csi-plugin.md
```diff
@@ -96,7 +96,6 @@ Implementation of `cinder-csi-plugin` relies on following OpenStack services.
 
 | Service                        | API Version(s) | Deprecated | Required |
 |--------------------------------|----------------|------------|----------|
```
jichenjc, line 98: just curious we  are talking about block storage version? why remove keystone version here?

ramineni: @jichenjc its also outdated , so removed it. Ill update the description . Thanks

PR #1783, opened 2022-02-08
[occm] The service monitor svc should select the pods of the occm DS

pnordh:
The `occm.labels` helm helper in the occm chart doesn't match the pods of the DS as they instead uses the `occm.controllermanager.labels` helper. By changing the service selector to use the `occm.controllermanager.matchLabels` helper we ensure that it matches the occm DS pods.

### DIFF
--- a/charts/openstack-cloud-controller-manager/templates/service-sm.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/service-sm.yaml
@@ -11,5 +11,5 @@ spec:
     port: 10258
     protocol: TCP
   selector:
-    {{- include "occm.labels" . | nindent 4 }}
+    {{- include "occm.controllermanager.matchLabels" . | nindent 4 }}
 {{- end }}


pnordh:
A similar change is probably needed (but in reverse) for the ServiceMonitor as well as it currently matches for endpoints with the `occm.controllermanager.labels` labelset but the Service (and thus endpoints) uses `occm.labels`.

Any particular reason why there are two different labelset in use for different resources? Should we perhaps consolidate to just one?

MPV:
Looping in @eumel8 & @morremeyer who worked on this when these changes were added back in https://github.com/kubernetes/cloud-provider-openstack/pull/1458

morremeyer:
I won't be able to do something meaningful here for the foreseeable future.

jichenjc:
seems servicemonitor.yaml and daemonset.yaml both use `occm.controllermanager.matchLabels` 

/lgtm

jichenjc:
>>>Should we perhaps consolidate to just one?

seems we should ..

PR #1785, opened 2022-02-09
[cinder-csi-plugin] cherry-pick of #1747

MPV:
Cherry-picks https://github.com/kubernetes/cloud-provider-openstack/pull/1747 into the `release-1.23` branch.

See https://github.com/kubernetes/cloud-provider-openstack/pull/1747 for description.

**Release note**:

```release-note
[cinder-csi-plugin] Make CLOUD_CONFIG secret keys consistent in helm chart for better re-usability
```

### DIFF
--- a/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
+++ b/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
@@ -113,7 +113,7 @@ spec:
             - name: CSI_ENDPOINT
               value: unix://csi/csi.sock
             - name: CLOUD_CONFIG
-              value: /etc/kubernetes/cloud-config
+              value: /etc/kubernetes/cloud.conf
             - name: CLUSTER_NAME
               value: "{{ .Values.clusterID }}"
           ports:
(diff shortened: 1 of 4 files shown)


lingxiankong:
What would happen when the user is upgrading the chart?

MPV:
> What would happen when the user is upgrading the chart?

@lingxiankong I suggested a change here to make it backwards-compatible: https://github.com/kubernetes/cloud-provider-openstack/pull/1747/files#r803493214. And here's a separate PR for it, if we think it makes sense to adopt: https://github.com/kubernetes/cloud-provider-openstack/pull/1789

lingxiankong:
> @lingxiankong I suggested a change here to make it backwards-compatible: https://github.com/kubernetes/cloud-provider-openstack/pull/1747/files#r803493214. And here's a separate PR for it, if we think it makes sense to adopt: #1789

make sense to me.

/lgtm

ramineni:
This shouldn't be required if https://github.com/kubernetes/cloud-provider-openstack/pull/1789 is merged right

MPV:
> This shouldn't be required if #1789 is merged right

You’re almost right. 
Rather, this PR should include a back port of #1789 also.

ramineni:
@MPV I dont see reason why we need backport for this. The charts still work with existing change, so ,its not a bug rather convenient if we have same filenames.

/hold

ramineni:
Not a bug, not a potential backport. so closing this.

/close

PR #1788, opened 2022-02-10
[manila-csi-plugin] Add extra labels to k8s resources

### LINKED ISSUE #1786: [Manila-CSI-plugin] Add extra labels to k8s resources in helm chart

>
> /kind bug
kind feature

**What happened**:
I reckon that having the possibility to add optoinal extra labels to the K8s resources would add flexibility to the Helm chart. 

It is possible to add arbitrary labels to the k8s resources in an optional way.

**How to reproduce it**:
N/A

I've already written the code for that, I will submit it so you can evaluate and approve if you consider it appropriate

**Environment**:
- OpenStack version: N/A
- Others: N/A

JPRbrs:
Extra labels are optional and applied to all resources already sporting labels

Having the possibility of adding optional extra labels to K8s resources would make the Helm chart more flexible.

fixes #1786

I've tested it setting extraValues.enabled to true and adding some dummy labels as well as with no values and extraValues.enabled set to false. Both times I've run helm3 template . 

**Release note**:
```release-note
[Manila-CSI-plugin]
Adding optional extra labels to be set in .Values.extraLabels.labels. Enable the usage of extra labels setting .Values.extraLabels.enable to true
```

### DIFF
--- a/charts/manila-csi-plugin/values.yaml
+++ b/charts/manila-csi-plugin/values.yaml
@@ -15,9 +15,7 @@ shareProtocols:
  #     dir: /var/lib/kubelet/plugins/cephfs.csi.ceph.com
  #     sockFile: csi.sock
 
-extraLabels:
-  enabled: false
-  labels: {}
+extraLabels: {}
 # CSI Manila spec
 csimanila:
   # Set topologyAwarenessEnabled to true to enable topology awareness
--- a/charts/manila-csi-plugin/values.yaml
+++ b/charts/manila-csi-plugin/values.yaml
@@ -15,6 +15,9 @@ shareProtocols:
  #     dir: /var/lib/kubelet/plugins/cephfs.csi.ceph.com
  #     sockFile: csi.sock
 
+extraLabels:
+  enabled: false
+  labels: {}
 # CSI Manila spec
 csimanila:
   # Set topologyAwarenessEnabled to true to enable topology awareness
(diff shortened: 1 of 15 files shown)


charts/manila-csi-plugin/values.yaml
```diff
@@ -15,6 +15,9 @@ shareProtocols:
  #     dir: /var/lib/kubelet/plugins/cephfs.csi.ceph.com
  #     sockFile: csi.sock
 
+extraLabels:
+  enabled: false
```
gman0, line 19: Is `enabled` needed? Won't just leaving the default value at `{}` do the job?

gman0:
@JPRbrs thank you for your PR! One comment in-line, otherwise lgtm. Also please add `[manila-csi-plugin] ` prefix to the release note.

I think it would be good if we kept Cinder's and Manila's charts more or less in sync. Would you mind pushing a follow up PR to add this `extraLabels` to Cinder helm chart too?

charts/manila-csi-plugin/values.yaml
JPRbrs: you're right, I'l change it, thanks!


> @JPRbrs thank you for your PR! One comment in-line, otherwise lgtm. Also please add `[manila-csi-plugin] ` prefix to the release note.
> 
> I think it would be good if we kept Cinder's and Manila's charts more or less in sync. Would you mind pushing a follow up PR to add this `extraLabels` to Cinder helm chart too?

Definitely, no problem at all!

JPRbrs:
comments addressed @gman0.

Thanks!

gman0 (approved):
All good, @JPRbrs. Thank you!

/lgtm
/approve

gman0:
@jichenjc can you please /approve if this looks ok to you?

gman0:
@JPRbrs I believe new charts are only pushed along with regular CPO releases -- unless it's a bugfix patch release.

I'm not CPO repo maintainer but I'm not sure this PR is eligible for a bugfix release. Maybe others and @lingxiankong can tell you more?

JPRbrs:
Many thanks for your quick response! I'll wait to see what other people say, thanks again!

PR #1789, opened 2022-02-10
[cinder-csi-plugin] Configurable secret filename

MPV:
Making the change from #1747 backwards-compatible, when using `secret.enabled: true` and `secret.create: false`, and thus creating the secret externally (referred from `secret.name`). Some users might even be sharing the same secret between OCCM and Cinder CSI out of convenience.

#### Before #1747:

```yaml
$ helm template cinder-csi-plugin charts/cinder-csi-plugin | grep "value: /etc/kubernetes"
              value: /etc/kubernetes/cloud-config
              value: /etc/kubernetes/cloud-config
```

#### After #1747:

```yaml
$ helm template cinder-csi-plugin charts/cinder-csi-plugin | grep "value: /etc/kubernetes"
              value: /etc/kubernetes/cloud.conf
              value: /etc/kubernetes/cloud.conf
```

#### After this PR (by default):

```yaml
$ helm template cinder-csi-plugin charts/cinder-csi-plugin | grep "value: /etc/kubernetes"
              value: /etc/kubernetes/cloud.conf
              value: /etc/kubernetes/cloud.conf
```

#### After this PR (when configured for backwards-compatibility):

```yaml
$ helm template --set secret.filename="cloud-config" cinder-csi-plugin charts/cinder-csi-plugin | grep "value: /etc/kubernetes"
              value: /etc/kubernetes/cloud-config
              value: /etc/kubernetes/cloud-config
```

**Release note**:

```release-note
[cinder-csi-plugin] Make CLOUD_CONFIG secret filename configurable in helm chart for backwards-compatibility
```

### DIFF
--- a/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
+++ b/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
@@ -113,7 +113,7 @@ spec:
             - name: CSI_ENDPOINT
               value: unix://csi/csi.sock
             - name: CLOUD_CONFIG
-              value: /etc/kubernetes/cloud.conf
+              value: /etc/kubernetes/{{ .Values.secret.filename }}
             - name: CLUSTER_NAME
               value: "{{ .Values.clusterID }}"
           ports:
--- a/charts/cinder-csi-plugin/templates/nodeplugin-daemonset.yaml
+++ b/charts/cinder-csi-plugin/templates/nodeplugin-daemonset.yaml
@@ -65,7 +65,7 @@ spec:
             - name: CSI_ENDPOINT
               value: unix://csi/csi.sock
             - name: CLOUD_CONFIG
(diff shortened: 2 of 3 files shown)


MPV:
@jichenjc Is the gcloud with in CI broken in general at the moment? 

Seeing same failure in both this and another PR: `Your credentials are invalid. Please run gcloud auth login`

jichenjc:
@MPV not sure , I knew gcloud CI has your issue time to time but should be not that frequent..

/lgtm

MPV:
@jichenjc Thanks, looks like the tests passed now. 👏

MPV:
/assign @lingxiankong 

...as you asked in https://github.com/kubernetes/cloud-provider-openstack/pull/1785#issuecomment-1034326036:
> What would happen when the user is upgrading the chart?

PR #1792, opened 2022-02-16
[Cinder-csi-plugin] Provide capability to add extra labels for Cinder

JPRbrs:
As per comment in PR#1788 (https://github.com/kubernetes/cloud-provider-openstack/pull/1788#issuecomment-1038868091) it's been considered a good idea to provide extraLabels capability yo Cinder plugin too.

N/A

This has been tested templating the Cinder plugin Helm chart with and without a test.yaml file containing extra labels

```helm3 template . -f test.yaml```

**Release note**:

```release-note
[Cinder-CSI-plugin]
Adding optional extra labels to be set in .Values.extraLabels. Enable the usage of extra labels adding the desired labels to Values.extraLabels
```

### DIFF
--- a/charts/cinder-csi-plugin/templates/_helpers.tpl
+++ b/charts/cinder-csi-plugin/templates/_helpers.tpl
@@ -62,14 +62,14 @@ Create unified labels for cinder-csi components
 {{- define "cinder-csi.common.matchLabels" -}}
 app: {{ template "cinder-csi.name" . }}
 release: {{ .Release.Name }}
-{{- if .Values.extraLabels }}
-{{ toYaml .Values.extraLabels -}}
-{{- end }}
 {{- end -}}
 
 {{- define "cinder-csi.common.metaLabels" -}}
 chart: {{ template "cinder-csi.chart" . }}
 heritage: {{ .Release.Service }}
+{{- if .Values.extraLabels }}
+{{ toYaml .Values.extraLabels -}}
+{{- end }}
 {{- end -}}
 
 {{- define "cinder-csi.controllerplugin.matchLabels" -}}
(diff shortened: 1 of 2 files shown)


JPRbrs:
@gman0 here is the PR to align Manila CSI with Cinder CSI on the extraLabels feature. Thanks in advance!

charts/cinder-csi-plugin/templates/_helpers.tpl
```diff
@@ -62,6 +62,9 @@ Create unified labels for cinder-csi components
 {{- define "cinder-csi.common.matchLabels" -}}
 app: {{ template "cinder-csi.name" . }}
 release: {{ .Release.Name }}
+{{- if .Values.extraLabels }}
```
gman0, line 65: Hm I wonder if this shouldn't go in `cinder-csi.common.metaLabels` instead? Of course it depends on what exactly we want to label, but https://github.com/kubernetes/cloud-provider-openstack/pull/1788 adds `extraLabels` into `metadata.labels`, not `spec.selector.matchLabels`. What do @ramineni @jichenjc think?

gman0:
@JPRbrs thank you for following this up!

charts/cinder-csi-plugin/templates/_helpers.tpl
JPRbrs: I reckon you're right @gman0. Happy to move it to that template if you all agree

ramineni: +1 it should be in metaLabels

JPRbrs:
Comments addressed @ramineni @gman0  Thanks!

gman0:
Perfect, thanks!

/lgtm
/assign @ramineni

jichenjc:
/approve 

thanks ~

PR #1793, opened 2022-02-16
[OCCM] fix listener pop during iteration

FlorinPeter:
Move the remove of the listener at the end of the loop.
Since the listener is a pointer and needed during the loop the removal will mess up the logic.
We observed on a k8s service with 4 ports that a load balancer listener pool were updated on every reconcile loop.

fixes https://github.com/kubernetes/cloud-provider-openstack/issues/1795

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1874,9 +1874,6 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 				return nil, err
 			}
 
-			// After all ports have been processed, remaining listeners are removed if they were created by this Service.
-			curListeners = popListener(curListeners, listener.ID)
-
 			pool, err := lbaas.ensureOctaviaPool(loadbalancer.ID, cutString(fmt.Sprintf("pool_%d_%s", portIndex, lbName)), listener, service, port, nodes, svcConf)
 			if err != nil {
 				return nil, err
@@ -1885,6 +1882,11 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 			if err := lbaas.ensureOctaviaHealthMonitor(loadbalancer.ID, cutString(fmt.Sprintf("monitor_%d_%s", portIndex, lbName)), pool, port, svcConf); err != nil {
 				return nil, err
 			}
+
+			// After all ports have been processed, remaining listeners are removed if they were created by this Service.
+			// The remove of the listener must always happen at the end of the loop to avoid wrong assignment.
+			// Modifying the curListeners would also change the mapping.
+			curListeners = popListener(curListeners, listener.ID)
 		}
 
 		// Deal with the remaining listeners, delete the listener if it was created by this Service previously.


jichenjc:
thanks for the PR 

do you mind create an issue with background and impact etc ( which is in your commit msg basically) so we can track the changes etc, thanks

jichenjc:
and please update the header of PR to `[OCCM] fix listener pop during iteration `
which indicate this is for OCCM change.. thanks

FlorinPeter:
@jichenjc thx for your help the issue is created

pkg/openstack/loadbalancer.go
```diff
@@ -1885,6 +1882,9 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 			if err := lbaas.ensureOctaviaHealthMonitor(loadbalancer.ID, cutString(fmt.Sprintf("monitor_%d_%s", portIndex, lbName)), pool, port, svcConf); err != nil {
 				return nil, err
 			}
+
+			// After all ports have been processed, remaining listeners are removed if they were created by this Service.
+			curListeners = popListener(curListeners, listener.ID)
```
lingxiankong, line 1887: I don't quite understand, why putting this line to the end could solve the issue?

FlorinPeter: because the listener pointer is being used in in the next function but the pointer is poped before so the reference is not longer the right listener.
if you debug the code you will see how the reference is shifted to the wrong listener while the next function needs the right one

jichenjc: when you say `next function` are you saying `deleteOctaviaListeners` which is next function?

if so I am also curious why move to here can help ..thanks

FlorinPeter: No, I mean ```lbaas.ensureOctaviaPool(loadbalancer.ID, cutString(fmt.Sprintf("pool_%d_%s", portIndex, lbName)), listener, service, port, nodes, svcConf)``` 
The listener pointer is being used there but the reference is not longer correct!

FlorinPeter: I will look into it in the next days and try to give you feedback

FlorinPeter: it turns out the issue was on the OpenStack API because the backend was in trouble and records where mixed up.
Sorry for the noise, I will close the issue.

FlorinPeter:
@lingxiankong @jichenjc I was able to reproduce the issue again and I will try to provide a many informations I can.
On OpenStack side there is one LoadBalancer with 2 listeners.
Bildschirmfoto 2022-03-30 um 14 19 42
While iterating over ```service.Spec.Ports``` the ```lbaas.ensureOctaviaListener```is called to get the listener from ```curListenerMapping``` map
Bildschirmfoto 2022-03-30 um 14 22 26
Right before the ```popListener``` call the listener pointing to the Name ```listener_0_kube_service_g04_kube-system_ingress-nginx-controller```
Bildschirmfoto 2022-03-30 um 14 28 41
After the ```popListener``` call, the listener is pointing to the Name ```listener_1_kube_service_g04_kube-system_ingress-nginx-controller``` and here is the problem, because we are still on the ```portIndex``` = 0 but the listener is now pointing to the wrong one and the function ```ensureOctaviaPool``` is mixing members up!

FlorinPeter:
here https://go.dev/play/p/OXmDUNxHIVj you can also see the issue if you don't trust my debugger ;)

ramineni:

lingxiankong:
> here https://go.dev/play/p/OXmDUNxHIVj you can also see the issue if you don't trust my debugger ;)

Thanks for the investigation @FlorinPeter, I see the issue now, the problem is caused by the line:

```
curListenerMapping[key] = &curListeners[i]
```

Here the value of curListenerMapping is a pointer which is going to be changed during the ports iteration.

So rather than moving `curListeners = popListener(curListeners, listener.ID)`, we should change the definition of curListenerMapping to `curListenerMapping := make(map[listenerKey]listeners.Listener)` to avoid the potential issues in the future.

Changed version: https://go.dev/play/p/45IYjqFA2fZ

FlorinPeter:
> So rather than moving curListeners = popListener(curListeners, listener.ID), we should change the definition of curListenerMapping to curListenerMapping := make(map[listenerKey]listeners.Listener) to avoid the potential issues in the future.

sorry but this is a massive change and there is absolutely no benefit doing it instead of moving this function to the end! 
I attached the diff of a possible change in order that you understand what implications it will have and besides that there are still ```if listener == nil {``` checks that needs to changed....
I have no clue why we need to overcomplicate such a small fix...

```
diff --git a/pkg/openstack/loadbalancer.go b/pkg/openstack/loadbalancer.go
index 4eabdc4..9620e20 100644
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1205,7 +1205,7 @@ func (lbaas *LbaasV2) buildMonitorCreateOpts(svcConf *serviceConfig, port corev1
 }
 
 // Make sure the pool is created for the Service, nodes are added as pool members.
-func (lbaas *LbaasV2) ensureOctaviaPool(lbID string, name string, listener *listeners.Listener, service *corev1.Service, port corev1.ServicePort, nodes []*corev1.Node, svcConf *serviceConfig) (*v2pools.Pool, error) {
+func (lbaas *LbaasV2) ensureOctaviaPool(lbID string, name string, listener listeners.Listener, service *corev1.Service, port corev1.ServicePort, nodes []*corev1.Node, svcConf *serviceConfig) (*v2pools.Pool, error) {
 	pool, err := openstackutil.GetPoolByListener(lbaas.lb, lbID, listener.ID)
 	if err != nil && err != openstackutil.ErrNotFound {
 		return nil, fmt.Errorf("error getting pool for listener %s: %v", listener.ID, err)
@@ -1334,7 +1334,7 @@ func (lbaas *LbaasV2) buildBatchUpdateMemberOpts(port corev1.ServicePort, nodes
 }
 
 // Make sure the listener is created for Service
-func (lbaas *LbaasV2) ensureOctaviaListener(lbID string, name string, curListenerMapping map[listenerKey]*listeners.Listener, port corev1.ServicePort, svcConf *serviceConfig, _ *corev1.Service) (*listeners.Listener, error) {
+func (lbaas *LbaasV2) ensureOctaviaListener(lbID string, name string, curListenerMapping map[listenerKey]listeners.Listener, port corev1.ServicePort, svcConf *serviceConfig, _ *corev1.Service) (listeners.Listener, error) {
 	listener, isPresent := curListenerMapping[listenerKey{
 		Protocol: getListenerProtocol(port.Protocol, svcConf),
 		Port:     int(port.Port),
@@ -1349,7 +1349,7 @@ func (lbaas *LbaasV2) ensureOctaviaListener(lbID string, name string, curListene
 		var err error
 		listener, err = openstackutil.CreateListener(lbaas.lb, lbID, listenerCreateOpt)
 		if err != nil {
-			return nil, fmt.Errorf("failed to create listener for loadbalancer %s: %v", lbID, err)
+			return listeners.Listener{}, fmt.Errorf("failed to create listener for loadbalancer %s: %v", lbID, err)
 		}
 
 		klog.V(2).Infof("Listener %s created for loadbalancer %s", listener.ID, lbID)
@@ -1417,7 +1417,7 @@ func (lbaas *LbaasV2) ensureOctaviaListener(lbID string, name string, curListene
 		if listenerChanged {
 			klog.InfoS("Updating listener", "listenerID", listener.ID, "lbID", lbID, "updateOpts", updateOpts)
 			if err := openstackutil.UpdateListener(lbaas.lb, lbID, listener.ID, updateOpts); err != nil {
-				return nil, fmt.Errorf("failed to update listener %s of loadbalancer %s: %v", listener.ID, lbID, err)
+				return listeners.Listener{}, fmt.Errorf("failed to update listener %s of loadbalancer %s: %v", listener.ID, lbID, err)
 			}
 			klog.InfoS("Updated listener", "listenerID", listener.ID, "lbID", lbID)
 		}
@@ -1728,7 +1728,7 @@ func (lbaas *LbaasV2) checkService(service *corev1.Service, nodes []*corev1.Node
 }
 
 // checkListenerPorts checks if there is conflict for ports.
-func (lbaas *LbaasV2) checkListenerPorts(service *corev1.Service, curListenerMapping map[listenerKey]*listeners.Listener, isLBOwner bool, lbName string) error {
+func (lbaas *LbaasV2) checkListenerPorts(service *corev1.Service, curListenerMapping map[listenerKey]listeners.Listener, isLBOwner bool, lbName string) error {
 	for _, svcPort := range service.Spec.Ports {
 		key := listenerKey{Protocol: listeners.Protocol(svcPort.Protocol), Port: int(svcPort.Port)}
 
@@ -1856,10 +1856,10 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 	// This is an existing load balancer, either created by occm for other Services or by the user outside of cluster.
 	if !createNewLB {
 		curListeners := loadbalancer.Listeners
-		curListenerMapping := make(map[listenerKey]*listeners.Listener)
+		curListenerMapping := make(map[listenerKey]listeners.Listener)
 		for i, l := range curListeners {
 			key := listenerKey{Protocol: listeners.Protocol(l.Protocol), Port: l.ProtocolPort}
-			curListenerMapping[key] = &curListeners[i]
+			curListenerMapping[key] = curListeners[i]
 		}
 		klog.V(4).InfoS("Existing listeners", "portProtocolMapping", curListenerMapping)
 
@@ -1874,6 +1874,9 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 				return nil, err
 			}
 
+			// After all ports have been processed, remaining listeners are removed if they were created by this Service.
+			curListeners = popListener(curListeners, listener.ID)
+
 			pool, err := lbaas.ensureOctaviaPool(loadbalancer.ID, cutString(fmt.Sprintf("pool_%d_%s", portIndex, lbName)), listener, service, port, nodes, svcConf)
 			if err != nil {
 				return nil, err
@@ -1882,9 +1885,6 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 			if err := lbaas.ensureOctaviaHealthMonitor(loadbalancer.ID, cutString(fmt.Sprintf("monitor_%d_%s", portIndex, lbName)), pool, port, svcConf); err != nil {
 				return nil, err
 			}
-
-			// After all ports have been processed, remaining listeners are removed if they were created by this Service.
-			curListeners = popListener(curListeners, listener.ID)
 		}
 
 		// Deal with the remaining listeners, delete the listener if it was created by this Service previously.
@@ -2100,10 +2100,10 @@ func (lbaas *LbaasV2) ensureLoadBalancer(ctx context.Context, clusterName string
 	if err != nil {
 		return nil, fmt.Errorf("error getting LB %s listeners: %v", loadbalancer.Name, err)
 	}
-	curListenerMapping := make(map[listenerKey]*listeners.Listener)
+	curListenerMapping := make(map[listenerKey]listeners.Listener)
 	for i, l := range oldListeners {
 		key := listenerKey{Protocol: listeners.Protocol(l.Protocol), Port: l.ProtocolPort}
-		curListenerMapping[key] = &oldListeners[i]
+		curListenerMapping[key] = oldListeners[i]
 	}
 	for portIndex, port := range ports {
 		key := listenerKey{Protocol: listeners.Protocol(port.Protocol), Port: int(port.Port)}
@@ -2746,7 +2746,7 @@ func (lbaas *LbaasV2) updateOctaviaLoadBalancer(ctx context.Context, clusterName
 			return fmt.Errorf("loadbalancer %s does not contain required listener for port %d and protocol %s", loadbalancer.ID, port.Port, port.Protocol)
 		}
 
-		_, err := lbaas.ensureOctaviaPool(loadbalancer.ID, cutString(fmt.Sprintf("pool_%d_%s", portIndex, loadbalancer.Name)), &listener, service, port, nodes, svcConf)
+		_, err := lbaas.ensureOctaviaPool(loadbalancer.ID, cutString(fmt.Sprintf("pool_%d_%s", portIndex, loadbalancer.Name)), listener, service, port, nodes, svcConf)
 		if err != nil {
 			return err
 		}
diff --git a/pkg/util/openstack/loadbalancer.go b/pkg/util/openstack/loadbalancer.go
index 605a307..ddca8b9 100644
--- a/pkg/util/openstack/loadbalancer.go
+++ b/pkg/util/openstack/loadbalancer.go
@@ -314,18 +314,18 @@ func UpdateListener(client *gophercloud.ServiceClient, lbID string, listenerID s
 }
 
 // CreateListener creates a new listener
-func CreateListener(client *gophercloud.ServiceClient, lbID string, opts listeners.CreateOpts) (*listeners.Listener, error) {
+func CreateListener(client *gophercloud.ServiceClient, lbID string, opts listeners.CreateOpts) (listeners.Listener, error) {
 	mc := metrics.NewMetricContext("loadbalancer_listener", "create")
 	listener, err := listeners.Create(client, opts).Extract()
-		return nil, err
+		return listeners.Listener{}, err
 	}
 
 	if err := WaitLoadbalancerActive(client, lbID); err != nil {
-		return nil, fmt.Errorf("failed to wait for load balancer %s ACTIVE after creating listener: %v", lbID, err)
+		return listeners.Listener{}, fmt.Errorf("failed to wait for load balancer %s ACTIVE after creating listener: %v", lbID, err)
 	}
 
-	return listener, nil
+	return *listener, nil
 }
 
 // DeleteListener deletes a listener.
```

pkg/openstack/loadbalancer.go
lingxiankong: Hi @FlorinPeter, seems the changes hare is more reasonable for now, but could you please add a comment that describes the relations between curListenerMapping and curListeners, e.g. modifying the curListeners would also change the mapping.

FlorinPeter:
@lingxiankong I updated the comment, can you please override the openstack-cloud-controller-manager-e2e-conformance-test and bring this to an end?
I also checked the previous versions and only 1.23 is affected as far I can see.
Is there a planned patch release or how is the patches release cadence?

lingxiankong:
No idea why the occm pod failed to create (we need to get the pod logs when it fails), but given this change shouldn't affect the deployment, so I would override the job.

/override openstack-cloud-controller-manager-e2e-test

jichenjc:
/approve
/lgtm

a separate topic: suffering this error 
`/usr/local/bin/runner.sh: line 114: ./tests/ci-occm-e2e-conformance.sh: No such file or directory`
guess this PR suffering same per above?

jichenjc:
@lingxiankong the test failed, can you help to grant me owner so I can overwrite as well? Thanks

lingxiankong:

PR #1794, opened 2022-02-16
[manila-csi-plugin] added helper named templates for labels

gman0:
This PR moves various `labels` listings from individual Kubernetes component manifests into common named templates, into `_helpers.tpl`. This is inspired by (borrowed / stolen from :)) cinder-csi's chart that does this too, and I think is nicer than what manila-csi currently does.

No effective change in rendered YAML output, please see https://gist.github.com/gman0/5b711294cdbbbc7445f426bfd0b44c3f (only some key-value pairs are reordered but that has no real effect).

fixes #

**Release note**:

```release-note
NONE
```


jacksgt (approved):
Nice cleanup!

Issue #1722, opened 2021-12-27
[openstack-cloud-controller-manager] Serviceaccount cannot patch resource "services"


jichenjc:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
>
> /kind bug
> /kind feature

**What happened**:

create a LB follow 
https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md

```
kind: Service
apiVersion: v1
metadata:
  name: hello-world
  annotations:
    service.beta.kubernetes.io/brightbox-load-balancer-healthcheck-request: /
spec:
  type: LoadBalancer
  selector:
    app: hello-world
  ports:
    - name: http
      protocol: TCP
      port: 80
      targetPort: web
```

in logs seems 
```
E1226 09:11:24.522851       1 controller.go:307] error processing service example/hello-world (will retry): failed to ensure load balancer: failed to patch service object example/hello-world: services "hello-world" is forbidden: User "system:serviceaccount:kube-system:cloud-provider-openstack" cannot patch resource "services" in API group "" in the namespace "example"
I1226 09:11:24.523070       1 event.go:291] "Event occurred" object="example/hello-world" kind="Service" apiVersion="v1" type="Warning" reason="SyncLoadBalancerFailed" message="Error syncing load balancer: failed to ensure load balancer: failed to patch service object example/hello-world: services \"hello-world\" is forbidden: User \"system:serviceaccount:kube-system:cloud-provider-openstack\" cannot patch resource \"services\" in API group \"\" in the namespace \"example\""
```

it's weird why `"system:serviceaccount:kube-system:cloud-provider-openstack\"` is used

I created following command
```
kubectl create clusterrolebinding my-1 --clusterrole=system:cloud-controller-manager --serviceaccount=kube-system:cloud-provider-openstack
```

and it works fine in the LB after the clusterrolebinding created 
it's weird why `kube-system:cloud-provider-openstack` is created 
**What you expected to happen**:

**How to reproduce it**:

**Anything else we need to know?**:

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version:
- OpenStack version:
- Others:

jichenjc:
opening this to track the problem as this is first time I saw this kind of issue

jacksgt:
Hello,

[...]

PR #1802, opened 2022-02-28
[occm] Allow service status changes in ClusterRole

jacksgt:
This modification allows the Openstack Cloud Controller Manager to send `PATCH` requests specifically for the Status field of a Service (across all namespaces).
This is required when the OCCM uses it's own service account to modify the Service resource (i.e. with `--use-service-account-credentials=false`).

Without this permission I encountered the following error message when installing the Helm chart (from master) and using `--use-service-account-credentials=false`:

```
E0228 11:04:53.120584       1 controller.go:310] error processing service default/my-lb (will retry): failed to add load balancer cleanup finalizer: services "my-lb" is forbidden: User "system:serviceaccount:NAMESPACE:openstack-cloud-controller-manager" cannot patch resource "services/status" in API group "" in the namespace "default"
```

This was discussed in https://github.com/kubernetes/cloud-provider-openstack/issues/1722#issuecomment-1054037751

**Release note**:

```release-note
[occm] Add PATCH services/status permission for ServiceAccount in Helm chart.
```

### DIFF
--- a/charts/openstack-cloud-controller-manager/templates/clusterrole.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/clusterrole.yaml
@@ -43,6 +43,12 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - services/status
+  verbs:
+  - patch
+- apiGroups:
+  - ""
+  resources:
   - serviceaccounts/token
   verbs:
   - create
--- a/manifests/controller-manager/cloud-controller-manager-roles.yaml
+++ b/manifests/controller-manager/cloud-controller-manager-roles.yaml
@@ -45,6 +45,12 @@ items:
   - apiGroups:
     - ""
     resources:
+    - services/status
+    verbs:
+    - patch
+  - apiGroups:
+    - ""
+    resources:
     - serviceaccounts
     verbs:
     - create


jichenjc:
I lgtm but another thought is this mandatory for all configuration (true/false of use-service-account-credentials)  can we make it optional?

I am thinking that the manifests in manifests/controller-manager with # before (so someone can uncomment it as well?)

jacksgt:
Thanks for taking a look @jichenjc !
> I lgtm but another thought is this mandatory for all configuration (true/false of use-service-account-credentials) can we make it optional?

We could do this if the Helm chart had a separate value for the use-service-account-credentials options. However, at the moment it can only be set via the extraArgs array, so we cannot automatically detect if the value is set or not.
In any case, I don't think this additional permission is super critical compared to all the others in the cluster role.

> I am thinking that the manifests in manifests/controller-manager with # before (so someone can uncomment it as well?)

Good point, I have updated my commit.

jacksgt:
@jichenjc I updated the YAML manifests in manifests/controller-manager in accordance with the Helm chart changes.
Could you take a look again and approve? Thanks.

jacksgt:
@jichenjc This test case failed, but I don't think it has anything to do with this PR:
```
/usr/local/bin/runner.sh: line 114: ./tests/ci-occm-e2e-conformance.sh: No such file or directory

```

ramineni:
> @jichenjc This test case failed, but I don't think it has anything to do with this PR:
> 
> ```
> /usr/local/bin/runner.sh: line 114: ./tests/ci-occm-e2e-conformance.sh: No such file or directory
> ```

@jacksgt ya, it has nothing to do with this PR, new job is being added , so please ignore it.

jacksgt:
But someone still needs to merge this :-)

ramineni:

jacksgt:
Thanks everyone!

PR #1809, opened 2022-03-09
[OCCM] Add --cluster-name=$(CLUSTER_NAME) support

### LINKED ISSUE #1807: [occm][lb] Incorrect cluster name
/kind bug

**What happened**:

https://github.com/kubernetes/cloud-provider/blob/7f90ee552e3b169ccbdd131cce3ce3574f0afc8b/cloud.go#L160

The above spec says that the cluster name should come from the `kube-control-manager`

I have a cluster (via CAPO) which has `kube-control-manager` arg:

`      --cluster-name=istio-2`

However, when occm creates a LB, I see in the logs:

```console
I0304 14:29:55.159348       1 loadbalancer.go:1915] EnsureLoadBalancer(kubernetes, istio-system/istio-ingressgateway)
```
That is, the cluster name is `kubernetes` not `istio-2`

This causes an issue when you have multiple clusters and have same LB service (e.g. istio-ingressgateway) is some of the clusters. Only one LB is created and it is shared between the clusters. Obviously this does not work.

The cluster name comes what the `kube-control-manager` says it is. Each cluster gets it's own LB even if the LB service namespace/name is the same.

**How to reproduce it**:

Always. This was not an issue in 1.20

**Environment**:
1.22

jichenjc:

fixes #1807 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
@@ -37,6 +37,7 @@ spec:
             - /bin/openstack-cloud-controller-manager
             - --v={{ .Values.logVerbosityLevel }}
             - --cloud-config=$(CLOUD_CONFIG)
+            - --cluster-name=$(CLUSTER_NAME)
             - --cloud-provider=openstack
             - --use-service-account-credentials=true
             - --controllers={{- trimAll "," (include "occm.enabledControllers" . ) -}}
@@ -79,6 +80,8 @@ spec:
           env:
             - name: CLOUD_CONFIG
               value: /etc/config/cloud.conf
+            - name: CLUSTER_NAME
+              value: {{ .Values.cluster.name }}
       {{- if .Values.extraInitContainers }}
       initContainers: {{ toYaml .Values.extraInitContainers | nindent 6 }}
       {{- end }}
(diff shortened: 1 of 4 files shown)


charts/openstack-cloud-controller-manager/templates/daemonset.yaml
```diff
@@ -79,6 +80,10 @@ spec:
           env:
             - name: CLOUD_CONFIG
               value: /etc/config/cloud.conf
+            - name: CLUSTER_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
```
lingxiankong, line 86: I don't think using nodeName is a good idea, maybe just set a default value `kubernetes`?

ramineni:
Thanks.

/approve

powellchristoph:
Could we get a new Chart version? I've run into this exact problem.

jichenjc:
@lingxiankong @ramineni does this need a bump of OCCM version (e.g 1.23=>1.23.1 ) 
or a separate flow of release a new chart? I didn't do it personally before and I can do it if any doc we have

lingxiankong:
As we are still in the dev cycle, it's not possible to do a chart release given the code is not officially released, and this change is kind of "new feature", so it doesn't fit into the backport policy neither. We have to wait until the next release.

PR #1822, opened 2022-04-06
[occm] Get occm pod logs when CI job fails

lingxiankong:
Currently when the CI job fails, we can't see the pod logs, making it hard for troubleshooting.

This PR also fixes the image tag issue.

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/playbooks/roles/install-cpo-occm/defaults/main.yaml
+++ b/tests/playbooks/roles/install-cpo-occm/defaults/main.yaml
@@ -8,5 +8,5 @@ build_image: false
 run_e2e: false
 # Used for access the private registry image from k8s
 remote_registry_host: "{{ ansible_default_ipv4.address }}"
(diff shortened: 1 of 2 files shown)


lingxiankong:
> /test openstack-cloud-controller-manager-e2e-test

Found the issue, we were using PR number as the image tag which is not allowed any more after you fixed the version issue.

jichenjc:
>>>as the image tag which is not allowed any more after you fixed the version issue.

ok ,didn't expect such side effect
so we need additional fix to the version thing .. let me know anything I can help

lingxiankong:
> ok ,didn't expect such side effect
> so we need additional fix to the version thing .. let me know anything I can help

All good now, I added the fix in this PR.

ramineni:
Thanks.
/lgtm

PR #1828, opened 2022-04-13
cinder-csi-plugin: Use os-volume_attachments API to poll detach

### LINKED ISSUE #1645: [cinder-csi-plugin] Use Nova's os-volume_attachments API to poll attach/detach completion
/kind feature

**What happened**:

The current volume attach and detach code within the cinder-csi-plugin uses Nova's os-volume_attachments API [1] to initially attach [2] and later detach [3] a volume from an instance. As noted in [1] both of these APIs are async and need to be polled (with GETs against /servers/{server_id}/os-volume_attachments) after calling to determine when an operation has completed fully.

Instead in both situations cinder-csi appears to poll Cinder's volume API [4] checking the attachments listed there instead. As discussed upstream in OpenStack [5][6] there are some race conditions within Cinder at present that could allow cinder-csi to request that a volume be deleted *before* the original request to detach it had actually been completed. Cinder then incorrectly accepting the request to delete the volume and borking the original request to detach late in the flow.

Ultimately cinder-csi just needs to poll Nova's os-volume_attachments API to determine when an operation has completed.

[1] https://docs.openstack.org/api-ref/compute/#servers-with-volume-attachments-servers-os-volume-attachments
[2] https://github.com/kubernetes/cloud-provider-openstack/blob/61f2129345c6ab854ec464bc85e0b558f011b08e/pkg/csi/cinder/openstack/openstack_volumes.go#L164-L198
[3] https://github.com/kubernetes/cloud-provider-openstack/blob/61f2129345c6ab854ec464bc85e0b558f011b08e/pkg/csi/cinder/openstack/openstack_volumes.go#L259-L287
[4] https://github.com/kubernetes/cloud-provider-openstack/blob/61f2129345c6ab854ec464bc85e0b558f011b08e/pkg/csi/cinder/openstack/openstack_volumes.go#L370-L381
[5] https://review.opendev.org/c/openstack/cinder/+/801913
[6] https://bugs.launchpad.net/cinder/+bug/1937084

Ultimately cinder-csi just needs to poll Nova's os-volume_attachments API to determine when an operation has completed.

**How to reproduce it**:

I've marked this as a feature but honestly it could just be a bug so feel free to change.

Cause volumes to be attached or detached to an instance and note that Nova's os-volume_attachments API isn't used to poll the state of either operation.

**Environment**:
- OpenStack version: N/A (Nova's os-volume_attachments API has been around for all of v2)
- Others:

lyarwood:

As set out in the issue this is the correct OpenStack API to use when
polling volume attachment and detachment from an instance.

fixes #1645

I couldn't see any unit tests for this code, existing functional tests should hopefully already cover the use case.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/openstack/openstack_volumes.go
+++ b/pkg/csi/cinder/openstack/openstack_volumes.go
@@ -369,16 +369,26 @@ func (os *OpenStack) GetMaxVolLimit() int64 {
 
 // diskIsAttached queries if a volume is attached to a compute instance
 func (os *OpenStack) diskIsAttached(instanceID, volumeID string) (bool, error) {
-	volume, err := os.GetVolume(volumeID)
-	if err != nil {
-		return false, err
-	}
-	for _, att := range volume.Attachments {
-		if att.ServerID == instanceID {
-			return true, nil
+	foundAttachment := false
+	err := volumeattach.List(os.compute, instanceID).EachPage(func(page pagination.Page) (bool, error) {
+		attachments, err := volumeattach.ExtractVolumeAttachments(page)
+		if err != nil {
+			return false, err
 		}
+		for _, attachment := range attachments {
+			if attachment.VolumeID == volumeID && attachment.ServerID == instanceID {
+				klog.V(4).Infof("Found volume attachment %s for server %s and volume %s", attachment.ID, instanceID, volumeID)
+				foundAttachment = true
+				return false, nil
+			}
+		}
+		klog.V(4).Infof("Unable to find volume attachment for server %s and volume %s", instanceID, volumeID)
+		return false, nil
+	})
+	if err != nil {
+		return foundAttachment, err
 	}
-	return false, nil
+	return foundAttachment, nil
 }
 
 // diskIsUsed returns true a disk is attached to any node.


jichenjc:
looks like devstack has flaky problem

jichenjc:
@lyarwood looks like this time the test real executed and 2 failed cases , not sure it's related but let's give another try

jichenjc:
still some test case failed, I doubt the change trigger the error ...

jichenjc:
```
Apr 14 14:12:07.273: INFO: At 2022-04-14 14:07:03 +0000 UTC - event for pod-b24d99fc-971c-4e94-8df6-e8b27b7953a9: {kubelet k3s-master} FailedMount: Unable to attach or mount volumes: unmounted volumes=[volume1], unattached volumes=[volume1 kube-api-access-swxsd]: volume volume1 has volumeMode Block, but is specified in volumeMounts
Apr 14 14:12:07.273: INFO: At 2022-04-14 14:07:03 +0000 UTC - event for pod-b24d99fc-971c-4e94-8df6-e8b27b7953a9: {attachdetach-controller } FailedAttachVolume: AttachVolume.Attach failed for volume "pvc-7205016b-7c10-4c62-a438-dd4bb2b33eb3" : rpc error: code = Internal desc = [ControllerPublishVolume] failed to get device path of attached volume : can not get device path of volume pvc-7205016b-7c10-4c62-a438-dd4bb2b33eb3, its status is reserved 
Apr 14 14:12:07.273: INFO: At 2022-04-14 14:07:04 +0000 UTC - event for pod-b24d99fc-971c-4e94-8df6-e8b27b7953a9: {attachdetach-controller } FailedAttachVolume: AttachVolume.Attach failed for volume "pvc-7205016b-7c10-4c62-a438-dd4bb2b33eb3" : rpc error: code = Internal desc = [ControllerPublishVolume] Attach Volume failed with error failed to attach 329c9594-e71b-4de5-ac8b-2a4859886356 volume to c54ae2b9-1de2-451c-b9b3-3a8925281d3c compute: Bad request with: [POST http://10.0.2.15/compute/v2.1/servers/c54ae2b9-1de2-451c-b9b3-3a8925281d3c/os-volume_attachments], error message: {"badRequest": {"code": 400, "message": "Invalid volume: volume 329c9594-e71b-4de5-ac8b-2a4859886356 is already attached to instances: c54ae2b9-1de2-451c-b9b3-3a8925281d3c"}}
```

looks like this is the root cause , but not sure why it's happening

lyarwood:
@jichenjc any idea where I could find the broken down logs or do we only get a single test log with prow?

jichenjc:
@lyarwood we were supposed to have logs if run to error or success

https://github.com/kubernetes/cloud-provider-openstack/blob/master/tests/playbooks/roles/install-csi-cinder/tasks/main.yaml#L151

in your case I don't know why it stuck for 3 hours then quit without any info ,need check more about it ..

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -378,16 +378,26 @@ func (os *OpenStack) GetMaxVolLimit() int64 {
...
 // diskIsAttached queries if a volume is attached to a compute instance
 func (os *OpenStack) diskIsAttached(instanceID, volumeID string) (bool, error) {
-	volume, err := os.GetVolume(volumeID)
-	if err != nil {
-		return false, err
-	}
-	for _, att := range volume.Attachments {
-		if att.ServerID == instanceID {
-			return true, nil
+	var foundAttachment bool = true
```
chrigl, line 381: Is this the correct default? If I understand the code correct, this way ` foundAttachment` would be always `true`, or did I miss something?

lyarwood: Yeah apologies I think I confused this with the behaviour of EachPage below for some reason. I'll update and try again now.

drencrom:
I tested this code but does not work for me. If I try to unattach a volume after this version of diskIsAttached returns true I get this error: 
`Invalid volume: Invalid input received: Invalid volume: Unable to detach volume. Volume status must be 'in-use' and attach_status must be 'attached' to detach`

jichenjc:
> I tested this code but does not work for me. If I try to unattach a volume after this version of diskIsAttached returns true I get this error: `Invalid volume: Invalid input received: Invalid volume: Unable to detach volume. Volume status must be 'in-use' and attach_status must be 'attached' to detach`

not sure this code is still under work now.. maybe you can provide your fix in case you can? Thanks @drencrom

PR #1829, opened 2022-04-13
[occm] Remove finalizer from Service

MaysaMacedo:
When the load-balancer creation is not handled
by the CCM, the Service still gets a finalizer
named "service.kubernetes.io/load-balancer-cleanup"
added to it. The finalizer existence is blocking
Services from being deleted as the finalizer is not
cleaned up when an exception is raised of load-balancer
not found. This commit fixes the issue by ensuring not
found exceptions are ignored and allowing the
finalizer to be removed.

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -359,7 +359,7 @@ func getLoadbalancerByName(client *gophercloud.ServiceClient, name string, legac
 		Name: name,
 	}
 	allLoadbalancers, err := openstackutil.GetLoadBalancers(client, opts)
-	if err != nil {
+	if err != nil && !cpoerrors.IsNotFound(err) {
 		return nil, err
 	}
 
@@ -370,7 +370,7 @@ func getLoadbalancerByName(client *gophercloud.ServiceClient, name string, legac
 				Name: legacyName,
 			}
 			allLoadbalancers, err = openstackutil.GetLoadBalancers(client, opts)
-			if err != nil {
+			if err != nil && !cpoerrors.IsNotFound(err) {
 				return nil, err
 			}
 		} else {


jichenjc:
openstack-cloud-controller-manager-e2e-conformance-test is irrelevant, ignore it for now

I am requesting to be granted to override the failure , will do after that

mdbooth (commented):
Did you work out how GetLoadBalancers() is able to return IsNotFound?

mdbooth:
I think the issue can be described as:

occm will always add a finalizer to a Service object, even when `enabled = False` is set in `[LoadBalancers]`. This behaviour is defined in `cloud-provider`, and it is not desirable to change it.

When the service object is deleted, `GetLoadBalancers()` unexpectedly returns an `IsNotFound` error rather than the expected empty list. This causes the finalizer to fail, and therefore the finalizer is never removed from the Service object, which therefore can't be deleted.

---

I'm eager to understand how GetLoadBalancers() can return IsNotFound, because it's not at all obvious how that is possible.

MaysaMacedo:
> Did you work out how GetLoadBalancers() is able to return IsNotFound?

@mdbooth Yes, I double checked that this exception was raised in the CCM code and I Octavia logs had the not found.

```
2022-04-13 22:09:39.497 25 DEBUG octavia.api.v2.controllers.base [req-911a7958-2b41-4ebc-a2df-2b5f1f3740c2 - a64676dfa4b24cc9adfb620fef7b6506 - default default] Load Balancer 7ef868b6-80eb-440f-9f01-01bebf7e173e not found _get_db_obj /usr/lib/python3.6/site-packages/octavia/api/v2/controllers/base.py:71
2022-04-13 22:09:39.497 25 DEBUG wsme.api [req-911a7958-2b41-4ebc-a2df-2b5f1f3740c2 - a64676dfa4b24cc9adfb620fef7b6506 - default default] Client-side error: Load Balancer 7ef868b6-80eb-440f-9f01-01bebf7e173e not found. format_exception /usr/lib/python3.6/site-packages/wsme/api.py:222
```

mdbooth:
To summarise an offline discussion we had about this:

The error is raised because GetLoadBalancers() is trying to hit the neutron lbaas endpoint, which no longer exists. Therefore the 404 is because the endpoint doesn't exist, not because the resource doesn't exist.

GetLoadBalancers() is trying to hit the neutron endpoint instead of the Octavia endpoint because we are explicitly setting `use-octavia = False` in cloud.conf. This was how we disabled the service handling behaviour of KCM using the legacy cloud provider. We recognised at the time that it was 'a bit of a hack', which is why @dulek added the `enabled` option in the first place.

---

I think there are a few issues here:

Firstly I don't believe we should ignore the 404 from GetLoadBalancers(): it is reporting a missing endpoint rather than a missing resource, so it would be misleading at best to ignore it here. I think this is a genuine failure which should cause the finalizer to exit unsuccessfully. Additionally, for the default configuration where load balancer handling is enabled, ignoring this 404 would erroneously cause the finalizer to succeed without cleaning up the load balancer during a temporary outage of the Octavia endpoint.

Ideally we would not call this code at all. I understand that we deliberately decided to call the finalizer even when `enabled = False` because we want to handle the case of load balancers which were created by a previous configuration before it was disabled. I wonder if there is any reliable (including *historically* reliable) way to determine from the Service object if we previously created an LB. e.g. Did we add an annotation or update something in the Status?

As it stands, the use-octavia flag currently *must* reflect the underlying cloud. Therefore setting `use-octavia = false` on a cloud which is new enough to no longer have the neutron lbaas endoint is not correct.

Additionally, given that neutron lbaas has now gone beyond deprecation to removal, it might be worth deprecating this option and documenting that Octavia is required?

jichenjc:
>>>Firstly I don't believe we should ignore the 404 from GetLoadBalancers(): it is reporting a missing endpoint rather than a missing resource, so it would be misleading at best to ignore it here. I think this is a genuine failure which should cause the finalizer to exit unsuccessfully.

https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md

we stated: 
```NOTE: Now, the openstack-cloud-controller-manager implementation is based on OpenStack Octavia, Neutron-LBaaS has been deprecated in OpenStack since Queens release and no longer maintained in openstack-cloud-controller-manager. So make sure to use Octavia if upgrade to the latest openstack-cloud-controller-manager docker image.```

so I doubt whether this is a valid use case for latest OCCM

mdbooth:
> > > > Firstly I don't believe we should ignore the 404 from GetLoadBalancers(): it is reporting a missing endpoint rather than a missing resource, so it would be misleading at best to ignore it here. I think this is a genuine failure which should cause the finalizer to exit unsuccessfully.
> 
> https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
> 
> we stated: `NOTE: Now, the openstack-cloud-controller-manager implementation is based on OpenStack Octavia, Neutron-LBaaS has been deprecated in OpenStack since Queens release and no longer maintained in openstack-cloud-controller-manager. So make sure to use Octavia if upgrade to the latest openstack-cloud-controller-manager docker image.`
> 
> so I doubt whether we should do any change in the OCCM side other than ask for upgrade from lbaas to octavia now

Interestingly we're not really trying to use neutron lbaas here! Setting `use-octavia=false` was simply a useful way to disable the legacy cloud provider's Service controller when we were managing Services with an external controller (in this case, Kuryr). However, we now have `enabled=false` which is explicit, so `use-octavia=false` should be redundant and we can stop setting it.

We do still have a problem of trying to finalize Services which we didn't create. We can probably ignore that for now, though, as long as we aren't hitting this 404 error.

MaysaMacedo:
I'm closing this fix since `use-octavia` shouldn't be set as false.

PR #1838, opened 2022-04-18
[cinder-csi-plugin][charts] add metadata.namespace

### LINKED ISSUE #1837: [cinder-csi-plugin] namespaced resource template should has metadata.namespace explicitly

/kind bug

**What happened**:

daemonsets and deployments are represented without metadata.namespace.

-   daemonsets
-   deployments

are namespaced resource.

so `nodeplugin-daemonset` and `controllerplugin-deployment` resource should represent metadata.namespace if given.

**How to reproduce it**:

    $ helm template cinder-csi cpo/openstack-cinder-csi -n kube-system

<https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-uris>

    $ kubectl api-resources | egrep true | egrep 'daemonset|deployments'
    daemonsets                        ds           apps/v1                                    true         DaemonSet
    deployments                       deploy       apps/v1                                    true         Deployment

**Environment**:
- OpenStack version:
- Others:

cinder-csi-plugin chart version v2.0.0

aanoaa:
cause Deployment and DaemonSet are namespaced resource

fixes #1837 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
+++ b/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
@@ -2,6 +2,7 @@ kind: Deployment
 apiVersion: apps/v1
 metadata:
   name: {{ include "cinder-csi.name" . }}-controllerplugin
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "cinder-csi.controllerplugin.labels" . | nindent 4 }}
 spec:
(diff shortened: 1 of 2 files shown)


jichenjc:
I searched https://github.com/helm/helm/issues/5465 and found seems omit the ns is best practice ..

so can you help check whether above comment make sense and use helm --namespace can solve you problem?

aanoaa:
In my case, I'm not using `kubectl` or `helm` cli.

Instead, create raw manifests through helm pkg(go) and perform server side apply through dynamic client of client-go.
At this time, metadata.namespace is empty in namespaced resources such as deployments and fails.

jichenjc:
ok, seems reasonable use case to me 
I am not helm expert and I think this change not blocking --namespace usage ,so seems reasonable unless someone can comment

PR #1843, opened 2022-04-22
[cinder-csi-plugin][charts] Add volumeMount for the cloud-config secret or hostPath volume

tcassaert:

Currently a volume is created for the cloud-config secret or hostPath volume, but it is never mounted into the pod. This PR adds a mount for this volume.

**Release note**:

```release-note
NONE
```


tcassaert:
I made a mistake in my usage of the chart. This pull request is invalid. Sorry for the inconvenience.

PR #1854, opened 2022-05-12
Fix for Authentication Dependencies for Application Credentials

### LINKED ISSUE #1857: Authentication Dependencies for Application Credentials are missing

> /kind bug

**What happened**:

It is not possible to use the the authentication with application credentials only.

That I can configure the authentication with application credentials and that the configuration is satisfied.

**How to reproduce it**:

Try to configure authentication with authentication credentials

The Application Credentials are missing in the `depdendsOn` in the `Auth` configuration.

There is a already a PR https://github.com/kubernetes/cloud-provider-openstack/pull/1854 for this

shaardie:
This patch extend the authentication configuration by the possibility to use
application credentials.

fix_auth_dependencies_for_application_credentials

Signed-off-by: Sven Haardiek <sven.haardiek@uni-muenster.de>

This patch extend the authentication configuration by the possibility to use
application credentials.

Without this PR the dependencies are not properly satisfied, if application credentials are used.

fixes #1857

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -38,17 +38,17 @@ import (
 )
 
 type AuthOpts struct {
-	AuthURL          string                   `gcfg:"auth-url" mapstructure:"auth-url" name:"os-authURL" dependsOn:"os-password|os-trustID|os-applicationCredentialSecret|os-clientCertPath"`
+	AuthURL          string                   `gcfg:"auth-url" mapstructure:"auth-url" name:"os-authURL" dependsOn:"os-password|os-trustID|os-applicationCredentialSecret|os-applicationCredentialID|os-applicationCredentialName|os-clientCertPath"`
 	UserID           string                   `gcfg:"user-id" mapstructure:"user-id" name:"os-userID" value:"optional" dependsOn:"os-password"`
 	Username         string                   `name:"os-userName" value:"optional" dependsOn:"os-password"`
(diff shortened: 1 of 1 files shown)


jichenjc:
can you help create an issue so that we can track the purpose of change etc in the issue? Thanks

shaardie:
> can you help create an issue so that we can track the purpose of change etc in the issue? Thanks

I created an issue and referenced it in the PR Text.

pkg/client/client.go
```diff
@@ -38,17 +38,17 @@ import (
 )
 
 type AuthOpts struct {
-	AuthURL          string                   `gcfg:"auth-url" mapstructure:"auth-url" name:"os-authURL" dependsOn:"os-password|os-trustID|os-applicationCredentialSecret|os-clientCertPath"`
+	AuthURL          string                   `gcfg:"auth-url" mapstructure:"auth-url" name:"os-authURL" dependsOn:"os-password|os-trustID|os-applicationCredentialSecret|os-applicationCredentialName|os-clientCertPath"`
```
chrigl, line 41: Is there a need to include `applicationCredentialID` here as well?

shaardie: Yes, you are correct. I updated the PR.

shaardie:
Seems like there is a authentication problem in the CI, I can not really do anything about.

PR #1862, opened 2022-05-14
[e2e] Factor out module test to simplify tooling

### LINKED ISSUE #1861: [e2e] Split out module tests/e2e

/kind cleanup

**What happened**:

`go.mod`  is pretty crowded, mostly due to the fact that e2e tests depend on `k8s.io/kubernetes`. This is not just being crowded, but also makes linting and unit-testing more difficult than necessary. e.g. a plain `go test` starts e2e tests, making development in IDEs hard.  

* `go test` just runs unit tests
* `go.mod` only contains "real" dependencies
* `golangci-lint` does not lint e2e tests by default
* seamless usage of go tool and IDEs

**How to reproduce it**:

[cloud-provider-vsphere](https://github.com/kubernetes/cloud-provider-vsphere/tree/master/test/e2e) and
[cloud-provider-aws](https://github.com/kubernetes/cloud-provider-aws/tree/master/tests/e2e) also factored out an e2e module.

**Environment**:
- OpenStack version:
- Others:

chrigl:

There are two "new" modules tests/e2e and tests/sanity. With them
separated, the main go.mod is much cleaner and only contains
dependencies for the "real" binaries. This makes maintenance a bit
easier.

Because with the separation, tools work much more staightforward,
withouth the need to call out for make, which in turn makes development
using a IDE just that bit simpler. e.g. go test ./... now just works.

This commit also removes the kubernetes feature gates from occm, since
they are not in use, and don't make much sense.

fixes #1861 

**Release note**:

```release-note
NONE
```


chrigl:
/hold
/milestone v1.25

Let's first cut `release-1.24` !?

jichenjc:
yes, we should have our 1.24 release first

tests/playbooks/roles/install-csi-cinder/tasks/main.yaml
```diff
@@ -142,7 +142,7 @@
       set -e
       set -o pipefail
 
-      cd $GOPATH/src/k8s.io/cloud-provider-openstack
+      cd $GOPATH/src/k8s.io/cloud-provider-openstack/tests/e2e
       mkdir /var/log/csi-pod
       go test -v ./cmd/tests/cinder-csi-e2e-suite/cinder_csi_e2e_suite_test.go -ginkgo.v -ginkgo.progress -ginkgo.skip="\[Disruptive\]" -ginkgo.focus="\[cinder-csi-e2e\]" -ginkgo.noColor -timeout=0 -report-dir="/var/log/csi-pod" | tee "/var/log/csi-pod/cinder-csi-e2e.log"
```
jichenjc, line 147: will the folder need update ?

chrigl: Yes, because the go tools expect commands like build and  run to be executed within a module.

I'm also thinking about moving to use upstream kubernetes e2e.test to replace our binary when I find some time. It is configurable, so that the same set of tests will run.

Issue #1868, opened 2022-05-18
[cinder-csi-plugin] testsuite failing since update to v1.24


chrigl:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind bug

**What happened**:

Test suite for cinder-csi fails because of missing files.

See [pr-logs](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/cloud-provider-openstack/1865/openstack-cloud-csi-cinder-e2e-test/1526759486381887488)
```
{ Failure /root/pkg/mod/k8s.io/kubernetes@v1.24.0/test/e2e/storage/testsuites/provisioning.go:246
May 18 03:53:34.256: reading manifest file: Test file "test/e2e/testing-manifests/storage-csi/any-volume-datasource/crd/populator.storage.k8s.io_volumepopulators.yaml" was not found.
Test files are expected in "../../../../manifests/cinder-csi-plugin" = "/root/src/k8s.io/manifests/cinder-csi-plugin".

/root/pkg/mod/k8s.io/kubernetes@v1.24.0/test/e2e/storage/utils/create.go:58}
```

**What you expected to happen**:

Test passes

**How to reproduce it**:

run
```
go test \
  -v ./cmd/tests/cinder-csi-e2e-suite/cinder_csi_e2e_suite_test.go \
  -ginkgo.v -ginkgo.progress \
  -ginkgo.skip="\[Disruptive\]" \
  -ginkgo.focus="\[Driver:\s+cinder.csi.openstack.org\]\s+\[Testpattern:\s+Dynamic\s+PV\s+\(default\s+fs\)\]\s+provisioning\s+should\s+mount\s+multiple\s+PV\s+pointing\s+to\s+the\s+same\s+storage\s+on\s+the\s+same\s+node" \
  -ginkgo.noColor -timeout=0 -report-dir="/var/log/csi-pod"
```

**Anything else we need to know?**:

The needed test files are located in [k/k and are new in v1.24](https://github.com/kubernetes/kubernetes/tree/release-1.24/test/e2e/testing-manifests/storage-csi/any-volume-datasource).

Kubernetes releases provide a binary `e2e.test`, that embed all needed testfiles for a particular release. So my idea is to switch to that binary rather than running and maintaining our own ones.

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version:
- OpenStack version:
- Others:

PR #1894, opened 2022-05-26
Improve manila csi driver chart adding fsGroupPolicy

### LINKED ISSUE #1893: [manila-csi-plugin] Cannot customize csi driver options

/kind feature

**What happened**:
I can't enable fsGroup policy for CSIDriver

Features supported by Kubernetes CSI are available to customize CSIDriver deployed using helm template defined in https://github.com/kubernetes/cloud-provider-openstack/blob/master/charts/manila-csi-plugin/templates/csidriver.yaml

**How to reproduce it**:
Deploy CSIDriver using template https://github.com/kubernetes/cloud-provider-openstack/blob/master/charts/manila-csi-plugin/templates/csidriver.yaml

There is no option to configure fsGroup policy https://kubernetes-csi.github.io/docs/support-fsgroup.html#supported-modes

I have tried to modify CSIDriver manually and fsGroup policy feature works fine but there is no way to configure it to be deployed using templates.

As I have tested it manually and it works, I would say it is possible to modify helm template to allow enabling options defined by Kubernetes CSI.

If you agree, I can create the PR to allow customising csidriver.

**Environment**:
- OpenStack version: N/A
- Others: N/A

fmount:
The fsGroupPolicy option can now be defined per protocol,
enforcing a different behavior to the driver [1]

[1] https://kubernetes-csi.github.io/docs/support-fsgroup.html

Signed-off-by: Francesco Pantano <fpantano@redhat.com>

fixes #1893

**Release note**:

```release-note
Set fsGroupPolicy (None by default) for the CSIDriver object
```

### DIFF
--- a/charts/manila-csi-plugin/templates/csidriver.yaml
+++ b/charts/manila-csi-plugin/templates/csidriver.yaml
@@ -6,5 +6,6 @@ metadata:
 spec:
   attachRequired: false
   podInfoOnMount: false
+  fsGroupPolicy: {{ printf "%s" .fsGroupPolicy }}
 ---
 {{- end }}
--- a/charts/manila-csi-plugin/values.yaml
+++ b/charts/manila-csi-plugin/values.yaml
@@ -7,6 +7,7 @@ driverName: manila.csi.openstack.org
 # Enabled Manila share protocols
 shareProtocols:
   - protocolSelector: NFS
+    fsGroupPolicy: None
     fwdNodePluginEndpoint:
       dir: /var/lib/kubelet/plugins/csi-nfsplugin
       sockFile: csi.sock
--- a/tests/playbooks/roles/install-csi-manila/tasks/main.yaml
+++ b/tests/playbooks/roles/install-csi-manila/tasks/main.yaml
@@ -126,6 +126,7 @@
           tag: {{ github_pr }}
       shareProtocols:
         - protocolSelector: NFS
+          fsGroupPolicy: None
           fwdNodePluginEndpoint:
             dir: /var/lib/kubelet/plugins/csi-nfsplugin
             sockFile: csi.sock


ramineni:
@fmount this is enabled just in charts , does the driver take care of enabling in the code?

gman0:
@ramineni

> @fmount this is enabled just in charts , does the driver take care of enabling in the code?

Is there anything to be enabled in the driver code? AFAIK kubelet does all the work for us, it's just that the CSIDriver object needs to specify what policy is acceptable for the underlying storage system.

fmount:
> @ramineni
> 
> > @fmount this is enabled just in charts , does the driver take care of enabling in the code?
> 
> Is there anything to be enabled in the driver code? AFAIK kubelet does all the work for us, it's just that the CSIDriver object needs to specify what policy is acceptable for the underlying storage system.
Yeah, in my understanding this is enough and kubelet takes care about processing the policy according to the CSIDriver object and apply the corresponding behavior [1].
We should be good here!

[1] https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/csi/csi_mounter.go

gouthampacha:
/retest

failure seems unrelated; Missing a manifest related to a new e2e test:

https://github.com/kubernetes/kubernetes/commit/cd634b59ac69145860712d56fbb976dedb940b1e
https://github.com/kubernetes/kubernetes/blob/v1.24.0/test/e2e/testing-manifests/storage-csi/any-volume-datasource/crd/populator.storage.k8s.io_volumepopulators.yaml

fmount:
> The issue repeats. This is the same as the issue that was affecting the cinder-csi e2e tests: #1868

Hey @gouthampacha thanks for mentioning that issue, I'll try to move manila to the new approach as well and update the PR accordingly.

gouthampacha:
/lgtm

@fmount thank you for this change, if you make another patch, the only thing i'd clarify is the release note:

`Set fsGroupPolicy (None by default) for the Manila CSIDriver object`

gman0:
/lgtm
/approve

Thank you, @fmount!

PR #1882, opened 2022-05-24
Remove neutron lbaas v1 version code

jichenjc:
**What this PR does / why we need it**:

**Which issue this PR fixes(if applicable)**:
fixes #1833

**Special notes for reviewers**:

**Release note**:

```release-note
neutron lbaas is not supported anymore, we should not keep those data, removed all neutron lbaas v1 code along with the UseOctavia flag, action required.
```

PR #1901, opened 2022-05-31
[occm] Add initial support for dual-stack k8s services

### LINKED ISSUE #1897: [OCCM] Support IPv4/IPv6 dual-stack services
/kind feature

**What happened**:
The OCCM does not support dual-stack (IPv4 + IPv6) clusters very well. This is a list of the existing issues:
- IPv4 or IPv6 VIP should be allocated based on the preference defined in service's `.spec.ipFamilies`
- for dual-stack services, both IPv4 and IPv6 load-balancers should be created
- floating IP should not be allocated for IPv6 VIP
- for IPv6 LB, the default `spec.loadBalancerSourceRanges` should be `::/0`
- LB member IP should be IPv4 / IPv6 based on the address family of the VIP
- security group rules should have etherType set based on the LB member address family
- OCCM should support LBs in different subnets (as there will be always 1 IPv4 and 1 IPv6 subnet at least)

In a dual-stack k8s cluster, users can create both IPv4 and IPv6 LoadBalancer services and dual-stack LoadBalancer services.

rastislavs:
Signed-off-by: Rastislav Szabo <rastislav@kubermatic.com>

Adds initial support for [IPv4 / IPv6 dual-stack](https://kubernetes.io/docs/concepts/services-networking/dual-stack/) kubernetes services. The aim was to create a minimal change that would fix as much as possible without the need for a larger code refactor. Therefore, with this change, OCCM is able to create either IPv4 or IPv6 load-balancer in a dual-stack clusters, based on the preference indicated in service's `spec.ipFamilies`:

 - If `IPv4` is the first, it would create an `IPv4` load-balancer.
 - If `IPv6` is the first, it would create an `IPv6` load-balancer.
 - This PR does not address creating 2 load-balancers if both address families are specified in `spec.ipFamilies` as that would require a larger code change.

fixes #1897 (partially, a new issue for full dual-stack support with 2 LBs in case of 2 `spec.ipFamilies` to be created after this is merged)

- The change should be backward compatible and should not affect single-stack clusters.
- Only the code path for Octavia LB has been covered, legacy code path does not pass any `preferredIPFamily` to the methods shared with the Octavia code path - and should not be affected.
- The globally shared `lbaas.opts.SubnetID`  should not be used to store subnet ID in the Octavia code path. With dual-stack, clusters have at last 2 subnets (one IPv4 and one IPv6), and therefore storing it in a globally shared variable does not work.

**Release note**:

```release-note
[occm] Add initial support for dual-stack k8s services.
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -48,6 +48,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/klog/v2"
+	netutils "k8s.io/utils/net"
 
 	"k8s.io/cloud-provider-openstack/pkg/metrics"
 	cpoutil "k8s.io/cloud-provider-openstack/pkg/util"
@@ -59,10 +60,11 @@ import (
 // Note: when creating a new Loadbalancer (VM), it can take some time before it is ready for use,
 // this timeout is used for waiting until the Loadbalancer provisioning status goes to ACTIVE state.
 const (
-	servicePrefix                   = "kube_service_"
-	defaultLoadBalancerSourceRanges = "0.0.0.0/0"
-	activeStatus                    = "ACTIVE"
-	annotationXForwardedFor         = "X-Forwarded-For"
+	servicePrefix                       = "kube_service_"
+	defaultLoadBalancerSourceRangesIPv4 = "0.0.0.0/0"
+	defaultLoadBalancerSourceRangesIPv6 = "::/0"
+	activeStatus                        = "ACTIVE"
+	annotationXForwardedFor             = "X-Forwarded-For"
 
 	ServiceAnnotationLoadBalancerInternal             = "service.beta.kubernetes.io/openstack-internal-load-balancer"
 	ServiceAnnotationLoadBalancerConnLimit            = "loadbalancer.openstack.org/connection-limit"
@@ -346,6 +348,7 @@ type serviceConfig struct {
 	healthMonitorDelay      int
 	healthMonitorTimeout    int
 	healthMonitorMaxRetries int
+	preferredIPFamily       corev1.IPFamily
 }
 
 type listenerKey struct {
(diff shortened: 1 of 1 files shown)


jichenjc:
THanks for the PR, really helpful feature~

minor question, we should have some kind of CI cases in follow up PR so we will make sure no break by other PRs later on?

pkg/openstack/loadbalancer.go
```diff
@@ -1549,6 +1567,10 @@ func (lbaas *LbaasV2) checkService(service *corev1.Service, nodes []*corev1.Node
 		return fmt.Errorf("no service ports provided")
 	}
 
+	if len(service.Spec.IPFamilies) > 0 {
```
jichenjc, line 1570: I assume this should doc somewhere as it's a logic that embeded in the code
first service IP family will be used?

pkg/openstack/loadbalancer.go
```diff
@@ -1932,7 +1956,7 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 	}
 
 	if lbaas.opts.ManageSecurityGroups {
-		err := lbaas.ensureSecurityGroup(clusterName, service, nodes, loadbalancer)
+		err := lbaas.ensureSecurityGroup(clusterName, service, nodes, loadbalancer, svcConf.preferredIPFamily, svcConf.lbMemberSubnetID)
```
jichenjc, line 1959: never used IPv6 + sec group , assume it works , though

pkg/openstack/loadbalancer.go
rastislavs: Agree, let me add some comments and maybe also a dual-stack section into `docs/openstack-cloud-controller-manager)/expose-applications-using-loadbalancer-type-service.md`

rastislavs:
> THanks for the PR, really helpful feature~
> 
> minor question, we should have some kind of CI cases in follow up PR so we will make sure no break by other PRs later on?

Yes, definitely. I think I will create 2 new issues for follow-up:
 - CI task
 - dual load-balancer support

pkg/openstack/loadbalancer.go
rastislavs: done

jichenjc:
/lgtm

@chrigl @zetaab maybe you can give some insight? THanks

chrigl:
I love it!

I will give it a try on my OpenStack over the weekend. From my point of view we can extend occm testing until v1.25. At the moment we lack some testing due to the openlab -> prow migration.

This PR will conflict #1882 , however, will not be great of a problem.

chrigl:
Works here as well. v4 is also still working

@lingxiankong Do you have any concerns? Maybe something I missed?

chrigl:
Works here as well. v4 is also still working

@lingxiankong Do you have any concerns? Maybe something I missed?

rastislavs:
Hey @chrigl / @jichenjc / @zetaab is there anything I can do to move this forward?

PR #1904, opened 2022-06-01
[cinder-csi-plugin] Remove downstream from probe

### LINKED ISSUE #1903: [cinder-csi-plugin] spam against OpenStack APIs increases heavily if using livenessProbes

/kind bug

**What happened**:

we are seeing HEAVY increasement in cinder API calls after 1.24.0.

With 1.23.0:
csi-cinder-controllerplugin made GET https://xxx:13776/ (cinder api in our case) calls 1/minute
csi-cinder-nodeplugin made GET https://xxx:13776/ (cinder api in our case) calls 0/minute

With 1.24.0:
csi-cinder-controllerplugin makes GET https://xxx:13776/ (cinder api in our case) calls 1/minute
csi-cinder-nodeplugin makes GET https://xxx:13776/ (cinder api in our case) calls 6/minute (once every 10 seconds)

Screenshot 2022-06-01 at 15 03 01

Here is screenshot from OpenStack loadbalancer statistics. We started updating our clusters to 1.24.0 15 May (as you can see from the graphs). 

Image Pasted at 2022-6-1 14-57

I expect that csi-cinder-nodeplugin does not make GET call to cinder API root path every 10 seconds.

**How to reproduce it**:

Start 1.24.0 csi-cinder-nodeplugin with -v=6 and you can see the behaviour.

this happens every 10 seconds to each pod in each node:

```
I0601 11:52:31.267565       1 utils.go:88] [ID:12] GRPC call: /csi.v1.Identity/Probe
I0601 11:52:31.267580       1 utils.go:89] [ID:12] GRPC request: {}
I0601 11:52:31.267695       1 openstack_volumes.go:54] OpenStack Request URL: GET https://xxx:13776/
```

**Environment**:
- OpenStack version:
- Others:

chrigl:

Remove all sorts of downstream dependency on cinder from probe. Probe is
the one called by the liveness sidecar to determine the health of the
csi plugin. Including downstream services in a health check can result
in unfortunate behavior in cluster. For example, over and over restarts
of daemonsets/deployments just because the downstream api is not
available at the moment... e.g. because of a maintenance.

On the other hand, just because the downstream api is available doesn't
tell us anything about whether or not the downstream works as expected.
Therefore there are facilities in place that reports errors when they
happen all over the place.

fixes #1903

**Release note**:

```release-note
[cinder-csi-plugin] Remove dependency on openstack api from csi Probes
```

### DIFF
--- a/pkg/csi/cinder/identityserver.go
+++ b/pkg/csi/cinder/identityserver.go
@@ -21,7 +21,6 @@ import (
 	"golang.org/x/net/context"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-	"k8s.io/cloud-provider-openstack/pkg/csi/cinder/openstack"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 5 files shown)


zetaab (approved):
/approve

ping @jichenjc what you think?

jichenjc:
@chrigl @zetaab we can overwrite the CI failure as the PR less likely cause issue and don't know when we can fix the gate ..

PR #1907, opened 2022-06-02
[ci] Pin k3s to v1.23.6+k3s1

### LINKED ISSUE #1902: cinder-csi-plugin: gate failure

>
> /kind bug
> /kind feature

**What happened**:

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/cloud-provider-openstack/1898/openstack-cloud-csi-cinder-e2e-test/1531482333746565120/build-log.txt

A pattern 

```
May 31 05:09:28.735: FAIL: "echo PymaVGuuyCbDj6LBtMWeRPL9dgJlDiFHDYT1ZTjD0Gylfb7tjBUgd092PaCn0zgDK3WH0mRVGyw14xAabOaCDw== | base64 -d | dd of=/mnt/volume1/file1.txt  bs=64 count=1" should succeed, but failed with error message "error dialing backend: EOF"
```

looks like this kind of error is random , as some pod with such POST action success ,so not sure what's the real problem here

**How to reproduce it**:

**Environment**:
- OpenStack version:
- Others:

chrigl:

This PR pinns k3s to v1.23.6+k3s1. We stick with that as long as we
can't reliably say that newer versions work well enough for us.
In the future, we must change code for k3s-v1.24.0 anywayse, because
docker is not supported there anyways. Until we change away from docker
we can't update to v1.24.

We must have a fixed version for now, to unblock our gates.

fixes #1902

I deliberately removed the guessing of the correct k3s release, because
any change in dependencies might cause trouble in our gating.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/playbooks/roles/install-k3s/defaults/main.yaml
+++ b/tests/playbooks/roles/install-k3s/defaults/main.yaml
@@ -1,5 +1,5 @@
 ---
-k8s_branch: "master"
+k3s_release: "v1.23.6+k3s1"
 worker_node_count: 1
 cluster_token: "9a08jv.c0izixklcxtmnze7"
 devstack_workdir: "{{ ansible_user_dir }}/devstack"
--- a/tests/playbooks/roles/install-k3s/tasks/main.yaml
+++ b/tests/playbooks/roles/install-k3s/tasks/main.yaml
@@ -7,23 +7,6 @@
     packages:
       - jq
 
-- name: Get k3s release
-  shell:
-    executable: /bin/bash
-    cmd: |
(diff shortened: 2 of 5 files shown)


jichenjc:
/lgtm

@chrigl maybe we need run multiple CI with this change?

chrigl:

Let's see if I can trigger that one :)

chrigl:
From my point of view the failed manila test-case is ok. That's the same one that fails on cinder as well, and there were changes in v1.24 test suite. I would disable it in #1887 because I am working on the tests there anyways.

PR #1909, opened 2022-06-06
Add region to providerID magic string.

### LINKED ISSUE #1900: [OCCM] Add Region code to ProviderID
/kind feature

**What happened**:

Hello. I have a multi-region OpenStack setup for one Kubernetes cluster. Each region has its independent OpenStack installation. 

Currently, OCCM does not support multi-region installation. 
And I can launch OCCM separately in each zone, but they will affect each other. One OCCM wouldn't find an instance from another region (it will try to delete the node from different region). 

All works fine if ProviderID has a ```region``` code.

Here a few changes in OCCM https://github.com/sergelogvinov/cloud-provider-openstack/commit/c102acbb0f404f64d78c92826ce308d225b9c215 (it support old and new ProviderID format)

So can I add this feature to the OCCM ?

OLD string:  ProviderID =```${ProviderName}:///${instance-id}```
NEW string: ProviderID =```${ProviderName}://${REGION}/${instance-id}```

**Environment**:
- OpenStack version:
- Others:

sergelogvinov:

Change ProviderID to magic sting ```${ProviderName}://${REGION}/${instance-id}```
Which helps to find an instance in a multi-region setup.

It also support old magic string (without region). 
In this case OCCM works as before (can be used only in the one-zone cluster).

fixes #1900

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/ingress/utils/utils.go
+++ b/pkg/ingress/utils/utils.go
@@ -74,14 +74,17 @@ func stringSlicesEqual(x, y []string) bool {
 
 // GetNodeID get instance ID from the Node spec.
 func GetNodeID(node *apiv1.Node) (string, error) {
-	var providerIDRegexp = regexp.MustCompile(`^openstack:///([^/]+)$`)
+	var providerIDRegexp = regexp.MustCompile(`^openstack://([^/]*)/([^/]+)$`)
 
 	matches := providerIDRegexp.FindStringSubmatch(node.Spec.ProviderID)
-	if len(matches) != 2 {
+	switch len(matches) {
+	case 2:
+		return matches[1], nil
+	case 3:
+		return matches[2], nil
+	default:
 		return "", fmt.Errorf("failed to find instance ID from node provider ID %s", node.Spec.ProviderID)
 	}
-
-	return matches[1], nil
 }
 
 // Convert2Set converts a string list to string set.
--- a/pkg/openstack/instances.go
+++ b/pkg/openstack/instances.go
@@ -20,7 +20,7 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"os"
+	sysos "os"
 	"regexp"
 	"sort"
 	"strings"
@@ -46,13 +46,16 @@ import (
 
 // Instances encapsulates an implementation of Instances for OpenStack.
 type Instances struct {
-	compute        *gophercloud.ServiceClient
-	opts           metadata.Opts
-	networkingOpts NetworkingOpts
+	compute          *gophercloud.ServiceClient
+	region           string
+	regionProviderID bool
+	opts             metadata.Opts
+	networkingOpts   NetworkingOpts
 }
 
 const (
-	instanceShutoff = "SHUTOFF"
+	instanceShutoff       = "SHUTOFF"
+	RegionalProviderIDEnv = "OS_CCM_REGIONAL"
 )
 
 var _ cloudprovider.Instances = &Instances{}
@@ -77,10 +80,17 @@ func (os *OpenStack) instances() (*Instances, bool) {
 		return nil, false
 	}
 
+	regionalProviderID := false
+	if isRegionalProviderID := sysos.Getenv(RegionalProviderIDEnv); isRegionalProviderID == "true" {
+		regionalProviderID = true
+	}
+
 	return &Instances{
-		compute:        compute,
-		opts:           os.metadataOpts,
-		networkingOpts: os.networkingOpts,
+		compute:          compute,
+		region:           os.epOpts.Region,
+		regionProviderID: regionalProviderID,
+		opts:             os.metadataOpts,
+		networkingOpts:   os.networkingOpts,
 	}, true
 }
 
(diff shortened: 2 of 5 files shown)


pkg/openstack/instances.go
```diff
@@ -335,23 +357,28 @@ func isValidLabelValue(v string) bool {
...
 
 	// https://github.com/kubernetes/kubernetes/issues/85731
 	if providerID != "" && !strings.Contains(providerID, "://") {
 		providerID = ProviderName + "://" + providerID
 	}
 
 	matches := providerIDRegexp.FindStringSubmatch(providerID)
-	if len(matches) != 2 {
-		return "", fmt.Errorf("ProviderID \"%s\" didn't match expected format \"openstack:///InstanceID\"", providerID)
+	switch len(matches) {
```
jichenjc, line 374: can we extend existing UT to cover this case and logic?

pkg/openstack/instances.go
```diff
@@ -128,14 +130,18 @@ func (i *Instances) NodeAddresses(ctx context.Context, name types.NodeName) ([]v
...
 
-	instanceID, err := instanceIDFromProviderID(providerID)
+	instanceID, instanceRegion, err := instanceIDFromProviderID(providerID)
 
 	if err != nil {
 		return []v1.NodeAddress{}, err
 	}
 
+	if instanceRegion != "" && instanceRegion != i.region {
+		return []v1.NodeAddress{}, nil
```
jichenjc, line 142: do we need a log for such error?

and return nil means success? should we return error because 
we are getting a provider on different region in such case?

sergelogvinov: Thank you,
I've added logs and error messages where needed. 

Here, I return an empty node-IP list to hide error messages from the main cloud-control module. It does nothing it NodeAddress empty.

jichenjc:
@chrigl @zetaab maybe you want to take a quick look at this?

zetaab:
this needs release-note. I know that at least we have component that are reading this providerid and using it. So we need to modify that in another applications. I have no idea will this break something in kubernetes itself

jichenjc:
>>>this needs release-note. I know that at least we have component that are reading this providerid and using it. So we need to modify that in another applications. I have no idea will this break something in kubernetes itself

good point , I agree we need give a thorough thought on the backend compatible 
one possible solution is to expose 2 files (w/ and w/o ) or we have a bool to control this 
we don't provide any API (leverage cloud provider) so ,need find how to make the contract compatible in such case

sergelogvinov:
Good catch. 
Let's add the environment like ```OS_CCM_REGIONAL``` to allow OCCM to add the region to the ProviderID string. By default, it will be ```false```. OCCM will support both type of string. But if OS_CCM_REGIONAL=false, OCCM creates the ProviderID without region code.

And in the future, we'll switch it to true.

sergelogvinov:
> this needs release-note. I know that at least we have component that are reading this providerid and using it. So we need to modify that in another applications. I have no idea will this break something in kubernetes itself

Finally, I found that Ingress and OCCM use ProdiverID sting. 
I've added:
* support for both sting types.
* optional environment to switch on/off region code in ProviderID sting.

sergelogvinov:
@jichenjc @zetaab Hi. I've updated the documentation and this feature can be activated by the environment value.

zetaab (approved):
/approve

pkg/ingress/utils/utils.go
```diff
@@ -74,14 +74,17 @@ func stringSlicesEqual(x, y []string) bool {
 
 // GetNodeID get instance ID from the Node spec.
 func GetNodeID(node *apiv1.Node) (string, error) {
-	var providerIDRegexp = regexp.MustCompile(`^openstack:///([^/]+)$`)
+	var providerIDRegexp = regexp.MustCompile(`^openstack://([^/]*)/([^/]+)$`)
 
 	matches := providerIDRegexp.FindStringSubmatch(node.Spec.ProviderID)
-	if len(matches) != 2 {
+	switch len(matches) {
```
anguslees, line 80: This is a bit weird. Just use a non-capturing regex group (`(?: ... )`) above instead, and then always use the original len == 2 code here.

`^openstack://(?:[^/]*)/([^/]+)$`

pkg/openstack/instances.go
```diff
@@ -335,23 +372,28 @@ func isValidLabelValue(v string) bool {
...
 	// https://github.com/kubernetes/kubernetes/issues/85731
 	if providerID != "" && !strings.Contains(providerID, "://") {
 		providerID = ProviderName + "://" + providerID
 	}
 
 	matches := providerIDRegexp.FindStringSubmatch(providerID)
-	if len(matches) != 2 {
-		return "", fmt.Errorf("ProviderID \"%s\" didn't match expected format \"openstack:///InstanceID\"", providerID)
+	switch len(matches) {
+	case 2:
```
anguslees, line 390: Does this case=2 ever happen?

I don't think so: https://go.dev/play/p/HLwE1ezK1TB

sergelogvinov: nice catch, thank you.

jichenjc: thanks for the catch ~

curious why our CI didn't find this ..

jichenjc: @sergelogvinov I assume you will provide a fix later on? Thanks

Issue #1910, opened 2022-06-07
OpenSSL PEM generates PKCS #8 keys


laurencegill:
[octavia-ingress-controller] cannot parse PKCS8 private key

**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

**What happened**:

Error from the ingress controller when trying to use a private key in pkcs8 encoding:
`Failed to update openstack resources for ingress XXX: failed to create Barbican secret: Cannot decode supplied PEM data`

Once the key was converted it worked:
`openssl rsa -in pkcs8.key -out pkcs1ca.key`

Log message:
`level=info msg="secret created in Barbican"`

**What you expected to happen**:
Decode the PKCS8 key

**How to reproduce it**:
Use a PCKS8 private key beginning with:
`-----BEGIN PRIVATE KEY-----`

**Anything else we need to know?**:
Issue seems to also be in master as well as 1.18

This seems relevant:
https://github.com/golang/go/blob/master/src/crypto/tls/tls.go#L336

**Environment**:
- octavia-ingress-controller:v1.18.0
- OpenStack version:v1.18.0

jichenjc:
addressed in related PR, close it

PR #1911, opened 2022-06-07
OpenSSL PEM generates PKCS #8 keys

laurencegill:
[octavia-ingress-controller] PKCS8 keys

#1910 

#1910 (I am not suggesting you use this code)

This is probably more relevant:
https://github.com/golang/go/blob/master/src/crypto/tls/tls.go#L336

**Release note**:

```release-note
NONE
```


jichenjc:
thanks for the PR, can you help fix the CLA issue?

jichenjc:
the fix looks good, need fix the check failure though..

PR #1946, opened 2022-07-21
[occm] Support specifying a sort order for node addresses

ederst:
This introduces a configuration key which influences the way the provider reports the node addresses to the Kubernetes node resource.

The default order depends on the hard-coded order the provider queries the addresses and what the cloud returns, which does not guarantee a specific order.

To override this behavior it is possible to specify a comma separated list of CIDRs. Essentially, this will sort and group all addresses matching a CIDR in a prioritized manner, where the first item having a higher priority than the last. All non-matching addresses will remain in the same order they are already in.

For example, this option can be useful when having multiple or dual-stack interfaces attached to a node and needing a user-controlled, deterministic way of sorting the addresses.

Fixes at least my issue with occm reporting either IPv6 or IPv4 when trying to access Pods and not needing to deactivate IPv6 reporting or specifying the IPv6 network as an external network (which is only possible if the IPv6 is from a different network anyhow).

None so far.

**Release note**:
```release-note
[openstack-cloud-controller-manager] Support specifying a sort order for node addresses
```


jichenjc:
/lgtm
/approve

Thanks for the PR, I met this issue before so I Think it's helpful feature

PR #1964, opened 2022-08-25
[occm] Use healthz Endpoint in HTTP Monitor

shaardie:
This patch introduces the usage of the /healthz Endpoint in the monitor.

Documentation for this is hard to find, but you can see in the kubernetes
repository
https://github.com/kubernetes/kubernetes/blob/26842bb0c8fcf89885798a2ebd9a97cd104e78bf/staging/src/k8s.io/cloud-provider/service/helpers/helper.go#L87
that /healthz is the prefered endpoint here.

I did not create a issues for this, since it is not really a bug or a feature, but I can, if needed.

I did not touch the legacy code not necessary for octavia load balancer.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1204,19 +1204,22 @@ func (lbaas *LbaasV2) ensureOctaviaHealthMonitor(lbID string, name string, pool 
 
 // buildMonitorCreateOpts returns a v2monitors.CreateOpts without PoolID for consumption of both, fully popuplated Loadbalancers and Monitors.
 func (lbaas *LbaasV2) buildMonitorCreateOpts(svcConf *serviceConfig, port corev1.ServicePort) v2monitors.CreateOpts {
-	monitorProtocol := string(port.Protocol)
-	if port.Protocol == corev1.ProtocolUDP {
-		monitorProtocol = "UDP-CONNECT"
-	}
-	if svcConf.healthCheckNodePort > 0 {
-		monitorProtocol = "HTTP"
-	}
-	return v2monitors.CreateOpts{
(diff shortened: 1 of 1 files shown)


pkg/openstack/loadbalancer.go
```diff
@@ -1204,19 +1204,22 @@ func (lbaas *LbaasV2) ensureOctaviaHealthMonitor(lbID string, name string, pool
...
 		Delay:      svcConf.healthMonitorDelay,
 		Timeout:    svcConf.healthMonitorTimeout,
 		MaxRetries: svcConf.healthMonitorMaxRetries,
 	}
+	if port.Protocol == corev1.ProtocolUDP {
+		opts.Type = "UDP-CONNECT"
+	}
+	if svcConf.healthCheckNodePort > 0 {
+		opts.Type = "HTTP"
+		opts.URLPath = "/healthz"
```
jichenjc, line 1218: This /healthz is used for pod healthy status check
but you are proposing to do this for openstack? Not against it but want to make sure I understand the reason
and how its used..

shaardie: This is the health check on the nodes. So, if a load balancer service in Kubernetes has `healthCheckNodePort` set, this port can be used to check if the service on the node port is healthy or not. When creating a health monitor in Openstack, the load balancer uses this port to check, if the service is healthy on the node.

This seems to have worked fine with a request to `/`, but as you can see in the linked Kubernetes code, `/healthz` is the preferred endpoint here.

jichenjc: got it ,thanks for the detailed info~

PR #1968, opened 2022-08-25
[occm] Introduce Hostname Annotation

### LINKED ISSUE #1859: [occm] Workaround with hostname `ip`.nip.io does not work with IPv6

> /kind feature

**What happened**:

We are using the settings `enable-ingress-hostname=true`, because we are using proxy-protocoll and therefore have to route all Traffic over the loadbalancer (as described in the docs https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md#use-proxy-protocol-to-preserve-client-ip).

But unfortunately this workaround does not work with IPv6, since XXX:XXX:XXX.nip.io is not a valid dns name and the controller failed to patch the service

```
E0513 08:39:48.529201       1 controller.go:310] error processing service xxx (will retry): failed to update load balancer status: Service "xxx" is invalid: status.loadBalancer.ingress[0].hostname: Invalid value: XXX:XXX:XXX.nip.io": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

Honestly, I think the hostname workaround with `ip`.nip.io does not work properly I would prefer a proper solution with an annotation `service.beta.kubernetes.io/do-loadbalancer-hostname`, like already described in https://github.com/kubernetes/cloud-provider-openstack/issues/1287#issuecomment-716618469.

I know this is a temporary workaround, but with the current solution, I do not know how to properly use IPv6.

**How to reproduce it**:

Use the `enable-ingress-hostname=true` for an IPv6 loadbalancer.

We are also using the workaround described in #1086.

shaardie:

This patch makes it possible to explicitly configure the hostname of a
load balancer via annotation.

This hostname is used in the load balancer instead of the IP address or the
*fake* hostname.

Similar to `enabie-ingress-hostname`, this can be used with proxy protocol to preserve the IP, **but** this also works for IPv6 load balancer (<ipv6>.nip.io is no valid hostname).

https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md#use-proxy-protocol-to-preserve-client-ip

fixes #1859

**Release note**:

```release-note
Add option to set hostname on load balancer via annotation
```

### DIFF
--- a/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
+++ b/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
@@ -219,6 +219,10 @@ Request Body:
 
   If this annotation is specified, the other annotations which define the load balancer features will be ignored.
 
+- `loadbalancer.openstack.org/hostname`
+
+  This annotations explicitly sets a hostname in the status of the load balancer service.
+
 ### Switching between Floating Subnets by using preconfigured Classes
 
 If you have multiple `FloatingIPPools` and/or `FloatingIPSubnets` it might be desirable to offer the user logical meanings for `LoadBalancers` like `internetFacing` or `DMZ` instead of requiring the user to select a dedicated network or subnet ID at the service object level as an annotation.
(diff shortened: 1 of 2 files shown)


zetaab (approved):
This looks great!

PR #1969, opened 2022-08-25
[occm] Ensure namespaced objects have namespace set

yankcrime:

Namespaced API resources should have the namespace value set explicitly.

This commit updates the OCCM Helm chart to add this field to each template, inheriting the value of `Release.Namespace`.

**Release note**:

```release-note
NONE
```

Signed-off-by: Nick Jones <nick@dischord.org>

### DIFF
--- a/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   name: {{ include "occm.name" . }}
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "occm.labels" . | nindent 4 }}
 spec:
--- a/charts/openstack-cloud-controller-manager/templates/secret.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/secret.yaml
@@ -3,6 +3,7 @@ apiVersion: v1
 kind: Secret
 metadata:
  name: {{ .Values.secret.name | default "cloud-config" }}
+ namespace: {{ .Release.Namespace }}
 type: Opaque
 data:
  {{ if .Values.cloudConfigContents -}}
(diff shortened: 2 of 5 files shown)


jichenjc:
I use manifest mostly , so not sure helm has this issue 
just to confirm ,so without fix , we will deploy to `default` then with the PR we can deploy to selected ns?

yankcrime:
I think it depends on the client.  It's normally not a problem as without these Helm will create objects in the namespace set by the current config context, or you can override with the `--namespace` option and it'll do the right thing.  However, I actually hit a problem when rendering the manifests with `helm template` and then generating a ConfigMap for use by CAPI and CRS.  In that case, when CAPI tries to apply the resources you see errors along the lines of:

```
E0826 09:03:49.258292       1 clusterresourceset_controller.go:348] "failed to apply ClusterResourceSet resource" err="[failed to create object /v1, Kind=Secret /cloud-config: an empty namespace may not be set during creation, failed to create object /v1, Kind=ServiceAccount /openstack-cloud-controller-manager: an empty namespace may not be set during creation, failed to create object apps/v1, Kind=DaemonSet /openstack-cloud-controller-manager: an empty namespace may not be set during creation]" controller="clusterresourceset" controllerGroup="addons.cluster.x-k8s.io" controllerKind="ClusterResourceSet" clusterResourceSet="default/openstack-ccm" namespace="default" name="openstack-ccm" reconcileID=74ffb57b-2e74-44be-96ca-135681b3e46a cluster="test" Resource kind="ConfigMap" Resource name="openstack-ccm"
```

The solution is to make sure the namespace is set for these objects.

yankcrime:
To answer your question a bit more clearly - it doesn't change the behaviour right now, you can already install via Helm chart to a namespace of your choosing.  But this fixes a problem in cases where the client tries to create these objects without specifying the target namespace.

jichenjc:
/lgtm
/approve

thanks for the detailed update, apprecaite~

PR #1970, opened 2022-08-26
fix: Add region to providerID magic string

sergelogvinov:
Fix definition instanceID from providerID.
Revert to the previous version with small changes.

Based on the conversation #1909.
Thanks @anguslees


anguslees (approved):
/lgtm

PR #1972, opened 2022-08-26
[occm] Add security rule for the HealthCheckNodePort

### LINKED ISSUE #1971: [occm] Security Rule for HealthCheckNodePort Port missing

>
/kind bug
> /kind feature

**What happened**:

When using a load balancer service with `healthCheckNodePort` there is no rule for this Port in the load balancer security group.

The security group should also allow ingress traffic to the `healthCheckNodePort` port.

**How to reproduce it**:

Create a load balancer with `healthCheckNodePort` != 0 and check the load balancer security group rules.

I am working on a PR for this.

**Environment**:
- OpenStack version: near current 33f0db0e1c459564b39773a606459135482d91e1
- Others:

shaardie:

This PR adds a security rule to the load balancer security group for the `healthCheckNodePort`, if present.
WIthout this patch you need to open the whole range of possible node ports on all potential nodes, if using `healthCheckNodePort`.
fixes #1971 

* I created a new function `ensureSecurityRule` to reduce code duplication
* I moved the calculation of `etherType` and `subnet` out of the loop, since it is the same for each iteration.
* During development I encounterd a bug in `applyNodeSecurityGroupIDForLB`. It tries to add the same security group multiple times and fails. So I added a small check to only apply missing security groups.

**Release note**:

```release-note
* healthCheckNodePort ports are handled from occm
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -858,6 +858,20 @@ func applyNodeSecurityGroupIDForLB(compute *gophercloud.ServiceClient, network *
 		}
 
 		for _, port := range allPorts {
+			// If the Security Group is already present on the port, skip it.
+			// As soon as this only supports Go 1.18, this can be replaces by
+			// slices.Contains.
+			if func() bool {
+				for _, currentSG := range port.SecurityGroups {
+					if currentSG == sg {
+						return true
+					}
+				}
+				return false
+			}() {
+				continue
+			}
+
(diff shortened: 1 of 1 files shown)


shaardie:
I realized that the logic for `updateSecurityGroup` is still missing for this PR.

Also this is somehow attached to the Issue  #1976 for supporting `manage-security-groups` with Octavia load balancer.

PR #1982, opened 2022-09-01
Revert "[ci] bump golang version to 1.19 in install-golang role"

jichenjc:
Reverts kubernetes/cloud-provider-openstack#1981

looks like this causing issue now ..

```release-note
None
```


jichenjc:
@gman0 looks like PR causing issue in gate, I think we need revert this first
e.g https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/cloud-provider-openstack/1979/openstack-cloud-controller-manager-e2e-test/1565146390990950400/build-log.txt

PR #1975, opened 2022-08-31
Bump k8s deps to 1.25

olemarkus:
* openstack-cloud-controller-manager (occm)
* cinder-csi-plugin
* manila-csi-plugin
* k8s-keystone-auth
* client-keystone-auth
* octavia-ingress-controller
* magnum-auto-healer
* barbican-kms-plugin

**What this PR does / why we need it**:

Bumps all the k8s dependencies to 1.25. Also pulls in a couple of other deps upgrades.

**Release note**:

-->
```release-note
NONE
```

PR #1983, opened 2022-09-01
[CI] bump golang version to 1.19 in install-golang role

gman0:

This PR sets golang version to 1.19 in install-golang role. The download location was also changed because 1.19 doesn't seem to be available in https://storage.googleapis.com/golang.

We need 1.19 golang toolchain because k/k 1.25 dependencies are already using some parts of its std lib which are not available in previous verisions (see https://github.com/kubernetes/cloud-provider-openstack/pull/1975#issuecomment-1232711142).

fixes #

This PR contains a fixed version of https://github.com/kubernetes/cloud-provider-openstack/pull/1981. There was a typo in the download URL.

This is the URL that gets generated + wget call to test

```
wget https://go.dev/dl/go1.19.linux-amd64.tar.gz
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/playbooks/roles/install-golang/defaults/main.yml
+++ b/tests/playbooks/roles/install-golang/defaults/main.yml
@@ -1,5 +1,5 @@
 ---
-go_version: '1.17.4'
+go_version: '1.19'
 arch: 'amd64'
 go_tarball: 'go{{ go_version }}.linux-{{ arch }}.tar.gz'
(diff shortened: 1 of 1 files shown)


gman0:
PTAL @olemarkus @jichenjc

jichenjc:
/lgtm
/approve

thanks for the update~ ,CI passed :)

PR #1984, opened 2022-09-01
[occm] Manage Security Groups for Octavia

### LINKED ISSUE #1971: [occm] Security Rule for HealthCheckNodePort Port missing

### LINKED ISSUE #1976: [occm] Manage Security Groups for Octavia Load Balancer

>
> /kind bug
/kind feature

**What happened**:

According to the documentation the configuration option `manage-security-groups` does not work with Octavia.
But we tried it, since we do not want to open all Node Ports and it works for most cases.

Obviously it would be nice to use  `manage-security-groups` with Octavia load balancer.

**How to reproduce it**:

Enable  `manage-security-groups` with Octavia load balancer.

It would be nice to have this as a fully working feature and I would be willing to implement this (In fact I already started with https://github.com/kubernetes/cloud-provider-openstack/pull/1972).

As far as I can see the only part not working properly is the `updateSecurityGroup` which adds the security rules to every security group on the nodes. But this could be fixed using a similar logic than in `ensureSecurityGroup`.

Also I want to note that there were some attempts to implement this and you can still see this in the code. E.g. `ensureSecurityGroup` already has the distinction between octavia and non octavia load balancer, see https://github.com/kubernetes/cloud-provider-openstack/blob/d1196878af7af85bc2aea06c8fa37d27ba4a1f37/pkg/openstack/loadbalancer.go#L2576

If this is something you want to have upstream, maybe we can discuss how to submit this.

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: Current master (d1196878af7af85bc2aea06c8fa37d27ba4a1f37)
- OpenStack version:
- Others:

shaardie:

This patch makes it possible to use the manage-security-groups with octavia
load balancer.

It creates a new function `ensureAndUpdateOctaviaSecurityGroup` which is called
on `EnsureLoadBalancer` and `UpdateLoadBalancer` and handles the security group
and security rule creation for Octavia Load Balancer.

To make the changes not bigger than needed, it hooks into the
`updateSecurityGroup` and `ensureSecurityGroup` function first, checks if
Octavia is used and uses the new code. Otherwise the legacy code is called. So
the removal of the legacy code should be quite easy.

Also it adds a missing security rule for the HealthCheckNodePort and therefore supersedes  PR #1972.

fixes #1976 
fixes #1971

This PR supersedes #1972.

**Release note**:

```release-note
Make `manage-security-groups` available for Octavia load balancer
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -862,6 +862,20 @@ func applyNodeSecurityGroupIDForLB(compute *gophercloud.ServiceClient, network *
 		}
 
 		for _, port := range allPorts {
+			// If the Security Group is already present on the port, skip it.
+			// As soon as this only supports Go 1.18, this can be replaces by
+			// slices.Contains.
+			if func() bool {
+				for _, currentSG := range port.SecurityGroups {
+					if currentSG == sg {
+						return true
+					}
+				}
+				return false
+			}() {
+				continue
+			}
+
 			newSGs := append(port.SecurityGroups, sg)
 			updateOpts := neutronports.UpdateOpts{SecurityGroups: &newSGs}
 			mc := metrics.NewMetricContext("port", "update")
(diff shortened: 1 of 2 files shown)


pkg/openstack/loadbalancer.go
```diff
@@ -2567,10 +2581,64 @@ func (lbaas *LbaasV2) getSubnet(subnet string) (*subnets.Subnet, error) {
...
+		Protocol:       protocol,
+		PortRangeMax:   portRangeMin,
+		PortRangeMin:   portRangeMax,
+		RemoteIPPrefix: remoteIPPrefix,
+		SecGroupID:     secGroupID,
+		EtherType:      etherType,
+	}
+
+	mc := metrics.NewMetricContext("security_group_rule", "create")
+	_, err = rules.Create(lbaas.network, sgRuleCreateOpts).Extract()
```
jichenjc, line 2621: @shaardie I saw the creation and update here ,maybe I missed something, those rules/sec group will be deleted 
at EnsureSecurityGroupDeleted so we get it for free?

shaardie: Yes, in  `EnsureSecurityGroupDeleted` all Rules in the Security Group are already deleted. So no need to write a function for that. 

This function to create or update Rules was extracted from the existing code and only slightly adapted. Because we create Rules now in different parts of the code (Rules for the Node Port and Rules for the Health Checks Port), I thought it was better to extract a function instead of duplicating the code.

Sorry I had directly resolved this. No intention, only clumsy fingers.

jichenjc: works for me , thanks for the info :)

shaardie:
Anything blocking this?

pkg/openstack/loadbalancer.go
dulek: @shaardie: Sorry for only noticing this when merged. How do we handle removal of SG rules in situations like a port getting edited on the .spec.ports of a Service? At a glance it seems to me a new SG rule will get created, while the old one will stay.

shaardie: I guess you are correct. I did not write any logic to remove the rules. So rules should only be deleted, if the security group itself is deleted at the moment.

Since this PR is already merged, maybe it makes sense to open a new Issue for that?

dulek: I opened #2003.

PR #1988, opened 2022-09-06
[occm] Protect from `AllocateLoadBalancerNodePorts=false`

### LINKED ISSUE #1996: [occm] `AllocateLoadBalancerNodePorts=false` makes cloud provider to send nonsense requests to Octavia
/kind bug

**What happened**:
As explained in [K8s docs](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-nodeport-allocation) it is possible to set `spec.allocateLoadBalancerNodePorts` to `false` for a `type=LoadBalancer` Service which will disable creation of the NodePorts for that Service. Effectively this means `.spec.ports[*].nodePort` on that Service will be set to `0`. The cloud-provider-openstack doesn't really account for that and just uses this value when creating loadbalancer members. Obviously that fails because Octavia won't accept port number 0.

Creation of members with a port `0` is skipped, so that Service can be handled fully without errors.

**How to reproduce it**:
Create a `type=LoadBalancer` Service with `spec.allocateLoadBalancerNodePorts=false` and observe events on the Service.

This won't make such Services to suddenly start to work as in direct traffic correctly without NodePorts, but rather protect from sending nonsense requests to Octavia.

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: current
- OpenStack version: Doesn't matter.
- Others:

dulek:
When aforementioned option is set to `false` on the Service `.spec`, the
NodePorts stop being allocated and set for that Service, effectively
defaulting all `.spec.ports[*].nodePort` to 0. Similar situation happens
with `healthCheckNodePort`. When cloud provider tries to create a member
or SG with such a port, Octavia fails.

This commit makes sure we are not adding members when `nodePort` value
is 0. We also omit the SG manipulations for such ports.

It's to make sure we are not making requests to Octavia that it cannot fulfill.

fixes #1996

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1365,29 +1365,31 @@ func (lbaas *LbaasV2) buildBatchUpdateMemberOpts(port corev1.ServicePort, nodes 
 			}
 		}
 
-		member := v2pools.BatchUpdateMemberOpts{
-			Address:      addr,
-			ProtocolPort: int(port.NodePort),
-			Name:         &node.Name,
-			SubnetID:     &svcConf.lbMemberSubnetID,
-		}
-		if svcConf.healthCheckNodePort > 0 {
-			useHealthCheckNodePort := true
-			if lbaas.opts.LBProvider == "ovn" {
-				// ovn-octavia-provider doesn't support HTTP monitors at all, if we have it we got to rely on NodePort
-				// and UDP-CONNECT health monitor.
-				useHealthCheckNodePort = false
-			} else if port.Protocol == "UDP" {
-				// Older Octavia versions doesn't support HTTP monitors on UDP pools. If we have one like that, we got
-				// to rely on checking the NodePort instead.
-				useHealthCheckNodePort = openstackutil.IsOctaviaFeatureSupported(lbaas.lb, openstackutil.OctaviaFeatureHTTPMonitorsOnUDP, lbaas.opts.LBProvider)
+		if port.NodePort != 0 { // It's 0 when AllocateLoadBalancerNodePorts=False
+			member := v2pools.BatchUpdateMemberOpts{
+				Address:      addr,
+				ProtocolPort: int(port.NodePort),
+				Name:         &node.Name,
+				SubnetID:     &svcConf.lbMemberSubnetID,
 			}
-			if useHealthCheckNodePort {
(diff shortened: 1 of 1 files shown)


dulek:
@mdbooth, you might want to take a quick look here.

jichenjc:
>>>When aforementioned option is set on the Service .spec, the NodePorts
stop being set, effectively defaulting .spec.ports[*].nodePort to 0. When
cloud provider tries to create a member with such a port, Octavia fails.

do you mind create a bug to track this? just want to keep record and know the impact /reason etc as reference

jichenjc:
code looks reasonable @dulek 
but I guess if there's more info in the issue talk about AllocateLoadBalancerNodePorts in commit msg will be helpful 
as I think it's come from k8s core defintion which lack some background in the PR itself

jichenjc:
@dulek could you please help rebase this? Thanks

mandre (commented):
Don't we also need to add a guard to [ensureAndUpdateOctaviaSecurityGroup()](https://github.com/kubernetes/cloud-provider-openstack/blob/95e2de6/pkg/openstack/loadbalancer.go#L3237-L3256) as well?

dulek:
> Don't we also need to add a guard to [ensureAndUpdateOctaviaSecurityGroup()](https://github.com/kubernetes/cloud-provider-openstack/blob/95e2de6/pkg/openstack/loadbalancer.go#L3237-L3256) as well?
> 
> **edit**: replying to myself, yes, most likely. The code was added in #1984 after you first submitted your PR.

Done, thanks for spotting this!

pkg/openstack/loadbalancer.go
```diff
@@ -3292,6 +3305,9 @@ func (lbaas *LbaasV2) updateSecurityGroup(clusterName string, apiService *corev1
 	}
 
 	for _, port := range ports {
+		if port.NodePort == 0 {
```
mandre, line 3308: Nit: not sure if you meant to leave a comment about NodePort being 0 here too.

dulek: Sure, a comment never hurts, done.

mandre (approved):
LGTM

dulek:
@jichenjc: Hi! I've updated the commit message a bit, can you take another look at this?

PR #2000, opened 2022-09-23
[occm] Do not create HTTP monitors for UDP on old Octavia

### LINKED ISSUE #1999: [occm] Monitors created incorrectly for UDP pools on older Octavia versions
/kind bug

**What happened**:
If you create an `ETP=Local` `type=LoadBalancer` Service you got to have the Octavia health monitors created in order for traffic to only be directed to nodes hosting the Service pods. Moreover you'll be getting `healthCheckNodePort` allocated and an HTTP endpoint spawned there. This means that for such Services OCCM will create Octavia health monitors of type HTTP pointing to `healthCheckNodePort` value. This works great unless you're using older Octavia (API version lower than 2.16) and the Service is UDP as [such Octavia doesn't support HTTP monitors on UDP pools](https://github.com/openstack/octavia/blob/master/octavia/api/root_controller.py#L115-L117).

Now an interesting part is that even though this is not supported, Octavia doesn't complain when OCCM does such a request. This is because OCCM uses the API that creates whole LB in a single call and there's a bug in Octavia causing health monitors validation to be bypassed in such cases.

What we end up with is that health monitor gets created, but it's 100% not functional and the calls to the loadbalancer are distributed to all the members, even these that doesn't host Services pods, so Service is unreliable.

OCCM should be smart enough to detect Octavia API version and fallback to using UDP-CONNECT health monitors when API version is lower than 2.16.

**How to reproduce it**:
On a cluster with several nodes, hosted on OpenStack with Octavia older than 2.16 create an `ETP=Local` UDP Service with lower number of Pods than the number of nodes. Then test the reliablity of the LB connections.

**Environment**:
- OpenStack version: Train
- Others:

dulek:
Octavia API before 2.16 doesn't support HTTP monitors on UDP pools. It will however allow creation of such if all-in-one LB creation call is used. This is a problem for UDP `ETP=Local` Services, where `healthCheckNodePort` is allocated and CCM will try to create HTTP monitor there. This monitor will be completely not functional.

OVN Octavia provider doesn't support HTTP health monitors either.

This commit fixes that by making sure we're checking Octavia API version
and if it's lower than 2.16 we fall back to creating UDP-CONNECT health
monitors on a regular port instead. Moreover same is done when provider
is set to "ovn".

fixes #1999

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1232,10 +1232,21 @@ func (lbaas *LbaasV2) buildMonitorCreateOpts(svcConf *serviceConfig, port corev1
 		opts.Type = "UDP-CONNECT"
 	}
 	if svcConf.healthCheckNodePort > 0 {
-		opts.Type = "HTTP"
-		opts.URLPath = "/healthz"
-		opts.HTTPMethod = "GET"
-		opts.ExpectedCodes = "200"
+		setHTTPHealthMonitor := true
+		if lbaas.opts.LBProvider == "ovn" {
+			// ovn-octavia-provider doesn't support HTTP monitors at all. We got to avoid creating it with ovn.
+			setHTTPHealthMonitor = false
+		} else if opts.Type == "UDP-CONNECT" {
+			// Older Octavia versions or OVN provider doesn't support HTTP monitors on UDP pools. We got to check if that's the case.
+			setHTTPHealthMonitor = openstackutil.IsOctaviaFeatureSupported(lbaas.lb, openstackutil.OctaviaFeatureHTTPMonitorsOnUDP, lbaas.opts.LBProvider)
+		}
+
+		if setHTTPHealthMonitor {
+			opts.Type = "HTTP"
+			opts.URLPath = "/healthz"
+			opts.HTTPMethod = "GET"
+			opts.ExpectedCodes = "200"
+		}
 	}
 	return opts
 }
@@ -1361,7 +1372,19 @@ func (lbaas *LbaasV2) buildBatchUpdateMemberOpts(port corev1.ServicePort, nodes 
 			SubnetID:     &svcConf.lbMemberSubnetID,
 		}
 		if svcConf.healthCheckNodePort > 0 {
-			member.MonitorPort = &svcConf.healthCheckNodePort
+			useHealthCheckNodePort := true
+			if lbaas.opts.LBProvider == "ovn" {
+				// ovn-octavia-provider doesn't support HTTP monitors at all, if we have it we got to rely on NodePort
+				// and UDP-CONNECT health monitor.
+				useHealthCheckNodePort = false
+			} else if port.Protocol == "UDP" {
+				// Older Octavia versions doesn't support HTTP monitors on UDP pools. If we have one like that, we got
+				// to rely on checking the NodePort instead.
+				useHealthCheckNodePort = openstackutil.IsOctaviaFeatureSupported(lbaas.lb, openstackutil.OctaviaFeatureHTTPMonitorsOnUDP, lbaas.opts.LBProvider)
+			}
+			if useHealthCheckNodePort {
+				member.MonitorPort = &svcConf.healthCheckNodePort
+			}
 		}
 		members = append(members, member)
 		newMembers.Insert(fmt.Sprintf("%s-%s-%d-%d", node.Name, addr, member.ProtocolPort, svcConf.healthCheckNodePort))
--- a/pkg/util/openstack/loadbalancer.go
+++ b/pkg/util/openstack/loadbalancer.go
@@ -42,6 +42,7 @@ const (
 	OctaviaFeatureFlavors           = 2
 	OctaviaFeatureTimeout           = 3
 	OctaviaFeatureAvailabilityZones = 4
+	OctaviaFeatureHTTPMonitorsOnUDP = 5
 
 	waitLoadbalancerInitDelay   = 1 * time.Second
 	waitLoadbalancerFactor      = 1.2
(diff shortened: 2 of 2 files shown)


mavynet:
@dulek Sorry if offtopic, but how did you manage to make the loadbalancer work on UDP?
I'm getting the same error in the OCCM logs each time saying that UDP is not supported: 

failed to ensure load balancer: only TCP LoadBalancer is supported for openstack load balancers

thanks

dulek:
> @dulek Sorry if offtopic, but how did you manage to make the loadbalancer work on UDP? I'm getting the same error in the OCCM logs each time saying that UDP is not supported:
> 
> failed to ensure load balancer: only TCP LoadBalancer is supported for openstack load balancers

It's offtopic, but you're setting `use-octavia: false` and that's why you see this. Neutron-lbaas doesn't support anything else than TCP.

dulek:
@jichenjc: Can you help me here? Seems like there's some vendoring issue that can be fixed by `go mod vendor`, but running it doesn't produce anything commitable, so I'm pretty confused here.

jichenjc:
>>>Can you help me here? Seems like there's some vendoring issue that can be fixed by go mod vendor, but running it doesn't produce anything commitable, so I'm pretty confused here.

I think we don't use `vendor` any more? 
https://maelvls.dev/go111module-everywhere/ 
maybe `go mod tidy` can help or with `GO111MODULE=on ` ?

dulek:
> > > > Can you help me here? Seems like there's some vendoring issue that can be fixed by go mod vendor, but running it doesn't produce anything commitable, so I'm pretty confused here.
> 
> I think we don't use `vendor` any more? https://maelvls.dev/go111module-everywhere/ maybe `go mod tidy` can help or with `GO111MODULE=on ` ?

Okay, it was so simple, I just had to remove vendor directory. Thank you!

pkg/openstack/loadbalancer.go
```diff
@@ -306,6 +310,73 @@ func (s *floatingSubnetSpec) Matcher(tag bool) (matcher, error) {
 	return match, nil
 }
 
+func initGetMaxOctaviaAPIVersion() func(client *gophercloud.ServiceClient) (*semver.Version, error) {
+	// Why all the timing cache hassle? The underlying OpenStack cloud might get updated at some point
+	// and we probably want to notice that.
```
jichenjc, line 315: um... this is general update of openstack? if so I think worthy in utils instead of here

and the use case is openstack cloud got upgraded w/o notify CPO?

dulek: Yeah, it's about the underlying OpenStack cloud. The K8s VMs are only a workload for OpenStack, so there's nothing that could notify CPO. Technically VMs will get restarted when compute nodes are upgraded but:
1. This might happen in the future, in bigger clouds computes might be upgraded one-by-one.
2. The VM CPO was running on could have been live-migrated, so never restarted.

pkg/openstack/loadbalancer.go
```diff
@@ -306,6 +310,73 @@ func (s *floatingSubnetSpec) Matcher(tag bool) (matcher, error) {
...
+
+		cachedVersion = max
+		cachedTime = time.Now()
+		return max, nil
+	}
+}
+
+var getMaxOctaviaAPIVersion = initGetMaxOctaviaAPIVersion()
+
+func (lbaas *LbaasV2) isOctaviaVersionSupported(client *gophercloud.ServiceClient, constraint string) (bool, error) {
```
mandre, line 365: Perhaps it's just me but I find this function hard to read.
How about something like that (not tested)?

```go
func (lbaas *LbaasV2) isOctaviaVersionSupported(client *gophercloud.ServiceClient, minimumVersion string) (bool, error) {
	remoteOctaviaVersion := semver.MustParse("2.0")
	if lbaas.opts.UseOctavia {
		var err error
		remoteOctaviaVersion, err = getMaxOctaviaAPIVersion(client)
		if err != nil {
			return false, err
		}
	}

	constraint, err := semver.NewConstraint(">= " + minimumVersion)
	if err != nil {
		return false, err
	}

	valid, _ := constraint.Validate(remoteOctaviaVersion)

	return valid, nil
}
```

pkg/openstack/loadbalancer.go
jichenjc: thanks for the info

that's the reason I am suggesting move to utils function instead of put here,,

pkg/openstack/loadbalancer.go
jichenjc: I actually have a PR to remove non-Octaiva case .. it will make those easier
but never got review and merge..

pkg/openstack/loadbalancer.go
dulek: Oh, we even have a function for that in there already. Okay, I'll just use that.

pkg/openstack/loadbalancer.go
dulek: I'll just use the version from util module instead.

pkg/util/openstack/loadbalancer.go
```diff
@@ -42,6 +42,7 @@ const (
 	OctaviaFeatureFlavors           = 2
 	OctaviaFeatureTimeout           = 3
 	OctaviaFeatureAvailabilityZones = 4
+	OctaviaFeatureHTTPMonitorsOnUDP = 8
```
mandre, line 45: Unless I'm missing something, we're doing arithmetic based on these constants so it should be safe to have picked `5`.

mandre (commented):
It's a lot simpler now, thanks for the suggestion from @jichenjc. LGTM.

pkg/util/openstack/loadbalancer.go
dulek: I have no idea why I missed there's a `3` in there already, thanks for catching this. :smiley:

mandre (approved):
LGTM

PR #2014, opened 2022-10-11
[occm]  Improve retrival of external Network

### LINKED ISSUE #2015: [occm] No external Network detected for Floating IP creation

/kind bug

**What happened**:

When multiple External Networks are available in the cluster, the occm attempts to get the first external Network that has any IPv4 Subnet. However, if the Network is not shared, the tenant is not able to list and filter the IPv4 Subnets of that Network, resulting in no Network detection and no Floating IP created, even when the Network is valid to be used (external with IPv4 subnet, but not shared).

Attempt to use any external Network that has IPv4 Subnet or any Network that has Subnets that no details could be gathered about.

**How to reproduce it**:

Create an external network accessible to the tenant with Subnets not accessible to tenant

MaysaMacedo:

When attempting to fetch an external Network with IPv4 Subnets to
create a Floating IP, it's possible that no Subnets from that Network
are shared with the tenant, consequently it would not be possible to
filter them by IPVersion and no Network would be used. This commit
improve the detection of external Network by only allowing usage of
Networks that is possible to detect at least one IPV4 Subnet or that
is not possible to detect any info about the Subnets.

fixes #2015

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/util/openstack/network.go
+++ b/pkg/util/openstack/network.go
@@ -27,6 +27,7 @@ import (
 	"github.com/gophercloud/gophercloud/pagination"
 
 	"k8s.io/cloud-provider-openstack/pkg/metrics"
+	cpoerrors "k8s.io/cloud-provider-openstack/pkg/util/errors"
 )
 
 // GetNetworkExtensions returns an extension map.
@@ -89,7 +90,7 @@ func GetFloatingNetworkID(client *gophercloud.ServiceClient) (string, error) {
 		networks.Network
 		external.NetworkExternalExt
 	}
-	var externalNetworks []NetworkWithExternalExt
+	var allNetworks []NetworkWithExternalExt
 
 	mc := metrics.NewMetricContext("network", "list")
 	page, err := networks.List(client, networks.ListOpts{}).AllPages()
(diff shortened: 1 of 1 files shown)


pkg/util/openstack/network.go
```diff
@@ -101,24 +102,41 @@ func GetFloatingNetworkID(client *gophercloud.ServiceClient) (string, error) {
 	if err != nil {
 		return "", mc.ObserveRequest(err)
 	}
+
 	for _, externalNet := range externalNetworks {
-		if externalNet.External {
+		if externalNet.External && len(externalNet.Subnets) > 0 {
 			mc := metrics.NewMetricContext("subnet", "list")
-			page, err := subnets.List(client, subnets.ListOpts{NetworkID: externalNet.ID, IPVersion: 4}).AllPages()
+			page, err := subnets.List(client, subnets.ListOpts{NetworkID: externalNet.ID}).AllPages()
```
chrigl, line 109: Does that change any behavior for v6 LBs? Not sure if this is a thing.

pkg/util/openstack/network.go
```diff
@@ -101,24 +102,41 @@ func GetFloatingNetworkID(client *gophercloud.ServiceClient) (string, error) {
 	if err != nil {
 		return "", mc.ObserveRequest(err)
 	}
+
 	for _, externalNet := range externalNetworks {
```
mandre, line 106: Hmm, not sure why the variable is called `externalNetworks` because it looks to me like it contains _all_ networks, regardless of whether they're external or not.

We do not filter anything in on line 96: https://github.com/kubernetes/cloud-provider-openstack/pull/2014/files#diff-a89aa055a2316c0db33d258ec1c17fc3948b8f53fbc3a475d5a6602025096b33R96

MaysaMacedo: Right. In [gophercloud](https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/external/doc.go#L19) it's documented as you mentioned. Let me fix that.

mandre (commented):
Sweet, thanks for updating the variable name. This looks good to me.

jichenjc:
though the CI of cinder is not related, let me trigger it again

PR #2023, opened 2022-11-01
[manila-csi-plugin] Add csi-provisioner's --extra-create-metadata as share metadata

gman0:

This PR adds support for recognizing volume parameters when csi-provisioner sidecar is run with `--extra-create-metadata`, and adds those as share metadata.

```
$ kubectl get pvc
NAME                   STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS            AGE
new-cephfs-share-pvc   Bound    pvc-755d6df7-b29c-40c6-9df6-9008c8db5339   1Gi        RWX            geneva-cephfs-testing   56s

$ openstack share show pvc-755d6df7-b29c-40c6-9df6-9008c8db5339 -c properties -f value
{'csi.storage.k8s.io/pv/name': 'pvc-755d6df7-b29c-40c6-9df6-9008c8db5339', 'csi.storage.k8s.io/pvc/name': 'new-cephfs-share-pvc', 'csi.storage.k8s.io/pvc/namespace': 'default', 'manila.csi.openstack.org/cluster': '968241dc-0820-44b0-9f6c-3c78cf0dd0b7'}
```

**Release note**:

```release-note
manila-csi-plugin: Add csi-provisioner's --extra-create-metadata as share metadata. Disabled by default.
```

### DIFF
--- a/pkg/csi/manila/controllerserver.go
+++ b/pkg/csi/manila/controllerserver.go
@@ -43,6 +43,15 @@ type controllerServer struct {
 var (
 	pendingVolumes   = sync.Map{}
 	pendingSnapshots = sync.Map{}
+
+	// Recognized volume parameters passed by Kubernetes csi-provisioner sidecar
+	// when run with --extra-create-metadata flag. These are added to metadata
+	// of newly created shares if present.
+	recognizedCSIProvisionerParams = []string{
+		"csi.storage.k8s.io/pvc/name",
+		"csi.storage.k8s.io/pvc/namespace",
+		"csi.storage.k8s.io/pv/name",
+	}
 )
 
 func getVolumeCreator(source *csi.VolumeContentSource, shareOpts *options.ControllerVolumeContext, compatOpts *options.CompatibilityOptions) (volumeCreator, error) {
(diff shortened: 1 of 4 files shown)


pkg/csi/manila/controllerserver.go
```diff
@@ -43,6 +43,15 @@ type controllerServer struct {
 var (
 	pendingVolumes   = sync.Map{}
 	pendingSnapshots = sync.Map{}
+
+	// Recognized volume parameters passed by Kubernetes csi-provisioner sidecar
+	// when run with --extra-create-metadata flag. These are added to metadata
+	// of newly created shares if present.
+	recognizedCSIProvisionerParams = []string{
```
jichenjc, line 50: looks like cinder CSI currently always honor the param now ...

jichenjc:
test fail is valid.. wait for the fix then merge

pkg/csi/manila/controllerserver.go
gman0: @jichenjc thanks for checking this! I wanted to retain the previous behavior -- no PV/PVC info in volume metadata. But I guess enabling it by default shouldn't cause issues. @gouthampacha do you think this is ok?

PR #2033, opened 2022-11-11
[occm] Remove unneeded security rules during Update

### LINKED ISSUE #2003: [occm] SG rules not removed with `manage-security-groups` and Octavia

/kind bug

**What happened**:
(I only see that in the code, haven't reproduced it in practice yet)
Seems like with #1984 once you enable `manage-security-groups` with Octavia, the SG rules are never removed. This means that each edit of the Service's .spec.ports will create a new SG rule, while outdated ones will not be removed, potentially leading to unexpected traffic being open.

The SG rules got to be fully reevaluated each time Service is changed.

**How to reproduce it**:
Create a Service when `manage-security-groups` is on. Modify the Service's `.spec.ports` changing a port number. Observe that new SG rule is created, but the one with old port number stays.

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: Current
- OpenStack version: Any with Octavia

shaardie:

This patch adds the functionality to delete old security rules from the security group during the load balancer update.

fixes #2003

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -2677,13 +2677,14 @@ func (lbaas *LbaasV2) getSubnet(subnet string) (*subnets.Subnet, error) {
 
 // ensureSecurityRule ensures to create a security rule for a defined security
 // group, if it not present.
+// Returns the either the found or the new created security rule.
 func (lbaas *LbaasV2) ensureSecurityRule(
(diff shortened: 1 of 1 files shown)


dulek:
/remove-lifecycle rotten

I'm going to have time to work on this fairly soon.

PR #2042, opened 2022-11-25
Concat AZ name with volume type

Demcheck:
cinder-csi-plugin

This PR gives opportuinity to concatenate availability zone name with type of volume.
Motivation:
We have kubernetes cluster on the cloud that use openstack under the hood.
This cloud has few availability zones. For every AZ we have volume type with AZ suffix.
For example:
We has k8s at AZs called: eu1a, eu1b, eu1c. For every zone we have volume types: fastssd.eu1a, fastssd.eu1b, etc.
To be able to have disks in every zone we need to concat type of volume and AZ name together and pass to openstack server.

fixes #

**Release note**:

```release-note
NONE
```


zetaab:
openstack cinder supports volume types, why that is configured to region name?

Demcheck:
> openstack cinder supports volume types, why that is configured to region name? you are going to have problems with this kind of setup in ~all tools

Hello! Yeah, I know that cinder supports volume types but for all availability zones we have different volume types with suffix in the end of volume type.
For example: availability zone has name eu-1a and volume type name fastssd.eu-1a, availability zone has name eu-1b and volume type name fastssd.eu-1b and etc. 

We have one kubernetes cluster in defferent availability zones, but in the Storage Class we can use only one name for volume type. It means that we are able to create disks only in one AZ. With my fix we have managed to create disks in defferent zones via concatanation of volume type and zone name.

zetaab:
the problem here is that why upstream should keep maintaining these kind of hacks? All code that is contributed to upstream must be maintained and as I see it: none else has possibility really use than your setup.

ramineni:
Agree with @zetaab . This looks more of hack specific to your setup, so unless there is general usability of the feature, it cant be merged upstream
/hold

PR #2067, opened 2023-01-03
[occm] Make sure LB status is updated immediately

### LINKED ISSUE #2066: [occm] LB `provisioningStatus` check always fails on LB creation

/kind bug

**What happened**:
`createFullyPopulatedOctaviaLoadBalancer()` ends with a call to `WaitLoadbalancerActive()`. It'll only exit without error if that function succeeded. But later on `ensureOctaviaLoadBalancer()` checks the LB's `ProvisioningStatus` again and exists if it's not ACTIVE. The problem is that the comparison is done on LB instance returned from POST call and not from one of the GET calls done in
`WaitLoadbalancerActive()`, so it's always `PENDING_CREATE`. This means that for LB to be completely created, `ensureOctaviaLoadBalancer()` always needs to be called twice, increasing number of Octavia calls made, time for LBs to be created and user experience by creating false positive warning Event.

In case multiple LBs are created in a short period - `Service.Status` of the first one will only be populated once all of them are created as events are processed in order. This increases the time needed to create LBs even more.

LB creation continues without error Event creation.

**How to reproduce it**:
Create LB, observe Events on Service.

**Environment**:
- Others:

dulek:
`createFullyPopulatedOctaviaLoadBalancer()` ends with a call to `WaitLoadbalancerActive()`. It'll only exit without error if that function succeeded. But later on `ensureOctaviaLoadBalancer()` checks the LB's `ProvisioningStatus` again and exits if it's not ACTIVE. The problem is that the comparison is done on LB instance returned from POST call and not from one of the GET calls done in
`WaitLoadbalancerActive()`, so it's always `PENDING_CREATE`. This means that for LB to be completely created, `ensureOctaviaLoadBalancer()` always needs to be called twice, increasing number of Octavia calls made, time for LBs to be created and user experience by creating false positive warning Event.

In case multiple LBs are created in a short period - `Service.Status` of the first one will only be populated once all of them are created as events are processed in order. This increases the time needed to create LBs even more.

This commit solves that by making sure `WaitLoadbalancerActive()` returns last instance of `LoadBalancer` struct it got from Octavia API and that that object is later used by `ensureLoadBalancer()` logic.

fixes #2066

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -549,7 +549,7 @@ func (lbaas *LbaasV2) createFullyPopulatedOctaviaLoadBalancer(name, clusterName 
 		svcConf.lbMemberSubnetID = loadbalancer.VipSubnetID
 	}
 
-	if err := openstackutil.WaitLoadbalancerActive(lbaas.lb, loadbalancer.ID); err != nil {
+	if loadbalancer, err = openstackutil.WaitLoadbalancerActive(lbaas.lb, loadbalancer.ID); err != nil {
 		return nil, err
 	}
 
--- a/pkg/util/openstack/loadbalancer.go
+++ b/pkg/util/openstack/loadbalancer.go
@@ -159,7 +159,7 @@ func IsOctaviaFeatureSupported(client *gophercloud.ServiceClient, feature int, l
 	return false
 }
 
-func WaitLoadbalancerActive(client *gophercloud.ServiceClient, loadbalancerID string) error {
+func WaitLoadbalancerActive(client *gophercloud.ServiceClient, loadbalancerID string) (*loadbalancers.LoadBalancer, error) {
 	klog.InfoS("Waiting for load balancer ACTIVE", "lbID", loadbalancerID)
 	backoff := wait.Backoff{
 		Duration: waitLoadbalancerInitDelay,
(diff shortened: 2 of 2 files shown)


zetaab:
looks ok for me

/approve

mandre (commented):
Makes sense to me.

pkg/openstack/loadbalancer.go
```diff
@@ -549,7 +549,7 @@ func (lbaas *LbaasV2) createFullyPopulatedOctaviaLoadBalancer(name, clusterName
 		svcConf.lbMemberSubnetID = loadbalancer.VipSubnetID
 	}
 
-	if err := openstackutil.WaitLoadbalancerActive(lbaas.lb, loadbalancer.ID); err != nil {
+	if loadbalancer, err = openstackutil.WaitLoadbalancerActive(lbaas.lb, loadbalancer.ID); err != nil {
```
jichenjc, line 552: I would prefer a quick update on the function name .. will submit a nit PR for this

dulek: You probably mean to make it `LoadBalancer`?

jichenjc: openstackutil.WaitLoadbalancerActive  => openstackutil.WaitLoadbalancerActiveAndGetLB

PR #2068, opened 2023-01-04
[OCCM]nit: rename function WaitLoadbalancerActive

jichenjc:
**What this PR does / why we need it**:

**Which issue this PR fixes(if applicable)**:
fixes #

**Special notes for reviewers**:

**Release note**:

```release-note
NONE
```

PR #2069, opened 2023-01-04
[occm] Rename WaitLoadbalancerActive

dulek:
As `WaitLoadbalancerActive()` function now returns the LB as well, it's renamed to `WaitLoadBalancerActiveAndGetLB()`.

```release-note
NONE
```


dulek:
Ah, there's #2068 already.

PR #2079, opened 2023-01-25
[occm] Allow user to specify common annotations

dioguerra:

When deploying all the services from one cluster with a central chart, it might be that applications can alter the lifecicle deployment of helm  due to the use of helm hooks. 

Some examples of this applications include:
- kube-prometheus-stack
- ingress-nginx

A chart cannot change it's subcharts hook implementations. Because occm is required to untaint nodes from `node.cloudprovider.kubernetes.io/uninitialized`, the other applications will stall helm deployment and eventually deployment will fail due to timeout (race condition)

One way to circunvent this, is by making occm be deployed as a critical component, previous to all other components by using a hook with higher priority. See [upstream docs](https://helm.sh/docs/topics/charts_hooks/#hooks-and-the-release-lifecycle)

To do this, all the components required for the occm app to work need to be launched as a helm hook resource and thus need to be annotated.

This patch allows a user to add commonAnnotations into all manifests. 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/openstack-cloud-controller-manager/Chart.yaml
+++ b/charts/openstack-cloud-controller-manager/Chart.yaml
@@ -4,7 +4,7 @@ description: Openstack Cloud Controller Manager Helm Chart
 icon: https://object-storage-ca-ymq-1.vexxhost.net/swift/v1/6e4619c416ff4bd19e1c087f27a43eea/www-images-prod/openstack-logo/OpenStack-Logo-Vertical.png
 home: https://github.com/kubernetes/cloud-provider-openstack
 name: openstack-cloud-controller-manager
-version: 1.4.0
+version: 1.5.0
 maintainers:
   - name: morremeyer
     email: kubernetes@maurice-meyer.de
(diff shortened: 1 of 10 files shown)


zetaab:
@dioguerra great! The problem that I have at least is that we are not using helm at all, so I am just trying to update those manifests when changes to happen, but no deeper idea how those should be done. So thanks for fixing these

dioguerra:
Tried to validate this on one of my clusters, and the weights are not working as expected or im missing something?
Follow: https://github.com/helm/helm/issues/11770 for info

dioguerra:
Hey, can we merge this? Tested this fully in my deployment and works like a charm :)

zetaab (approved):
/approve

zetaab:
@dioguerra can you rebase this against master branch to get test fixes?

so something like
```
git checkout master
git pull origin (modify if needed) master 
git checkout manifest-wide-annotations
git rebase master
git push origin (modify if needed) manifest-wide-annotations -f
```

dioguerra:
@zetaab done.

PR #2090, opened 2023-02-08
Optimize route reconciliation logic

### LINKED ISSUE #2089: [oocm] generates an excessive number of os-interface api requests
> /kind bug

**What happened**:

My first suggestion would be:

**How to reproduce it**:

**Environment**:

kayrus:

the `openstack.Routes`' `ListRoutes` method must not loop over all project servers just to resolve a couple of route's next hops to node names. This PR resolves the node name only when it's found in the route's next hop using cached k8s API calls for nodes listing.

fixes #2089 

**Release note**:

```release-note
Improve performance of the `ListRoutes` method, especially when an OpenStack project contains hundreds of non-k8s nodes
```

### DIFF
--- a/pkg/openstack/openstack.go
+++ b/pkg/openstack/openstack.go
@@ -34,6 +34,8 @@ import (
 	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/klog/v2"
 
+	"k8s.io/client-go/informers"
+	coreinformers "k8s.io/client-go/informers/core/v1"
 	"k8s.io/cloud-provider-openstack/pkg/client"
 	"k8s.io/cloud-provider-openstack/pkg/metrics"
 	"k8s.io/cloud-provider-openstack/pkg/util"
@@ -146,9 +148,11 @@ type OpenStack struct {
 	metadataOpts   metadata.Opts
 	networkingOpts NetworkingOpts
 	// InstanceID of the server where this OpenStack object is instantiated.
-	localInstanceID string
-	kclient         kubernetes.Interface
-	useV1Instances  bool // TODO: v1 instance apis can be deleted after the v2 is verified enough
+	localInstanceID       string
+	kclient               kubernetes.Interface
+	useV1Instances        bool // TODO: v1 instance apis can be deleted after the v2 is verified enough
+	nodeInformer          coreinformers.NodeInformer
+	nodeInformerHasSynced func() bool
 }
 
 // Config is used to read and store information from the cloud configuration file
@@ -455,13 +459,7 @@ func (os *OpenStack) Routes() (cloudprovider.Routes, bool) {
 		return nil, false
 	}
 
-	compute, err := client.NewComputeV2(os.provider, os.epOpts)
-	if err != nil {
-		klog.Errorf("Failed to create an OpenStack Compute client: %v", err)
-		return nil, false
-	}
-
-	r, err := NewRoutes(compute, network, os.routeOpts, os.networkingOpts)
+	r, err := NewRoutes(os, network)
 	if err != nil {
 		klog.Warningf("Error initialising Routes support: %v", err)
 		return nil, false
@@ -469,4 +467,12 @@ func (os *OpenStack) Routes() (cloudprovider.Routes, bool) {
 
 	klog.V(1).Info("Claiming to support Routes")
 	return r, true
+}
+
+// SetInformers implements InformerUser interface by setting up informer-fed caches to
+// leverage Kubernetes API for caching
+func (os *OpenStack) SetInformers(informerFactory informers.SharedInformerFactory) {
+	klog.V(1).Infof("Setting up informers for Cloud")
+	os.nodeInformer = informerFactory.Core().V1().Nodes()
+	os.nodeInformerHasSynced = os.nodeInformer.Informer().HasSynced
 }
--- a/pkg/openstack/routes.go
+++ b/pkg/openstack/routes.go
@@ -19,13 +19,14 @@ package openstack
 import (
 	"context"
 	"net"
+	"sync"
 
 	"github.com/gophercloud/gophercloud"
-	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/routers"
-	neutronports "github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
-	"github.com/gophercloud/gophercloud/pagination"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
 
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/cloud-provider-openstack/pkg/metrics"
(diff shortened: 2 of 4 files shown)


jichenjc:
still fixing gate, we can continue the code review before that..

zetaab:
/override openstack-cloud-csi-manila-e2e-test

manila is broken currently

kayrus:
Manila e2e tests are broken. What is the process for the PR review?

jichenjc:
/override openstack-cloud-csi-manila-e2e-test

that's something we need fix later, for now, let's ignore this

kayrus:
@jichenjc could you please review this PR? Or do I need to wait for assigned reviewers approval?

jichenjc:
no, I will read this along with other reviewers, will do soon, thanks for the reminder

kayrus:
@anguslees @FengyunPan2 @zetaab @jichenjc kindly ping

pkg/openstack/routes.go
```diff
@@ -180,22 +211,29 @@ func updateAllowedAddressPairs(network *gophercloud.ServiceClient, port *neutron
 
 // CreateRoute creates the described managed route
 func (r *Routes) CreateRoute(ctx context.Context, clusterName string, nameHint string, route *cloudprovider.Route) error {
-	klog.V(4).Infof("CreateRoute(%v, %v, %v)", clusterName, nameHint, route)
-
-	onFailure := newCaller()
+	r.Lock()
```
jichenjc, line 214: do we have any criteria for this lock? 
e.g what the resource action should be lock/unlock? e.g all POST/UPDATE operation?

pkg/openstack/openstack.go
```diff
@@ -461,3 +459,11 @@ func (os *OpenStack) Routes() (cloudprovider.Routes, bool) {
 	klog.V(1).Info("Claiming to support Routes")
 	return r, true
 }
+
+// SetInformers implements InformerUser interface by setting up informer-fed caches to
+// leverage Kubernetes API for caching
+func (os *OpenStack) SetInformers(informerFactory informers.SharedInformerFactory) {
```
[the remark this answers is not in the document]
kayrus: this is needed for https://github.com/kubernetes/cloud-provider/blob/11e49706691b05e67ac6ca4acae4a2607a739352/cloud.go#L71-L74 interface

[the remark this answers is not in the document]
jichenjc: ok, will read more about the inform logic in cloud provider , that's something not pay attention before , thanks for the link

jichenjc:
/lgtm

ok, might need more time to consume the code logic for informer and lock but let's get this merge to test
maybe @zetaab can help review as well

kayrus:
@zetaab kindly ping

pkg/openstack/routes.go
kayrus: @jichenjc gophercloud now supports atomic routes update. I submitted a new draft PR to use this feature: https://github.com/kubernetes/cloud-provider-openstack/pull/2134
It's based on this PR. Are there any updates on when this PR can be merged?

kayrus:
@jichenjc @zetaab I had to rebase

kayrus:
@jichenjc @zetaab kindly ping

kayrus:
cc @lingxiankong @anguslees @FengyunPan2

pkg/openstack/routes.go
```diff
@@ -35,97 +36,127 @@ import (
...
+	pages, err := ports.List(network, opts).AllPages()
+	if mc.ObserveRequest(err) != nil {
+		return nil, err
+	}
+	ports, err := ports.ExtractPorts(pages)
+	if err != nil {
+		return nil, err
+	}
+
+	var networkIDs []string
```
zetaab, line 117: this does not contain unique values for networkid?

pkg/openstack/routes.go
```diff
@@ -35,97 +36,127 @@ import (
...
+			TargetNode:      nodeName, //contains the nexthop address if node name was not found
 			Blackhole:       !foundNode,
 			DestinationCIDR: item.DestinationCIDR,
 		}
 		routes = append(routes, &route)
 	}
 
+	// detect router's private network ID for further VM ports filtering
+	r.Lock()
+	r.networkIDs, err = getRouterNetworkIDs(r.network, r.os.routeOpts.RouterID)
```
zetaab, line 94: is this really changing so often that we need to fetch this all the time? In general networks are not changing quite often? (spoiler: I am not using this feature at all)

zetaab (commented):
some comments, but like mentioned - I do not use this feature so I have not 100% understanding what is even needed

pkg/openstack/routes.go
kayrus: duplicate values are less probable, since router cannot handle the same network more than once.

pkg/openstack/routes.go
kayrus: There is a low probability that this may happen. Nevertheless, I decided to be on the safe side, since it's just one API call.

kayrus:
>  I do not use this feature so I have not 100% understanding what is even needed

hm. IMHO this is used in every openstack based k8s cluster, and the `router-id` is a required config option.

zetaab:
its not? example from kops https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/openstack/cloud.go#L835-L892

kayrus:
I was confused by https://github.com/kubernetes/cloud-provider-openstack/blob/31f475117755f5dcf55f89ab2f340b55c8ae32cb/pkg/openstack/openstack.go#L132 and https://github.com/kubernetes/cloud-provider-openstack/blob/31f475117755f5dcf55f89ab2f340b55c8ae32cb/pkg/openstack/routes.go#L49 comment. 

And apologies, I've never used kops.

zetaab:
Interesting, why that is required? I have used cloud provider openstack for 7 years without that. 

```
I0302 21:25:15.943948       1 openstack.go:439] openstack.Routes() called
W0302 21:25:16.355725       1 openstack.go:466] Error initialising Routes support: router-id not set in cloud provider config
```

but everything works like should?

kayrus:
> but everything works like should?

in my test environment the amount of API calls has been decreased by an order of magnitude, if not by two.

Could you please skip `openstack-cloud-csi-manila-e2e-test`?

zetaab:
/override openstack-cloud-csi-manila-e2e-test

yeah but I mean, the interesting question is that why that is marked as required if its not really needed.

kayrus:
> yeah but I mean, the interesting question is that why that is marked as required if its not really needed.

I'll cover this in my next PR: #2134

kayrus:
@zetaab @jichenjc what is the process for backporting this fix to earlier releases?

jichenjc:
just cherry pick ,but we ususally backport a bug instead of a feature ..

kayrus:
@jichenjc well, our API endpoints are overloaded by useless API requests, which cause nasty performance issues and sometimes outages. Let me try to cherry-pick and let's see how will it go.

pkg/openstack/routes.go
```diff
@@ -35,97 +36,127 @@ import (
...
+			for _, v := range node.Status.Addresses {
+				if v.Type == v1.NodeInternalIP {
+					ip := net.ParseIP(v.Address)
+					if ip == nil {
+						continue
+					}
+					if needIPv6 && ip.To4() == nil {
+						return v.Address
+					}
+					return v.Address
```
majewsky, line 152: This if-branch is ineffective because the return value is the same regardless of whether the condition is true.

kayrus: Thanks, good catch!
Fixed in atomic routes PRs.

PR #2121, opened 2023-02-20
[occm] Reintroduce seeding of math/rand source

### LINKED ISSUE #2120: [occm] Rand not being initialised any more with a seed
/kind bug

**What happened**:
https://github.com/kubernetes/cloud-provider-openstack/pull/2100 changed the use of `rand.Seed`, which seeds the global PRNG, with `rand.NewSource`, that **returns** a new initialised PRNG. With that change, that line becomes a noop (because the generated PRNG is never assigned to a variable). As a result, as stated by [the documentation of math/rand](https://pkg.go.dev/math/rand@go1.19.6#Seed):

> If Seed is not called, the generator behaves as if seeded by Seed(1)

As a consequence, all random numbers generated by the binary will always be the same in the same order, which doesn't look like the expected behaviour of the PR.

https://github.com/kubernetes/cloud-provider-openstack/blob/1bf5a4a67701234ccdf88621ffb6078bccdee16b/cmd/openstack-cloud-controller-manager/main.go#L47

`rand.Seed(time.Now().UnixNano())` is left untouched, or is somehow replaced in the whole binary codebase by using the result of `rand.NewSource(time.Now().UnixNano())`.

**How to reproduce it**:

**Environment**:
- OpenStack version:
- Others:

pierreprinetti:
In Go v1.20, `rand.Seed` is deprecated and `math/rand`'s is automatically seeded. However, that doesn't happen in Go v1.19 that is used to build this branch.

With this change, the global source for `math/rand` is seeded again in `main.go`.

In a separate commit, this PR bumps `golangci-lint` to v1.51.2, which handles the deprecation more gracefully.

Partially reverts https://github.com/kubernetes/cloud-provider-openstack/pull/2102

rand.Seed was prematurely removed in a previous commit.

fixes #2120

**Release note**:
```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,7 +155,7 @@ build-cmd-%: work $(SOURCES)
 test: unit functional
 
 check: work
-	go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.51.1 run ./...
+	go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.51.2 run ./...
 
 unit: work
 	go test -tags=unit $(shell go list ./... | sed -e '/sanity/ { N; d; }' | sed -e '/tests/ {N; d;}') $(TESTARGS)
--- a/cmd/openstack-cloud-controller-manager/main.go
+++ b/cmd/openstack-cloud-controller-manager/main.go
@@ -44,7 +44,7 @@ import (
 )
 
 func main() {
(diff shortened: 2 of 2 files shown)


zetaab (approved):
/approve
/lgtm

Issue #1540, opened 2021-05-24
[all] Consider switching to (mirroring to) GCR


ialidzhikov:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind feature
/sig cloud-provider
/area provider/openstack

**What happened**:

Currently cloud provider openstack related images are maintained in Docker Hub. 

In some environments we face Docker Hub rate limits while pulling OCCM or the Cinder CSI driver:

```
  Normal   Pulling    12s              kubelet            Pulling image "docker.io/k8scloudprovider/cinder-csi-plugin:v1.19.0"
  Warning  Failed     11s              kubelet            Failed to pull image "docker.io/k8scloudprovider/cinder-csi-plugin:v1.19.0": rpc error: code = Unknown desc = Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
  Warning  Failed     11s              kubelet            Error: ErrImagePull
```

Last year Docker Hub restricted the amount of image pulls for anonymous and free users. From https://www.docker.com/increase-rate-limits:

> On November 20, 2020, rate limits anonymous and free authenticated use of Docker Hub went into effect. Anonymous and Free Docker Hub users are limited to 100 and 200 container image pull requests per six hours.

After this announcement multiple projects switched to GCR. For example as a mitigation to the Docker Hub rate limits, Istio provides official mirrors to GCR - see https://istio.io/latest/blog/2020/docker-rate-limit/.

I am not sure why cloud provider openstack maintains the images in Docker Hub, probably there are multiple reasons that I am not aware of. From outside, as user of the cloud provider openstack images, it looks a little bit confusing. Currently main K8s components are maintained under GCR (for example `k8s.gcr.io/kube-apiserver`, `k8s.gcr.io/kube-controller-manager`, etc.). I see that K8s CSI sidecars relatively recently switched to GCR as well (`k8s.gcr.io/sig-storage/csi-provisioner`, `k8s.gcr.io/sig-storage/csi-attacher`, etc.). I see that the AWS EBS CSI driver is maintained under `k8s.gcr.io` as well - `k8s.gcr.io/provider-aws/aws-ebs-csi-driver`.
I am not sure whether there is central initiative to maintain the images in the official K8s GCR. But it would be great if cloud provider openstack images are maintained in GCR as well. In this way end users won't need to tackle Docker Hub rate limit issues.

[...]

PR #2125, opened 2023-02-21
Add cloudbuild.yaml for staging

mnaser:
Adds a `cloudbuild.yaml` so we can start pushing things out of DockerHub

related to https://github.com/kubernetes/cloud-provider-openstack/issues/1540

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,23 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
+
+# this must be specified in seconds. If omitted, defaults to 600s (10 mins)
+timeout: 1200s
+# this prevents errors if you don't use both _GIT_TAG and _PULL_BASE_REF,
+# or any new substitutions added in the future.
+options:
+  substitution_option: ALLOW_LOOSE
+steps:
+  - name: gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20220609-2e4c91eb7e
+    entrypoint: make
+    env:
+    - REGISTRY=gcr.io/k8s-staging-cloud-provider-os
+    - VERSION=$_GIT_TAG
+    args:
+    - upload-images
+substitutions:
+  # _GIT_TAG will be filled with a git-based tag for the image, of the form
+  # vYYYYMMDD-hash, and can be used as a substitution
+  _GIT_TAG: '12345'
+  # _PULL_BASE_REF will contain the ref that was pushed to to trigger this
+  # build - a branch like 'master' or 'release-0.2', or a tag like 'v0.2'.
+  _PULL_BASE_REF: 'master'


jichenjc:
could you please help elaborate the way to solve the issue you pointed out with this PR? Thanks 

@mnaser

zetaab:
@mnaser could you please open little bit how this is supposed to work? As we are using github actions in this repo and so on, we do not have any cloudbuild access.

cloudbuild.yaml
```diff
@@ -0,0 +1,23 @@
...
+timeout: 1200s
+# this prevents errors if you don't use both _GIT_TAG and _PULL_BASE_REF,
+# or any new substitutions added in the future.
+options:
+  substitution_option: ALLOW_LOOSE
+steps:
+  - name: gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20220609-2e4c91eb7e
+    entrypoint: make
+    env:
+    - REGISTRY=gcr.io/k8s-staging-cloud-provider-os
```
mdbooth, line 13: AFAICT this is the correct registry, defined here: https://github.com/kubernetes/k8s.io/blob/main/k8s.gcr.io/manifests/k8s-staging-cloud-provider-os/promoter-manifest.yaml

cloudbuild.yaml
```diff
@@ -0,0 +1,23 @@
...
+    - VERSION=$_GIT_TAG
+    args:
+    - upload-images
+substitutions:
+  # _GIT_TAG will be filled with a git-based tag for the image, of the form
+  # vYYYYMMDD-hash, and can be used as a substitution
+  _GIT_TAG: '12345'
+  # _PULL_BASE_REF will contain the ref that was pushed to to trigger this
+  # build - a branch like 'master' or 'release-0.2', or a tag like 'v0.2'.
+  _PULL_BASE_REF: 'master'
```
mdbooth, line 23: We're not using this. However, as it seems to be a standard substitution probably simpler to leave it in to reduce friction if we want to use it in the future.

mdbooth (commented):
Here's what we currently do for release builds:

https://github.com/kubernetes/cloud-provider-openstack/blob/847eadddd838a6abad23a12b5f96a41dc52085a2/.github/workflows/release-cpo.yaml#L36

This looks correct to me and I recall this being a required part of the release puzzle. I can't find the docs, though. A link in the PR would be nice.

/lgtm

zetaab:
I can easily add our images to k8s.io registry. However, I do not have credentials or contact where to ask those. So if someone could help with that.

mdbooth:
> I can easily add our images to k8s.io registry. However, I do not have credentials or contact where to ask those. So if someone could help with that.

I think Prow already has them? IIUC they are managed by the k8s org.

mdbooth:
@zetaab @mnaser Found the reference: https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md

E.g. Here's how it works in CAPO, which is in kubernetes-sigs but it works the same way:

There's a postsubmit job which runs after commits to main or release branches: https://github.com/kubernetes/test-infra/blob/c7adcaabf4800775a2b3ccfdff5545c4119a5c43/config/jobs/image-pushing/k8s-staging-cluster-api.yaml#L184-L208

It automatically pushes a new image to the staging repo by running the cloud build. Maintainers then promote staging images to production manually by pushing a PR to update https://github.com/kubernetes/k8s.io/blob/main/k8s.gcr.io/images/k8s-staging-capi-openstack/images.yaml

It's documented in our release doc: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/RELEASE.md

mdbooth:
> I can easily add our images to k8s.io registry. However, I do not have credentials or contact where to ask those. So if someone could help with that.

Incidentally, I believe the answer to this is suggested in the image-pushing job:

https://github.com/kubernetes/test-infra/blob/e8be84d7787b4864ac5b53130defa743625eecc9/config/jobs/image-pushing/k8s-staging-cloud-provider-openstack.yaml#L5

```yaml
      cluster: k8s-infra-prow-build-trusted
```

My guess is that the reason you aren't permitted to run arbitrary code in the trusted cluster is that it has these credentials. Guessing this also implies they won't share them except via prow.

cloudbuild.yaml
jichenjc: seems correct? CPO didn't switch to main currently, still using master as primary dev branch..

jichenjc:
@mdbooth thanks for the detailed info

To confirm my understanding, So we need get this PR merged then we have a pre-requirement satisfied 
then get https://github.com/kubernetes/test-infra/pull/28817/files merged 
after that we will have a similar output / method of release corresponding to CAPO ?

mnaser:
Based on my understanding is that yes, this is the first step to being able to upload into the staging repo.

mdbooth:
@mnaser Probably want to remove the word 'fixes' before https://github.com/kubernetes/cloud-provider-openstack/issues/1540 because IIUC it'll close the issue again. I think this is just step 1.

jichenjc:
I updated the commit message and let's get this

PR #2134, opened 2023-02-27
occm: implement an optional atomic routes update

### LINKED ISSUE #2089: [oocm] generates an excessive number of os-interface api requests

kayrus:

An extra improvement for the route reconciliation logic. This PR allows to use a single API call to add or remove routes instead of fetching existing routes and combining them with new or removing old routes with a further update. This change allows concurrent calls without consistency issues.

fixes #2089 
based on #2090

**Release note**:

```release-note
Implement an optional `extraroute-atomic` boolean flag for more efficient routes updates.
```


pkg/openstack/openstack.go
```diff
@@ -129,7 +131,8 @@ type NetworkingOpts struct {
 
 // RouterOpts is used for Neutron routes
 type RouterOpts struct {
-	RouterID string `gcfg:"router-id"` // required
+	RouterID         string `gcfg:"router-id"` // required
```
zetaab, line 134: in my opinion this `required` should be removed because its not really required. Like mentioned in another PR - we have been using OCCM for 7 years without using this at all.

zetaab:
please rebase this after that another PR is merged

kayrus:
This is still WIP. I'm in contact with our neutron team. I'd like to remove the manual config toggle, based on extensions API. So there should be no need for an extra option.
I'll also remove the `required` mark from the struct member.

zetaab:
@kayrus if you have time could you explain why this routes feature is needed? Our architecture has always been that we have router in network with dhcp and router will handle the traffic automatically, no need to add routes.

kayrus:
@zetaab I updated my comment above. See the bold text.

docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
```diff
@@ -174,6 +174,13 @@ The options in `Global` section are used for openstack-cloud-controller-manager
   For example, this option can be useful when having multiple or dual-stack interfaces attached to a node and needing a user-controlled, deterministic way of sorting the addresses.
   Default: ""
 
+### Router
+
+* `router-id`
+  Required. Specifies the Neutron router ID to manage Kubernetes cluster routes.
+* `extraroute-atomic`
```
jichenjc, line 181: maybe in the future we can check through `neutron ext-list` to detect whether it's enabled or not ?

kayrus: that's what I'm investigating right now.

kayrus: done. my tests show overall routes reconciliation latency improvements from 2-5 minutes down to 10 seconds for a cluster of 16 nodes.

ready for review.

kayrus:
@zetaab @jichenjc ready for review

kayrus:
@zetaab @jichenjc kindly ping
see also #2150 and #2152.

zetaab (approved):
/approve
/lgtm

PR #2139, opened 2023-03-02
[all] Use multi-stage docker builds and buildx for multi-arch

### LINKED ISSUE #2136: [all] cloudbuild failing
Example: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/cloud-provider-openstack-push-images/1630385626224267264

The immediate cause seems to be an incorrect go version, but as the build method looks highly unusual it is probably simpler to move to a more orthodox build method. Specifically, the docker builds should be self-contained. They currently build a rootfs.tar from the Makefile and copy it into an alpine image.

I suggest copying the structure of a Dockerfile from another project would be simplest. We will likely have to update cloudbuild.yaml afterwards as well.

This impacts #1540

mdbooth:
This is a rewrite of the image build process. The primary motivation for it is in support of https://github.com/kubernetes/cloud-provider-openstack/issues/1540. It makes 2 primary changes:

* Container builds are self-contained multi-stage builds
* Multi-arch builds use docker buildx

The former greatly simplifies the build process. It makes the build independent of the CI image executing the job, which in turns makes the builds easier to integrate into other CI systems. In this case, the primary motivator is k8s prow.

The latter is a simplification enabled by the former. As both github actions and k8s prow already support buildx for multi-arch it allows us to remove the custom scaffolding in the Makefile.

There are some 'user-facing' changes to the Makefile with many targets removed and a few added.

* Targets removed:
  * image-* (individual targets for each image)
  * push-manifest-%
  * push-all-manifest
  * build-images
  * push-images
  * image-%
  * push-image-%
  * images
  * images-all-arches
  * build-arch-image-%
  * upload-image-%
  * upload-images
  * build-cmd-%
* Targets added:
  * build-local-image-%
  * build-local-images
  * push-multiarch-image-%
  * push-multiarch-images

The 2 new build targets are `build-local-image` and `push-multiarch-image`. The former is intended for local development only. It builds a target for the default local architecture. The resulting image is available in the local docker and is not pushed. The latter builds a target for all defined architectures and pushes to remote repository. Note that there is not currently a build-multiarch-image because buildx doesn't support pushing a multiarch build to the local docker. It's also probably not useful in practise.

There are also some changes to the resulting images. Firstly we no longer produce arch-specific images called `[target]-[arch]` e.g. `openstack-cloud-controller-manager-arm64`. The only image produced is called `[target]` e.g. `openstack-cloud-controller-manager` and it supports all architectures. This is a breaking change for anybody using the arch-specific images, but the only required change is to remove the arch component from the image name. The correct arch-specific image will be pulled automatically.

The following images change from Alpine to Distroless:
* k8s-keystone-auth
* magnum-auto-healer
* octavia-ingress-controller
* openstack-cloud-controller-manager

barbican-kms-plugin continues to use Alpine because it uses a shell-form CMD.

cinder-csi-plugin continues to use Debian because it installs multiple additional packages.

manila-csi-plugin continue to use Alpine because it installs jq and curl.

Image CMDs and ENTRYPOINTs are unchanged, but note these have become inconsistent. This might be worth a separate cleanup, but that would be a breaking change.

fixes #2136

**Release note**:

```release-note
* Images are now available at registry.k8s.io/k8s-staging-cloud-provider-os for development builds and registry.k8s.io/cloud-provider-os for releases.
* Single architecture images are no longer provided. All images are multi-arch.
```


zetaab:
/test openstack-cloud-controller-manager-e2e-test

Thanks for making PR! Tried to look can I see anything to mention, but I could not. I am thinking could we just make some "test release" to test the CI

mdbooth:
I'll try to take a look at the test failures tomorrow.

I'll also have a think about how we can test the github action. Possibly on a fork?

mdbooth:
Unless I'm missing it, csi-manila and csi-cinder didn't get as far as building an image. Possibly an unrelated issue?

mdbooth:
Looks like OCCM failure was because I accidentally changed CMD to ENTRYPOINT 🤦

mdbooth:
Lets try these again:

mdbooth:
```
goroutine 1 [running]:
k8s.io/component-base/metrics.parseVersion({{0x0, 0x0}, {0x0, 0x0}, {0x192c3f0, 0x4}, {0x16de09d, 0xb}, {0x0, 0x0}, ...})
	k8s.io/component-base@v0.26.0/metrics/version_parser.go:47 +0x274
k8s.io/component-base/metrics.newKubeRegistry({{0x0, 0x0}, {0x0, 0x0}, {0x192c3f0, 0x4}, {0x16de09d, 0xb}, {0x0, 0x0}, ...})
	k8s.io/component-base@v0.26.0/metrics/registry.go:349 +0x119
k8s.io/component-base/metrics.NewKubeRegistry()
	k8s.io/component-base@v0.26.0/metrics/registry.go:363 +0x78
k8s.io/component-base/metrics/legacyregistry.init()
	k8s.io/component-base@v0.26.0/metrics/legacyregistry/registry.go:30 +0x1d
panic: version string "2139" doesn't match expected regular expression: "^v(\d+\.\d+\.\d+)"
```

Looks like it doesn't like the embedded version.

mdbooth:
Issues resolved (so far, anyway) in tests:
* The manila test seems to be broken anyway. I've done my best to update it correctly,  but haven't attempted to fix the underlying issues.
* The tests were previously explicitly referencing a `-amd64` image which we no longer build.
* The tests were previously building the cinder and manila csi images with a VERSION that wasn't a valid semver. This breaks now we're setting it in `version.gitVersion` in the build. I have updated these tests to set the same VERSION as was already used by openstack-cloud-controller-manager, which is a valid semver.

zetaab:
@mdbooth `latest` would be nice from master, but if it takes like one hour always to build all images its too much (old solution takes something like 35-40 mins to build all images)

mdbooth:
@zetaab I pushed some custom changes to my fork and used that to push images to `quay.io/mbooth/*` using the github release workflow. All the images should be public.

The branch is here: https://github.com/mdbooth/cloud-provider-openstack/tree/dockerbuild-test

I haven't done any testing or even cursory inspection on them except to note that they exist.

mdbooth:
> @mdbooth `latest` would be nice from master, but if it takes like one hour always to build all images its too much (old solution takes something like 35-40 mins to build all images)

Worried about running out of GH actions quota?

zetaab:
I do not know is there quota :)

jichenjc:
>>>The manila test seems to be broken anyway. I've done my best to update it correctly, but haven't attempted to fix the underlying issues.

ignore this for now, there are various issue for this already tracking now 

>>>Worried about running out of GH actions quota?

I knew GH has quota on API request, but seems our previous build process do not have such

cluster/images/barbican-kms-plugin/Dockerfile
```diff
@@ -10,23 +10,67 @@
...
+# barbican-kms-plugin uses ALPINE instead of distroless because its entrypoint
+# uses a shell for environment substitution. If there are no other uses this
+# could be replaced by callers passing arguments explicitly.
+
+FROM --platform=${TARGETPLATFORM} ${ALPINE_IMAGE}
+
+ARG COMPONENT
+
+COPY --from=builder /build/$COMPONENT /bin/$COMPONENT
+COPY --from=certs /etc/ssl/certs /etc/ssl/certs
```
jichenjc, line 65: I compared previous code a little bit ,just curious ,the reason of copy those certs is because alpine doen'st have those certs so we have to copy for ssl connection?

mdbooth: Actually I originally wrote it for the distroless occm image. I kept it the same for all images just to keep it the same, tbh. I was thinking of a potential situation where Debian and Alpine CA are briefly out of sync and occm can connect to OpenStack but cinder-csi can't. I figured they could all have exactly the same ca-certs from the same source copied in the same way.

mdbooth:
I might be able to DRY this a bit by making a single Dockerfile with multiple targets. This could also speed the build up slightly.

mdbooth:
I've refactored all the Dockerfiles into a single Dockerfile. This has the dual advantages of de-duplicating them, as well as significantly improving the build speed. All images and archs built in github actions in under 20 minutes.

mdbooth:
Also, I'll just leave this here:

```
+264 −496
```

mdbooth:
There's a bug in that last build, btw: the arm64 cinder-csi-plugin image seems to have been built from the amd64 debian image. I don't think I should be declaring the default TARGETPLATFORM like that at the top level. I'm going to take it out and re-test.

mdbooth:
I've fixed the TARGETPLATFORM issue and tested locally that both the built binary and an installed executable are foreign when built for linux/arm64:

```
> docker run --platform linux/arm64 --rm -ti quay.io/mbooth/cinder-csi-plugin:v1.26.1-mdbooth-test6-dirty
exec /bin/cinder-csi-plugin: exec format error
> docker run --platform linux/arm64 --rm -ti quay.io/mbooth/cinder-csi-plugin:v1.26.1-mdbooth-test6-dirty /bin/mount
exec /bin/mount: exec format error
```

Dockerfile
```diff
@@ -0,0 +1,200 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+################################################################################
+##                               BUILD ARGS                                   ##
+################################################################################
+# This build arg allows the specification of a custom Golang image.
+ARG GOLANG_IMAGE=golang:1.19.6
```
wwentland, line 17: Could use `golang@sha256:1a86aa60f99a253ad32ec0e8820f5813d557b1735ac5cc32e042397379d57fb2` here

Dockerfile
```diff
@@ -0,0 +1,200 @@
...
+
+# The distroless image on which the CPI manager image is built.
+#
+# Please do not use "latest". Explicit tags should be used to provide
+# deterministic builds. Follow what kubernetes uses to build
+# kube-controller-manager, for example for 1.23.x:
+# https://github.com/kubernetes/kubernetes/blob/release-1.24/build/common.sh#L94
+ARG DISTROLESS_IMAGE=k8s.gcr.io/build-image/go-runner:v2.3.1-go1.19.6-bullseye.0
```
wwentland, line 25: Explicit digest: `k8s.gcr.io/build-image/go-runner@sha256:fb66eab3c3723d9706971ed37c29dfea5addc945c7e2b66b60b76bbf8186c71c`

Dockerfile
```diff
@@ -0,0 +1,200 @@
...
+
+# We use Alpine as the source for default CA certificates and some output
+# images
+ARG ALPINE_IMAGE=alpine:3.15.4
```
wwentland, line 29: By digest: `alpine@sha256:4edbd2beb5f78b1014028f4fbb99f3237d9561100b6881aabbf5acce2c4f9454`

Dockerfile
```diff
@@ -0,0 +1,200 @@
...
+
+# cinder-csi-plugin uses Debian as a base image
+ARG DEBIAN_IMAGE=registry.k8s.io/build-image/debian-base:bullseye-v1.4.3
```
wwentland, line 32: By digest: `registry.k8s.io/build-image/debian-base@sha256:3fcf9bcc4949392155cde1f96591ba3eda092f4735a9cc856bf526d71f48af47`

wwentland (commented):
I like these changes, as the build process is *much* clearer with this.

I am, however, a bit surprised by the number of base images that are in use here and was wondering if it would make sense to standardise on a single one with `scratch` being my preferred option.

I'd also suggest to reference images by digest rather than tags and have added appropriate comments.

mdbooth:
> I like these changes, as the build process is _much_ clearer with this.
> 
> I am, however, a bit surprised by the number of base images that are in use here and was wondering if would make sense to standardise on a single one with `scratch` being my preferred option.
> 
> I'd also suggest to reference images by digest rather than tags and have added comments with the them.

I agree with nearly all[1] of those suggestions and have more wrt our inconsistent use of CMD and ENTRYPOINT for the various images.

However, my goal here is to fix the build system whilst making minimal substantive changes to the resulting images, so I'd very much like to push that work to other PRs.

[1] Using the k8s project distroless image probably makes the most sense here, if only for consistency within the project.

wwentland:
> However, my goal here is to fix the build system whilst making minimal substantive changes to the resulting images, so I'd very much like to push that work to other PRs.

Certainly, happy to follow up on this PR with one that references images by digest. The build process is certainly a lot clearer.

jichenjc:
/lgtm

I think we should be good to go for now ,if anything need improve, let's continue add PR

PR #2141, opened 2023-03-03
Optimize route reconciliation logic cherry-pick to 1.26 (#2090)

### LINKED ISSUE #2089: [oocm] generates an excessive number of os-interface api requests

kayrus:

Cherry pick of the #2090 to 1.26 release

fixes #2089

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/instances.go
+++ b/pkg/openstack/instances.go
@@ -504,14 +504,6 @@ func mapNodeNameToServerName(nodeName types.NodeName) string {
 	return string(nodeName)
 }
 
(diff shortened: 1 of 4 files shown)


kayrus:
@zetaab @jichenjc could you please trigger the tests?

kayrus:
@zetaab @jichenjc I don't know why do tests fail, but could you please approve this PR?

PR #2145, opened 2023-03-07
[all] Fix cloudbuild

### LINKED ISSUE #2143: [all]: push images job fails
/kind bug

**What happened**:

push images job fails: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/cloud-provider-openstack-push-images/1631431436202414080

push images should work without issues

**How to reproduce it**:

See recent commits status: https://github.com/kubernetes/cloud-provider-openstack/commits/master

mdbooth:
cloudbuild is failing, e.g.: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/cloud-provider-openstack-push-images/1632991558871552000

This PR addresses 2 issues with cloudbuild.

The immediate cause of the above error is a quirk in the build environment: HOME is set to /builders/root but buildx is installed in /root, so we need to set HOME correctly for buildx to work. [Discussion](https://kubernetes.slack.com/archives/CCK68P2Q2/p1678184220721889).

Secondly we need to initialise the buildx environment before running make. If we were using the image's buildx entrypoint it would do this for us. We don't want to do that, though: because we re-use the compilation build stage for all image targets, building them all in separate builders would mean we could not take advantage of caching this stage, which would increase build times by approximately the number of images, which is currently 7. To re-use the initialisation which the image owners maintain in that entrypoint we call it once manually before invoking make.

fixes #2143

I have managed to give this a measure of local testing by running the following:

```
> docker run --privileged --rm -ti \
    --entrypoint bash -e HOME=/root \
    -v (pwd):/workspace \
    -v /var/run/docker.sock:/var/run/docker.sock \
    --entrypoint bash \
    gcr.io/k8s-testimages/gcb-docker-gcloud@sha256:13bb9f426fba44b7e482c19f62796c44a1598d5bc9617aaccea1aa8f6086d1ae \
  -c "
      # Run the image's buildx entrypoint to initialise the build environment
      # appropriately for the image before running make
      /buildx-entrypoint version

      make push-multiarch-images \
        REGISTRY=gcr.io/$PROJECT_ID \
        VERSION=$_SHORT_TAG
"
```

This runs far enough to fail when it doesn't have credentials to push to the registry, so for the first time in this series of PRs I am hopeful it might actually work.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,8 +32,7 @@ TAR_FILE	?= rootfs.tar
 
 GOOS		?= $(shell go env GOOS)
 GOPROXY		?= $(shell go env GOPROXY)
-GIT_VERSION     := $(shell git describe --dirty --tags --match='v*')
-VERSION         ?= $(GIT_VERSION)
+VERSION         ?= $(shell git describe --dirty --tags --match='v*')
 GOARCH		:=
 GOFLAGS		:=
 TAGS		:=
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -9,16 +9,29 @@ options:
   machineType: 'N1_HIGHCPU_8'
 steps:
   - name: gcr.io/k8s-testimages/gcb-docker-gcloud
-    entrypoint: make
+    entrypoint: bash
     env:
-    - REGISTRY=gcr.io/$PROJECT_ID
-    - VERSION=$_GIT_TAG
+    # Required because the default HOME is /builder/root but buildx is
+    # installed in /root/.docker. Not setting this results in docker not
+    # finding buildx during make.
+    - HOME=/root
     args:
-    - push-multiarch-images
+    - -c
+    - |
+      # Run the image's buildx entrypoint to initialise the build environment
+      # appropriately for the image before running make
+      /buildx-entrypoint version
+
+      make push-multiarch-images \
+        REGISTRY=gcr.io/$PROJECT_ID \
+        VERSION=$_SHORT_TAG
 substitutions:
   # _GIT_TAG will be filled with a git-based tag for the image, of the form
   # vYYYYMMDD-hash, and can be used as a substitution
-  _GIT_TAG: '12345'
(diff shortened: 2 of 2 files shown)


mdbooth:
@zetaab @jichenjc This could be the one 🤞

mdbooth:
Eurgh, nope: https://storage.googleapis.com/kubernetes-jenkins/logs/cloud-provider-openstack-push-images/1633087765925072896/build-log.txt

Looks like VERSION wasn't passed correctly.

zetaab:
@mdbooth ERROR: invalid tag "gcr.io/k8s-staging-provider-openstack/openstack-cloud-controller-manager:": invalid reference format

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/cloud-provider-openstack-push-images/1633087765925072896/

mdbooth:
This is the cloudbuild command:
```
gcloud builds submit --verbosity info --config /home/prow/go/src/github.com/kubernetes/cloud-provider-openstack/cloudbuild.yaml --substitutions _PULL_BASE_REF=master,_GIT_TAG=v20230307-v1.26.0-27-g61713e20 --project k8s-staging-provider-openstack --gcs-log-dir gs://k8s-staging-provider-openstack-gcb/logs --gcs-source-staging-dir gs://k8s-staging-provider-openstack-gcb/source gs://k8s-staging-provider-openstack-gcb/source/204ff3b2-4452-410e-b0ab-d236b2dc098b.tgz
```

We can see here why we need the _SHORT_TAG substitution, btw: if we pass `v20230307-v1.26.0-27-g61713e20` directly it will fail the regexp in k8s.io/component-base.

mdbooth:
Unfortunately I don't have access to a GCE project which can submit cloud builds, so I'm not sure how I can test this other than raising PRs and hoping.

mdbooth:
I've asked for help: https://kubernetes.slack.com/archives/CCK68P2Q2/p1678196100322849

PR #2151, opened 2023-03-08
[occm] Don't seed math/rand

mdbooth:
As the linter notes, `rand.Seed` is deprecated. We don't use `math/rand` directly, and if we did we would use `crypto/rand` instead if we needed cryptographically unpredictable numbers.

**Release note**:

```release-note
NONE
```


mdbooth:
This is going to have a merge conflict when https://github.com/kubernetes/cloud-provider-openstack/pull/2149 lands, btw.

PR #2154, opened 2023-03-09
[all] Switch cloudbuild image to gcr.io/cloud-builders/docker

mdbooth:
This PR makes 2 principal changes. Firstly it moves us to a vanilla docker cloud-builder image as it already includes everything we need. gcr.io/k8s-testimages/gcb-docker-gcloud also includes go, which we don't need because we have a completely containerized build. This allows us to ignore the quirks of how buildx is installed in gcb-docker-gcloud.

Secondly it adds an explicit `gcloud auth configure-docker`. This will hopefully fix the docker authentication error, and seems to be widely used in other k8s images, e.g. https://github.com/kubernetes/release/blob/b1394e2f1d533de1cb30307b1b3bd74ab7ae1a4c/images/build/cross/cloudbuild.yaml#L31-L32

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -9,21 +9,19 @@ options:
   substitution_option: ALLOW_LOOSE
   machineType: 'N1_HIGHCPU_8'
 steps:
-  - name: gcr.io/k8s-testimages/gcb-docker-gcloud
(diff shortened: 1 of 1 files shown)


zetaab (approved):
/approve
/lgtm

PR #2161, opened 2023-03-13
[all] Tag pushes to master as :latest

mdbooth:
When the cloudbuild job pushes to staging it will tag the image with the `git describe` of the commit. With this PR, pushes to the `master` branch will additionally be tagged with `latest`.

The change contains logic to detect pushes to `master` which is technically redundant. We could simply hardcode this in `cloudbuild.yaml` in the master branch and either remove it or hardcode something else when creating a release branch. I am personally undecided what the best approach is.

**Release note**:

```release-note
The lastest images pushed to master are now available from registry.k8s.io/k8s-staging-provider-os/[image-name]:latest
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,9 @@ TAR_FILE	?= rootfs.tar
 GOOS		?= $(shell go env GOOS)
 GOPROXY		?= $(shell go env GOPROXY)
 VERSION         ?= $(shell git describe --dirty --tags --match='v*')
+# IMAGE_TAGS is a space separated list of one or more tags to apply to built
+# images. Unless overridden it will be set the value of VERSION.
+IMAGE_TAGS	?= $(VERSION)
 GOARCH		:=
 GOFLAGS		:=
 TAGS		:=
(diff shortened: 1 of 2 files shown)


zetaab:
`latest` tags are not supported in registry.k8s.io, they do not allow mutable tags. However, the latest can be available in gcr.io registry. However, not sure is it good practice that we will share registry.k8s.io for stable releases and then gcr.io for latest?

mdbooth:
> `latest` tags are not supported in registry.k8s.io, they do not allow mutable tags. However, the latest can be available in gcr.io registry. However, not sure is it good practice that we will share registry.k8s.io for stable releases and then gcr.io for latest?

Oh, interesting. I only thought of this at all because [cloud-provider-aws adds one](https://github.com/kubernetes/cloud-provider-aws/blob/982e654c1a637dde00f397309a9f56eec6069025/cloudbuild.yaml#L14), but I agree the combination of registries is untidy. Do you have a docs reference for that, btw?

FWIW I'm ambivalent about this change and don't mind just dropping it. It doesn't add much complexity, but if we don't think anybody's actually going to use it we shouldn't add it.

mdbooth:
> I do not have docs but https://kubernetes.slack.com/archives/CCK68P2Q2/p1678370317184379?thread_ts=1678364564.820099&cid=CCK68P2Q2
> 
> yeah AWS is building `latest` to gcr. However, they do not have `latest` through registry.k8s.io https://github.com/kubernetes/k8s.io/blob/main/k8s.gcr.io/images/k8s-staging-provider-aws/images.yaml
> 
> but we can do it in similar way if we want. We just must use gcr.io with latest instead of registry.k8s.io
> 
> My opinion is that `latest` is not great thing because some people are just fetching manifests from `master` and then the situation is that they have production running with latest tag. However, some people could benefit of having possibility to verify images before real releases. But I think that can be done already? Build logs have the image information
> 
> https://storage.googleapis.com/kubernetes-jenkins/logs/cloud-provider-openstack-push-images/1633894945653788672/artifacts/build.log
> 
> for instance current master is `gcr.io/k8s-staging-provider-os/openstack-cloud-controller-manager:v1.27.0-alpha.0-1-gb95c82d6`

Ok, I'll drop this. It's very easy to add if we find a definite use case for it.

PR #2166, opened 2023-03-15
[octavia-ingress-controller] Add annotations to keep floating IP and/or specify an existing floating IP

ccleouf66:

Add features:
- allow to keep floating IP when delete ingress
- allow to use an existing available floating **IP**

none

The doc is also updated to indicated how to.
During the ingress creation only add two annotations inspired from [openstack-cloud-controller](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md#creating-service-by-specifying-a-floating-ip):
```yaml
annotations:
  octavia.ingress.kubernetes.io/keep-floatingip: "true"        # floating ip will not be deleted when ingress is deleted
  octavia.ingress.kubernetes.io/floatingip: "122.112.219.229"  # define the floating to use 
```

keep-floatingip:  default value **false**
If octavia.ingress.kubernetes.io/floatingip is sepcified and an error occurred (fip not available, not exist, not valid) fallback to create new fip. 

**Release note**:

```release-note
NONE
```


dulek (changes requested):
Some tiny remarks inline.

I wonder if we could have approached this as we do in the controller for Services. There we don't need the keep-ip annotation because we simply check the description of the FIP and in case it's not created by us, we do not delete it.

This would make the cloud provider more consistent, question is - isn't using implicit annotation better?

pkg/ingress/controller/openstack/neutron.go
```diff
@@ -100,9 +100,41 @@ func (os *OpenStack) EnsureFloatingIP(needDelete bool, portID string, floatingIP
...
+				PortID:      &portID,
+				Description: &description,
+			}
+			// attach fip to lb vip
+			fip, err = floatingips.Update(os.neutron, osFips[0].ID, updateOpts).Extract()
+			if err != nil {
+				return "", err
+			}
+		} else {
+			fip, err = floatingips.Create(os.neutron, floatIPOpts).Extract()
```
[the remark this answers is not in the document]
ccleouf66: I do not ignore IP, if we reach else, that mean no IP address is provided by annotation, so I keep the default mechanism that is to create a new one.

pkg/ingress/controller/controller.go
```diff
@@ -579,15 +589,26 @@ func (c *Controller) deleteIngress(ing *nwv1.Ingress) error {
...
+	// Manage the floatingIP
+	keepFloatingSetting := getStringFromIngressAnnotation(ing, IngressAnnotationLoadBalancerKeepFloatingIP, "false")
+	keepFloating, err := strconv.ParseBool(keepFloatingSetting)
+	if err != nil {
+		return fmt.Errorf("unknown annotation %s: %v", IngressAnnotationLoadBalancerKeepFloatingIP, err)
 	}
 
-	logger.WithFields(log.Fields{"lbID": loadbalancer.ID}).Info("VIP or floating IP deleted")
+	if keepFloating {
+		logger.Debug("keep floating IP")
```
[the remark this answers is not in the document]
ccleouf66: I'm sorry, I don't understand your question. The idea here is, by default, when you delete your ingress, the Octavia LB and attached FIP are deleted.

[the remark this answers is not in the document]
ccleouf66: In the log message ? On openstack side the fip had a comment with ingress name and cluster name and this comment is kept as is, needed to be updated ?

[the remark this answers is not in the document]
dulek: My point is that this looks like a leftover log from debugging. If it's not and you want to keep it I'm fine, but it has to include more info (which Ingress and FIP is affected) and maybe the reason (because Ingress has keep-IP annotation).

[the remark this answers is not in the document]
jichenjc: +1 

given we have chance to update this code, better to enhance this 
current log seems useless

pkg/ingress/controller/openstack/neutron.go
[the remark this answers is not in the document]
dulek: Sure, but I'd probably expect a try to create a FIP with specified IP. Check `FloatingIP` on the `floatingips.CreateOpts`.

dulek:
> Some tiny remarks inline.
> 
> I wonder if we could have approached this as we do in the controller for Services. There we don't need the keep-ip annotation because we simply check the description of the FIP and in case it's not created by us, we do not delete it.
> 
> This would make the cloud provider more consistent, question is - isn't using implicit annotation better?

Thinking about this more now - if we have two separate annotations now (`floatingip` and `keep-floatingip`), we have to consider if it makes sense to use them separately. An Ingress with just `floatingip` annotation would expect the FIP to exist beforehand and then it will remove that FIP on deletion. Then before it's recreated (K8s reconciliation pattern can trigger that automatically) it would need the FIP to be recreated by the user. This just doesn't fit K8s model.

Then using just `keep-floatingip` annotation would make the FIPs orphaned on every Ingress recreation. This again doesn't fit K8s model well, when resources might be recreated fairly often and without consequences.

IMO that's a strong argument to just implement the `floatingip` annotation and instead of relying on `keep-floatingip` annotation, just check the FIPs description to see if it was created by the Ingress controller or not. See a similar thing in CCM [1]. 

I now see that CCM has a `keep-floatingip` annotation support too. I don't think that's a good idea in general, unless there's a use case I don't see.

[1] https://github.com/kubernetes/cloud-provider-openstack/blob/2231e343644690f3e672c024af89d695fe20ac3c/pkg/openstack/loadbalancer.go#L2335-L2339

ccleouf66:
Okay thanks for precision ! The use case I have in mind is: when you are running a kube cluster on publuci cloud provider based on openstack like OVH or t-system, the first time when you create your ingress, you don't have yet an existing FIP (except if you created it manually before) so you probably want to let your Cloud provider to create it for you. 

Now, when you delete your ingress, maybe you want to keep the FIP to not have to recreate DNS record for example and having place the keep annotations can allow you to do that. 

You probably right, their is a better implementation possible, juste for the moment I don't have it.

dulek:
I think that at a minimum I'd add that part about checking if description indicates if FIP was created by Ingress controller before deleting it like it's done in CCM. Then `keep-floatingip` annotation has a precedent in CCM, so I guess it can be added here too. Thing is - it's much easier to add a new functional annotation than to remove it later, so I'm trying to consider all option.

pkg/ingress/controller/openstack/neutron.go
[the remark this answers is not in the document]
ccleouf66: Hey @dulek, I removed debug logs but for this I let it like that.
I think, if user set annotation `existingfloatingIP`that means he want to reuse an existing and not create new one (maybe additional fees on public cloud providers ?). I preferred to rise an error like 'specified FIP not existing' instead to take initiative to create new FIP and potentially generate additional fees.
What do you think ?

pkg/ingress/controller/controller.go
```diff
@@ -101,6 +101,16 @@ const (
...
 
+	// IngressAnnotationLoadBalancerKeepFloatingIP is the annotation used on the Ingress
+	// to indicate that we want to keep the floatingIP after the ingress deletion. The Octavia LoadBalancer will be deleted
+	// but not the floatingIP. That mean this floatingIP can be reused on another ingress without editing the dns area or update the whitelist.
+	// Default to false.
+	IngressAnnotationLoadBalancerKeepFloatingIP = "octavia.ingress.kubernetes.io/keep-floatingip"
+
+	// IngressAnnotationFloatingIp is the key of the annotation on an ingress to set floating IP that will be associated to LoadBalancers.
+	// If the floatingIP is not available, a new one will be created.
+	IngressAnnotationFloatingIp = "octavia.ingress.kubernetes.io/floatingip"
```
dulek, line 112: Please capitalize `IP`.

pkg/ingress/controller/controller.go
```diff
@@ -579,15 +589,24 @@ func (c *Controller) deleteIngress(ing *nwv1.Ingress) error {
...
+	keepFloating, err := strconv.ParseBool(keepFloatingSetting)
+	if err != nil {
+		return fmt.Errorf("unknown annotation %s: %v", IngressAnnotationLoadBalancerKeepFloatingIP, err)
 	}
 
-	logger.WithFields(log.Fields{"lbID": loadbalancer.ID}).Info("VIP or floating IP deleted")
+	if !keepFloating {
+		// Delete the floating IP for the load balancer VIP. We don't check if the Ingress is internal or not, just delete
+		// any floating IPs associated with the load balancer VIP port.
+		logger.WithFields(log.Fields{"lbID": loadbalancer.ID, "vIP": loadbalancer.VipAddress}).Info("deleting VIP or floating")
```
dulek, line 602: I think it should be `"VIP"`.

pkg/ingress/controller/openstack/neutron.go
```diff
@@ -100,9 +100,41 @@ func (os *OpenStack) EnsureFloatingIP(needDelete bool, portID string, floatingIP
...
+			if err != nil {
+				return "", err
+			}
+			if len(osFips) != 1 {
+				return "", err
+			}
+			// check if fip is used
+			if osFips[0].PortID != "" {
+				return "", fmt.Errorf("floating IP %s already used by port %s", osFips[0].FloatingIP, osFips[0].PortID)
+			}
```
dulek, line 123: What if `osFips[0].PortID == portID` already? That should be considered success, right?

pkg/ingress/controller/openstack/neutron.go
```diff
@@ -100,9 +100,41 @@ func (os *OpenStack) EnsureFloatingIP(needDelete bool, portID string, floatingIP
 			FloatingNetworkID: floatingIPNetwork,
 			Description:       description,
 		}
```
dulek, line 102: floatIPOpts should now be declared in the part in which we create the IP.

dulek (changes requested):
I added a bunch of comments regarding the logic around this. I do not think this should attempt to create a random FIP if user-specified one doesn't exist. If one is attempted to be created it has to request a user-requested address too (even if it's admin API).

pkg/ingress/controller/controller.go
```diff
@@ -909,15 +928,21 @@ func (c *Controller) ensureIngress(ing *nwv1.Ingress) error {
...
 		if err != nil {
-			return fmt.Errorf("failed to create floating IP: %v", err)
+			return fmt.Errorf("unknown annotation %s: %v", IngressAnnotationFloatingIP, err)
 		}
 
-		logger.WithFields(log.Fields{"fip": address}).Info("floating IP created")
+		description := fmt.Sprintf("Floating IP for Kubernetes ingress %s in namespace %s from cluster %s", ingName, ingNamespace, clusterName)
+
+		if floatingIPSetting != "" {
+			logger.Info("try to use floating IP : ", floatingIPSetting)
```
dulek, line 940: Missing `%s`? And unnecessary space before colon?

dulek (commented):
The logic seems to make sense now, but the code of `EnsureFloatingIP()` is structured weirdly. IMO it should:

1. Get list of FIPs attached to the VIP port.
2. if `needDelete`.
   1. Delete them and return.
3. Check if `len(fips) > 1`, return error if so.
4. `if existingFloatingIP == ""`:
   1. `if len(fips) == 1` return success.
   2. `else`
      1. Create new FIP and save it into `fip`
5. `else`:
   1. `if len(fips) == 1` check if attached FIP has that address and network. If so return success. If not, return error. _We might consider if here we shouldn't detach that FIP instead of returning error_.
   2. `else` find FIP by address, check if it's free. If not error. Save it to `fip`.
6. Attach `fip` to the VIP port.

pkg/ingress/controller/controller.go
ccleouf66: For `%s` missing, nop seem to be ok:
image

For space before colon, ok fixed :)

Ok for the weirdly EnsureFloatingIP() structure, I will try to improve it tomorrow.

ccleouf66: Should be better now, I let you check

ccleouf66:
> The logic seems to make sense now, but the code of `EnsureFloatingIP()` is structured weirdly. IMO it should:
> 
> 
> 
> 1. Get list of FIPs attached to the VIP port.
> 
> 2. if `needDelete`.
> 
>    1. Delete them and return.
> 
> 3. Check if `len(fips) > 1`, return error if so.
> 
> 4. `if existingFloatingIP == ""`:
> 
>    1. `if len(fips) == 1` return success.
> 
>    2. `else`
> 
>       1. Create new FIP and save it into `fip`
> 
> 5. `else`:
> 
>    1. `if len(fips) == 1` check if attached FIP has that address and network. If so return success. If not, return error. _We might consider if here we shouldn't detach that FIP instead of returning error_.
> 
>    2. `else` find FIP by address, check if it's free. If not return error. Save it to `fip`.
> 
> 6. Attach `fip` to the VIP port.

I let you validate and test or need something else ?

kayrus:
I'm working on #2377 and I think it makes sense to wait until it's merged and then try to sync/reuse the logic in this PR.

ccleouf66:
> I'm working on #2377 and I think it makes sense to wait until it's merged and then try to sync/reuse the logic in this PR.

Ok, thanks for update :)

AlexonOliveiraRH:
May we get this merged? I have a scenario where this could be very suitable.

dulek:
I think this looks good now. @kayrus, do you think we can merge this now? Please note that this modifies how Octavia Ingress Controller handles LBs, so #2377 shouldn't affect this?

/lgtm

pkg/ingress/controller/controller.go
```diff
@@ -909,15 +928,24 @@ func (c *Controller) ensureIngress(ing *nwv1.Ingress) error {
...
+		description := fmt.Sprintf("Floating IP for Kubernetes ingress %s in namespace %s from cluster %s", ingName, ingNamespace, clusterName)
+
+		if floatingIPSetting != "" {
+			logger.Info("try to use floating IP: ", floatingIPSetting)
+		} else {
+			logger.Info("creating new floating IP")
+		}
+		address, err = c.osClient.EnsureFloatingIP(false, lb.VipPortID, floatingIPSetting, c.config.Octavia.FloatingIPNetwork, description)
+		if err != nil {
+			return fmt.Errorf("failed to use provided floating IP %s : %v", floatingIPSetting, err)
```
jichenjc, line 946: is this true? maybe the create floating ip can fail?

pkg/ingress/controller/openstack/neutron.go
```diff
@@ -47,6 +47,33 @@ func (os *OpenStack) getFloatingIPs(listOpts floatingips.ListOpts) ([]floatingip
 	return allFIPs, nil
 }
 
+func (os *OpenStack) createFloatingIP(portID string, floatingNetworkID string, description string) (*floatingips.FloatingIP, error) {
```
jichenjc, line 50: https://github.com/kubernetes/cloud-provider-openstack/blob/35ff40533448368dc1c0de8259aea5754ba5df36/pkg/openstack/loadbalancer.go#L561-L581

should we reuse those functions?

dulek: I'd vote to not reuse them, just because these are separate controllers with separate codepaths and maintaining them through two LB controllers might be cumbersome.

ccleouf66: So we duplicate this function and add it in this project ?

pkg/ingress/controller/controller.go
ccleouf66: You right, the error message is not correct the error don't cover the creation of a new fip. 
That is what you are pointing ?

jouvin:
I just started with Octavia ingress controller and faced the problem adressed by this issue: use/reuse an externally defined FIP to ensure that a service is exposed with a static address. Any chance to see this feature released soon?

PR #2168, opened 2023-03-15
[occm] Remove FIP on internal annotation addition

### LINKED ISSUE #2167: [occm] FIP not removed from LB when Service is annotated with internal annotation

/kind bug

**What happened**:
When the service.beta.kubernetes.io/openstack-internal-load-balancer annotation gets added to a Service we do not reconcile in terms of removal of a FIP that may already exist.

FIP getting deleted.

**How to reproduce it**:
Create Service, wait for FIP to be attached to the created LB, add service.beta.kubernetes.io/openstack-internal-load-balancer annotation to a Service, observe that FIP is still kept there.

**Environment**:
- Others:

dulek:
When the service.beta.kubernetes.io/openstack-internal-load-balancer annotation gets added to a Service we do not reconcile in terms of removal of a FIP that may already exist. This commit fixes that by making sure FIP attached to an LB VIP is deleted.

fixes #2167

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -908,26 +908,74 @@ func (lbaas *LbaasV2) createFloatingIP(msg string, floatIPOpts floatingips.Creat
 	return floatIP, err
 }
 
-// Priority of choosing VIP port floating IP:
-// 1. The floating IP that is already attached to the VIP port.
-// 2. Floating IP specified in Spec.LoadBalancerIP
-// 3. Create a new one
-func (lbaas *LbaasV2) getServiceAddress(clusterName string, service *corev1.Service, lb *loadbalancers.LoadBalancer, svcConf *serviceConfig, isLBOwner bool) (string, error) {
-
-	if svcConf.internal {
-		return lb.VipAddress, nil
+func (lbaas *LbaasV2) updateFloatingIP(floatingip *floatingips.FloatingIP, portID *string) (*floatingips.FloatingIP, error) {
+	floatUpdateOpts := floatingips.UpdateOpts{
+		PortID: portID,
 	}
+	if portID != nil {
+		klog.V(4).Infof("Attaching floating ip %q to loadbalancer port %q", floatingip.FloatingIP, portID)
+	} else {
+		klog.V(4).Infof("Detaching floating ip %q from port %q", floatingip.FloatingIP, floatingip.PortID)
+	}
+	mc := metrics.NewMetricContext("floating_ip", "update")
+	floatingip, err := floatingips.Update(lbaas.network, floatingip.ID, floatUpdateOpts).Extract()
+	if mc.ObserveRequest(err) != nil {
+		return nil, fmt.Errorf("error updating LB floatingip %+v: %v", floatUpdateOpts, err)
+	}
+	return floatingip, nil
+}
 
-	var floatIP *floatingips.FloatingIP
+// ensureFloatingIP manages a FIP for a Service and returns the address that should be advertised in the
+// .Status.LoadBalancer. In particular it will:
+//  1. Lookup if any FIP is already attached to the VIP port of the LB.
+//     a) If it is and Service is internal, it will attempt to detach the FIP and delete it if it was created
+//     by cloud provider. This is to support cases of changing the internal annotation.
+//     b) If the Service is not the owner of the LB it will not contiue to prevent accidental exposure of the
+//     possible internal Services already existing on that LB.
+//     c) If it's external Service, it will use that existing FIP.
+//  2. Lookup FIP specified in Spec.LoadBalancerIP and try to assign it to the LB VIP port.
+//  3. Try to create and assign a new FIP:
+//     a) If Spec.LoadBalancerIP is not set, just create a random FIP in the external network and use that.
+//     b) If Spec.LoadBalancerIP is specified, try to create a FIP with that address. By default this is not allowed by
+//     the Neutron policy for regular users!
+func (lbaas *LbaasV2) ensureFloatingIP(clusterName string, service *corev1.Service, lb *loadbalancers.LoadBalancer, svcConf *serviceConfig, isLBOwner bool) (string, error) {
 	serviceName := fmt.Sprintf("%s/%s", service.Namespace, service.Name)
 
-	// first attempt: fetch floating IP attached to load balancer's VIP port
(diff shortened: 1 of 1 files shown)


dulek:
/hold

There are a few questions to answer here:
1. Do we differentiate Services that had predefined FIP, i.e. do we still remove that FIP?
2. What if multiple FIPs are attached to the VIP port? Should we delete all?
3. Do we support users adding FIPs manually?

dulek:
> /hold
> 
> There are a few questions to answer here:
> 
> 1. Do we differentiate Services that had predefined FIP, i.e. do we still remove that FIP?
This is fixed now, we don't delete FIPs that weren't created by us or if there's keep-floatingip annotation. We still detach them.

> 2. What if multiple FIPs are attached to the VIP port? Should we delete all?
Still no idea, but not a big deal I think, eventually all will get detached.

> 3. Do we support users adding FIPs manually?
Okay, this is now solved, in such case we don't delete FIP but still detach it.

New question - shared Servcies, what if we mix internal and external.

pkg/openstack/loadbalancer.go
```diff
@@ -2268,6 +2303,26 @@ func (lbaas *LbaasV2) EnsureLoadBalancerDeleted(ctx context.Context, clusterName
 	return mc.ObserveReconcile(err)
 }
 
+func (lbaas *LbaasV2) deleteFIPIfCreatedByProvider(fip *floatingips.FloatingIP, portID string, service *corev1.Service) (bool, error) {
+	matched, err := regexp.Match("Floating IP for Kubernetes external service", []byte(fip.Description))
+	if err != nil {
+		return false, err
```
jichenjc, line 2309: curious how this can be a error :) 

if it's error ,should we just return `false, nil` ?

dulek: Yeah, my mistake.

dulek: I basically moved it from [1] making no changes. The only error can come from Compile of the regex, so I guess it's deterministic. If you want I can make it ignore the error.

[1] https://github.com/kubernetes/cloud-provider-openstack/blob/2231e343644690f3e672c024af89d695fe20ac3c/pkg/openstack/loadbalancer.go#L2336-L2339

dulek: But I think I'd rather return the `err`, as when it happens it means we haven't properly processed the FIP and we shouldn't make the workflow to progress.

jichenjc: just curious the check here, I think we can take it without change

dulek:
I'm keeping the `hold` as I'd like to understand how to handle shared LBs and FIPs first. I'm experimenting with coding that, but have nothing that makes sense yet.

pkg/openstack/loadbalancer.go
```diff
@@ -908,26 +908,65 @@ func (lbaas *LbaasV2) createFloatingIP(msg string, floatIPOpts floatingips.Creat
...
+		if floatIP != nil {
+			keepFloatingAnnotation := getBoolFromServiceAnnotation(service, ServiceAnnotationLoadBalancerKeepFloatingIP, false)
+			fipDeleted := false
+			if !keepFloatingAnnotation {
+				klog.V(4).Infof("Deleting floating IP %v attached to loadbalancer port id %q for internal service %s", floatIP, portID, serviceName)
+				fipDeleted, err = lbaas.deleteFIPIfCreatedByProvider(floatIP, portID, service)
+				if err != nil {
+					return "", err
+				}
+			}
```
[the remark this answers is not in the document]
dulek: I shared that feeling when writing this. Honestly function named `getServiceAddress()` shouldn't create FIP either. Anyway the refactoring for that could just be to rename it into `ensureFIPHandled()` or something like this.

pkg/openstack/loadbalancer.go
```diff
@@ -908,26 +908,74 @@ func (lbaas *LbaasV2) createFloatingIP(msg string, floatIPOpts floatingips.Creat
...
+	return floatingip, nil
+}
+
+// ensureFloatingIP manages a FIP for a Service and returns the address that should be advertised in the
+// .Status.LoadBalancer. In particular it will:
+//  1. Lookup if any FIP is already attached to the VIP port of the LB.
+//     a) If it is and Service is internal, it will attempt to detach the FIP and delete it if it was created
+//     by cloud provider. This is to support cases of changing the internal annotation.
+//     b) If the Service is not the owner of the LB it will not contiue to prevent accidental exposure of the
+//     possible internal Services already existing on that LB.
```
mdbooth, line 934: You might have addressed this recently. Is it possible to:
* Create service A as internal
* Create shared service B as internal
* Remove the internal annotation from service A, exposing service B?

dulek: No, the second Service won't get created, we don't allow secondary Services to be internal starting from #2190.

pkg/openstack/loadbalancer.go
```diff
@@ -908,26 +908,74 @@ func (lbaas *LbaasV2) createFloatingIP(msg string, floatIPOpts floatingips.Creat
...
+			}
+		}
+		return lb.VipAddress, nil
+	}
+
+	// first attempt: fetch floating IP attached to load balancer's VIP port
+	// we've fetched it in the beginning of the function, operating on that here
+	if floatIP != nil {
+		klog.V(4).Infof("Found floating ip %v by loadbalancer port id %q", floatIP, portID)
+	}
```
mdbooth, line 978: nit: I'd move this up to where the floating ip was fetched. This log message is also relevant in the case we delete or update the fip.

mdbooth (commented):
There's a nit but it's not important. I wonder about this wisdom of this interface for shared loadbalancers at all. My separate comment about that remains, but I don't think it impacts this PR, which is strictly an improvement over what currently exists.

/lgtm

dulek:
> There's a nit but it's not important. I wonder about this wisdom of this interface for shared loadbalancers at all. My separate comment about that remains, but I don't think it impacts this PR, which is strictly an improvement over what currently exists.
> 
> /lgtm

The more I analyze this idea, the more I want to remove it. It needs the user to be cooperative, it might be risky in terms of security and it might have more unexpected side effects than I found.

mdbooth (commented):
/lgtm

PR #2182, opened 2023-03-21
[cinder-csi-plugin] Add cascade to volumes deletion

### LINKED ISSUE #2181: [cinder-csi-plugin] Cannot delete volume associated to a snapshot
/kind bug

**What happened**:
When a volume with the snapshot is restored by another volume it's possible that the deletion of a volume will be blocked failing with the following message until the snapshot is deleted.
`Volume status must be available or error or error_restoring or error_extending or error_managing and must not be migrating, attached, belong to a group, have snapshots or be disassociated from snapshots after volume transfer.` 

The volume would be deleted and the tests would pass.

**How to reproduce it**:

Run the following tests scenario:

-  External Storage [Driver: cinder.csi.openstack.org] [Testpattern: Dynamic PV (xfs)][Slow] multiVolume [Slow] should concurrently access the volume and restored snapshot from pods on the same node [LinuxOnly][Feature:VolumeSnapshotDataSource][Feature:VolumeSourceXFS]
-   External Storage [Driver: cinder.csi.openstack.org] [Testpattern: Dynamic PV (ext4)] multiVolume [Slow] should concurrently access the volume and restored snapshot from pods on the same node [LinuxOnly][Feature:VolumeSnapshotDataSource][Feature:VolumeSourceXFS] 
-   External Storage [Driver: cinder.csi.openstack.org] [Testpattern: Dynamic PV (block volmode)] multiVolume [Slow] should concurrently access the volume and restored snapshot from pods on the same node [LinuxOnly][Feature:VolumeSnapshotDataSource][Feature:VolumeSourceXFS] 

More information around the test can be found [here](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/multivolume.go#L319).

MaysaMacedo:

When a volume with the snapshot is restored by another volume it's possible that the deletion of a volume will be blocked until the snapshot is deleted. To overcome such scenario, this commit enables the cascade deletion of a volume to allow the deletion of any snapshot associated to the volume.

fixes #2181

**Release note**:

```release-note
[cinder-csi-plugin] Add cascade to volumes deletion by default. Before the deletion of volume could be blocked if it had a snapshot associated. With the cascade deletion the volume and any snapshot associated with it will be deleted.
```

### DIFF
--- a/pkg/csi/cinder/openstack/openstack_volumes.go
+++ b/pkg/csi/cinder/openstack/openstack_volumes.go
@@ -142,7 +142,7 @@ func (os *OpenStack) GetVolumesByName(n string) ([]volumes.Volume, error) {
 	return vols, nil
 }
 
-// DeleteVolume delete a volume
+// DeleteVolume delete a volume and any snapshots associated with it
 func (os *OpenStack) DeleteVolume(volumeID string) error {
 	used, err := os.diskIsUsed(volumeID)
 	if err != nil {
@@ -153,7 +153,8 @@ func (os *OpenStack) DeleteVolume(volumeID string) error {
 	}
 
 	mc := metrics.NewMetricContext("volume", "delete")
-	err = volumes.Delete(os.blockstorage, volumeID, nil).ExtractErr()
+	opts := volumes.DeleteOpts{Cascade: true}
+	err = volumes.Delete(os.blockstorage, volumeID, opts).ExtractErr()
 	return mc.ObserveRequest(err)
 }
 


jichenjc:
you mentioned 3 test in the https://github.com/kubernetes/cloud-provider-openstack/issues/2181

can you help open those test case after the fix of this issue and give a CI test run?

MaysaMacedo:
Those test failures were actually fixed with:
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/multivolume.go#L321

Basically this makes sure we give time for the snapshot to be deleted before the volume deletion is attempted.
I believe this PR can be closed.

jichenjc:
ok, so please close this PR in case there's nothing to be done

PR #2186, opened 2023-03-30
update: update gophercloud

jichenjc:

steal from CAPO, make sense to upgrade gophercloud time to time in order to catchup latest stuff
to both fix but and potential for other code change as base
fixes #

**Release note**:

```release-note
NONE
```


jichenjc:
/approve

2 lgtm already

PR #2188, opened 2023-04-06
[barbican-kms-plugin] Add manifest for barbikan-kms daemoset

berghaus:
This manifest deploys a barbican-kms-plugin on every control plane node. This is desirable in Kubernetes configurations with multiple control plane nodes to have a KMS endpoint on each control plane node.

This is just an additional kubernetes manifest.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/manifests/barbican-kms/ds.yaml
+++ b/manifests/barbican-kms/ds.yaml
@@ -1,0 +1,65 @@
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: barbican-kms
+  namespace: kube-system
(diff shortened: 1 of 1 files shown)


dulek:
/lgtm

No need to run tests here, I think.

jichenjc:
>>>This manifest deploys a barbican-kms-plugin on every control plane node. This is desirable in Kubernetes configurations with multiple control plane nodes to have a KMS endpoint on each control plane node.

just curious, someone need consume this manifest, correct? seems no doc update point to the new manifest ..

PR #2189, opened 2023-04-07
[occm] Improve no fip message

dulek:
If user specifies an address in `.spec.loadBalancerIP`,
cloud-provider-openstack expects it to be precreated by the user and
never attempts to actually create the FIP with that address, because
default Neutron policy forbids that action for non-admin users.

While we can argue if CPO should actually try that expecting a
non-default policy, for now we can at least communicate this to the
users clearly. This commits makes sure the message explains the reason
behind LB creation failure better. The event should be then logged by
the controller above with `SyncLoadBalancerFailed` event.

**Release note**:
```release-note
NONE
```


dulek:
/hold

Wait, this is wrong.

dulek:
I'll need to test this more, but this seems to be wrong.

PR #2190, opened 2023-04-07
[occm] Don't allow internal Services to share an LB

### LINKED ISSUE #2174: [occm] Internal Service might be exposed when LB is shared

/kind bug

**What happened**:
When you create a Service with `service.beta.kubernetes.io/openstack-internal-load-balancer: "true"` annotation and then create another regular non-internal Service that's supposed to share the LB with the previous one, CPO will allow that and will attach the FIP to the load balancer. This will expose the both Services outside, including the one that supposed to only be internal. While a bit obscure, it's a potential security risk that a confused user might create without knowing.

Either sharing internal LB with an external Service is blocked or there's a way to only expose one listener from a share LB. I don't think the latter is possible.

**How to reproduce it**:
Explained above.

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version:  master
- Others:

dulek:
Feature of sharing LBs seems to have some logic flaws. One of them is the problem of mixing internal and external Services on a single LB. As a FIP is tied to the LB and not individual listener, it means that if one Service attached the FIP to the Service, all the other Services will be available on that FIP. This may lead to accidental exposure of an internal Service which can potentially be pretty bad.

This commits attempts to limit the number of cases when the user can shoot themselves in the foot and makes it impossible for the internal Services to be secondary on a share load balancer. Moreover a condition is added that prevents secondary Services to create FIPs at all, so that accidental exposure of primary internal service by a secondary external one is solved.

There is no other way to reliably do that prevention as we only save truncated name into the tags of the LB, so there's no way to get list of all Services in a load balancer without listing all of them (in every namespace) and looking through their LB ID annotation. Even with that it's still really complicated to make decisions.

Cases (P - primary, S - secondary, E - external, I - internal):

```
P S
E E - all good
E I - prevented, all good
I I - prevented, not great, because it's technically a working
      combination, but it's a cost
I E - external one won't create FIP, good

Updates (after #2168 merges):
E E -> E I - we'll get error on the second Service, good
E E -> I E - FIP gets detached, E won't be able to readd it, good.
```

fixes #2174

Cases are listed

**Release note**:
```release-note
[openstack-cloud-controller-manager] OCCM will no longer allow internal Services to share a load balancer with another Service.
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -912,7 +912,8 @@ func (lbaas *LbaasV2) createFloatingIP(msg string, floatIPOpts floatingips.Creat
 // 1. The floating IP that is already attached to the VIP port.
 // 2. Floating IP specified in Spec.LoadBalancerIP
 // 3. Create a new one
-func (lbaas *LbaasV2) getServiceAddress(clusterName string, service *corev1.Service, lb *loadbalancers.LoadBalancer, svcConf *serviceConfig) (string, error) {
+func (lbaas *LbaasV2) getServiceAddress(clusterName string, service *corev1.Service, lb *loadbalancers.LoadBalancer, svcConf *serviceConfig, isLBOwner bool) (string, error) {
+
 	if svcConf.internal {
 		return lb.VipAddress, nil
 	}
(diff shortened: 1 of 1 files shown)


pkg/openstack/loadbalancer.go
```diff
@@ -928,6 +929,13 @@ func (lbaas *LbaasV2) getServiceAddress(clusterName string, service *corev1.Serv
 	}
 	klog.V(4).Infof("Found floating ip %v by loadbalancer port id %q", floatIP, portID)
 
+	// we cannot add a FIP to a shared LB when we're a secondary Service or we risk adding it to an internal
+	// Service and exposing it to the world unintentionally.
```
jichenjc, line 933: ```adding it to an internal``` ===> curious line 917 we return directly if it's internal.. so will it possible that described scenario can happen?

dulek: It's the case when we create an internal Service A as an LB owner and then a secondary external Service B on the same LB. B is external, so normally it would attempt to create the FIP for the LB, exposing A externally in the process. The idea is to prevent secondary Services from creating FIPs.

It has to be done this way, because when we're a secondary Service, there is no reliable way to get the primary Service besides listing all the Services and looking for LB ID in the annotations. The LB name and tags are not reliable because they're truncated to 255 characters.

jichenjc: ok, thanks for the detailed info

PR #2191, opened 2023-04-11
[occm] Remove intermediate funcs on Ensure* interfaces

dulek:
I think these were relics of the times when we had different codepaths for Octavia and Neutron LBaaS. The latter is not supported anymore, so `Ensure*()` functions can call `ensureOctavia*()` directly.

**Release note**:
```release-note
NONE
```


jichenjc:
that's correct, we used to support both and neutron-lbaas was removed months before

/lgtm
/approve

PR #2218, opened 2023-04-27
[util] Ignore proxies when calling Nova Metadata

### LINKED ISSUE #2217: Volume gets attached to wrong VM when behind proxy cinder-csi-plugin
/kind bug

**What happened**:
I created a volume and tried to attach it to one pod but it tried to attach it to a machine in another project that acts as a proxy for all our outgoing requests. I don't know how it works behind the scene but since the request to the openstack api comes from our proxy, I guess that it thinks that it's the one doing the request. So question is how I can tell the openstack api that there is an proxy involved?

I get the folllowing error:
`AttachVolume.Attach failed for volume "pvc-ef7afed3-07ae-4478-b995-83240963ed97" : rpc error: code = NotFound desc = [ControllerPublishVolume] Instance 67cf04ca-9b1a-4c4b-85e0-33f0e9dc5d9b not found`

Disk to get attached to the machine requesting the volume

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: 1.6.2

pierreprinetti:
Prior to this change, the call to the Nova Metadata service used HTTP proxies as directed by the environment variables HTTP_PROXY, HTTPS_PROXY and NO_PROXY (or the lowercase versions thereof).

However, the call to the Metadata well-known IP must originate from the requesting server directly to be correctly served.

With this change, calls to the Nova Metadata service ignore proxy settings.

fixes #2217

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/util/metadata/metadata_test.go
+++ b/pkg/util/metadata/metadata_test.go
@@ -18,6 +18,9 @@ package metadata
 
 import (
 	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"os"
 	"strings"
 	"testing"
 )
@@ -147,4 +150,28 @@ func TestCheckMetaDataOpts(t *testing.T) {
 				testcase.name, testcase.expectedError, err)
 		}
 	}
+}
+
+func TestGetFromMetadataService(t *testing.T) {
+	t.Run("ignores HTTP_PROXY", func(t *testing.T) {
+		// Here I spin up an HTTP server, set it as HTTP_PROXY, and
+		// assert that the request to the Metadata server doesn't hit
+		// it.
+		fakeProxy := httptest.NewServer(http.HandlerFunc(func(_ http.ResponseWriter, _ *http.Request) {
+			t.Errorf("the call to Metadata hit the proxy server")
+		}))
+		defer fakeProxy.Close()
+
+		// defer resetting HTTP_PROXY to whatever it was before this test
+		defer func(originalValue string, wasSet bool) {
+			if wasSet {
+				os.Setenv("HTTP_PROXY", originalValue)
+			} else {
+				os.Unsetenv("HTTP_PROXY")
+			}
+		}(os.LookupEnv("HTTP_PROXY"))
+
+		os.Setenv("HTTP_PROXY", fakeProxy.URL)
+		_, _ = getFromMetadataService("")
+	})
 }
--- a/pkg/util/metadata/metadata.go
+++ b/pkg/util/metadata/metadata.go
@@ -189,11 +189,17 @@ func getFromConfigDrive(metadataVersion string) (*Metadata, error) {
 	return parseMetadata(f)
 }
 
+func noProxyHTTPClient() *http.Client {
+	noProxyTransport := http.DefaultTransport.(*http.Transport).Clone()
+	noProxyTransport.Proxy = nil
+	return &http.Client{Transport: noProxyTransport}
+}
+
 func getFromMetadataService(metadataVersion string) (*Metadata, error) {
 	// Try to get JSON from metadata server.
 	metadataURL := getMetadataURL(metadataVersion)
-	klog.V(4).Infof("Attempting to fetch metadata from %s", metadataURL)
-	resp, err := http.Get(metadataURL)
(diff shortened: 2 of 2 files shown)


pkg/util/metadata/metadata_test.go
```diff
@@ -148,3 +151,23 @@ func TestCheckMetaDataOpts(t *testing.T) {
...
+			t.Errorf("the call to Metadata hit the proxy server")
+		}))
+		defer fakeProxy.Close()
+
+		// defer resetting HTTP_PROXY to whatever it was before this test
+		defer func(reset string) {
+			os.Setenv("HTTP_PROXY", reset)
+		}(os.Getenv("HTTP_PROXY"))
+
+		os.Setenv("HTTP_PROXY", fakeProxy.URL)
```
mdbooth, line 170: It *probably* doesn't matter, but as environment variables are global and therefore potentially 'spooky action at a distance' I think it's worth being strict here: this doesn't strictly put things back as they were. In all likelihood `HTTP_PROXY` wasn't previously set, but after execution this will leave it set to the empty string.

May I suggest a helper function something like:
```go
testSetEnv(t *testing.T, name, value string) {
  orig, ok := os.LookupEnv(name)
  t.Cleanup(func() {
    if ok {
      os.Setenv(name, orig)
    } else {
      os.Unsetenv(name)
    }
  })

  os.Setenv(name, value)
}
```

You should be able to just call this from inside the test and it will do the right thing including cleaning itself up on test exit.

mdbooth (changes requested):
We should consider if there's potentially a problem hitting a random real metadata service wherever the test is running. I don't think so, but could anybody else reading this also consider 'what could possibly go wrong'?

This is tidy! I think we should be stricter with the environment variable cleanup, though. Suggestion below.

pkg/util/metadata/metadata_test.go
pierreprinetti: Amended the patch to unset the variable in case it was previously not set. Thank you.

pkg/util/metadata/metadata_test.go
```diff
@@ -148,3 +151,27 @@ func TestCheckMetaDataOpts(t *testing.T) {
...
+		defer fakeProxy.Close()
+
+		// defer resetting HTTP_PROXY to whatever it was before this test
+		defer func(originalValue string, wasSet bool) {
+			if wasSet {
+				os.Setenv("HTTP_PROXY", originalValue)
+			} else {
+				os.Unsetenv("HTTP_PROXY")
+			}
+		}(os.LookupEnv("HTTP_PROXY"))
```
mdbooth, line 172: This is better than my suggestion.

jichenjc:
seems the test fail is valid  ? maybe @pierreprinetti you can check whether we need update ? 
if it's not related, I can override this test
```
 --- FAIL: TestDriver (0.00s)
    sanity_test.go:51: failed to initialize CSI Manila driver: failed to initialize proxied CSI driver: connecting to unix:///fake-fwd-endpoint endpoint failed: failed to build resolver: passthrough: received empty target in Build() 
```

mdbooth:
> seems the test fail is valid ? maybe @pierreprinetti you can check whether we need update ? if it's not related, I can override this test
> 
> ```
>  --- FAIL: TestDriver (0.00s)
>     sanity_test.go:51: failed to initialize CSI Manila driver: failed to initialize proxied CSI driver: connecting to unix:///fake-fwd-endpoint endpoint failed: failed to build resolver: passthrough: received empty target in Build() 
> ```

@jichenjc That test hasn't passed since 2nd March: https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/openstack-cloud-csi-manila-sanity-test

jichenjc:
/override openstack-cloud-csi-manila-sanity-test
/override openstack-cloud-csi-manila-e2e-test

ok, I didn't pay attention (maybe we didn't trigger the manila test mostly) on this 
thanks for the reminder, not sure @gman0 you still work on this area, FYI

gman0:
@jichenjc thanks, I'm away for a week, but I'll have a look when I get back.

PR #2223, opened 2023-04-28
[octavia-ingress-controller] Increase timeout for LB to get to ACTIVE state

### LINKED ISSUE #2222: [octavia-ingress-controller] timeout too low when creating LBs
/kind bug

**What happened**:
When creating an ingress it fails as LB is in PENDING_CREATE status, this times out after I think 154 seconds based on 
19 attempts increasing wait by factor of 1.2 starting at 1 second. In our cloud a load balancer can take around 5 minutes sometimes to provision.

The timeout needs to allow for envs that take longer for LBs to provision

**How to reproduce it**:
Create an ingress in an openstack env when LBs are slower to create

**Environment**:
Using octavia Zed

sorrison:
Up the timeout for LB create, in some environments this can take a bit longer than ~150s increase it to ~320s

Fixes #2222

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/util/openstack/loadbalancer.go
+++ b/pkg/util/openstack/loadbalancer.go
@@ -46,7 +46,7 @@ const (
 
 	waitLoadbalancerInitDelay   = 1 * time.Second
 	waitLoadbalancerFactor      = 1.2
-	waitLoadbalancerActiveSteps = 19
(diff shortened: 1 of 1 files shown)


zetaab (approved):
/approve

PR #2227, opened 2023-05-02
Use ci-built occm in all e2e tests

### LINKED ISSUE #2220: cinder e2e tests permafailing?
TL;DR: e2e tests in [this PR](https://github.com/kubernetes/cloud-provider-openstack/pull/2218) systematically fail.

While investigating with the help of @mdbooth, we found that the occm release manifests [are still referring to v1.27.0](https://github.com/kubernetes/cloud-provider-openstack/blob/52b4f44b7429c95be4f83fb2a68c5792747e9cb9/manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml#L41). That version is broken IIUC, and that looks like a good root cause candidate.

One other thing that surprised me is that we apparently run release images as a base for presubmits in this job. Is that intentional? Shall we be running a CI build in the e2e CSI presubmit jobs?

CC @zetaab @mdbooth

mdbooth:
Previously in e2e tests we only used ci-built occm for the occm e2e test. The cinder and manila e2e tests used the latest release occm image. With this change all e2e tests deploy a ci-built occm.

This came about because we were surprised that cinder and manila e2e tests were broken by an issue in the release images.

There are pros and cons to doing this:

Pros:
* You can test changes which may affect the interaction of CPO and CSI, e.g. node labels.

Cons:
* We might accidentally merge commits which require the simultaneous upgrade of both CSI and OCCM.

Ideally we'd test both, of course. If we pick one, which is better? If we decided not to merge this, we should at least add documentation and/or comments explaining the decision for posterity.

fixes #2220

**Release note**:
```release-note
NONE
```

/cc @pierreprinetti

### DIFF
--- a/tests/playbooks/roles/install-cpo-occm/defaults/main.yaml
+++ b/tests/playbooks/roles/install-cpo-occm/defaults/main.yaml
@@ -4,7 +4,7 @@ devstack_workdir: "{{ ansible_user_dir }}/devstack"
 
 # Used for uploading image to local registry.
 image_registry_host: localhost
-build_image: false
+build_image: true
 run_e2e: false
 # Used for access the private registry image from k8s
 remote_registry_host: "{{ ansible_default_ipv4.address }}"
--- a/tests/playbooks/test-csi-cinder-e2e.yaml
+++ b/tests/playbooks/test-csi-cinder-e2e.yaml
@@ -26,7 +26,6 @@
       cert_hosts: ' ["{{ ansible_default_ipv4.address }}"]'
     - role: install-cpo-occm
       run_e2e: false
-      build_image: false
       environment: "{{ global_env }}"
     - role: install-csi-cinder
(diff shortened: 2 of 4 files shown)


jichenjc:
I am in favor of this PR but I am open for folks with different opinion or know the history..

/approve

pierreprinetti:
Sounds reasonable, thanks @mdbooth!
Deferring to the owners for a LGTM.

mdbooth:
Just a reminder that the manila e2e tests are permafailing and should be ignored: https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/openstack-cloud-csi-manila-e2e-test

mdbooth:
@zetaab Do you have an opinion on this?

dulek:
My experiences with OpenStack development are that this is a correct approach. Leaving LGTM to @zetaab.

PR #2228, opened 2023-05-02
[occm] Get IP addresses of neutron subports

jingczhang:
Those VMs are plugged into flat networks and use neutron trunk.

Currently, getAttachedInterfacesByID() only returns neutron ports directly attached to VM.
For vlan-aware VMs, IP addresses are assigned on neutron subports.
Subports are attached to the trunk; they are not attached to the VM directly.
This pull request improves getAttachedInterfacesByID() to return IP addresses of neutron subports when they exist.
Without this change, Kubernetes is unable to get IP addresses of vlan-aware VMs.
This change is transparent to VMs not using neutron trunk.

This is a re-attempt of the following PR which was idle out:
https://github.com/kubernetes/cloud-provider-openstack/pull/1535
The code has been in production for two years now.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.20
 require (
 	github.com/container-storage-interface/spec v1.8.0
 	github.com/go-chi/chi/v5 v5.0.8
-	github.com/gophercloud/gophercloud v1.3.0
+	github.com/gophercloud/gophercloud v1.4.0
 	github.com/gophercloud/utils v0.0.0-20230330070308-5bd5e1d608f8
 	github.com/hashicorp/go-version v1.6.0
 	github.com/kubernetes-csi/csi-lib-utils v0.13.0
--- a/go.sum
+++ b/go.sum
@@ -225,8 +225,9 @@ github.com/google/uuid v1.3.0/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+
 github.com/googleapis/gax-go/v2 v2.0.4/go.mod h1:0Wqv26UfaUD9n4G6kQubkQ+KchISgw+vpHVxEJEs9eg=
 github.com/googleapis/gax-go/v2 v2.0.5/go.mod h1:DWXyrwAJ9X0FpwwEdw+IPEYBICEFu5mhpdKc/us6bOk=
 github.com/googleapis/google-cloud-go-testing v0.0.0-20200911160855-bcd43fbb19e8/go.mod h1:dvDLG8qkwmyD9a/MJJN3XJcT3xFxOKAvTZGvuZmac9g=
-github.com/gophercloud/gophercloud v1.3.0 h1:RUKyCMiZoQR3VlVR5E3K7PK1AC3/qppsWYo6dtBiqs8=
 github.com/gophercloud/gophercloud v1.3.0/go.mod h1:aAVqcocTSXh2vYFZ1JTvx4EQmfgzxRcNupUfxZbBNDM=
+github.com/gophercloud/gophercloud v1.4.0 h1:RqEu43vaX0lb0LanZr5BylK5ICVxjpFFoc0sxivyuHU=
+github.com/gophercloud/gophercloud v1.4.0/go.mod h1:aAVqcocTSXh2vYFZ1JTvx4EQmfgzxRcNupUfxZbBNDM=
 github.com/gophercloud/utils v0.0.0-20230330070308-5bd5e1d608f8 h1:K9r5WEeAiaEgFZsuOP0OYjE4TtyFcCLG1nI08t9AP6A=
 github.com/gophercloud/utils v0.0.0-20230330070308-5bd5e1d608f8/go.mod h1:VSalo4adEk+3sNkmVJLnhHoOyOYYS8sTWLG4mv5BKto=
 github.com/gorilla/websocket v1.4.2 h1:+/TMaTYc4QFitKJxsQ7Yye35DkWvkdLcvGKqM+x0Ufc=


pkg/openstack/instances.go
```diff
@@ -677,26 +721,91 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+	allTrunks, err := trunks.ExtractTrunks(allPages)
+	if err != nil {
+		klog.Errorf("Failed to extract trunks: %v", err)
+		return interfaces, err
+	}
+
+	// Get subports attached to the trunk
+	var subports []trunks.Subport
+	for _, trunk := range allTrunks {
+		for _, iface := range interfaces {
```
mdbooth, line 758: Interfaces is still empty? That would make the allTrunks loop a no-op.

pkg/openstack/instances.go
```diff
@@ -677,26 +721,91 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+					return interfaces, err
+				}
+				subports = append(subports, s...)
+			}
+		}
+	}
+	klog.V(5).Infof("subports %v", subports)
+
+	// Arrange subports as for directly attached ports
+	for _, sport := range subports {
```
mdbooth, line 772: Given that the previous loop looks like a no-op I think subports is guaranteed to be empty here.

pkg/openstack/instances.go
```diff
@@ -677,26 +721,91 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+		p, err := ports.Get(network, sport.PortID).Extract()
+		if err != nil {
+			klog.Errorf("Failed to get port info for subport %s: %v", sport.PortID, err)
+			return interfaces, err
+		}
+		n, err := networks.Get(network, p.NetworkID).Extract()
```
mdbooth, line 778: A network query per subport doesn't seem like a good idea.

mdbooth (commented):
Apologies if I've missed something really obvious, but this doesn't look right to me. I can't see how `getSubInterfaces()` in particular can work. Is there some code missing which populates interfaces? Even If it did work it would have potentially catastrophic performance impacts on some clouds due to the nested API calls. I don't think we should merge this with its current code structure.

I wonder if we can do something much more efficient with https://docs.openstack.org/api-ref/network/v2/index.html?expanded=list-trunks-detail,show-trunk-details-detail#show-trunk-details?

@dulek Can you take a look at this when you're around?

pkg/openstack/instances.go
```diff
@@ -677,26 +721,91 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
 
+// getSubInterfaces
+func getSubInterfaces(network *gophercloud.ServiceClient) ([]attachinterfaces.Interface, error) {
+	var interfaces []attachinterfaces.Interface
+
+	// Check if trunk ports are attached
+	listOpts := trunks.ListOpts{}
+	allPages, err := trunks.List(network, listOpts).AllPages()
+	if err != nil {
+		klog.Errorf("Failed to list trunks: %v", err)
```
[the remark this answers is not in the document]
stephenfin: You need to check for this extension, ideally once at startup. I don't know if gophercloud does this for us. I imagine not.

pkg/openstack/instances.go
```diff
@@ -677,26 +721,91 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+		return interfaces, err
+	}
+	allTrunks, err := trunks.ExtractTrunks(allPages)
```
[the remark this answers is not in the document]
stephenfin: As an aside, neutron has exceptionally good filtering capabilities and the ability to request specific fields. I don't know if that's exposed by gophercloud but we should probably be using that. It's used extensively in nova for performance reasons.

pkg/openstack/instances.go
```diff
@@ -677,26 +721,91 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
-	interfaces, err := getAttachedInterfacesByID(client, srv.ID)
+	interfaces, err := getAttachedInterfacesByID(compute, srv.ID, network)
 	if err != nil {
 		return nil, err
 	}
 
 	return nodeAddresses(&srv.Server, interfaces, networkingOpts)
 }
 
+// getSubInterfaces
```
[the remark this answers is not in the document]
jingczhang: comment taken, will add short description of the function.

pkg/openstack/instances.go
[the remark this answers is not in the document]
jingczhang: Neutron trunk had been existing for many releases, before CCM? I would expect API error, Openstack instances we have including N-k (k is big) all support neutron trunk, let alone the Openstack instances running Kubernetes.

jingczhang:
> and do you think we can consider add a functional test to this?

Yes, we will look into adding test case.

pkg/openstack/instances.go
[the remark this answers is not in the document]
jingczhang: Good point. I will look into filtering with parent ports.

pkg/openstack/instances.go
```diff
@@ -677,26 +721,91 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+		klog.Errorf("Failed to extract trunks: %v", err)
+		return interfaces, err
+	}
+
+	// Get subports attached to the trunk
+	var subports []trunks.Subport
+	for _, trunk := range allTrunks {
+		for _, iface := range interfaces {
+			if iface.PortID == trunk.PortID {
+				s, err := trunks.GetSubports(network, trunk.ID).Extract()
```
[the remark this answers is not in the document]
jingczhang: Acked, will handle this with the comment before.

pkg/openstack/instances.go
jingczhang: Sharp eyes! My mistake when creating the PR, a rebase error, the functional prototype should be the following, I will correct it.
// getSubInterfaces
func getSubInterfaces(interfaces []attachinterfaces.Interface, network *gophercloud.ServiceClient) ([]attachinterfaces.Interface, error)

pkg/openstack/instances.go
jingczhang: Acked, will handle with the above comment.

pkg/openstack/instances.go
jingczhang: Need to get the network name for filtering on internal-network-name later on.
https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#networking

pkg/openstack/instances.go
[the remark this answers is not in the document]
jingczhang: > This is pulling back all trunks without filtering. I don't think this is a safe thing to do in general.

Thank you very much for all the detailed comments. I will try to update the PR to address the set of comments related to this point first.

pkg/openstack/instances.go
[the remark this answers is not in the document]
jichenjc: sorry, I mean some openstack didn't have this extension ... it's common that some plugins are not enabled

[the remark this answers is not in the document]
jingczhang: Good point, we will check and ensure code change is transparent when trunk plugin is not configured.

pkg/openstack/instances.go
jingczhang: Point taken, we will keep this mind when working on PR update.

pkg/openstack/instances.go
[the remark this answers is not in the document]
jingczhang: I have updated the PR yesterday taking care of comments for getSubInterfaces() function, please review.
Re-arrange getSubInterface() per review comments:
Check for trunks attached to VM interfaces only
Collapse two loop (trunks and subports) into one loop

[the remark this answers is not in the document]
jingczhang: Next PR update is planned for (1) additional code comments (2) adding unit test, plus we will confirm code transparence when trunk plugin is not enabled in neutron

mdbooth (changes requested):
This definitely needs unit tests and comments. It's very hard to understand what it's doing.

pkg/openstack/instances.go
```diff
@@ -735,68 +735,67 @@ func getAddressesByName(compute *gophercloud.ServiceClient, name types.NodeName,
...
-			NetID:     n.Name,
-			MACAddr:   p.MACAddress,
-		}
-		for _, ip := range p.FixedIPs {
-			var ip2 = attachinterfaces.FixedIP{
-				SubnetID:  ip.SubnetID,
-				IPAddress: ip.IPAddress,
+		// Get subports attached to the trunks
+		for _, trunk := range allTrunks {
+			subports, err := trunks.GetSubports(network, trunk.ID).Extract()
```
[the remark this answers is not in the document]
jingczhang: you are right, point taken

pkg/openstack/instances.go
```diff
@@ -613,17 +634,40 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+	}
+	klog.V(5).Infof("Node '%s' extraPrivates '%s'", srv.Name, extraPrivates)
+	for k, v := range extraPrivates {
+		v1, ok := addresses[k]
+		if !ok {
+			addresses[k] = v
+		} else {
+			addresses[k] = append(v1, v...)
+		}
+	}
```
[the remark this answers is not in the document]
jingczhang: This piece of code is to work with the initial version of getSubInterfaces(), i.e., to filter out the subports not used by the VM. It is no longer needed, I will remove it.

pkg/openstack/instances.go
```diff
@@ -613,17 +634,40 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
-		Addr   string
-	}
-
 	var addresses map[string][]Address
 	err := mapstructure.Decode(srv.Addresses, &addresses)
 	if err != nil {
 		return nil, err
 	}
 
+	// add subports if exist to the server
```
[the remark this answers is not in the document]
jingczhang: Related to the above comment, yes, you are absolutely right, the work has been done in the modified getSubInterfaces(). no longer needed, will be removed.

pkg/openstack/instances.go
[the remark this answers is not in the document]
jingczhang: We have checked a deployment where trunk plugin is not configured, and see the API error as expected:
2023-05-08T20:35:11.179365006+03:00 stderr F E0508 17:35:11.179178       1 instances.go:734] Failed to list trunks: Resource not found    
We will change the code to ignore the error as "no trunk" is a valid configuration.

jingczhang:
Unit test TestNodeAddressesWithSubports() is added, it will be in the next PR update. I will try to upload the update soon.

jingczhang:
PR is updated to address all the comments received so far. Please kindly review to make sure.
Jing

jingczhang:
Hi @mdbooth, I think the last PR update has addressed all your comments, can you please check?
Hi all reviewers, please let me know if you have additional comments? also, I think I have replied all comments, if there is pending comments I haven't addressed (esp not addressed from the review procedure point of view) please let me know.

pkg/openstack/instances.go
```diff
@@ -613,17 +634,47 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+					}
+				}
+			}
+			// All the addresses in the private network are not directly connected
+			// Save the private network
+			if !ok {
+				extraPrivates[k] = v
+			}
+		}
+		klog.V(5).Infof("Node '%s' extraPrivates '%s'", srv.Name, extraPrivates)
```
dulek, line 667: This feels like a leftover from debugging. We should name `extraPrivates` in a meaningful way for logs. How about simply "extra private interfaces '%s'"?

pkg/openstack/instances.go
```diff
@@ -677,26 +728,86 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+			klog.Errorf("Failed to list trunks: %v", err)
+			return subports, nil
+		}
+		allTrunks, err := trunks.ExtractTrunks(allPages)
+		if err != nil {
+			klog.Errorf("Failed to extract trunks: %v", err)
+			return subports, err
+		}
+		// Get subports attached to the trunks
+		for _, trunk := range allTrunks {
```
dulek, line 766: Would it be easier to use trunk_details instead?

```
{
  "port": {
    <snip>
    "id": "713cc70f-fb49-4f7a-ae33-efb9eab93cb3",
    <snip>
    "trunk_details": {
      "sub_ports": [
        {
          "mac_address": "fa:16:3e:5b:bf:e1",
          "port_id": "0cd81ae6-df5d-47f6-ae14-a7ba706f4f23",
          "segmentation_id": 123,
          "segmentation_type": "vlan"
        }
      ],
      "trunk_id": "c264b0a2-b7d6-49a4-82b7-d1b42862d065"
    },
    <snip>
  }
}
```

At a glance Gophercloud doesn't support it, but we can still implement that support there if so. @pierreprinetti, do you find it feasible? How do we implement optional extension fields for Ports in Gophercloud? I feel like it should be possible by extending the Port structure we use?

pkg/openstack/instances.go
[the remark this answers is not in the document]
dulek: I still think a better way would be to check for `trunk` extension on startup and cache that.

pkg/openstack/instances.go
dulek: We could probably list networks and then use that list to find needed info. It has downsides too.

pkg/openstack/instances.go
```diff
@@ -677,26 +728,86 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+				n, err := networks.Get(network, p.NetworkID).Extract()
+				if err != nil {
+					klog.Errorf("Failed to get network info for subport %s: %v", sport.PortID, err)
+					return subports, err
+				}
+				var sface = attachinterfaces.Interface{
+					PortState: "ACTIVE",
+					FixedIPs:  []attachinterfaces.FixedIP{},
+					PortID:    p.ID,
+					NetID:     n.Name,
```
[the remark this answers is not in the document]
dulek: Aw, but it's probably correct given that we later filter by network name…?

dulek (commented):
I think this looks fairly okay after the round of Matt's comments getting addressed. I'm dropping a few comments, but nothing too serious from my side.

pkg/openstack/instances.go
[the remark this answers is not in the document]
jingczhang: yes, this is to cope with nodeAddresses() filtering on network name

pkg/openstack/instances.go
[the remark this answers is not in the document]
jingczhang: good point, I ack it

pkg/openstack/instances.go
```diff
@@ -677,26 +728,86 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+					klog.Errorf("Failed to get port info for subport %s: %v", sport.PortID, err)
+					return subports, err
+				}
+				n, err := networks.Get(network, p.NetworkID).Extract()
+				if err != nil {
+					klog.Errorf("Failed to get network info for subport %s: %v", sport.PortID, err)
+					return subports, err
+				}
+				var sface = attachinterfaces.Interface{
+					PortState: "ACTIVE",
```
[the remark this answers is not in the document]
jingczhang: yes, I ack it, the port state is irrelevant here

pkg/openstack/instances.go
jingczhang: good point, I ack it, will do in the next PR update as for the aboves.

pkg/openstack/instances.go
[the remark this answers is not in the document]
jingczhang: Actually, nodeAddresses() filters on port state ACTIVE, so need to set the port state to align with that.

pkg/openstack/instances.go
```diff
@@ -613,17 +634,47 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
 	if err != nil {
 		return nil, err
 	}
 
+	// Get the private networks that are not directly connected
+	if len(addresses) < len(allPrivates) {
+		extraPrivates := make(map[string][]Address)
+		// For each private network
+		for k, v := range allPrivates {
+			ok := false
```
mdbooth, line 648: Suggest ok -> found, because that's what it means here.

pkg/openstack/instances.go
```diff
@@ -613,17 +634,47 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+						if a.Addr == a1.Addr {
+							ok = true
+							break
+						}
+					}
+				}
+			}
+			// All the addresses in the private network are not directly connected
+			// Save the private network
+			if !ok {
```
mdbooth, line 663: Then this becomes:

```go
if !found {
```

which is much more readable.

pkg/openstack/instances.go
```diff
@@ -677,26 +728,86 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
 	return addrs, nil
 }
 
-func getAddressesByName(client *gophercloud.ServiceClient, name types.NodeName, networkingOpts NetworkingOpts) ([]v1.NodeAddress, error) {
-	srv, err := getServerByName(client, name)
+func getAddressesByName(compute *gophercloud.ServiceClient, name types.NodeName, networkingOpts NetworkingOpts, network *gophercloud.ServiceClient) ([]v1.NodeAddress, error) {
```
mdbooth, line 731: nit: I'd personally group similar arguments here and both both gophercloud.ServiceClient arguments first.

pkg/openstack/instances.go
mdbooth: I thought about this, but I think it's the same number of API calls? We need to fetch the subports anyway in order to get the network name, so whether we get them from the trunk or the parent port doesn't make any difference, I think.

pkg/openstack/instances.go
```diff
@@ -677,26 +728,86 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+		allTrunks, err := trunks.ExtractTrunks(allPages)
+		if err != nil {
+			klog.Errorf("Failed to extract trunks: %v", err)
+			return subports, err
+		}
+		// Get subports attached to the trunks
+		for _, trunk := range allTrunks {
+			if len(trunk.Subports) == 0 {
+				continue
+			}
```
mdbooth, line 769: This is redundant: if Subports is empty the loop below is already a no-op.

pkg/openstack/instances.go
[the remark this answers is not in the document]
mdbooth: Incidentally, adding this network name here is the only reason for this additional round of API calls. If we aren't using that network name, which could be checked statically from config, we can omit all of this in favour of getting the rest of the info from trunk details as @dulek suggested above.

Lets not go there with this PR, though. This is good, imho.

pkg/openstack/instances.go
```diff
@@ -677,26 +728,86 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+		}
+	}
+	klog.V(5).Infof("Node has %d sub-interfaces '%v'", len(subports), subports)
+	return subports, nil
+}
+
 // getAttachedInterfacesByID returns the node interfaces of the specified instance.
-func getAttachedInterfacesByID(client *gophercloud.ServiceClient, serviceID string) ([]attachinterfaces.Interface, error) {
+func getAttachedInterfacesByID(compute *gophercloud.ServiceClient, serviceID string, network *gophercloud.ServiceClient) ([]attachinterfaces.Interface, error) {
```
mdbooth, line 806: nit: Again, I'd group the service clients.

pkg/openstack/instances.go
```diff
@@ -677,26 +728,86 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+func getSubInterfaces(interfaces []attachinterfaces.Interface, network *gophercloud.ServiceClient) ([]attachinterfaces.Interface, error) {
+	klog.V(5).Infof("Node has %d interfaces '%v'", len(interfaces), interfaces)
+
+	var subports []attachinterfaces.Interface
+	for _, iface := range interfaces {
+		// Check if trunk ports are attached
+		listOpts := trunks.ListOpts{
+			PortID: iface.PortID,
+		}
+		allPages, err := trunks.List(network, listOpts).AllPages()
```
mdbooth, line 755: We still need to check if trunks are enabled.

pkg/openstack/instances.go
```diff
@@ -708,6 +819,10 @@ func getAttachedInterfacesByID(client *gophercloud.ServiceClient, serviceID stri
 	if mc.ObserveRequest(err) != nil {
 		return interfaces, err
 	}
-
+	subInterfaces, err := getSubInterfaces(interfaces, network)
```
mdbooth, line 822: I want this to be free if trunks aren't in use, but it's not: it's still adding at least 1 additional API call per attached port (list trunks by port id). This seems like a very high cost for a feature most people aren't going to be using.

I wonder if we do need to refactor all of this to fetch ports by device id instead of attachinterfaces, which would allow us to use trunk details.

pkg/openstack/openstack_test.go
```diff
@@ -878,6 +878,233 @@ func TestNodeAddressesWithAddressSortOrderOptions(t *testing.T) {
 	}
 }
 
+func TestNodeAddressesWithSubports(t *testing.T) {
```
mdbooth, line 881: This isn't really testing the interesting code. In fact, nothing here necessarily relates to subports. We need test coverage for getAttachedInterfacesByID(), e.g.:

* Trunking is disabled
* Trunking is enabled but there are no trunks
* A single port has a single subport
* A single port has multiple subports
* Multiple ports have multiple subports

mdbooth (commented):
Still many things here, but the biggest one is: this code should not add cost for users who don't need it. Achieving that may require some refactoring.

pkg/openstack/instances.go
dulek: Yeah, I think so, we end up with the same number of calls.

pkg/openstack/instances.go
```diff
@@ -613,17 +634,47 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
-		Addr   string
-	}
-
 	var addresses map[string][]Address
 	err := mapstructure.Decode(srv.Addresses, &addresses)
 	if err != nil {
 		return nil, err
 	}
 
+	// Get the private networks that are not directly connected
```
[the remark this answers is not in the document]
jingczhang: make sense, will try this in next PR update.

pkg/openstack/instances.go
```diff
@@ -613,17 +634,47 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+	if len(addresses) < len(allPrivates) {
+		extraPrivates := make(map[string][]Address)
+		// For each private network
+		for k, v := range allPrivates {
```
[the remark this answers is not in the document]
jingczhang: comment taken, can not disagree.

pkg/openstack/instances.go
```diff
@@ -613,17 +634,47 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+			ok := false
+			// For each address in the private network
+			for _, a := range v {
```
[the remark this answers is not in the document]
jingczhang: comment taken

pkg/openstack/instances.go
```diff
@@ -613,17 +634,47 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+				// Check if the address is directly connected
+				for _, v1 := range addresses {
```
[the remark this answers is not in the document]
jingczhang: comment taken

pkg/openstack/instances.go
```diff
@@ -613,17 +634,47 @@ func nodeAddresses(srv *servers.Server, interfaces []attachinterfaces.Interface,
...
+					for _, a1 := range v1 {
```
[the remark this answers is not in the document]
jingczhang: comment taken

pkg/openstack/instances.go
jingczhang: comment taken,

pkg/openstack/instances.go
jingczhang: This is already checked, API error is returned if trunk plugin is not enable, the API error stops the flow and returns to caller with no subports.

pkg/openstack/instances.go
jingczhang: not really, the getSubInterfaces() returns upon the first trunk API error.

pkg/openstack/openstack_test.go
jingczhang: hmm...I will get back for this.

jingczhang:
Thanks @mdbooth and @dulek for the code comments. I will address them first. Additional unit tests will follow after that.

pkg/openstack/instances.go
jingczhang: sorry, reading this again and realizing you refer to CPU cycles spent on checking if trunk is configured, this cost depends on number of VMs and ports per VM

pkg/openstack/openstack_test.go
jingczhang: sure, will get resources to add those unit tests

pkg/openstack/instances.go
jingczhang: comment taken

pkg/openstack/instances.go
jingczhang: comment taken

pkg/openstack/instances.go
jingczhang: comment taken, it is added to avoid the trace for trunk has no subports

pkg/openstack/instances.go
mdbooth: Not so much CPU cycles as API calls. Remember we're running this loop for every node in every cluster every 10 minutes by default. At only 100 nodes that's still every 6 seconds on average. Additional API calls here are *expensive*.

jingczhang:
PR is updated per last round code comments.
Hi @mdbooth, I need pointer for writing the 1st unit test you've requested, it is not obvious to me how it could be done, please advise.
<===
We need test coverage for getAttachedInterfacesByID(), e.g.:
Trunking is disabled

mdbooth (commented):
The more I think about this, the more I think we should refactor the surrounding code before doing this. The way you've approached this is probably the only reasonable way to do this with the existing code structure. However, it adds an additional api call per attached interface for every node in your cloud, every 10 minutes, whether you're using subports or not. Given that nobody is currently using subports in this way and most people probably never will, this feels like an unreasonably large cost.

I think we need to fetch ports directly from neutron by deviceID in the first instance. This would also show us subports if there are any, meaning the cost of the feature would only be paid by systems which are using it.

In short:
* Refactor nodeAddresses() as above
* Rewrite this PR to use the refactored nodeAddresses

I think the feature is good, but I would prefer that this PR didn't merge in its current form due to its impact on the majority of users.

pkg/openstack/instances.go
jingczhang: You are right, thanks for the pointer. The patch is built for trunk plugin is configured (normal case) and trunk is used (vlan-aware VM use case), checking for if trunk plugin is configured was not needed so not considered. It seems this checking won't be necessary after baseline code restructure per @mdbooth.

jingczhang:
Hi @mdbooth, thank you very much for your review and plan. We have the interest to upstream this so as to get rid of the "out of tree" rebase trouble. Looking forward to retry this PR after your code restructure then.

mdbooth:
> 

Unfortunately although I think this code could do with a refactor with or without the addition of subports, it's not on my todo list right now 😞

mdbooth:
FYI I've added the following:
* [Fetch ports instead of interfaces in CPO](https://github.com/kubernetes/cloud-provider-openstack/pull/2250) (not yet merged)
* [Support trunk_details extension in gophercloud](https://github.com/gophercloud/gophercloud/pull/2625) (merged)

Additionally, @pierreprinetti has been noodling with a gophercloud/utils library to allow us to fetch multiple specified ports in a single API call: https://github.com/gophercloud/utils/pull/190.

With all of these in place, I believe:
* subport support adds no overhead for anybody not using it
* subport support requires only 1 additional API call per nodeAddresses() for anybody who is using it (to fetch all detected subports)

mdbooth:
@jingczhang both of the above mentioned changes are now merged, and I've bumped the gophercloud dep to 1.4.0 which includes the trunk_details patch. #2250 has almost certainly caused the merge conflict in this patch so you're probably going to have to re-work it even for your downstream fork. By using trunk_details you should now be able to implement this with zero cost for anybody not using the feature. Are you likely to update the PR?

jingczhang:
> @jingczhang both of the above mentioned changes are now merged, and I've bumped the gophercloud dep to 1.4.0 which includes the trunk_details patch. #2250 has almost certainly caused the merge conflict in this patch so you're probably going to have to re-work it even for your downstream fork. By using trunk_details you should now be able to implement this with zero cost for anybody not using the feature. Are you likely to update the PR?

Thank you for the information. Sure, I will update the PR.

PR #2232, opened 2023-05-04
[occm] Fixed the typo in the load balancing section in the README

armagankaratosun:
[occm] Fixed the typo in the load balancing section in the README

If you set the variable with a lowercase "b", the template will not catch that and the secret that is created via the helm chart will lack the [LoadBalancer] section. the correct value should be "loadBalancer", instead of "loadbalancer", as you can see from the values.yaml.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/openstack-cloud-controller-manager/Chart.yaml
+++ b/charts/openstack-cloud-controller-manager/Chart.yaml
@@ -4,7 +4,7 @@ description: Openstack Cloud Controller Manager Helm Chart
 icon: https://object-storage-ca-ymq-1.vexxhost.net/swift/v1/6e4619c416ff4bd19e1c087f27a43eea/www-images-prod/openstack-logo/OpenStack-Logo-Vertical.png
 home: https://github.com/kubernetes/cloud-provider-openstack
 name: openstack-cloud-controller-manager
-version: 2.28.0-alpha.5
+version: 2.28.0-alpha.6
 maintainers:
   - name: eumel8
     email: f.kloeker@telekom.de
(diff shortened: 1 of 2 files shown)


armagankaratosun:
sorry, i made a mistake while rebasing my previous pull request https://github.com/kubernetes/cloud-provider-openstack/pull/2183 and needed to reopen this.

jichenjc:
seems there are similiar 2 commits in the PR and  ,is it because of git rebase?

armagankaratosun:
> seems there are similiar 2 commits in the PR and ,is it because of git rebase?

I made a typo once, then fixed it again.

mdbooth:
These were broken until recently. Might pass now.

However, you are right that we don't need to run them against a README change. There's stuff we can configure in prow to do this, but we need to be careful with it.

jichenjc:
>>>However, you are right that we don't need to run them against a README change. There's stuff we can configure in prow to do this, but we need to be careful with it.

right , previously there are changes that make some file change didn't run CI.. so need to be careful

now I am confused why it reports issue against helm chart, maybe you can try rebase?
or we can override the failure if it's possible

armagankaratosun:
> > > > However, you are right that we don't need to run them against a README change. There's stuff we can configure in prow to do this, but we need to be careful with it.
> 
> right , previously there are changes that make some file change didn't run CI.. so need to be careful
> 
> now I am confused why it reports issue against helm chart, maybe you can try rebase? or we can override the failure if it's possible

ok, trying to rebase again

mdbooth:
An obvious fix.

/lgtm
/assign @zetaab

zetaab (approved):
/approve
thank you

mdbooth:
@armagankaratosun  I think that lint check means you need to bump a version number in the helm charts, but:
* honestly I haven't worked it out yet myself.
* I don't understand why that would be needed for this PR.

I think @zetaab has knowledge of the deep magic?

kayrus:
Since this README.md file was modified under the chart dir, it's required to manually bump the minor chart version before the merge. See https://github.com/kubernetes/cloud-provider-openstack/pull/2254/files for example.

armagankaratosun:
> Since this README.md file was modified under the chart dir, it's required to manually bump the minor chart version before the merge. See https://github.com/kubernetes/cloud-provider-openstack/pull/2254/files for example.

Oh, now it is clear, thanks! Then, I will rebase this pull request again, with the upgraded chart version version.  TBH, now that I understand what the issue is, it started to make perfect sense 😄

Issue #2148, opened 2023-03-08
[occm]: introduce readiness gates for pods


kayrus:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**What happened**:

Current LoadBalancer service implementation has a flaw: if during the deployment update a node's network is broken, but kubelet advertises that the pod is up and ready, the new deployment can cause an outage by shutting down old healthy pods.

**What you expected to happen**:

There should be a way to monitor pod's network readiness outside the k8s cluster using pods' [readiness gates](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-readiness-gate). See more details in this video: https://www.youtube.com/watch?v=Vw9GmSeomFg&t=289s

**Anything else we need to know?**:

In #1720 PR the `externalTrafficPolicy: Local` was introduced, which adds kube-proxy based monitors to LB pool members. We can keep this logic for OCCM configured without a router.

However for OCCM with a router configured, we can use service's endpoints instead of node ports and according to loadbalancer's member healthchecks patch the pods' readiness gates accordingly. This approach would increase a deployment update time (because of the LB healthchecks latency), but from other side this will increase the overall deployment accessibility. Additional set of advantages: more even traffic distribution between pods, and pod-based traffic affinity.

See also https://cloud.google.com/kubernetes-engine/docs/concepts/container-native-load-balancing#pod_readiness and https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.1/deploy/pod_readiness_gate/ descriptions.
Unlike GCE and AWS dedicated ingress/LB controllers, I'd like to implement readiness gates feature directly in the OCCM controller for seamless feature toggle, especially for existing deployments.

Feel free to provide your suggestions or objections.

cc @databus23 @jichenjc @zetaab

zetaab:
can current OCCM loadbalancer implementation route traffic to service endpoints? I cannot see anything in code, so I think its not working (yet?).

kayrus:
> can current OCCM loadbalancer implementation route traffic to service endpoints? I cannot see anything in code, so I think its not working (yet?).

if routes a defined on the router configured for the private network, then traffic from loadbalancer to a particular pod CIDR is routed through a router to a corresponding node.

> Another thing that I am thinking: when you create octavia amphora loadbalancer, how you are going to add routes to that?

routes are configured on the router, see above.

[...]

PR #2234, opened 2023-05-06
occm: LB: Fall back to TCP health mon w/o HTTP.

### LINKED ISSUE #2225: [occm] octavia: support fallback to TCP health monitor
/kind feature

The OVN provider load-balancer in OpenStack is getting in shape ... and in many ways a nice option when used in k8s, e.g. in front of an ingress controller:
- It's lightweight, fast to create, light on resource consumption (compared to amphorae)
- It preserves the client IPs without resorting to hacks like the PROXY protocol

Unfortunately, the health monitors for OVN LBs don't support HTTP mode, only TCP (and UDP-CONNECT). This may be resolved in OVN/octavia/neutron some day, but it does not appear trivial to me.

It would be nice if OCCM would allow using a TCP health-monitor instead of HTTP.

**Environment**:
OCCM-1.26.2
OpenStack Zed
OVN-23.03

garloff:
If the Loadbalancer does not support HTTP health-monitor, we currently fail to create one. This is currently the case with OVN loadbalancers, and knowledge that HTTP health-mon is unsupported is hardcoded into occm. While HTTP is preferable and is on the roadmap for OVN, it is not straight-forward and not likely to happen very soon.

TCP health-monitoring is supported and still detects the vast majority of cases with unavailable service, e.g. situations where a node is unavailable, the listening process has died or the node is severed from the network. So TCP health-mon is much better than no health-mon.

This patch implements an unconditional fall back to TCP health monitoring if HTTP health monitoring is not available. This is similar to UDP where we already use UDP-CONNECT health mon. The patch also outputs the type of health mon into the logs.

Tested with OpenStack Zed using OvS-3.1.1 and OVN 23.03 (SCS R4), k8s-v1.26 and -v1.27 and occm-v.26.2 and -1.27. Test scenario using nginx-ingress with externalTrafficPolicy: Local.

Without this patch, OVN provider octavia loadbalancers can not have health-monitors for TCP listeners, as OCCM wants to create an HTTP health-monitor, which OVN LBs don't currently support.
With it, we fall back to TCP health-monitors, which still cover the vast majority of situations with non-working backend member service and is certainly much better than no health-mon.
This makes OVN LBs a really attractive choice in front of a service with externalTrafficPolicy: local, as OVN LB does retain the original client IP (unlike the amphorae where you need to do workarounds with proxy protocol to get the information). 

fixes #2225

Containers to test are on:
registry.scs.community/occm-kg/garloff/openstack-cloud-controller-manager:v1.27.103 
registry.scs.community/occm-kg/garloff/openstack-cloud-controller-manager:v1.26.203 

I would like to see this patch applied not just to master, but also v1.27, v1.26 and v1.25 stable branches.

**Release note**:

```release-note
[occm] Support health-mon for OVN loadbalancers by falling back from HTTP to TCP.
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1156,7 +1156,7 @@ func (lbaas *LbaasV2) buildMonitorCreateOpts(svcConf *serviceConfig, port corev1
 			opts.URLPath = "/healthz"
 			opts.HTTPMethod = "GET"
 			opts.ExpectedCodes = "200"
-		} else {
+		} else if port.Protocol == corev1.ProtocolTCP  {
 			opts.Type = "TCP"
 		}
 	}
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1156,7 +1156,7 @@ func (lbaas *LbaasV2) buildMonitorCreateOpts(svcConf *serviceConfig, port corev1
 			opts.URLPath = "/healthz"
 			opts.HTTPMethod = "GET"
 			opts.ExpectedCodes = "200"
-		} else if port.Protocol == corev1.ProtocolTCP  {
+		} else if port.Protocol == corev1.ProtocolTCP {
 			opts.Type = "TCP"
 		}
 	}
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -527,7 +527,7 @@ func (lbaas *LbaasV2) createFullyPopulatedOctaviaLoadBalancer(name, clusterName 
 			opts := lbaas.buildMonitorCreateOpts(svcConf, port)
 			opts.Name = cutString(fmt.Sprintf("monitor_%d_%s", port.Port, name))
 			poolCreateOpt.Monitor = &opts
-			withHealthMonitor = " with healthmonitor"
+			withHealthMonitor = fmt.Sprintf(" with %s healthmonitor", opts.Type)
 		}
 
 		listenerCreateOpt.DefaultPool = &poolCreateOpt
@@ -1150,11 +1150,15 @@ func (lbaas *LbaasV2) buildMonitorCreateOpts(svcConf *serviceConfig, port corev1
 	if port.Protocol == corev1.ProtocolUDP {
 		opts.Type = "UDP-CONNECT"
 	}
-	if svcConf.healthCheckNodePort > 0 && lbaas.canUseHTTPMonitor(port) {
-		opts.Type = "HTTP"
-		opts.URLPath = "/healthz"
-		opts.HTTPMethod = "GET"
-		opts.ExpectedCodes = "200"
+	if svcConf.healthCheckNodePort > 0 {
+		if lbaas.canUseHTTPMonitor(port) {
+			opts.Type = "HTTP"
+			opts.URLPath = "/healthz"
+			opts.HTTPMethod = "GET"
+			opts.ExpectedCodes = "200"
+		} else {
+			opts.Type = "TCP"
+		}
 	}
 	return opts
 }


garloff:
This PR addresses issue #2225.

jichenjc:
/approve

@dulek please comment in case you have ,thanks

mdbooth:
I'm struggling to find any documentation for the other end of this health check, but this GKE documentation hints that this behaviour might not be correct: https://cloud.google.com/kubernetes-engine/docs/concepts/service-load-balancer-parameters#spd-hc-port

> Note: The kube-proxy always responds to packets sent to the health check node port. Health checks for LoadBalancer Services cannot be answered by serving Pods.

A TCP check would presumably always return success in this case.

dulek:
> I'm struggling to find any documentation for the other end of this health check, but this GKE documentation hints that this behaviour might not be correct: https://cloud.google.com/kubernetes-engine/docs/concepts/service-load-balancer-parameters#spd-hc-port
> 
> > Note: The kube-proxy always responds to packets sent to the health check node port. Health checks for LoadBalancer Services cannot be answered by serving Pods.
> 
> A TCP check would presumably always return success in this case.

I think this piece [1] would make the health monitor to actually check the nodePort and not healthCheckNodePort in this case.

[1] https://github.com/openshift/cloud-provider-openstack/blob/a7c2d10b1add321cf1bee5f8a2c7896d51e42a40/pkg/openstack/loadbalancer.go#L1288-L1290

pkg/openstack/loadbalancer.go
```diff
@@ -1150,11 +1150,15 @@ func (lbaas *LbaasV2) buildMonitorCreateOpts(svcConf *serviceConfig, port corev1
...
-		opts.ExpectedCodes = "200"
+	if svcConf.healthCheckNodePort > 0 {
+		if lbaas.canUseHTTPMonitor(port) {
+			opts.Type = "HTTP"
+			opts.URLPath = "/healthz"
+			opts.HTTPMethod = "GET"
+			opts.ExpectedCodes = "200"
+		} else {
+			opts.Type = "TCP"
+		}
```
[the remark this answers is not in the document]
garloff: > Do you think we could emit a warning event if we're doing a fallback to TCP?

That's why we output the type of health-monitor later on, so we can see it in the logs.
A warning might be more prominent, and I considered it, but we have not enough context in this function to output the affected load-balancer along with the warning.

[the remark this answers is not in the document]
garloff: > > Also changing this that way breaks the UDP fallback. This unconditional `else` in line 1159 will invalidate `UDP-CONNECT` fallback.
> 
> This should not happen and I thought it could not. Looking at the code again, I'm not convinced any more either. Thanks for carefully looking! I'll come back with an updated patch, if needed, sorry.

The `healthCheckNodePort == 0` for `UDP`, so this won't break `UDP` AFAICS.

mdbooth:
I still don't understand how this would work. [The documentation for the TCP health monitor](https://docs.openstack.org/octavia/latest/user/guides/basic-cookbook.html#other-health-monitors) says:

> TCP health monitors open a TCP connection to the back-end server’s protocol port. Your custom TCP application should be written to respond OK to the load balancer connecting, opening a TCP connection, and closing it again after the TCP handshake without sending any data.

But we don't have a custom TCP application running on `healthCheckNodePort` which returns `OK`, do we?

garloff:
> But we don't have a custom TCP application running on `healthCheckNodePort` which returns `OK`, do we?

And we should not need one. Normally TCP health checking means opening a TCP connection (SYN packet) and considering the connection to be successful, if the connection is established (SYN/ACK).
https://www.haproxy.com/documentation/aloha/latest/load-balancing/health-checks/tcp/
That is a proper L4 check, sending data or waiting for some specific "OK" would create a L7 check (and we should then use HTTP, of course.)
So I suspect the OpenStack docu to be misleading here. Will do some testing to be sure ...

garloff:
I pushed an explicit check to ensure we can only use TCP health-mon if we have a TCP LB.
(The UDP case was safe before, but let's make it explicit and robust.)
Testing: docker pull registry.scs.community/occm-kg/garloff/openstack-cloud-controller-manager@sha256:6c3d70c3a10e0411a6696c13634991ea033f1831f298e724c6d9343f7bc1e3fa//

mdbooth:
We have 2 major mismatches here:

We have Octavia documentation suggesting that a TCP check expects an `OK` response, but nothing will deliver that. The suggestion is that the documentation is wrong. If that's the case I'd prefer to see strong evidence of that, a long-form comment explaining this the the future similarly-confused maintainer, and preferably a linked docs patch against the Octavia docs to fix it.

If the Octavia documentation is not correct and the TCP check is indeed just a SYN/ACK check then we still have documentation (admittedly for GKE) suggesting that we will *always* get a response on this port whether a pod is present or not, suggesting that this isn't a valid health check.

We need code comments strongly backing up whatever choice we make here. As it stands I can't convince myself that this behaviour is correct. That's not to say it's incorrect, but it's sufficiently in doubt that we need to carefully document the assumptions we've made so they can be confidently challenged by future-us.

pkg/openstack/loadbalancer.go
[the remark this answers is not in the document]
dulek: No worries, this code is overly complicated, but so is caring about Octavia incompatibilities.

UDP can still have `healthCheckNodePort` if Service has ETP=Local. But I think with the check for `protocol=TCP` we're fine now.

dulek:
> We have 2 major mismatches here:
> 
> We have Octavia documentation suggesting that a TCP check expects an `OK` response, but nothing will deliver that. The suggestion is that the documentation is wrong. If that's the case I'd prefer to see strong evidence of that, a long-form comment explaining this the the future similarly-confused maintainer, and preferably a linked docs patch against the Octavia docs to fix it.

Sure, makes sense, anyway this wording sounds like it's just a connection being created:

> TCP health monitors open a TCP connection to the back-end server’s protocol port. Your custom TCP application should be written to respond OK to the load balancer connecting, opening a TCP connection, and closing it again after the TCP handshake **without sending any data**.

> If the Octavia documentation is not correct and the TCP check is indeed just a SYN/ACK check then we still have documentation (admittedly for GKE) suggesting that we will _always_ get a response on this port whether a pod is present or not, suggesting that this isn't a valid health check.

In this case we won't be calling that endpoint at all as we will not use `healthCheckNodePort` when `lbaas.canUseHTTPMonitor()` returns false [1]. Healthmonitor will call the Service NodePort directly and with ETP=Local that's a good source of truth about which nodes are hosting the Service pods.

> We need code comments strongly backing up whatever choice we make here. As it stands I can't convince myself that this behaviour is correct. That's not to say it's incorrect, but it's sufficiently in doubt that we need to carefully document the assumptions we've made so they can be confidently challenged by future-us.

+1 for more comments! I haven't added enough myself. We should especially add comment to [1].

[1] https://github.com/kubernetes/cloud-provider-openstack/blob/bd767ba5d4493d2aa7965d57286f1e6460e9dba9/pkg/openstack/loadbalancer.go#L1288-L1290

pkg/openstack/loadbalancer.go
```diff
@@ -1150,11 +1150,15 @@ func (lbaas *LbaasV2) buildMonitorCreateOpts(svcConf *serviceConfig, port corev1
...
-		opts.ExpectedCodes = "200"
+	if svcConf.healthCheckNodePort > 0 {
+		if lbaas.canUseHTTPMonitor(port) {
+			opts.Type = "HTTP"
+			opts.URLPath = "/healthz"
+			opts.HTTPMethod = "GET"
+			opts.ExpectedCodes = "200"
+		} else if port.Protocol == corev1.ProtocolTCP {
+			opts.Type = "TCP"
+		}
```
dulek, line 1161: I just realized that `opts.Type` here will already be `TCP` as set in line 1145. So… this should just work normally without this change…?

dulek (changes requested):

pkg/openstack/loadbalancer.go
zetaab: if I am reading the code correctly, if kubernetes service protocol is TCP this should work without change?

mdbooth:
@dulek This is approved but it sounds like we have doubts about it now?

dulek:
> @dulek This is approved but it sounds like we have doubts about it now?

Yes, we feel like the use case described in the PR and bug should just work in current state of the code. Waiting for @garloff input here.

mdbooth:
@dulek @jichenjc We should probably remove the approval. If the bot complains that I'm not allowed to do this, it's probably the right command :)

/approve cancel

pkg/openstack/loadbalancer.go
```diff
@@ -1150,11 +1150,15 @@ func (lbaas *LbaasV2) buildMonitorCreateOpts(svcConf *serviceConfig, port corev1
 	if port.Protocol == corev1.ProtocolUDP {
 		opts.Type = "UDP-CONNECT"
 	}
-	if svcConf.healthCheckNodePort > 0 && lbaas.canUseHTTPMonitor(port) {
-		opts.Type = "HTTP"
-		opts.URLPath = "/healthz"
-		opts.HTTPMethod = "GET"
-		opts.ExpectedCodes = "200"
+	if svcConf.healthCheckNodePort > 0 {
```
kayrus, line 1153: I'd make the opts return right away, e.g. 
```suggestion
	if svcConf.healthCheckNodePort == 0 {
		return opts
	}
```

and place the rest if/then outside the `if svcConf.healthCheckNodePort > 0 {` condition.

pkg/openstack/loadbalancer.go
ProbstDJakob: I can confirm this. We currently run an OKD with the cloud-provider forked for OpenShift based on the commit [https://github.com/openshift/cloud-provider-openstack/tree/6b8f88a9c34a541ffaf4b0abf1d8cd1acb55ff9b](6b8f88a) which is equal to this MRs base regarding the changes.

Although the official documentation of OpenStack states that health monitors are not supported (see [limitations](https://docs.openstack.org/networking-ovn/latest/contributor/design/loadbalancer.html#limitations)), the code exists and works. In our OpenStack cluster we do currently lack [a commit fixing an issue with the creation order of the health monitors and the floating ip](https://opendev.org/openstack/ovn-octavia-provider/commit/8eef0beb1d554f5bfb7426723a830033999fbec5). This means we must first create the service of type `LoadBalancer` and after successful creation of said in OpenStack we can add the annotation `loadbalancer.openstack.org/enable-health-monitor: "true"` (or if added via helm, after creation remove and re-add). Otherwise the internal IP of the OpenStack load balancer would correctly redirect to healthy nodes but the floating IP would redirect to every node in the pool.

The only weird thing we encountered was that the OpenStack pool marks nodes as healthy which lack the pod providing the service. In our case we deployed the nginx ingress controller (NIC) which creates a service of type `LoadBalancer` with port 80 and 443. Before knowing about the bug with the creation order both pools contained healthy marked nodes which did not run a NIC pod and if I remember correctly nodes running a NIC pod did not always get marked as healthy. Also contacting the floating IP was mostly rejected which later turned out to be caused by the described bug. Maybe this behaviour was also caused by the wild back and forth switching between options (we were testing how/if this works). After finding out about this bug I configured the NIC productively and freshly deployed it, removed the annotation and added it back after the health monitors have been removed in OpenStack. Now contacting the floating IP with netcat on both ports works repeatedly as expected, although in the pool for the port 80 is a node (`master-2`) which does not run a NIC pod. Opening a connection to `master-2` within the cluster from another node to the allocated `NodePort` for port 80 does result in a rejection, thus the implementation of the cloud provider/kubernetes works as expected and the issue seems to be on the OpenStack side. Also turning off `master-2` does not remove the healthy mark. But removing a pod and rescheduling it on another node results in a correct update (removing the healthy mark on the old node and adding it to the new node).

(Just to be clear, with the paragraph above I did not want to state that something is wrong with the cloud provider or that we need help/explanation, but only what we found out/which problems (probably only on OpenStack side) we have encountered.)

TL;DR: The provider seems to work without this MR and OpenStack successfully creates the health monitors which work as expected although the horizon UI (and CLI) sometimes marks nodes as healthy which are not running the service.

The only thing which should be updated is the documentation in my opinion. The note that the provider does not support OVN health monitors is proved to be wrong and the supported load balancer method (`lb-method`) for OVN should be changed to `SOURCE_IP_PORT`. Currently the documentation states that it must be `SOURCE_IP` which is not a valid option, at least for OVN.
Maybe if the configuration only contains the `lb-provider` set to `ovn` the code should not default to `ROUND_ROBIN` for the `lb-method`, but that would only be a nice to have feature.
Furthermore the documentation states that the only supported `lb-provider`s are `amphora` and `octavia`, but the code also contains `ovn` in the supported load balancer provider list, thus either the documentation is outdated or the code contains `ovn` although it should not.

dulek: > I can confirm this. We currently run an OKD with the cloud-provider forked for OpenShift based on the commit [https://github.com/openshift/cloud-provider-openstack/tree/6b8f88a9c34a541ffaf4b0abf1d8cd1acb55ff9b](6b8f88a) which is equal to this MRs base regarding the changes.
> 
> Although the official documentation of OpenStack states that health monitors are not supported (see [limitations](https://docs.openstack.org/networking-ovn/latest/contributor/design/loadbalancer.html#limitations)), the code exists and works. In our OpenStack cluster we do currently lack [a commit fixing an issue with the creation order of the health monitors and the floating ip](https://opendev.org/openstack/ovn-octavia-provider/commit/8eef0beb1d554f5bfb7426723a830033999fbec5). This means we must first create the service of type `LoadBalancer` and after successful creation of said in OpenStack we can add the annotation `loadbalancer.openstack.org/enable-health-monitor: "true"` (or if added via helm, after creation remove and re-add). Otherwise the internal IP of the OpenStack load balancer would correctly redirect to healthy nodes but the floating IP would redirect to every node in the pool.

It's this thing, it's fixed: https://bugs.launchpad.net/neutron/+bug/1997418

> The only weird thing we encountered was that the OpenStack pool marks nodes as healthy which lack the pod providing the service. In our case we deployed the nginx ingress controller (NIC) which creates a service of type `LoadBalancer` with port 80 and 443. Before knowing about the bug with the creation order both pools contained healthy marked nodes which did not run a NIC pod and if I remember correctly nodes running a NIC pod did not always get marked as healthy. Also contacting the floating IP was mostly rejected which later turned out to be caused by the described bug. Maybe this behaviour was also caused by the wild back and forth switching between options (we were testing how/if this works). After finding out about this bug I configured the NIC productively and freshly deployed it, removed the annotation and added it back after the health monitors have been removed in OpenStack. Now contacting the floating IP with netcat on both ports works repeatedly as expected, although in the pool for the port 80 is a node (`master-2`) which does not run a NIC pod. Opening a connection to `master-2` within the cluster from another node to the allocated `NodePort` for port 80 does result in a rejection, thus the implementation of the cloud provider/kubernetes works as expected and the issue seems to be on the OpenStack side. Also turning off `master-2` does not remove the healthy mark. But removing a pod and rescheduling it on another node results in a correct update (removing the healthy mark on the old node and adding it to the new node).

Take a look at https://github.com/kubernetes/cloud-provider-openstack/issues/2148 where we discuss this type of issues. TLDR - this is Octavia's implementation detail that members are considered up by default. Not great.

> (Just to be clear, with the paragraph above I did not want to state that something is wrong with the cloud provider or that we need help/explanation, but only what we found out/which problems (probably only on OpenStack side) we have encountered.)
> 
> TL;DR: The provider seems to work without this MR and OpenStack successfully creates the health monitors which work as expected although the horizon UI (and CLI) sometimes marks nodes as healthy which are not running the service.
> 
> The only thing which should be updated is the documentation in my opinion. The note that the provider does not support OVN health monitors is proved to be wrong and the supported load balancer method (`lb-method`) for OVN should be changed to `SOURCE_IP_PORT`. Currently the documentation states that it must be `SOURCE_IP` which is not a valid option, at least for OVN. Maybe if the configuration only contains the `lb-provider` set to `ovn` the code should not default to `ROUND_ROBIN` for the `lb-method`, but that would only be a nice to have feature. Furthermore the documentation states that the only supported `lb-provider`s are `amphora` and `octavia`, but the code also contains `ovn` in the supported load balancer provider list, thus either the documentation is outdated or the code contains `ovn` although it should not.

Would you be so kind to contribute a PR with docs fixes that you mention here?

jichenjc:
@mdbooth done I removed the approve just now

pkg/openstack/loadbalancer.go
ProbstDJakob: > It's this thing, it's fixed: https://bugs.launchpad.net/neutron/+bug/1997418

Thanks, we know but currently are unable to roll out this fix. I wrote it if someone else has similar problems and for explanation of the following findings.

> Take a look at https://github.com/kubernetes/cloud-provider-openstack/issues/2148 where we discuss this type of issues. TLDR - this is Octavia's implementation detail that members are considered up by default. Not great.

I'm not quite sure if this is related to our problem that the `master-2` stays up no matter what. I know that nodes are initially up and that is not great and learning by reading the issue that nodes will be marked healthy when creating a new one is irritating, but `master-2` is _healthy_ since the creation and that's a few days ago so I do not think those two issues are related or I misunderstood it. The only think I can think off is, that `master-2` does not get checked its health, but I am not sure how to look up when the last health check has been done.

> Would you be so kind to contribute a PR with docs fixes that you mention here?

Of course. Just to be clear I should do the following:
- remove the note that `ovn` does not support the annotation `loadbalancer.openstack.org/enable-health-monitor` within [expose-applications-using-loadbalancer-type-service.md](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md)
- remove the note that `ovn` has limited scope completely and add a note to the `lb-method` that `ovn` only supports `SOURCE_IP_PORT` within the [using-openstack-cloud-controller-manager.md](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md)
- add `SOURCE_IP_PORT` to the list of supported `lb-method`s within the [using-openstack-cloud-controller-manager.md](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md)
  \-
  Please correct me if I am wrong. As far as I know `ROUND_ROBIN`, `LEAST_CONNECTIONS`, and `SOURCE_IP` are valid for `amphora` and `octavia` and `SOURCE_IP_PORT` is only valid (and the only one) for `ovn`, right?
- add `ovn` as supported/tested `lb-provider` within the [using-openstack-cloud-controller-manager.md](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md)

Did I miss something?

dulek: > > It's this thing, it's fixed: https://bugs.launchpad.net/neutron/+bug/1997418
> 
> Thanks, we know but currently are unable to roll out this fix. I wrote it if someone else has similar problems and for explanation of the following findings.
> 
> > Take a look at #2148 where we discuss this type of issues. TLDR - this is Octavia's implementation detail that members are considered up by default. Not great.
> 
> I'm not quite sure if this is related to our problem that the `master-2` stays up no matter what. I know that nodes are initially up and that is not great and learning by reading the issue that nodes will be marked healthy when creating a new one is irritating, but `master-2` is _healthy_ since the creation and that's a few days ago so I do not think those two issues are related or I misunderstood it. The only think I can think off is, that `master-2` does not get checked its health, but I am not sure how to look up when the last health check has been done.

Oh, you're right here, sorry! I'm not sure how to debug that. One thing that we've seen is weird behavior with UDP health monitors, maybe you're using UDP there?

> > Would you be so kind to contribute a PR with docs fixes that you mention here?
> 
> Of course. Just to be clear I should do the following:
> 
> * remove the note that `ovn` does not support the annotation `loadbalancer.openstack.org/enable-health-monitor` within [expose-applications-using-loadbalancer-type-service.md](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md)
> * remove the note that `ovn` has limited scope completely and add a note to the `lb-method` that `ovn` only supports `SOURCE_IP_PORT` within the [using-openstack-cloud-controller-manager.md](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md)
> * add `SOURCE_IP_PORT` to the list of supported `lb-method`s within the [using-openstack-cloud-controller-manager.md](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md)
>   -
>   Please correct me if I am wrong. As far as I know `ROUND_ROBIN`, `LEAST_CONNECTIONS`, and `SOURCE_IP` are valid for `amphora` and `octavia` and `SOURCE_IP_PORT` is only valid (and the only one) for `ovn`, right?
> * add `ovn` as supported/tested `lb-provider` within the [using-openstack-cloud-controller-manager.md](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md)
> 
> Did I miss something?

No, this makes sense, we'll hash out any doubts on the PR itself.

dulek: > > Oh, you're right here, sorry! I'm not sure how to debug that. One thing that we've seen is weird behavior with UDP health monitors, maybe you're using UDP there?
> 
> No, we deployed a standard Nginx ingress controller thus both pools are TCP.

Sounds like an Octavia issue then, isn't it? Thinking about this more… Maybe your master-2 accepts the connections so that the TCP monitor is fine with it? That would be contrary to what we observed.

> > No, this makes sense, we'll hash out any doubts on the PR itself.
> 
> I have opened the PR #2489 and will look into EasyCLA.

ProbstDJakob: As the following test shows the `master-2` does act correct, thus the issue probably is on OpenStack's side.

```sh
[core@prod-k7xrn-master-0 ~]$ nc -zvw 2 prod-k7xrn-master-2 30507 # does NOT run a pod
nc: connect to prod-k7xrn-master-2 (10.0.3.15) port 30507 (tcp) failed: Connection refused
[core@prod-k7xrn-master-0 ~]$ nc -zvw 2 prod-k7xrn-worker-0-9fv4m 30507 # does run a pod
Connection to prod-k7xrn-worker-0-9fv4m (10.0.1.171) 30507 port [tcp/*] succeeded!
```

A screenshot of the members of the port 80 pool (in https://github.com/kubernetes/cloud-provider-openstack/pull/2489#discussion_r1422639496 more images show the state of the load balancer):
Pool for port 80

garloff:
This one is no longer needed.
Thanks for all the explanations and doc improvements!

PR #2238, opened 2023-05-09
[cinder-csi-plugin] shrink image, remove unnecessary utils (#2233)

### LINKED ISSUE #2233: [cinder-csi-plugin] Node plugin image with utils

>
> /kind bug
/kind feature

**What happened**:

The cinder-csi-plugin image runs as root in privileged mode (not a secrets) on each node.
And this image includes many packages inside ~126Mb.
CSI-Node-Plugin does not use those packages. Such as python, perl, edit, swapon, bash... 

GCP-CSI Plugin https://github.com/nberlee/gcp-compute-persistent-disk-csi-driver/blob/master/Dockerfile#L46 shrinks the image, removes most of the utils/libraries. 

So, what do you think? How we can repeat this idea?
Reuse the image: use gke.gcr.io/gcp-compute-persistent-disk-csi-driver, and replace csi plugin by openstack version.
Or make our own scripts, which does like the same? 

Thanks.

**How to reproduce it**:

**Environment**:
- OpenStack version:
- Others:

sergelogvinov:

fixes #2233

I've added small shell script `tools/csi-deps.sh`. 
It copies add utils which go-modules calls.
 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -106,11 +106,33 @@ CMD ["sh", "-c", "/bin/barbican-kms-plugin --socketpath ${socketpath} --cloud-co
 ##
 ## cinder-csi-plugin
 ##
-FROM --platform=${TARGETPLATFORM} ${DEBIAN_IMAGE} as cinder-csi-plugin
 
-# Install e4fsprogs for format
-RUN clean-install btrfs-progs e2fsprogs mount udev xfsprogs
+# step 1: copy all necessary files from Debian distro to /dest folder
+# all magic heppens in tools/csi-deps.sh
+FROM --platform=${TARGETPLATFORM} ${DEBIAN_IMAGE} as cinder-csi-plugin-utils
 
+RUN clean-install bash rsync mount udev btrfs-progs e2fsprogs xfsprogs
+COPY tools/csi-deps.sh /tools/csi-deps.sh
+RUN /tools/csi-deps.sh
+
+# step 2: check if all necessary files are copied and work properly
+# the build have to finish without errors, but the result image will not be used
+FROM --platform=${TARGETPLATFORM} ${DISTROLESS_IMAGE} as cinder-csi-plugin-utils-check
+
+COPY --from=cinder-csi-plugin-utils /dest /
+COPY --from=cinder-csi-plugin-utils /bin/sh /bin/sh
+COPY tools/csi-deps-check.sh /tools/csi-deps-check.sh
+
+SHELL ["/bin/sh"]
+RUN /tools/csi-deps-check.sh
+
+# step 3: build tiny cinder-csi-plugin image with only necessary files
+FROM --platform=${TARGETPLATFORM} ${DISTROLESS_IMAGE} as cinder-csi-plugin
+
+# Copying csi-deps-check.sh simply ensures that the resulting image has a dependency
+# on cinder-csi-plugin-utils-check and therefore that the check has passed
+COPY --from=cinder-csi-plugin-utils-check /tools/csi-deps-check.sh /bin/csi-deps-check.sh
+COPY --from=cinder-csi-plugin-utils /dest /
 COPY --from=builder /build/cinder-csi-plugin /bin/cinder-csi-plugin
 COPY --from=certs /etc/ssl/certs /etc/ssl/certs
 
--- a/tools/csi-deps.sh
+++ b/tools/csi-deps.sh
@@ -1,0 +1,85 @@
+#!/bin/sh
+
+# Copyright 2022 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+
+#
+# We will copy all dependencies for CSI Node driver to /dest directory
+# all utils are using by cinder-csi-plugin
+# to format/mount/unmount/resize the volumes.
+#
+# It is very important to have slim image,
+# because it runs as root (privileged mode) on the nodes
+#
+
+DEST=/dest
+
+copy_deps() {
+    PROG="$1"
+
+    mkdir -p "${DEST}$(dirname $PROG)"
+
+    if [ -d "${PROG}" ]; then
+        rsync -av "${PROG}/" "${DEST}${PROG}/"
+    else
+        cp -Lv "$PROG" "${DEST}${PROG}"
+    fi
+
+    if [ -x ${PROG} -o $(/usr/bin/ldd "$PROG" >/dev/null) ]; then
+        DEPS="$(/usr/bin/ldd "$PROG" | /bin/grep '=>' | /usr/bin/awk '{ print $3 }')"
+
+        for d in $DEPS; do
+            mkdir -p "${DEST}$(dirname $d)"
+            cp -Lv "$d" "${DEST}${d}"
+        done
+    fi
+}
+
+# Commmon lib /lib64/ld-linux-*.so.2
+# needs for all utils
+mkdir -p ${DEST}/lib64 && cp -Lv /lib64/ld-linux-*.so.2 ${DEST}/lib64/
+
+# This utils are using by
+# go mod k8s.io/mount-utils
+copy_deps /etc/mke2fs.conf
+copy_deps /bin/mount
+copy_deps /bin/umount
+copy_deps /sbin/blkid
+copy_deps /sbin/blockdev
+copy_deps /sbin/dumpe2fs
+copy_deps /sbin/fsck
+copy_deps /sbin/fsck.xfs
+cp /sbin/fsck* ${DEST}/sbin/
+copy_deps /sbin/e2fsck
+# from pkg e2fsprogs - e2image, e2label, e2scrub and etc.
+cp /sbin/e2* ${DEST}/sbin/
+copy_deps /sbin/mke2fs
+copy_deps /sbin/resize2fs
+cp /sbin/mkfs* ${DEST}/sbin/
+copy_deps /sbin/mkfs.xfs
+copy_deps /sbin/xfs_repair
+copy_deps /usr/sbin/xfs_growfs
+copy_deps /usr/sbin/xfs_io
+cp /usr/sbin/xfs* ${DEST}/usr/sbin/
+copy_deps /bin/btrfs
+cp /bin/btrfs* ${DEST}/bin/
+
+# This utils are using by
+# go mod k8s.io/cloud-provider-openstack/pkg/util/mount
+copy_deps /bin/udevadm
+copy_deps /lib/udev/rules.d
+copy_deps /bin/findmnt
--- a/tools/csi-deps-check.sh
+++ b/tools/csi-deps-check.sh
@@ -1,0 +1,41 @@
+#!/bin/sh
+
+# Copyright 2022 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+
+# We will check all necessary utils in the image.
+# They all have to launch without errors.
+
+# This utils are using by
+# go mod k8s.io/mount-utils
+/bin/mount -V
+/bin/umount -V
+/sbin/blkid -V
+/sbin/blockdev -V
+/sbin/dumpe2fs -V
+/sbin/fsck --version
+/sbin/mke2fs -V
+/sbin/mkfs.ext4 -V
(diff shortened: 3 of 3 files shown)


Dockerfile
```diff
@@ -106,11 +106,15 @@ CMD ["sh", "-c", "/bin/barbican-kms-plugin --socketpath ${socketpath} --cloud-co
...
-FROM --platform=${TARGETPLATFORM} ${DEBIAN_IMAGE} as cinder-csi-plugin
+FROM --platform=${TARGETPLATFORM} ${DEBIAN_IMAGE} as cinder-csi-plugin-utils
+
+RUN clean-install bash rsync mount udev btrfs-progs e2fsprogs xfsprogs
+COPY tools/csi-deps.sh /tools/csi-deps.sh
+RUN /tools/csi-deps.sh
 
-# Install e4fsprogs for format
-RUN clean-install btrfs-progs e2fsprogs mount udev xfsprogs
+FROM --platform=${TARGETPLATFORM} ${DEBIAN_IMAGE} as cinder-csi-plugin
```
mdbooth, line 115: IIUC the copy-deps is copying the binaries and all their linker dependencies. Do we still need to be using the debian base image here, or did you intend this to be something like distroless?

tools/csi-deps.sh
```diff
@@ -0,0 +1,68 @@
...
+        DEPS="$(/usr/bin/ldd "$PROG" | /bin/grep '=>' | /usr/bin/awk '{ print $3 }')"
+
+        for d in $DEPS; do
+            mkdir -p "${DEST}$(dirname $d)"
+            cp -Lv "$d" "${DEST}${d}"
+        done
+    fi
+}
+
+# k8s.io/mount-utils
```
mdbooth, line 43: Have you copied this from somewhere by any chance? Exactly how does the list below relate to mount-utils? It certainly looks like it does, but I wonder if you could add a comment explaining exactly how you came by it, which would make it easier to keep up to date in the future.

Dockerfile
sergelogvinov: I prefer `distroless` but i did not try yet to make a manual test...

It would be the best result if we finally switched to distroless image )

mdbooth:
I like where this is going. I wonder why the e2e tests aren't running against it. We definitely need to run those. I also think it's worth jumping directly to distroless.

mdbooth:
Ah: https://github.com/kubernetes/test-infra/blob/1c2c42ad7b86c1390b10b5e4a17f8ec09d09ad7d/config/jobs/kubernetes/cloud-provider-openstack/provider-openstack-presubmits-release-master-config.yaml#L12

```yaml
    run_if_changed: '^(pkg\/util\/|pkg\/openstack\/|cmd\/openstack-cloud-controller-manager\/|tests\/e2e\/cloudprovider\/)'
```

That should probably include Dockerfile changes. I'll submit a PR.

mdbooth:
Hopefully https://github.com/kubernetes/test-infra/pull/29485 would have made this run automatically.

mdbooth:
@sergelogvinov The csi-cinder e2e test passed using debian as the base image. Do you want to update to a distroless base and run it again?

sergelogvinov:
all images now from `DISTROLESS_IMAGE` 🎉

Dockerfile
```diff
@@ -106,11 +106,26 @@ CMD ["sh", "-c", "/bin/barbican-kms-plugin --socketpath ${socketpath} --cloud-co
...
 
+##
+FROM --platform=${TARGETPLATFORM} ${DISTROLESS_IMAGE} as cinder-csi-plugin-utils-check
+
+COPY --from=cinder-csi-plugin-utils /bin/sh /bin/sh
+COPY --from=cinder-csi-plugin-utils /dest /
+COPY tools/csi-deps-check.sh /tools/csi-deps-check.sh
+
+SHELL ["/bin/sh"]
+RUN /tools/csi-deps-check.sh
```
mdbooth, line 123: This looks weird to the casual fly-by maintainer because we're not actually using it. It's purpose is, I assume, to cause an error into the build process if csi-deps.sh didn't do its job properly?

Assuming I've got that right, please could you add the following comments:

* At the top of these section explaining that this is simply a sanity check on the previous build step and its output is not used.
* In csi-deps.sh explaining that any changes there should be reflected in csi-deps-check.sh

mdbooth (commented):
Woot! Please can I request a couple of additional comments to help future me who's completely forgotten what this was about?

Dockerfile
sergelogvinov: Thank you for advice. aseptically "future me" )

I've added a few comments to the shell file and Dockerfile + small refactoring.

mdbooth:
I've got some bad news: I'm pretty sure `cinder-csi-plugin-utils-check` is never built, meaning we're just adding dead code here. The reason is that we build each end-target by name. buildx seems to understand the dependency tree, and because `cinder-csi-plugin-utils-check` isn't a dependency of any of the targets we build, it will never execute and therefore never fail.

I suspected this might be the case and ran `make build-local-image-cinder-csi-plugin` locally and looked at the output. I see it build `cinder-csi-plugin-utils` and then `cinder-csi-plugin`. However, if you want to double check you could deliberately add a failure to csi-deps-check.sh and see if it causes anything to fail.

`csi-deps-check.sh` itself is only 1.1k and executing it doesn't add anything to the image. I wonder if we just move the execution of it into the release image instead. We probably shouldn't add dead code to the build, though.

sergelogvinov:
Yep, I wasn't clear here https://github.com/kubernetes/cloud-provider-openstack/issues/2233#issuecomment-1542153904, I thought we can add this stage to makefile (for ci) somewhere...

To run `csi-deps-check.sh` we need `/bin/sh` which i'd like to avoid.

As another idea:  Add COPY csi-deps-check.sh to image `cinder-csi-plugin`  from `cinder-csi-plugin-utils-check` stage (image) but it will not so clear "for us in future" )

mdbooth:
> Yep, I wasn't clear here [#2233 (comment)](https://github.com/kubernetes/cloud-provider-openstack/issues/2233#issuecomment-1542153904), I thought we can add this stage to makefile (for ci) somewhere...
> 
> To run `csi-deps-check.sh` we need `/bin/sh` which i'd like to avoid.

Of course.

> As another idea: Add COPY csi-deps-check.sh to image `cinder-csi-plugin` from `cinder-csi-plugin-utils-check` stage (image) but it will not so clear "for us in future" )
> 
> ```dockerfile
> # step 2: build tiny cinder-csi-plugin image with only necessary files
> FROM --platform=${TARGETPLATFORM} ${DISTROLESS_IMAGE} as cinder-csi-plugin
> 
> COPY --from=cinder-csi-plugin-utils /dest /
> COPY --from=cinder-csi-plugin-utils-check /tools/csi-deps-check.sh /bin/csi-deps-check.sh <--- here
> COPY --from=builder /build/cinder-csi-plugin /bin/cinder-csi-plugin
> COPY --from=certs /etc/ssl/certs /etc/ssl/certs
> ```

I actually like this idea. I'd just add a comment like:
```dockerfile
# Copying csi-deps-check.sh simply ensures that the resulting image has a dependency on cinder-csi-plugin-utils-check and therefore that the check has passed
COPY --from=cinder-csi-plugin-utils-check /tools/csi-deps-check.sh /bin/csi-deps-check.sh
```

You hopefully aren't going to delete that without reading the comment right next to it.

Can you confirm that it works, though? e.g. Push a change here with a broken csi-deps-check.sh so we can see it fail.

sergelogvinov:
Test case:

```shell
# echo "exit 1" >> tools/csi-deps-check.sh 
# make build-local-image-cinder-csi-plugin
docker buildx build --output type=docker \
                --build-arg VERSION=v1.27.0-alpha.0-32-g71c3849f-dirty \
                --tag registry.k8s.io/provider-os/cinder-csi-plugin:v1.27.0-alpha.0-32-g71c3849f-dirty \
                --target cinder-csi-plugin \
                .
[+] Building 1.0s (17/27)                                                                                                                                                                              
 => [internal] load build definition from Dockerfile                                                                                                                                              0.0s
 => => transferring dockerfile: 32B                                                                                                                                                               0.0s
 => [internal] load .dockerignore                                                                                                                                                                 0.0s
 => => transferring context: 2B                                                                                                                                                                   0.0s
 => [internal] load metadata for registry.k8s.io/build-image/debian-base:bullseye-v1.4.3                                                                                                          0.3s
 => [internal] load metadata for docker.io/library/alpine:3.17.3                                                                                                                                  0.0s
 => [internal] load metadata for registry.k8s.io/build-image/go-runner:v2.3.1-go1.20.3-bullseye.0                                                                                                 0.9s
 => [internal] load metadata for docker.io/library/golang:1.20.3                                                                                                                                  0.0s
 => [certs 1/2] FROM docker.io/library/alpine:3.17.3                                                                                                                                              0.0s
 => [builder 1/6] FROM docker.io/library/golang:1.20.3                                                                                                                                            0.0s
 => [cinder-csi-plugin-utils-check 1/5] FROM registry.k8s.io/build-image/go-runner:v2.3.1-go1.20.3-bullseye.0@sha256:ea665b97deab98ea4e373f9fe207852394b06e24caeacbd66ddf71f8632dec50             0.0s
 => [internal] load build context                                                                                                                                                                 0.1s
 => => transferring context: 11.41kB                                                                                                                                                              0.0s
 => [cinder-csi-plugin-utils 1/4] FROM registry.k8s.io/build-image/debian-base:bullseye-v1.4.3@sha256:3fcf9bcc4949392155cde1f96591ba3eda092f4735a9cc856bf526d71f48af47                            0.0s
 => CACHED [cinder-csi-plugin-utils 2/4] RUN clean-install bash rsync mount udev btrfs-progs e2fsprogs xfsprogs                                                                                   0.0s
 => CACHED [cinder-csi-plugin-utils 3/4] COPY tools/csi-deps.sh /tools/csi-deps.sh                                                                                                                0.0s
 => CACHED [cinder-csi-plugin-utils 4/4] RUN /tools/csi-deps.sh                                                                                                                                   0.0s
 => CACHED [cinder-csi-plugin-utils-check 2/5] COPY --from=cinder-csi-plugin-utils /dest /                                                                                                        0.0s
 => CACHED [cinder-csi-plugin-utils-check 3/5] COPY --from=cinder-csi-plugin-utils /bin/sh /bin/sh                                                                                                0.0s
 => [cinder-csi-plugin-utils-check 4/5] COPY tools/csi-deps-check.sh /tools/csi-deps-check.sh                                                                                                     0.0s
 => ERROR [cinder-csi-plugin-utils-check 5/5] RUN /tools/csi-deps-check.sh                                                                                                                        0.3s
------                                                                                                                                                                                                 
 > [cinder-csi-plugin-utils-check 5/5] RUN /tools/csi-deps-check.sh:
#0 0.252 mount from util-linux 2.36.1 (libmount 2.36.1: selinux, smack, btrfs, namespaces, assert, debug)
#0 0.253 umount from util-linux 2.36.1 (libmount 2.36.1: selinux, smack, btrfs, namespaces, assert, debug)
#0 0.254 blkid from util-linux 2.36.1  (libblkid 2.36.1, 16-Nov-2020)
#0 0.254 blockdev from util-linux 2.36.1
#0 0.255 dumpe2fs 1.46.2 (28-Feb-2021)
#0 0.255        Using EXT2FS Library version 1.46.2
#0 0.256 fsck from util-linux 2.36.1
#0 0.258 mke2fs 1.46.2 (28-Feb-2021)
#0 0.258        Using EXT2FS Library version 1.46.2
#0 0.259 mke2fs 1.46.2 (28-Feb-2021)
#0 0.259        Using EXT2FS Library version 1.46.2
#0 0.260 mkfs.xfs version 5.10.0
#0 0.261 xfs_io version 5.10.0
#0 0.262 xfs_repair version 5.10.0
#0 0.263 xfs_growfs version 5.10.0
#0 0.264 btrfs-progs v5.10.1 
#0 0.267 247
#0 0.268 findmnt from util-linux 2.36.1
------
ERROR: failed to solve: executor failed running [/bin/sh /tools/csi-deps-check.sh]: exit code: 1
make: *** [build-local-image-cinder-csi-plugin] Error 1
```

Dockerfile
```diff
@@ -106,11 +106,31 @@ CMD ["sh", "-c", "/bin/barbican-kms-plugin --socketpath ${socketpath} --cloud-co
...
+
+# step 3: build tiny cinder-csi-plugin image with only necessary files
+FROM --platform=${TARGETPLATFORM} ${DISTROLESS_IMAGE} as cinder-csi-plugin
+
+COPY --from=cinder-csi-plugin-utils /dest /
+COPY --from=cinder-csi-plugin-utils-check /tools/csi-deps-check.sh /bin/csi-deps-check.sh
```
mdbooth, line 133: Please can you put the comment about why this line is here immediately above it? We want a minimum distance from WTF to explanation!

sergelogvinov: Oh, sorry )))
Fixed.

mdbooth (commented):
/lgtm

mdbooth (commented):
I think we have to trust our CI here. If it turns out we don't have enough coverage, we'll need to address that.

Dockerfile
jichenjc: if line 135 is a check ,maybe do it at beginning ? e.g before line 132?

tools/csi-deps.sh
```diff
@@ -0,0 +1,84 @@
...
+copy_deps /bin/mount
+copy_deps /bin/umount
+copy_deps /sbin/blkid
+copy_deps /sbin/blockdev
+copy_deps /sbin/dumpe2fs
+copy_deps /sbin/fsck
+copy_deps /sbin/fsck.xfs
+cp /sbin/fsck* ${DEST}/sbin/
+copy_deps /sbin/e2fsck
+cp /sbin/e* ${DEST}/sbin/
```
jichenjc, line 68: I understand fsck, xfs etc
but e* seems need some explaination here about what will be copied..

sergelogvinov: i've added few comments.
And change `e*` -> `e2*`

mdbooth:
> /retest

Ignore the manila-e2e-test, it's broken.

@zetaab @jichenjc is this good to go?
/lgtm

sergelogvinov:
Thank you so much for taking the time to review and provide feedback. I genuinely appreciate it.

Issue #2142, opened 2023-03-03
[manila]: fix e2e tests


kayrus:
/kind bug

**What happened**:

Manila functional tests are failing with a python dependency problem:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
```

Example: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/cloud-provider-openstack/2141/openstack-cloud-csi-manila-e2e-test-release-126/1631577047371354112

**What you expected to happen**:

Manila tests should pass without errors

**How to reproduce it**:

Submit a PR

zetaab:
this issue does not fix the manila tests, see https://github.com/kubernetes/cloud-provider-openstack/pull/2108#issuecomment-1435650268

gouthampacha:
With pr #2244, we'll address the tests executing; but we have a test failure, and may still need pr #2065. Please keep this issue open until then..

gouthampacha:
fixed with #2244 and failing test was skipped with #2254 - this isn't a CSI driver issue, a bug has been reported against openstack/manila: https://bugs.launchpad.net/manila/+bug/2020816

kayrus:
https://bugs.launchpad.net/manila/+bug/2020816 is in progress now

dulek:
Pretty sure this is solved by now.

PR #2065, opened 2022-12-24
[manila-csi-plugin] e2e fix WIP

gman0:
e2e fix WIP

```release-note
NONE
```

PR #2244, opened 2023-05-13
manila-csi: Fix image tag in e2e test scripts

gouthampacha:

the csi driver images are built with a prefix "v0.0."; but the script isn't using this script when
starting the driver.

related to issue #2142 and pr #2065

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/playbooks/roles/install-csi-manila/tasks/main.yaml
+++ b/tests/playbooks/roles/install-csi-manila/tasks/main.yaml
@@ -123,7 +123,7 @@
       csimanila:
         image:
           repository: {{ remote_registry_host }}/manila-csi-plugin
-          tag: {{ github_pr }}
+          tag: v0.0.{{ github_pr }}
       shareProtocols:
         - protocolSelector: NFS
           fsGroupPolicy: None


zetaab:
```
Summarizing 1 Failure:
  [FAIL] [sig-storage] [manila-csi-e2e] CSI Volumes [Driver: nfs.manila.csi.openstack.org] [Testpattern: Dynamic PV (default fs)] provisioning [It] should provision correct filesystem size when restoring snapshot to larger size pvc [Feature:VolumeSnapshotDataSource]
  ```
  
this might be bug in current code because tests has not been running for long time?

jichenjc:
maybe ... we may think about bypass this test to make manila CI run again
and open a ticket to track the issue?

zetaab:
yeah that could be the way here

/approve

jichenjc:
@gouthampacha can you help comment the failed case and open a ticket to track 
then we can rerun test and get this merged? Thanks

gouthampacha:
Hi, (apologies, lost track here)

> @gouthampacha can you help comment the failed case and open a ticket to track then we can rerun test and get this merged? Thanks

Yes; I tried triaging the failure via the [logs here](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/cloud-provider-openstack/2244/openstack-cloud-csi-manila-e2e-test/1657231851871277056/artifacts/logs/manila-csi-e2e.log). The [test that failed](https://github.com/kubernetes/kubernetes/blob/bdbfbffef35c0ae46b9309fdab642f696e7a856a/test/e2e/storage/testsuites/provisioning.go#L452-L535) was added recently; so it's possible we've not run it here successfully before. I've commented on issue #2142. We can track the failure with that issue.

PR #2246, opened 2023-05-16
[octavia-ingress-controller] support wildcard in the host field

### LINKED ISSUE #2245: [Octavia-ingress-controller] Support wildcard supported by the host field in the ingress spec
/kind feature

Kubernetes ingress support wildcard at the beginning of the `host` field: https://kubernetes.io/docs/concepts/services-networking/ingress/#hostname-wildcards

The `octavia-ingress-controller` does not support wildcard today.

PS; I should have quick PR soon for that.

gmauleon:
Support wildcard for the `host` field in `ingress` resource

fixes #2245

**Release note**:
```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ vendor
 
 # Vscode files
 .vscode
+__debug_bin
 
 # This is where the result of the go build goes
 /output*/
--- a/.gitignore
+++ b/.gitignore
@@ -159,3 +159,7 @@ zz_generated.openapi.go
 
 # local gopath
 .go
+
+### Okteto ###
+okteto.yml
+.stignore
--- a/pkg/ingress/controller/controller.go
+++ b/pkg/ingress/controller/controller.go
@@ -26,6 +26,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -825,16 +826,21 @@ func (c *Controller) ensureIngress(ing *nwv1.Ingress) error {
 
 			if host != "" {
 
-				// Escape dot to be regex proof
-				hostRegex := strings.ReplaceAll(host, ".", "\\.")
+				hostRegex := "^"
+				if strings.HasPrefix(host, "*.") {
+					hostRegex += "[a-z0-9]([-a-z0-9]*[a-z0-9])?\\."
+					hostRegex += regexp.QuoteMeta(host[2:])
+				} else {
+					hostRegex += regexp.QuoteMeta(host)
+				}
 
-				// Handle wildcard
-				hostRegex = strings.ReplaceAll(hostRegex, "*", ".*")
+				hostRegex += fmt.Sprintf("(:%d)?$", port)
 
 				policyRules = append(policyRules, l7policies.CreateRuleOpts{
 					RuleType:    l7policies.TypeHostName,
 					CompareType: l7policies.CompareTypeRegex,
-					Value:       fmt.Sprintf("^%s(:%d)?$", hostRegex, port)})
+					Value:       hostRegex,
+				})
 			}
 
 			// make the pool name unique in the load balancer
--- a/pkg/ingress/controller/controller.go
+++ b/pkg/ingress/controller/controller.go
@@ -824,10 +824,17 @@ func (c *Controller) ensureIngress(ing *nwv1.Ingress) error {
 			var policyRules []l7policies.CreateRuleOpts
 
 			if host != "" {
+
+				// Escape dot to be regex proof
+				hostRegex := strings.ReplaceAll(host, ".", "\\.")
+
+				// Handle wildcard
+				hostRegex = strings.ReplaceAll(hostRegex, "*", ".*")
+
 				policyRules = append(policyRules, l7policies.CreateRuleOpts{
 					RuleType:    l7policies.TypeHostName,
 					CompareType: l7policies.CompareTypeRegex,
-					Value:       fmt.Sprintf("^%s(:%d)?$", strings.ReplaceAll(host, ".", "\\."), port)})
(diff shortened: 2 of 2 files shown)


pkg/ingress/controller/controller.go
```diff
@@ -824,10 +824,17 @@ func (c *Controller) ensureIngress(ing *nwv1.Ingress) error {
 			var policyRules []l7policies.CreateRuleOpts
 
 			if host != "" {
+
+				// Escape dot to be regex proof
+				hostRegex := strings.ReplaceAll(host, ".", "\\.")
+
+				// Handle wildcard
+				hostRegex = strings.ReplaceAll(hostRegex, "*", ".*")
```
mdbooth, line 832: I haven't looked at the full context, but at first glance this doesn't seem to do what it's supposed to. I'm assuming (again: sorry I haven't actually checked properly!) that the user writes `*.example.com`. They expect that to match `foo.example.com`, but do they expect it to match `foo.bar.example.com`? I suspect not, but this looks like it would match it.

pkg/ingress/controller/controller.go
```diff
@@ -824,10 +824,17 @@ func (c *Controller) ensureIngress(ing *nwv1.Ingress) error {
 			var policyRules []l7policies.CreateRuleOpts
 
 			if host != "" {
+
+				// Escape dot to be regex proof
+				hostRegex := strings.ReplaceAll(host, ".", "\\.")
```
mdbooth, line 829: I realise you've just moved this here, but if the goal is to escape a string which will later be interpreted by a regexp parser (which one, btw?) then this isn't it.

mdbooth (commented):
This feels like one of those areas where almost everything is a security landmine. I wonder if there's a library out there which does this safely?

At the very least there has to be something to safely escape a string literal for use in a regexp.

pkg/ingress/controller/controller.go
gmauleon: Fair point I'll see what I can do about that.

pkg/ingress/controller/controller.go
gmauleon: I'm not sure I follow, this was the existing "escaping" in place already and I suppose it was that simple because the `ingress` resource, sanitize its host field via API constraints, more precisely:
> a wildcard DNS-1123 subdomain must start with '*.', followed by a valid DNS subdomain, which must consist of lower case alphanumeric characters, '-' or '.' and end with an alphanumeric character (e.g. '*.example.com', regex used for validation is '\*\.[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')

If I only consider wildcard at the beginning of the host` string and sanitize the rest with something like https://pkg.go.dev/regexp#QuoteMeta, would that be acceptable?

.gitignore
```diff
@@ -159,3 +160,7 @@ zz_generated.openapi.go
 
 # local gopath
 .go
+
+### Okteto ###
+okteto.yml
+.stignore
```
jichenjc, line 166: I have no idea why we need add this?

pkg/ingress/controller/controller.go
```diff
@@ -824,10 +825,22 @@ func (c *Controller) ensureIngress(ing *nwv1.Ingress) error {
 			var policyRules []l7policies.CreateRuleOpts
 
 			if host != "" {
+
+				hostRegex := "^"
```
jichenjc, line 829: I think we need a comment to describe (maybe links as well) to describe why we do like this, maybe sample as well?

pkg/ingress/controller/controller.go
mdbooth: Should have been clearer in my comment: I know you didn't add this line; you just moved it, so this isn't a problem introduced by this PR.

It did make me notice it, though, which is why I commented!

.gitignore
```diff
@@ -24,6 +24,7 @@ vendor
 
 # Vscode files
 .vscode
+__debug_bin
```
mdbooth, line 27: Same as for the Okteto lines below. Could we pull these into a separate PR?

mdbooth:
For reference, here's the [relevant part of the definition of host](https://github.com/kubernetes/kubernetes/blob/189fe3f3e6d157bc63c2276f79c3de860d7d8c57/pkg/apis/networking/types.go#LL457C1-L468C1):
```golang
        // host can be "precise" which is a domain name without the terminating dot of
        // a network host (e.g. "foo.bar.com") or "wildcard", which is a domain name
        // prefixed with a single wildcard label (e.g. "*.foo.com").
        // The wildcard character '*' must appear by itself as the first DNS label and
        // matches only a single label. You cannot have a wildcard label by itself (e.g. Host == "*").
        // Requests will be matched against the host field in the following way:
        // 1. If host is precise, the request matches this rule if the http host header is equal to Host.
        // 2. If host is a wildcard, then the request matches this rule if the http host header
        // is to equal to the suffix (removing the first label) of the wildcard rule.
        // +optional
        Host string
```

pkg/ingress/controller/controller.go
mdbooth: Links to:
* Octavia rule creation API: https://docs.openstack.org/api-ref/load-balancer/v2/?expanded=create-an-l7-rule-detail#create-an-l7-rule
* Ingress Rule.Host: https://github.com/kubernetes/kubernetes/blob/189fe3f3e6d157bc63c2276f79c3de860d7d8c57/pkg/apis/networking/types.go#L444-L467C2

This would also be super simple to unit test if we pulled it into a small separate function, and it's probably complicated enough to deserve it.

mdbooth (commented):
Looks good to me, but agree with Ji's comments about comments and testing.

PR #2248, opened 2023-05-18
[occm] LoadBalancers: Remove dead SG code

dulek:
Seems like there was some dead code related to handling of the security groups in the implementation of the LoadBalancer Services support. This commit removes it. In particular:

* `LoadBalancerOpts.NodeSecurityGroupIDs` is never populated, so we can remove it as well as code using it.
* `IsAllowAll` function is never used. Moreover I have no idea why it was exported.

Please, please double-check if I'm right that this code is not used anywhere.

```release-note
`[LoadBalancer]NodeSecurityGroupIDs` is completely removed from the configuration options. The option was deprecated and unsupported, but remained accessible as `nodesecuritygroupids`. This won't be true anymore.
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -37,7 +37,6 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/rules"
 	neutronports "github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
-	"github.com/gophercloud/gophercloud/pagination"
 	secgroups "github.com/gophercloud/utils/openstack/networking/v2/extensions/security/groups"
 	"gopkg.in/godo.v2/glob"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/openstack/openstack.go
```diff
@@ -96,7 +96,6 @@ type LoadBalancerOpts struct {
 	MonitorTimeout        util.MyDuration     `gcfg:"monitor-timeout"`
 	MonitorMaxRetries     uint                `gcfg:"monitor-max-retries"`
 	ManageSecurityGroups  bool                `gcfg:"manage-security-groups"`
-	NodeSecurityGroupIDs  []string            // Do not specify, get it automatically when enable manage-security-groups. TODO(FengyunPan): move it into cache
```
mdbooth, line 99: I should point out that even though this has no `gcfg` tag and it's clearly intended to be deprecated, I'm pretty sure it is still possible for it to have been specified by the user in the config file using the name `nodesecuritygroupids`.

See https://pkg.go.dev/gopkg.in/gcfg.v1?utm_source=godoc#hdr-Data_structure for gcfg

See https://github.com/openshift/cloud-provider-openstack/blob/71c3849fcc6ea272c78b3fb0268d0f091428163f/pkg/openstack/openstack.go#L177-L178 for where we read this value from a file.

mdbooth (commented):
Probably requires a release note just in case? What would somebody use if there were using it and it went away?

The config parser would ignore it: https://github.com/openshift/cloud-provider-openstack/blob/71c3849fcc6ea272c78b3fb0268d0f091428163f/pkg/openstack/openstack.go#L215-L216

dulek:
Alright, I added the release note. I don't think the option was doing anything useful at that stage, but sure, it's worth to inform the users.

pkg/openstack/loadbalancer.go
```diff
@@ -2533,51 +2533,9 @@ func (lbaas *LbaasV2) EnsureSecurityGroupDeleted(_ string, service *corev1.Servi
...
-		klog.Warningf("Can not find node-security-group from all the nodes of this cluster when delete loadbalancer service %s/%s",
-			service.Namespace, service.Name)
-	} else {
-		// Delete the rules in the Node Security Group
-		for _, nodeSecurityGroupID := range lbaas.opts.NodeSecurityGroupIDs {
-			opts := rules.ListOpts{
-				SecGroupID:    nodeSecurityGroupID,
-				RemoteGroupID: lbSecGroupID,
-			}
-			secGroupRules, err := getSecurityGroupRules(lbaas.network, opts)
```
jichenjc, line 2549: the test shows `getSecurityGroupRules` need to be removed as well

dulek:
Ah, this interfered with my other patch, cool. More code to remove.

PR #2249, opened 2023-05-18
[occm] Make `ensureSecurityRule()` safely idempotent

dulek:
`ensureSecurityRule()` is used to add a rule to an existing security group. At first it was listing all the rules in the SG and then if the requested rule wasn't present it was adding it. This has some risk - the list of SG rules can be modified in-between by some other thread.

This commit changes the logic of that function to just try to create the rule. Neutron API will return 409 Conflict if the rule already exists and if it does OCCM will just ignore the error and move on.

/hold
I still got to test this a bit more. Will unhold once done.

**Release note**:
```release-note
NONE
```


pkg/util/errors/errors.go
```diff
@@ -77,3 +77,17 @@ func IsInvalidError(err error) bool {
...
+func IsConflictError(err error) bool {
+	if _, ok := err.(gophercloud.ErrDefault409); ok {
+		return true
+	}
+
+	if errCode, ok := err.(gophercloud.ErrUnexpectedResponseCode); ok {
+		if errCode.Actual == http.StatusConflict {
+			return true
+		}
+	}
```
mdbooth, line 90: What's this about, btw? Worth a code comment?

pkg/openstack/loadbalancer.go
```diff
@@ -2118,11 +2101,12 @@ func (lbaas *LbaasV2) ensureSecurityRule(
 	}
 
 	mc := metrics.NewMetricContext("security_group_rule", "create")
-	_, err = rules.Create(lbaas.network, sgRuleCreateOpts).Extract()
+	_, err := rules.Create(lbaas.network, sgRuleCreateOpts).Extract()
 	if mc.ObserveRequest(err) != nil {
```
mdbooth, line 2105: Note that by doing this we're generating an error in prometheus in all cases, even when the 'error' is expected as here. We have wrappers for this in CAPO: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/06f81e5b84c0fbeaed2bb71e8ec0ef6b36a90857/pkg/metrics/metrics.go#L56-L76

mdbooth (commented):
Nice cleanup! Safer and with 1 less API call.

pkg/util/errors/errors.go
dulek: It's copied from above. I could probably do some codegolf to avoid code duplication, let's see.

pkg/openstack/loadbalancer.go
dulek: > Note that by doing this we're generating an error in prometheus in all cases, even when the 'error' is expected as here. We have wrappers for this in CAPO: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/06f81e5b84c0fbeaed2bb71e8ec0ef6b36a90857/pkg/metrics/metrics.go#L56-L76

Ouch, you're right here, thanks!

pkg/util/errors/errors.go
dulek: Aw, I'd need to use reflections to pass type to a deduplicated function. Well, it's not Python, I bet it's not worth it.

mdbooth (commented):
/lgtm

dulek:
/hold cancel

I double checked that this code works fine too.

PR #2250, opened 2023-05-19
[occm] Replace call to Nova os-interfaces with direct Neutron call

mdbooth:
os-interfaces is a proxy API for neutron's ports list. This has 2 significant disadvantages to us:
* It adds the overhead of an extra API call every time it is used
* It doesn't contain all the information returned by Neutron

This is a non-functional change, although is does reduce the load on nova-api without reducing functionality.

The primary motivation for this is to enable additional filtering using information from ports, e.g. filtering by port tag or the addition of subports as in #2228.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/instances.go
+++ b/pkg/openstack/instances.go
@@ -27,9 +27,9 @@ import (
 	"strings"
 
 	"github.com/gophercloud/gophercloud"
-	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/attachinterfaces"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
 	"github.com/gophercloud/gophercloud/pagination"
 	"github.com/mitchellh/mapstructure"
 	v1 "k8s.io/api/core/v1"
@@ -43,11 +43,13 @@ import (
 	"k8s.io/cloud-provider-openstack/pkg/util"
 	"k8s.io/cloud-provider-openstack/pkg/util/errors"
 	"k8s.io/cloud-provider-openstack/pkg/util/metadata"
+	"k8s.io/cloud-provider-openstack/pkg/util/openstack"
 )
 
 // Instances encapsulates an implementation of Instances for OpenStack.
 type Instances struct {
 	compute          *gophercloud.ServiceClient
+	network          *gophercloud.ServiceClient
 	region           string
 	regionProviderID bool
 	opts             metadata.Opts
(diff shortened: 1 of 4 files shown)


mdbooth:
cc @stephenfin @dulek

stephenfin:
To the best of my knowledge, the only thing exposed by the nova `/server/{server}/os-interface` API that is not exposed by the `/ports?device_id={server}` API is the device tags. If you don't need those, this should be a no-op.

pkg/openstack/instances.go
```diff
@@ -148,13 +150,20 @@ func (os *OpenStack) instances() (*Instances, bool) {
 		return nil, false
 	}
 
+	network, err := client.NewNetworkV2(os.provider, os.epOpts)
+	if err != nil {
+		klog.Errorf("unable to access compute v2 API : %v", err)
```
dulek, line 155: That's network v2 API here.

pkg/openstack/instances.go
```diff
@@ -683,31 +692,19 @@ func getAddressesByName(client *gophercloud.ServiceClient, name types.NodeName,
...
-		}
-		interfaces = append(interfaces, s...)
-		return true, nil
-	})
-	if mc.ObserveRequest(err) != nil {
-		return interfaces, err
+// getAttachedPorts returns a list of ports attached to a server.
+func getAttachedPorts(client *gophercloud.ServiceClient, serverID string) ([]ports.Port, error) {
+	listOpts := ports.ListOpts{
+		DeviceID: serverID,
```
dulek, line 706: It's probably a minor use case, but I think we might want to filter by device_owner too even if Nova[^1] is arrogant enough to not consider other Neutron consumers. See https://bugs.launchpad.net/neutron/+bug/1982373 for context on how we hit that problem in the past. Even if it's "invalid" at the moment, I believe it should be fixed at some point.

[^1]: https://github.com/openstack/nova/blob/27.1.0/nova/api/openstack/compute/attach_interfaces.py#L74

pkg/openstack/instancesv2.go
```diff
@@ -56,13 +57,20 @@ func (os *OpenStack) instancesv2() (*InstancesV2, bool) {
 		return nil, false
 	}
 
+	network, err := client.NewNetworkV2(os.provider, os.epOpts)
+	if err != nil {
+		klog.Errorf("unable to access compute v2 API : %v", err)
```
dulek, line 62: That's network v2 API.

dulek (changes requested):

kayrus:
@mdbooth I don't think it's a good idea. We need to keep the interfaces order, listing neutron ports doesn't provide the exact order. See #863

mdbooth:
You're thinking of server.Addresses, which preserves order. This change doesn't affect that. The API it replaces also doesn't guarantee order.

kayrus:
@mdbooth interesting. so it looks like the neutron default order is a timestamp, right? When an interface is attached to a node, neutron creates a port and the port gets the recent timestamp. If this order corresponds to the os-interfaces order, then I'm fine with a change ;)

mdbooth:
@zetaab @jichenjc Could we merge this?

zetaab (approved):
wow this looks interesting. Thank you!

/approve

PR #2257, opened 2023-05-24
[occm]: fix blackhole route atomic delete logic

### LINKED ISSUE #2256: [occm]: orphaned blackhole routes are not correctly deleted when atomic routes extension is enabled

/kind bug

**What happened**:

the logic for atomic routes deletion doesn't set nexhop.

nexthop should be set

**How to reproduce it**:

remove a k8s node when atomic routes extension is enabled in neutron

#2089

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: 1.25.x, 1.26.x, 1.27.x
- OpenStack version: n/a
- Others: n/a

kayrus:

Orphaned blackhole routes are not deleted with atomic route update due to `NextHop` is empty.

fixes #2256

See #2134

**Release note**:

```release-note
Correctly delete oprhaned blackhole routes when atomic routes extension is enabled
```

### DIFF
--- a/pkg/openstack/routes.go
+++ b/pkg/openstack/routes.go
@@ -421,6 +421,9 @@ func (r *Routes) DeleteRoute(ctx context.Context, clusterName string, route *clo
 	} else {
 		// atomic route update
 		blackhole := route.Blackhole
+		if blackhole {
+			addr = string(route.TargetNode)
+		}
 		route := []routers.Route{{
 			DestinationCIDR: route.DestinationCIDR,
 			NextHop:         addr,


pkg/openstack/routes.go
```diff
@@ -421,6 +421,9 @@ func (r *Routes) DeleteRoute(ctx context.Context, clusterName string, route *clo
 	} else {
 		// atomic route update
 		blackhole := route.Blackhole
+		if blackhole {
+			addr = string(route.TargetNode)
+		}
```
mdbooth, line 426: Trying to get my read round this, but at first glance it doesn't look right to me.

When we add a route NextHop is an address. Why would it be a node name here, and why would it be different for blackhole routes?

kayrus: `route.TargetNode` stands to the node IP, when a node  name cannot be found:

https://github.com/kubernetes/cloud-provider-openstack/blob/5a9459935aa33136eb697a50f1c84502ad18a14b/pkg/openstack/routes.go#L86-L90

See also the logic, when there are no atomic routes:

https://github.com/kubernetes/cloud-provider-openstack/blob/5a9459935aa33136eb697a50f1c84502ad18a14b/pkg/openstack/routes.go#L396-L399

mdbooth: Ok, I see. As that context is in a comment 300 lines away it could probably do with another comment here:

```golang
// TargetNode was set to the NextHop of the route instead of the node name in ListRoutes() if the node was not found
```

kayrus:
> Hmm, no cherry-pick bot?

I already submitted cherry-pick's for releases starting from 1.25

mdbooth:
The problem with that is we need to manually ensure that the backports land in order, and are the same. I might look into enabling the cherry-pick bot.

zetaab (approved):
/approve
/lgtm

PR #2270, opened 2023-06-06
[occm] Get IP addresses of neutron subports v2

jingczhang:
Those VMs are plugged into flat networks and use neutron trunk:

Currenlty, getAttachedPorts() only returns neutron ports directly attached to VM. For vlan-aware VMs, IP addresses are assigned on neutron subports. Subports are attached to the trunk; they are not attached to the VM directly. This pull request changes getAttachedPorts() to return IP addresses of neutron subports when they exist. Without this change, Kubernetes is unable to IP addresses of vlan-aware VMs. This change is transparent to VMs not using neutron trunk.

This is a rebase of the following PR:
https://github.com/kubernetes/cloud-provider-openstack/pull/2228

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1393,8 +1393,10 @@ func (lbaas *LbaasV2) buildListenerCreateOpt(port corev1.ServicePort, svcConf *s
 		listenerCreateOpt.Protocol = listeners.ProtocolHTTP
 	}
 
-	if len(svcConf.allowedCIDR) > 0 {
-		listenerCreateOpt.AllowedCIDRs = svcConf.allowedCIDR
+	if openstackutil.IsOctaviaFeatureSupported(lbaas.lb, openstackutil.OctaviaFeatureVIPACL, lbaas.opts.LBProvider) {
+		if len(svcConf.allowedCIDR) > 0 {
+			listenerCreateOpt.AllowedCIDRs = svcConf.allowedCIDR
+		}
 	}
 	return listenerCreateOpt
 }
@@ -1751,18 +1753,19 @@ func (lbaas *LbaasV2) checkService(service *corev1.Service, nodes []*corev1.Node
 		svcConf.timeoutTCPInspect = getIntFromServiceAnnotation(service, ServiceAnnotationLoadBalancerTimeoutTCPInspect, 0)
 	}
 
-	var listenerAllowedCIDRs []string
 	sourceRanges, err := GetLoadBalancerSourceRanges(service, svcConf.preferredIPFamily)
 	if err != nil {
 		return fmt.Errorf("failed to get source ranges for loadbalancer service %s: %v", serviceName, err)
 	}
 	if openstackutil.IsOctaviaFeatureSupported(lbaas.lb, openstackutil.OctaviaFeatureVIPACL, lbaas.opts.LBProvider) {
 		klog.V(4).Info("LoadBalancerSourceRanges is suppported")
-		listenerAllowedCIDRs = sourceRanges.StringSlice()
+		svcConf.allowedCIDR = sourceRanges.StringSlice()
+	} else if lbaas.opts.LBProvider == "ovn" && lbaas.opts.ManageSecurityGroups {
+		klog.V(4).Info("LoadBalancerSourceRanges will be enforced on the SG created and attached to LB members")
+		svcConf.allowedCIDR = sourceRanges.StringSlice()
 	} else {
-		klog.Warning("LoadBalancerSourceRanges is ignored")
+		klog.Warning("LoadBalancerSourceRanges are ignored")
 	}
-	svcConf.allowedCIDR = listenerAllowedCIDRs
 
 	if openstackutil.IsOctaviaFeatureSupported(lbaas.lb, openstackutil.OctaviaFeatureFlavors, lbaas.opts.LBProvider) {
 		svcConf.flavorID = getStringFromServiceAnnotation(service, ServiceAnnotationLoadBalancerFlavorID, lbaas.opts.FlavorID)
@@ -2009,7 +2012,7 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 	status := lbaas.createLoadBalancerStatus(service, svcConf, addr)
 
 	if lbaas.opts.ManageSecurityGroups {
-		err := lbaas.ensureSecurityGroup(clusterName, service, nodes, loadbalancer, svcConf.preferredIPFamily, svcConf.lbMemberSubnetID)
+		err := lbaas.ensureAndUpdateOctaviaSecurityGroup(clusterName, service, nodes, svcConf)
 		if err != nil {
 			return status, fmt.Errorf("failed when reconciling security groups for LB service %v/%v: %v", service.Namespace, service.Name, err)
 		}
@@ -2068,14 +2071,6 @@ func (lbaas *LbaasV2) ensureSecurityRule(sgRuleCreateOpts rules.CreateOpts) erro
 		return fmt.Errorf("failed to create rule for security group %s: %v", sgRuleCreateOpts.SecGroupID, err)
 	}
 	return nil
-}
-
-// ensureSecurityGroup ensures security group exist for specific loadbalancer service.
-// Creating security group for specific loadbalancer service when it does not exist.
-func (lbaas *LbaasV2) ensureSecurityGroup(clusterName string, apiService *corev1.Service, nodes []*corev1.Node,
-	loadbalancer *loadbalancers.LoadBalancer, preferredIPFamily corev1.IPFamily, memberSubnetID string) error {
-
-	return lbaas.ensureAndUpdateOctaviaSecurityGroup(clusterName, apiService, nodes, memberSubnetID)
 }
 
 func (lbaas *LbaasV2) updateOctaviaLoadBalancer(ctx context.Context, clusterName string, service *corev1.Service, nodes []*corev1.Node) error {
@@ -2140,7 +2135,7 @@ func (lbaas *LbaasV2) updateOctaviaLoadBalancer(ctx context.Context, clusterName
 	}
 
 	if lbaas.opts.ManageSecurityGroups {
-		err := lbaas.updateSecurityGroup(clusterName, service, nodes, svcConf.lbMemberSubnetID)
+		err := lbaas.ensureAndUpdateOctaviaSecurityGroup(clusterName, service, nodes, svcConf)
 		if err != nil {
 			return fmt.Errorf("failed to update Security Group for loadbalancer service %s: %v", serviceName, err)
 		}
@@ -2204,7 +2199,7 @@ func getRulesToCreateAndDelete(wantedRules []rules.CreateOpts, existingRules []r
 }
 
 // ensureAndUpdateOctaviaSecurityGroup handles the creation and update of the security group and the securiry rules for the octavia load balancer
-func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, apiService *corev1.Service, nodes []*corev1.Node, memberSubnetID string) error {
+func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, apiService *corev1.Service, nodes []*corev1.Node, svcConf *serviceConfig) error {
 	// get service ports
 	ports := apiService.Spec.Ports
 	if len(ports) == 0 {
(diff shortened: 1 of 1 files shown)


jingczhang:
Hi @mdbooth, what is the minimum occm release version for getAttachedPorts() to work? 1.27?

zetaab:
@jingczhang new features are not backported to older releases, that is usually how new items are handled. Only bug fixes and security fixes should be backported if needed.

As I see it: this is feature, but it might take sometime to get 1.28 running, so at least I can accept 1.27 backport, but not perhaps older than that.

pkg/openstack/instances.go
```diff
@@ -622,17 +634,47 @@ func nodeAddresses(srv *servers.Server, ports []ports.Port, networkingOpts Netwo
...
+		extraPrivates := make(map[string][]Address)
+		// For each private network
+		for network, interfaceAddresses := range allPrivates {
+			found := false
+			// For each address in the private network
+			for _, interfaceAddress := range interfaceAddresses {
+				// Check if the address is directly assigned
+				for _, srvAddresses := range addresses {
+					for _, srvAddress := range srvAddresses {
+						if interfaceAddress.Addr == srvAddress.Addr {
```
zetaab, line 654: quite many for loops here, perhaps worth of using sub function(s) instead. It makes it more readable and easier to test if needed.

you could make function like `func findInterfaceAddresses(interfaceAddresses, addresses) bool`

pkg/openstack/instances.go
```diff
@@ -707,5 +749,54 @@ func getAttachedPorts(client *gophercloud.ServiceClient, serverID string) ([]por
...
+// getSubInterfaces returns the sub-interfaces attached via trunks to the specified interfaces
+func getSubInterfaces(client *gophercloud.ServiceClient, interfaces []ports.Port) ([]ports.Port, error) {
+	var subports []ports.Port
+	var portExt struct {
+		ports.Port
+		trunk_details.TrunkDetailsExt
+	}
+	for _, iface := range interfaces {
+		// Check if trunk is attached to the interface
+		err := ports.Get(client, iface.ID).ExtractInto(&portExt)
```
zetaab, line 774: could we do this trunk call already in row 752? include trunk info in GetPorts? By that way we could have the same info in single api call instead of own for each interface

zetaab: https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/util/openstack/network.go#L149

```
	allPorts, err := neutronports.ExtractPortsInto(allPages, structhere)
```

could do that.

zetaab: blah, looks like trunk information is available only in single GET call and not in list call at all... anyway I am not quite huge fan of this implementation. It will add unnecessary call for most of the users. Could we have somekind of way to get rid of this? Like config item in occm to enable trunk calls?

pkg/openstack/instances.go
jingczhang: Good point, will change.

pkg/openstack/instances.go
jingczhang: Good point, checking and will do.

zetaab (approved):
/approve

ping @mdbooth as you assigned this to yourself

pkg/openstack/instances.go
mdbooth: It is returned in list queries, as tested in the gophercloud acceptance test: https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/networking/v2/extensions/trunk_details/trunks_test.go

Only the MAC address is missing, and we don't need that here.

mdbooth (changes requested):
I don't understand why we need to fetch ports again when they should already contain trunk details, which also means we should not require an additional configuration option.

Can you explain why trunk details are not included?

pkg/openstack/instances.go
jingczhang: Apparently, I missed the extension to list, I will check and update.

zetaab: @mdbooth then perhaps something is missing from api docs, this is only which I found https://docs.openstack.org/api-ref/network/v2/index.html?expanded=show-trunk-details-detail and that is for single port. I do not have environment which contains trunk ports, so I cannot verify does it really work in list queries as well.

mdbooth: The gophercloud acceptance tests run against an installed OpenStack, so they reflect reality. They are returned for LIST.

docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
```diff
@@ -159,6 +159,8 @@ The options in `Global` section are used for openstack-cloud-controller-manager
 
 ###  Networking
 
+* `trunk-support-enabled`
+  Indicates whether or not IP addresses attached via trunk is supported. Default: false
```
mdbooth, line 163: As noted, this should not be required. The list of returned ports already includes subports. If they are present then trunk support is present and in use. If they are not present then trunk support is not in use.

Configuration is an API. We shouldn't add it where it is not required.

jingczhang: Sure.

pkg/openstack/instances.go
```diff
@@ -707,5 +753,47 @@ func getAttachedPorts(client *gophercloud.ServiceClient, serverID string) ([]por
...
+				continue
+			}
+			p.Status = "ACTIVE"
+			p.NetworkID = n.Name
+			interfaces = append(interfaces, *p)
+		}
+	}
+
+	klog.V(5).Infof("Node %s has %d interfaces '%v'", serverID, len(interfaces), interfaces)
+	return interfaces, nil
```
mdbooth, line 798: Looking at how they're used above, I wonder if we should return the return signature of this method to:

```go
func getAttachedPorts(client *gophercloud.ServiceClient, serverID string) (attachedPorts []ports.Port, subPorts []ports.Ports, err error)
```

And we return direct attached and subports separately. This would mean we wouldn't need the extremely confusing and error prone length conditional above. We could rename `addExtraNodeAddresses` to `addSubPortNodeAddresses` and call it unconditionally with the returned subports. If they're empty or nil it just wouldn't do anything.

mdbooth (changes requested):
This looks on the right lines to me. I still have major reservations about the maintainability of this version, though, but unlike the previous patches I think they're easily fixable. Nice work!

pkg/openstack/instances.go
jingczhang: comment taken.

pkg/openstack/instances.go
```diff
@@ -754,56 +753,47 @@ func getAttachedPorts(client *gophercloud.ServiceClient, networkingOpts Networki
 		DeviceOwner: "compute:nova",
 	}
 
-	interfaces, err := openstack.GetPorts(client, listOpts)
+	type portWithTrunkDetails struct {
+		ports.Port
+		trunk_details.TrunkDetailsExt
+	}
+
+	var interfaces []ports.Port
```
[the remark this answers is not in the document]
jingczhang: comment taken.

pkg/openstack/instances.go
```diff
@@ -707,5 +753,47 @@ func getAttachedPorts(client *gophercloud.ServiceClient, serverID string) ([]por
...
+			}
+			n, err := networks.Get(client, p.NetworkID).Extract()
+			if err != nil {
+				klog.Errorf("Failed to get subport %s network details: %v", subport.PortID, err)
+				continue
+			}
+			p.Status = "ACTIVE"
+			p.NetworkID = n.Name
+			interfaces = append(interfaces, *p)
+		}
```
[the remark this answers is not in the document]
jingczhang: agreed

pkg/openstack/instances.go
```diff
@@ -588,6 +595,10 @@ func nodeAddresses(srv *servers.Server, ports []ports.Port, networkingOpts Netwo
 							Address: fixedIP.IPAddress,
 						},
 					)
+					if port.NetworkID != "" {
```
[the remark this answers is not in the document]
jingczhang: no good reason, should be removed.

pkg/openstack/instances.go
```diff
@@ -577,6 +583,7 @@ func nodeAddresses(srv *servers.Server, ports []ports.Port, networkingOpts Netwo
 	addrs := []v1.NodeAddress{}
 
 	// parse private IP addresses first in an ordered manner
+	allPrivates := make(map[string][]Address)
```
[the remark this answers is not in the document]
jingczhang: I am sorry about that, will change.

pkg/openstack/instances.go
```diff
@@ -707,5 +753,47 @@ func getAttachedPorts(client *gophercloud.ServiceClient, serverID string) ([]por
...
+	}
+	err = ports.ExtractPortsInto(allPages, &allPorts)
+	if err != nil {
+		return interfaces, err
+	}
+
+	for _, portExt := range allPorts {
+		interfaces = append(interfaces, portExt.Port)
+		// Check if trunk is attached to the port
+		if portExt.TrunkDetails.TrunkID == "" {
```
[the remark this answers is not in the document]
jingczhang: comment taken.

pkg/openstack/openstack_test.go
```diff
@@ -878,6 +878,241 @@ func TestNodeAddressesWithAddressSortOrderOptions(t *testing.T) {
 	}
 }
 
+func TestNodeAddressesWithSubports(t *testing.T) {
```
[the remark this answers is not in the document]
jingczhang: We actually only had one bug in the solution, which was in the code added to filter out correct IP addresses in the nodeAddresses() function. I took the opportunity to write and test that use case here. I will simply this test to its key function.

pkg/openstack/instances.go
```diff
@@ -622,17 +633,18 @@ func nodeAddresses(srv *servers.Server, ports []ports.Port, networkingOpts Netwo
...
-	}
-
 	var addresses map[string][]Address
 	err := mapstructure.Decode(srv.Addresses, &addresses)
 	if err != nil {
 		return nil, err
 	}
 
+	// Add the private addresses that are not directly connected
+	if len(addresses) < len(allPrivates) {
```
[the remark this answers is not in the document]
jingczhang: the intend is to include the vlan networks, to which subports are attached, to the list of networks used to filter out the node's ip addresses per networking options. I will add the comment.

pkg/openstack/instances.go
```diff
@@ -686,6 +698,40 @@ func nodeAddresses(srv *servers.Server, ports []ports.Port, networkingOpts Netwo
...
+		if !found {
+			extraPrivates[network] = interfaceAddresses
+		}
+	}
+	for network, interfaceAddresses := range extraPrivates {
+		srvAddresses, ok := addresses[network]
+		if !ok {
+			addresses[network] = interfaceAddresses
+		} else {
+			addresses[network] = append(srvAddresses, interfaceAddresses...)
```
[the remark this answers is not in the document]
jingczhang: This was actually the only "bug" in our solution as I mentioned above. The original code does not have the "else" clause.

pkg/openstack/instances.go
```diff
@@ -686,6 +698,40 @@ func nodeAddresses(srv *servers.Server, ports []ports.Port, networkingOpts Netwo
...
+		// All the addresses in the private network are not directly assigned
+		// Save the private network
+		if !found {
+			extraPrivates[network] = interfaceAddresses
+		}
+	}
+	for network, interfaceAddresses := range extraPrivates {
+		srvAddresses, ok := addresses[network]
+		if !ok {
+			addresses[network] = interfaceAddresses
```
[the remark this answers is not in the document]
jingczhang: Network name is not available in subport info, node IP filtering is on network name, the interest is to get subports' IP. So, to me, this is a safe convenience, for sure, I am for adding network name to subport info.

pkg/openstack/instances.go
jingczhang: Hmm, since the caller of getAttachedPorts does not care the additional level of details I am not sure changing its signature is needed (right thing to do), I can for sure change addExtraNodeAddresses to addSubPortNodeAddresses, for code clarity.

mdbooth: The critical difference is that we subsequently treat subports differently to directly attached ports because the former are not included in server addresses but the latter are. By returning them as a single un-differentiated slice we're making our lives more complicated later when we have to determine which is which.

mdbooth: An alternative might be to not fetch subports here at all and just return directly attached ports. We could scan for subports later when we unconditionally call `addSubPortNodeAddresses`. That would require changing the return signature of this function to return `[]portWithTrunkDetails`.

jingczhang: The alternative sounds reasonable, will check and update.

jingczhang: The alternative seems a better idea, and it results in the same code structure as in the older releases, PR is updated.

dulek:
/retest

This looks great to me, are we waiting for anything else here?

pkg/openstack/instances.go
```diff
@@ -633,6 +635,34 @@ func nodeAddresses(srv *servers.Server, ports []ports.Port, networkingOpts Netwo
 		return nil, err
 	}
 
+	// Add the addresses assigned on subports via trunk
+	// This exposes the vlan networks to which subports are attached
+	if client != nil {
+		subports, err := getSubports(client, srv.ID)
```
mdbooth, line 641: Note that we're calling getSubports unconditionally for everybody here, but we're *not* passing in the list of ports that we already fetched, which already contains the subports.

pkg/openstack/instances.go
```diff
@@ -686,6 +716,51 @@ func nodeAddresses(srv *servers.Server, ports []ports.Port, networkingOpts Netwo
...
+		trunk_details.TrunkDetailsExt
+	}
+
+	var subports []ports.Port
+	var allPortDetails []portWithTrunkDetails
+
+	allPages, err := ports.List(client, listOpts).AllPages()
+	if err != nil {
+		return subports, err
+	}
```
mdbooth, line 737: We can't do this: it's adding the second port list for everybody. We've already done this query. We need to re-use the existing ports query.

mdbooth (changes requested):
This still isn't there. We need to avoid the second ports query.

jingczhang:
Sorry for the late response, I was on vacation, I will take care of the release and new comment soon.

pkg/openstack/instances.go
```diff
@@ -633,6 +635,34 @@ func nodeAddresses(srv *servers.Server, ports []ports.Port, networkingOpts Netwo
 		return nil, err
 	}
 
+	// Add the addresses assigned on subports via trunk
+	// This exposes the vlan networks to which subports are attached
+	if client != nil {
```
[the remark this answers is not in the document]
jingczhang: yes, this checking is added so unit test would not fail

pkg/openstack/instances.go
jingczhang: ok, same as above, rebase results in a new PR, will take care of the comments in the new PR:
https://github.com/kubernetes/cloud-provider-openstack/pull/2306

PR #2272, opened 2023-06-13
[WIP]: fix k3s installation issue

kayrus:

add cloudbuild.yaml from the master branch

fixes #

**Release note**:

```release-note
add cloudbuild.yaml from the master branch
```


zetaab:
this is not enough, you need to modify a lot more files. Like Makefile, dockerfiles etc. I did not cherry pick these changes because it is too much for old releases.

kayrus:
@zetaab can we use old-style build CI?

zetaab:
we can still use old CI for release-1.26 and older. The problem that I see is that we did like 10+ PRs to make release-1.27 and master to work with that. So I am pretty sure that if these are merged, the release branch will be broken after that.

kayrus:
> we can still use old CI for release-1.26 and older. The problem that I see is that we did like 10+ PRs to make release-1.27 and master to work with that. So I am pretty sure that if these are merged, the release branch will be broken after that.

can we at least give it a try?

kayrus:
1.25 is outdated

PR #2280, opened 2023-06-29
[occm] Wait for LB to be ACTIVE on HM update

### LINKED ISSUE #2267: [occm] Wait for ACTIVE load balancer, when updating health monitors

/kind bug

**What happened**:

Load Balancer reconciliation doesn't wait for load balancer to be ACTIVE when updating the health monitor.

When creating or deleting health monitors, we wait for the load balancer to show a provisioning status ACTIVE:
`WaitActiveAndGetLoadBalancer`
https://github.com/kubernetes/cloud-provider-openstack/blob/fd113d2b1d8f5b578ec84c9a9681a7f71ff48fcc/pkg/util/openstack/loadbalancer.go#L681-L709

For the update case, this seems to be missing:
https://github.com/kubernetes/cloud-provider-openstack/blob/fd113d2b1d8f5b578ec84c9a9681a7f71ff48fcc/pkg/util/openstack/loadbalancer.go#L670-L679

**How to reproduce it**:

Would you consider this a bug?

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: v1.26.2

<sub>Sean Schneeweiss <sean.schneeweiss@mercedes-benz.com>, Mercedes-Benz Tech Innovation GmbH, [Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)</sub>

dulek:
This commit makes sure that on exit from UpdateHealthMonitor() the LB is ACTIVE so that next operations can be called safely.

fixes #2267

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/util/openstack/loadbalancer.go
+++ b/pkg/util/openstack/loadbalancer.go
@@ -668,11 +668,15 @@ func CreateL7Rule(client *gophercloud.ServiceClient, policyID string, opts l7pol
 }
 
 // UpdateHealthMonitor updates a health monitor.
-func UpdateHealthMonitor(client *gophercloud.ServiceClient, monitorID string, opts monitors.UpdateOpts) error {
+func UpdateHealthMonitor(client *gophercloud.ServiceClient, monitorID string, opts monitors.UpdateOpts, lbID string) error {
 	mc := metrics.NewMetricContext("loadbalancer_healthmonitor", "update")
 	_, err := monitors.Update(client, monitorID, opts).Extract()
 	if mc.ObserveRequest(err) != nil {
 		return fmt.Errorf("failed to update healthmonitor: %v", err)
+	}
+
+	if _, err := WaitActiveAndGetLoadBalancer(client, lbID); err != nil {
+		return fmt.Errorf("failed to wait for load balancer %s ACTIVE after updating healthmonitor: %v", lbID, err)
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


dulek:
/retest

These look unrelated.

pkg/util/openstack/loadbalancer.go
```diff
@@ -668,13 +668,17 @@ func CreateL7Rule(client *gophercloud.ServiceClient, policyID string, opts l7pol
...
+func UpdateHealthMonitor(client *gophercloud.ServiceClient, monitorID string, opts monitors.UpdateOpts, lbID string) error {
 	mc := metrics.NewMetricContext("loadbalancer_healthmonitor", "update")
 	_, err := monitors.Update(client, monitorID, opts).Extract()
 	if mc.ObserveRequest(err) != nil {
 		return fmt.Errorf("failed to update healthmonitor: %v", err)
 	}
 
+	if _, err := WaitActiveAndGetLoadBalancer(client, lbID); err != nil {
+		return fmt.Errorf("failed to wait for load balancer %s ACTIVE after updating healthmonitor: %v", lbID, err)
+	}
```
mdbooth, line 680: Hmm, this is a bad controller pattern for at least 2 reasons:
* It ties up a controller thread waiting for a single LB to become active, which artificially limits concurrency.
* It takes no account of of a controller restart

Ideally, if an operation requires the LB to be active, we should check it immediately before whatever requires it be active and return from Reconcile() if it's not ready yet.

However, as this pattern seems to be idiomatic in this code base it's probably best to fix that all at once as a separate PR.

mdbooth (commented):
Slightly reluctantly

/lgtm

pkg/util/openstack/loadbalancer.go
dulek: I agree with you on this one, a while ago I was thinking we should change how we handle LB statuses, especially as with Amphora it can be 10s of seconds before LB is `ACTIVE` again. This patch only fixes the current state of things in which we could try listener creation on an LB that's still being `PENDING_UPDATE` because of the health monitor update. Such call would inevitably fail.

PR #2285, opened 2023-06-30
Inhibit Configuration Caching

### LINKED ISSUE #2284: [cinder-csi-plugin] Cloud Configs are Cached

>
> /kind bug
/kind feature

**What happened**:

When you change a password in OpenStack, any associated application credentials stop working.  So some times it's necessary (to be fair, good practice) to rotate the cloud config.  Sadly the CSI plugin (and I'll hazard a guess the rest of them), caches the old value, rather than going to the file system each time to check for an update, so it ceases working.

Everything should expect the cloud config to change, pick up new versions and keep working.

**How to reproduce it**:

Replace the cloud config with a new one, deleting the old application credential.  Provision a PVC and watch it complain the application credential is gone.  Kick the pod and will suddenly start working again.

I guess we could trick things into a reload by modifying the daemonset in some benign way, but it's sub optimal, and requires every client to remember to do this.

Also I'm up for doing the work, if the discussion goes well, seem easy enough on the surface.  Famous last words!

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version:v1.25.0
- OpenStack version: Zed
- Others:

spjmurray:

While caching is good from a performance perspective, I believe in this instance the benefits are miniscule, and it also doesn't handle credential rotation, which any security conscious organisation would tell you is a good thing, so exfiltration doesn't have a lasting impact.

fixes #2284

**Release note**:

```release-note
Prior to this release cloud configs were cached to improve performance, however this had the side-effect of ignoring credential rotation.
The new `--cache-cloud-config` flag allows this behavior to be tailored to your environment.
The default is `true` to maintain the current behavior for backward compatibility.
You can inhibit caching with the `--cache-cloud-config=false` CLI flag.
```


spjmurray:
Issue is open at https://github.com/kubernetes/cloud-provider-openstack/issues/2284 if you want to discuss this in more length

mdbooth:
/lgtm

Incidentally the primary cost of this is going to be in additional hits on keystone, although I think it's fair to let users make this determination themselves. The way we handle this in OpenShift is with [an annotation on the deployed pod based on a hash of the secret contents](https://github.com/openshift/library-go/blob/954ade536d6dd2e690a01921e8a73c67a3edfe46/pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go#L100-L116). This just means we create a new pod if the secret changes.

jichenjc:
are you intend to make this cinder only or the whole CPO (including manila, OCCM etc) ?
or we can merge this then do other update?

kayrus:
From my point of view this feature has to be supported on a lower level in `pkg/client/client.go`. I need some time to take a deeper look into this issue.

mdbooth:
> From my point of view this feature has to be supported on a lower level in `pkg/client/client.go`. I need some time to take a deeper look into this issue.

What are you thinking here? I don't think we'd ever want to update an existing client with new credentials: you'd always create a new one with the new credentials. I'm not convinced our client wrapper needs to support this directly.

kayrus:
> What are you thinking here?

I'm thinking about credentials rotation only when old credentials are failing. Do you have objections?

mdbooth:
> > What are you thinking here?
> 
> I'm thinking about credentials rotation only when old credentials are failing. Do you have objections?

That behaviour would be surprising to me, as well as complex to implement. I'm interested to know if this would be surprising to other users, or if these are just my expectations. I'm thinking that a credential rotation workflow is going to go something like:

1. Create new credentials
2. Update credentials in cluster
3. Ensure that everything is working correctly. Rollback if not.
4. Delete old credentials

I think a user expects that when they update cluster credentials they will be used immediately, which means that they have validated the new credentials in step 3. If we keep using the old credentials until they fail, the user has not validated the credentials in step 3. If there's a problem they might be surprised in step 4 when everything fails and there is no longer a rollback option.

To keep things simple to implement, test, and debug, I think it's simpler that if we get new credentials we use them immediately and start failing immediately if they don't work.

spjmurray:
Oh wow, I missed all this chatter, was commenting in the issue.  I did mention just now that we could do something a bit more clever by polling the file on disk, and comparing with the version that was used to initialize the client.  If that changes then we can perform an atomic update of the client.  That'll behave the same, but alleviate the load on Keystone.  Critiques?

kayrus:
I've spent a while to optimize auth logic from both the code and API performance perspective. Before each CPO component had its own auth logic and code duplication was insane. This PR doesn't look good because:

* In most cases the OCCM and other repo components share the same credentials. Implementing a single change only in the Cinder CSI will cause inconsistencies between CPO components.
* The `GetOpenStackProvider` func is called only once in `cmd/cinder-csi-plugin/main.go` and it's not looped, so the `cacheCloudConfig` global var doesn't make any sense.
* Credentials rotation must be done in a different way and the logic must consider all CPO components.

I'd close this PR and move the discussion on how to achieve a proper creds rotation into an issue.

Makefile
```diff
@@ -166,7 +166,7 @@ build-local-image-%:
 		.
 
 # Build all images locally
-build-local-images: $(addprefix build-image-,$(IMAGE_NAMES))
+build-local-images: $(addprefix build-local-image-,$(IMAGE_NAMES))
```
mdbooth, line 169: I'm guessing you didn't intend to commit this?

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -152,21 +202,72 @@ func InitOpenStackProvider(cfgFiles []string, httpEndpoint string) {
...
+	// without the intervention of the client application.
+	ctx := context.TODO()
+
+	go func() {
+		ticker := time.NewTicker(10 * time.Second)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ctx.Done():
```
mdbooth, line 240: I don't think anything will call this?

mdbooth (commented):
I would personally be ok with something like this in principal. I wonder if we could make it library code and re-use it for all the controllers?

Makefile
spjmurray: I couldn't get a local build to work without this, thought it was necessary, unless I've got to also specify some other bits and bobs.

spjmurray:
Yeah @mdbooth as mentioned, it's just a PoC.  I can see there are definitely efforts to coalesce the OS client code into a canonical blob, so was hoping the problem would go away eventually.  If I get some cycles free I may have a look to see if the CSI code can share what's been done with manilla et al.  CSI seems to be in its own little world at present.

mdbooth:
I'm really hoping @kayrus can provide some guidance here.

spjmurray:
I believe ostensibly we want to conceptualize this as how the typical controller-runtime app works.

* The filewatcher stuff acts as an API watch call
* You can either
** Read the config each time from the cache and create a new provider/client, but as we've discussed that's pretty crap because you need to do a token exchange, authenticate/get the catalog dance with Keystone.  This does have the benefit that _anything_ relying on the config will get updates.
** Cache clients as we do now to prevent the authentication dance, but use a "shared informer" aka the subscription stuff to do whatever updates we can.

kayrus:
I apologies for the delay. I'll try to take a look at this PR this week. Thanks for the patience.

spjmurray:
No rush!  I did have a thought though about `k8s.io/cloud-provider`.  Is that just legacy stuff from when it lived in tree, and thus could be removed?

kayrus (commented):
I noted the most obvious parts of this PR. Let's fix them first, then make a new review round.

kayrus:
`k8s.io/cloud-provider` is a common package for all other cloud providers, primarily golang interfaces.

pkg/util/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,171 @@
...
+
+			case <-ticker.C:
+				contents, hash, err := hashFile(f.path)
+				if err != nil {
+					log.Infof("Failed to read watched file: %v", err)
+					break
+				}
+
+				if hash == f.hash {
+					break
```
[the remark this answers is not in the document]
spjmurray: It'll break the `select`, not the `for` loop.

spjmurray:
Okay @kayrus I think we're starting to zero in on something good... let me know what you think!

spjmurray:
Of note, I've not committed it yet, but I found a solution to getting the OCCM to auto-rotate credentials.  Seems we register a cloud provider, then immediately call `InitCloudProvider` with a `*File`... from there, in the callback, we can type assert from `io.Reader` back to a file and extract the `Name()` and setup the polling file watcher.  But that's a pretty huge change as it affects a lot of the testing jazz, so would probably be better as follow up.  Like this a lot:

```
commit 2346c93af6e44f5de911861a527a05a08c25961c (HEAD -> occm_cloudconfig_port)
Author: Simon Murray <spjmurray@yahoo.co.uk>
Date:   Tue Aug 22 15:00:18 2023 +0100

    Port OCCM to use CloudConfigFactory

 pkg/openstack/openstack.go      | 152 ++++++++++++++++-----------------
 pkg/openstack/openstack_test.go | 295 ++++++++++++++++-------------------------------------------------
 pkg/openstack/routes_test.go    |  19 +++--
 3 files changed, 163 insertions(+), 303 deletions(-)
```

PR #2287, opened 2023-07-03
[occm] Delete unused SG rules with manage-security-groups

### LINKED ISSUE #2003: [occm] SG rules not removed with `manage-security-groups` and Octavia

dulek:
This is iteration of #2033. The PR makes sure unnecessary SG rules are removed when we're
reconciling a Service.

fixes #2003

You need to enable `[LoadBalancer]manage-security-groups` to test this.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -422,6 +422,15 @@ func getSecurityGroupName(service *corev1.Service) string {
 	return securityGroupName
 }
 
+func getSecurityGroupRules(client *gophercloud.ServiceClient, opts rules.ListOpts) ([]rules.SecGroupRule, error) {
+	mc := metrics.NewMetricContext("security_group_rule", "list")
+	page, err := rules.List(client, opts).AllPages()
+	if mc.ObserveRequest(err) != nil {
+		return nil, err
+	}
+	return rules.ExtractRules(page)
+}
+
 func getListenerProtocol(protocol corev1.Protocol, svcConf *serviceConfig) listeners.Protocol {
 	// Make neutron-lbaas code work
 	if svcConf != nil {
@@ -2060,30 +2069,16 @@ func (lbaas *LbaasV2) listSubnetsForNetwork(networkID string, tweak ...TweakSubN
 }
 
 // group, if it not present.
-func (lbaas *LbaasV2) ensureSecurityRule(
-	direction rules.RuleDirection,
-	protocol rules.RuleProtocol,
-	etherType rules.RuleEtherType,
-	remoteIPPrefix, secGroupID string,
-	portRangeMin, portRangeMax int,
-) error {
-	sgRuleCreateOpts := rules.CreateOpts{
-		Direction:      direction,
-		Protocol:       protocol,
-		PortRangeMax:   portRangeMin,
-		PortRangeMin:   portRangeMax,
-		RemoteIPPrefix: remoteIPPrefix,
-		SecGroupID:     secGroupID,
-		EtherType:      etherType,
-	}
-
+func (lbaas *LbaasV2) ensureSecurityRule(sgRuleCreateOpts rules.CreateOpts) error {
 	mc := metrics.NewMetricContext("security_group_rule", "create")
 	_, err := rules.Create(lbaas.network, sgRuleCreateOpts).Extract()
 	if err != nil && cpoerrors.IsConflictError(err) {
 		// Conflict means the SG rule already exists, so ignoring that error.
+		klog.Warningf("Security group rule already found when trying to create it. This indicates concurrent "+
+			"updates to the SG %s and is unexpected", sgRuleCreateOpts.SecGroupID)
 		return mc.ObserveRequest(nil)
 	} else if mc.ObserveRequest(err) != nil {
-		return fmt.Errorf("failed to create rule for security group %s: %v", secGroupID, err)
+		return fmt.Errorf("failed to create rule for security group %s: %v", sgRuleCreateOpts.SecGroupID, err)
 	}
 	return nil
 }
@@ -2174,6 +2169,50 @@ func (lbaas *LbaasV2) UpdateLoadBalancer(ctx context.Context, clusterName string
 	return mc.ObserveReconcile(err)
 }
 
+func compareSecurityGroupRuleAndCreateOpts(rule rules.SecGroupRule, opts rules.CreateOpts) bool {
+	return rule.Direction == string(opts.Direction) &&
+		rule.Protocol == string(opts.Protocol) &&
+		rule.EtherType == string(opts.EtherType) &&
+		rule.RemoteIPPrefix == opts.RemoteIPPrefix &&
+		rule.PortRangeMin == opts.PortRangeMin &&
+		rule.PortRangeMax == opts.PortRangeMax
+}
+
+func getRulesToCreateAndDelete(wantedRules []rules.CreateOpts, existingRules []rules.SecGroupRule) ([]rules.CreateOpts, []rules.SecGroupRule) {
+	toCreate := make([]rules.CreateOpts, 0, len(wantedRules))     // Max is all rules need creation
+	toDelete := make([]rules.SecGroupRule, 0, len(existingRules)) // Max will be all the existing rules to be deleted
+	// Surely this can be done in a more efficient way. Is it worth optimizing if most of
+	// the time we'll deal with just 1 or 2 elements in each array? I doubt it.
+	for _, existingRule := range existingRules {
+		found := false
+		for _, wantedRule := range wantedRules {
+			if compareSecurityGroupRuleAndCreateOpts(existingRule, wantedRule) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			// in existingRules but not in wantedRules, delete
+			toDelete = append(toDelete, existingRule)
+		}
+	}
+	for _, wantedRule := range wantedRules {
+		found := false
+		for _, existingRule := range existingRules {
+			if compareSecurityGroupRuleAndCreateOpts(existingRule, wantedRule) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			// in wantedRules but not in exisitngRules, create
+			toCreate = append(toCreate, wantedRule)
+		}
+	}
+
+	return toCreate, toDelete
+}
+
 // ensureAndUpdateOctaviaSecurityGroup handles the creation and update of the security group and the securiry rules for the octavia load balancer
 func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, apiService *corev1.Service, nodes []*corev1.Node, memberSubnetID string) error {
 	// get service ports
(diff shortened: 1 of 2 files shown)


pkg/openstack/loadbalancer.go
```diff
@@ -2078,14 +2101,21 @@ func (lbaas *LbaasV2) ensureSecurityRule(
 	}
 
 	mc := metrics.NewMetricContext("security_group_rule", "create")
-	_, err := rules.Create(lbaas.network, sgRuleCreateOpts).Extract()
+	sg, err := rules.Create(lbaas.network, sgRuleCreateOpts).Extract()
 	if err != nil && cpoerrors.IsConflictError(err) {
 		// Conflict means the SG rule already exists, so ignoring that error.
-		return mc.ObserveRequest(nil)
+		// We can get the rule ID from the message.
+		re := regexp.MustCompile("Rule id is ([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12})")
```
mdbooth, line 2108: Please put MustCompile() in a static context. If it's going to panic it should do it when the module is loaded. Also, parsing human readable error messages is nasty.

This looks like an effort to do atomic creation, which I'm normally very much in favour of, but I'm not sure it's serving us here. I think this code would be a lot cleaner if we fetched all rules first and did creation/deletion based on that. The controller should already prevent the race of simultaneous reconciliation of a single object, but even if it didn't we should be safe if we always fail on a 409 here.

pkg/openstack/loadbalancer.go
```diff
@@ -2256,11 +2291,31 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
 				"failed to apply security rule for port %d, %w",
 				port.NodePort, err)
 		}
+		wantedRuleIDs = append(wantedRuleIDs, id)
+	}
 
-		if err := applyNodeSecurityGroupIDForLB(lbaas.compute, lbaas.network, nodes, lbSecGroupID); err != nil {
-			return err
+	// delete unneeded rules
+	existingRules, err := getSecurityGroupRules(lbaas.network, rules.ListOpts{SecGroupID: lbSecGroupID})
```
mdbooth, line 2298: We could do this first and avoid the fragile error message parsing.

pkg/openstack/loadbalancer.go
```diff
@@ -2256,11 +2291,31 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
...
+	}
+	for _, existingRule := range existingRules {
+		needed := false
+		for _, wantedRuleID := range wantedRuleIDs {
+			if wantedRuleID == existingRule.ID {
+				needed = true
+				break
+			}
+		}
+		if !needed {
```
jichenjc, line 2311: can we at least log.v5 or v6 to record this deletion ? sometimes this will be helpful as well for debug..

pkg/openstack/loadbalancer.go
```diff
@@ -2256,11 +2291,31 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
...
+			rules.Delete(lbaas.network, existingRule.ID)
```
jichenjc, line 2312: will this API call 100% success or worthy a err check?

dulek: It definitely is!

dulek:
Making this a draft until I test it tomorrow morning, but all the comments should be addressed here.

pkg/openstack/loadbalancer.go
```diff
@@ -2220,47 +2236,98 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
 		etherType = rules.EtherType6
 	}
 
+	existingRules, err := getSecurityGroupRules(lbaas.network, rules.ListOpts{SecGroupID: lbSecGroupID})
+	if err != nil && !cpoerrors.IsNotFound(err) {
```
dulek, line 2240: I copied that IsNotFound clause here, but I'm fairly sure it's a past mistake here.

pkg/openstack/loadbalancer.go
```diff
@@ -2220,47 +2236,98 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
...
+			return fmt.Errorf("failed to apply security rule (%v), %w", opts, err)
 		}
+	}
 
-		if err := applyNodeSecurityGroupIDForLB(lbaas.compute, lbaas.network, nodes, lbSecGroupID); err != nil {
-			return err
+	// delete unneeded rules
+	for _, existingRule := range toDelete {
+		klog.Infof("Deleting rule %s from security group %s (%s)", existingRule.ID, existingRule.SecGroupID, lbSecGroupName)
+		err := rules.Delete(lbaas.network, existingRule.ID).ExtractErr()
```
dulek, line 2322: Should I add a metric to this?

dulek: I should, we do that elsewhere.

jichenjc: I guess so ..

dulek:
Alright, this is tested and ready for reviews now!

pkg/openstack/loadbalancer.go
```diff
@@ -2220,47 +2222,102 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
...
+				PortRangeMin:   int(port.NodePort),
+				PortRangeMax:   int(port.NodePort),
+			},
 		)
+	}
+
+	toCreate := make([]rules.CreateOpts, 0, len(ports)+1)         // Max is number of ports+1
+	toDelete := make([]rules.SecGroupRule, 0, len(existingRules)) // Max will be all the existing rules
+	// Surely this can be done in a more efficient way. Is it worth optimizing if most of
+	// the time we'll deal with just 1 or 2 elements in each array? I doubt it.
```
mdbooth, line 2269: 👍

pkg/openstack/loadbalancer.go
```diff
@@ -2220,47 +2222,102 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
...
+	for _, rule := range existingRules {
+		found := false
+		for _, opts := range wantedRules {
+			if compareSecurityGroupRuleAndCreateOpts(rule, opts) {
```
mdbooth, line 2273: nit: Suggest `existing` and `wanted` as variable names. These are a bit foobar. `opts` in particular made me think we were doing an OpenStack API call here at first glance.

pkg/openstack/loadbalancer.go
```diff
@@ -2220,47 +2222,102 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
...
-		if err := applyNodeSecurityGroupIDForLB(lbaas.compute, lbaas.network, nodes, lbSecGroupID); err != nil {
-			return err
+	// delete unneeded rules
+	for _, existingRule := range toDelete {
+		klog.Infof("Deleting rule %s from security group %s (%s)", existingRule.ID, existingRule.SecGroupID, lbSecGroupName)
+		mc := metrics.NewMetricContext("security_group_rule", "delete")
+		err := rules.Delete(lbaas.network, existingRule.ID).ExtractErr()
+		if err != nil && cpoerrors.IsNotFound(err) {
+			// ignore 404
+			return mc.ObserveRequest(nil)
```
mdbooth, line 2312: For this and the 419 above, actually hitting this should only be possible if we're making concurrent modifications elsewhere. I think we should log them both at a minimum, and possibly also fail. If we fail we'll just retry, presumably without conflicts.

mdbooth (commented):
Is there any way we could unit test the code which is just doing comparison of previously fetched values? It looks tricky enough to merit it.

pkg/openstack/loadbalancer.go
dulek: 419?

dulek: Oh, 409. Okay, makes sense. I'll stick with a warning. If rule exists already or is gone already, we're fine.

pkg/openstack/loadbalancer.go
```diff
@@ -2220,47 +2224,104 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
...
+				Protocol:       rules.ProtocolTCP,
+				EtherType:      etherType,
+				RemoteIPPrefix: subnet.CIDR,
+				SecGroupID:     lbSecGroupID,
+				PortRangeMin:   int(port.NodePort),
+				PortRangeMax:   int(port.NodePort),
+			},
 		)
+	}
+
```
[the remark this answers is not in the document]
dulek: Alright, done!

pkg/openstack/loadbalancer.go
```diff
@@ -2220,47 +2224,104 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
...
+			// ignore 404
+			klog.Warningf("Security group rule %s found missing when trying to delete it. This indicates concurrent "+
+				"updates to the SG %s and is unexpected", existingRule.ID, existingRule.SecGroupID)
+			return mc.ObserveRequest(nil)
+		} else if mc.ObserveRequest(err) != nil {
+			return fmt.Errorf("failed to delete security group rule %s: %w", existingRule.ID, err)
 		}
 	}
+
+	if err := applyNodeSecurityGroupIDForLB(lbaas.compute, lbaas.network, nodes, lbSecGroupID); err != nil {
```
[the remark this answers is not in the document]
dulek: True. This isn't the only problem with that code. Adding an SG is not an atomic operation in Neutron, so we should definitely make sure we're protecting ourselves from lost-update issues there. I have it on my TODO list.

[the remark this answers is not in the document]
dulek: https://github.com/kubernetes/cloud-provider-openstack/pull/2293

pkg/openstack/loadbalancer_test.go
```diff
@@ -133,3 +134,299 @@ func TestPopListener(t *testing.T) {
...
+type testGetRulesToCreateAndDelete struct {
+	wantedRules   []rules.CreateOpts
+	existingRules []rules.SecGroupRule
+	toCreate      []rules.CreateOpts
+	toDelete      []rules.SecGroupRule
+}
+
+func TestGetRulesToCreateAndDelete(t *testing.T) {
+	items := []testGetRulesToCreateAndDelete{
+		// Empty elements
```
mdbooth, line 147: nit: It's a common idiom to have the test description as part of the test. As an example, take a look at this randomly selected test in CAPO: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/9d183bd479fe9aed4f6e7ac3d5eee46681c518e7/pkg/cloud/services/compute/instance_test.go#L120-L121

Notice how the test name is part of the test data. Then when we run it we do:

```golang
	for _, tt := range tests {
		t.Run(tt.testName, func(t *testing.T) {
                  ...test function...
                }
        }
```

That way each invocation is a separate test with its own name, which is easier to read when it fails.

pkg/openstack/loadbalancer.go
```diff
@@ -2220,47 +2259,75 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
...
-			rules.RuleProtocol(port.Protocol),
-			etherType,
-			subnet.CIDR,
-			lbSecGroupID,
-			int(port.NodePort),
-			int(port.NodePort),
+		wantedRules = append(wantedRules,
+			rules.CreateOpts{
+				Direction:      rules.DirIngress,
+				Protocol:       rules.ProtocolTCP,
```
mdbooth, line 2293: It looks like this used to be `rules.RuleProtocol(port.Protocol)`. Why the change?

dulek: Ha, you're right, it's a mistake. Fixed now.

pkg/openstack/loadbalancer_test.go
dulek: Done!

mdbooth (commented):
/lgtm

PR #2291, opened 2023-07-07
[occm] Make manage-security-groups work with OVN

dulek:
OVN keeps the source IP of the incoming traffic. This means that the SG
on the members of the LB (nodes) needs to have a rule opening traffic
not only from inside the cluster, but also elsewhere. This can
restricted by using the `Service.Spec.LoadBalancerSourceRanges` and this
commit also makes sure the setting is taken into account.

fixes #

**Release note**:

```release-note
[openstack-cloud-controller-manager] Implemented support for `loadBalancerSourceRanges` for OVN Octavia Provider (with manage-security-groups enabled).
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1412,8 +1412,10 @@ func (lbaas *LbaasV2) buildListenerCreateOpt(port corev1.ServicePort, svcConf *s
 		listenerCreateOpt.Protocol = listeners.ProtocolHTTP
 	}
 
-	if len(svcConf.allowedCIDR) > 0 {
-		listenerCreateOpt.AllowedCIDRs = svcConf.allowedCIDR
+	if openstackutil.IsOctaviaFeatureSupported(lbaas.lb, openstackutil.OctaviaFeatureVIPACL, lbaas.opts.LBProvider) {
+		if len(svcConf.allowedCIDR) > 0 {
+			listenerCreateOpt.AllowedCIDRs = svcConf.allowedCIDR
+		}
 	}
 	return listenerCreateOpt
 }
@@ -1770,18 +1772,19 @@ func (lbaas *LbaasV2) checkService(service *corev1.Service, nodes []*corev1.Node
 		svcConf.timeoutTCPInspect = getIntFromServiceAnnotation(service, ServiceAnnotationLoadBalancerTimeoutTCPInspect, 0)
 	}
 
-	var listenerAllowedCIDRs []string
 	sourceRanges, err := GetLoadBalancerSourceRanges(service, svcConf.preferredIPFamily)
 	if err != nil {
 		return fmt.Errorf("failed to get source ranges for loadbalancer service %s: %v", serviceName, err)
 	}
 	if openstackutil.IsOctaviaFeatureSupported(lbaas.lb, openstackutil.OctaviaFeatureVIPACL, lbaas.opts.LBProvider) {
 		klog.V(4).Info("LoadBalancerSourceRanges is suppported")
-		listenerAllowedCIDRs = sourceRanges.StringSlice()
+		svcConf.allowedCIDR = sourceRanges.StringSlice()
+	} else if lbaas.opts.LBProvider == "ovn" && lbaas.opts.ManageSecurityGroups {
+		klog.V(4).Info("LoadBalancerSourceRanges will be enforced on the SG created and attached to LB members")
+		svcConf.allowedCIDR = sourceRanges.StringSlice()
 	} else {
-		klog.Warning("LoadBalancerSourceRanges is ignored")
+		klog.Warning("LoadBalancerSourceRanges are ignored")
 	}
-	svcConf.allowedCIDR = listenerAllowedCIDRs
 
 	if openstackutil.IsOctaviaFeatureSupported(lbaas.lb, openstackutil.OctaviaFeatureFlavors, lbaas.opts.LBProvider) {
 		svcConf.flavorID = getStringFromServiceAnnotation(service, ServiceAnnotationLoadBalancerFlavorID, lbaas.opts.FlavorID)
@@ -2028,7 +2031,7 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 	status := lbaas.createLoadBalancerStatus(service, svcConf, addr)
 
 	if lbaas.opts.ManageSecurityGroups {
-		err := lbaas.ensureSecurityGroup(clusterName, service, nodes, loadbalancer, svcConf.preferredIPFamily, svcConf.lbMemberSubnetID)
+		err := lbaas.ensureAndUpdateOctaviaSecurityGroup(clusterName, service, nodes, svcConf)
 		if err != nil {
 			return status, fmt.Errorf("failed when reconciling security groups for LB service %v/%v: %v", service.Namespace, service.Name, err)
 		}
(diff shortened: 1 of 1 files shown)


dulek:
This is based on #2287, I'll rebase it once it merges.

mdbooth:
Is this behaviour dependent on `externalTrafficPolicy`?

dulek:
> Is this behaviour dependent on `externalTrafficPolicy`?

Not really. Motivations for this PR:
* OVN needs different SGs as it keeps original source IP of the traffic.
* OVN does not implement support for `allowed_cidrs` property on the LB listeners. Instead we're supposed to set the SGs on the LB members to limit traffic. This commit makes sure we're using `loadBalancerSourceRanges` to set the correct CIDRs on these SG rules.

pkg/openstack/loadbalancer.go
```diff
@@ -2085,10 +2088,8 @@ func (lbaas *LbaasV2) ensureSecurityRule(sgRuleCreateOpts rules.CreateOpts) erro
 
 // ensureSecurityGroup ensures security group exist for specific loadbalancer service.
 // Creating security group for specific loadbalancer service when it does not exist.
-func (lbaas *LbaasV2) ensureSecurityGroup(clusterName string, apiService *corev1.Service, nodes []*corev1.Node,
-	loadbalancer *loadbalancers.LoadBalancer, preferredIPFamily corev1.IPFamily, memberSubnetID string) error {
-
-	return lbaas.ensureAndUpdateOctaviaSecurityGroup(clusterName, apiService, nodes, memberSubnetID)
+func (lbaas *LbaasV2) ensureSecurityGroup(clusterName string, apiService *corev1.Service, nodes []*corev1.Node, svcConf *serviceConfig) error {
```
mdbooth, line 2091: This isn't exported and it's just an alias for `ensureAndUpdateOctaviaSecurityGroup`. Can we delete it?

pkg/openstack/loadbalancer.go
```diff
@@ -2332,8 +2344,8 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
 }
 
 // updateSecurityGroup updating security group for specific loadbalancer service.
-func (lbaas *LbaasV2) updateSecurityGroup(clusterName string, apiService *corev1.Service, nodes []*corev1.Node, memberSubnetID string) error {
-	return lbaas.ensureAndUpdateOctaviaSecurityGroup(clusterName, apiService, nodes, memberSubnetID)
+func (lbaas *LbaasV2) updateSecurityGroup(clusterName string, apiService *corev1.Service, nodes []*corev1.Node, svcConf *serviceConfig) error {
```
mdbooth, line 2347: Ditto.

mdbooth (commented):
Doesn't this give unprivileged cluster users the ability to open the firewall to arbitrary ports on worker nodes? This doesn't sound like a good idea.

pkg/openstack/loadbalancer.go
```diff
@@ -2248,16 +2249,23 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
...
 
 	etherType := rules.EtherType4
 	if netutils.IsIPv6CIDRString(subnet.CIDR) {
 		etherType = rules.EtherType6
 	}
+	cidrs := []string{subnet.CIDR}
+	if lbaas.opts.LBProvider == "ovn" {
+		// OVN keeps the source IP of the incoming traffic. This means that we cannot just open the LB range, but we
+		// need to open for the whole world. This can be restricted by using the service.spec.loadBalancerSourceRanges.
+		// In general svcConf.allowedCIDR will give us this already calculated by GetLoadBalancerSourceRanges().
```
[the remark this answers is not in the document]
dulek: Sure, thanks for a careful review!

pkg/openstack/loadbalancer.go
dulek: Done!

dulek:
> Doesn't this give unprivileged cluster users the ability to open the firewall to arbitrary ports on worker nodes? This doesn't sound like a good idea.

Hm, good question… So this particular change does not really affect ports being open, the `[LoadBalancer]manage-security-groups` set to `true` does.

This change only makes sure OVN LBs will open what's required for them to work (unfortunately it's `0.0.0.0/0` as we don't know source IP) and use `LoadBalancerSourceRanges` to limit that range if specified. This doesn't affect how SG rules are created in Amphora use case. And without this PR OVN LBs would only be accessible from inside the cluster when `manage-security-groups` is enabled.

Now in regards to ports - without `manage-security-groups=true` you need to open the whole NodePorts range (30000-32767 by default, just below ephemeral ports range) on the nodes anyway to make your LoadBalancer Services working. Example in OpenShift [^1].

The idea of `manage-security-groups` is to only open these ports selectively when a Service is created with such a NodePort. While user can request a certain NodePort number when creating a Service instead of letting K8s assign it at random, it still has to be from the NodePort range:

```
# services "test2" was not valid:
# * spec.ports[0].nodePort: Invalid value: 22: provided port is not in the valid range. The range of valid ports is 30000-32767
```

So to me both this patch and `manage-security-groups=true` doesn't make the situation worse, but better.

[^1]: https://github.com/openshift/installer/blob/master/data/data/openstack/masters/sg-master.tf#L235-L257

pkg/openstack/loadbalancer.go
```diff
@@ -2248,16 +2243,23 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
...
-			"failed to find subnet %s from openstack: %v", memberSubnetID, err)
+			"failed to find subnet %s from openstack: %v", svcConf.lbMemberSubnetID, err)
 	}
 
 	etherType := rules.EtherType4
 	if netutils.IsIPv6CIDRString(subnet.CIDR) {
 		etherType = rules.EtherType6
 	}
+	cidrs := []string{subnet.CIDR}
+	if lbaas.opts.LBProvider == "ovn" {
```
mdbooth, line 2257: Do I vaguely recall a previous discussion about providers having arbitrary names? I don't remember who brought it up: either you or @kayrus I'm guessing?

dulek: Yes. Later on I consulted Octavia team to learn I was wrong:

> the providers should be explicitly listed in the octavia config file ([http://rhos-ci-logs.lab.eng.tlv2.redhat.com/logs/rcj/DFG-network-octavia-17.1_director-rhel-vi[…]ppet-generated/octavia/etc/octavia/octavia.conf.gz](http://rhos-ci-logs.lab.eng.tlv2.redhat.com/logs/rcj/DFG-network-octavia-17.1_director-rhel-virthost-3cont_2comp_1ipa-ipv4-geneve-tls/34/controller-0/var/lib/config-data/puppet-generated/octavia/etc/octavia/octavia.conf.gz))
> ```
> # Dictionary of enabled provider driver names and descriptions
> # A comma separated list of dictionaries of the enabled provider driver names
> # and descriptions.
> # enabled_provider_drivers = amphora:The Octavia Amphora driver.,octavia: \
> enabled_provider_drivers=amphora: The Octavia Amphora driver.,octavia: Deprecated alias of the Octavia Amphora driver.,ovn: Octavia OVN driver.
> #                            Deprecated alias of the Octavia Amphora driver.
> ```
> then ovn-provider defines an "official" name for its plugin: https://opendev.org/openstack/ovn-octavia-provider/src/branch/master/setup.cfg#L30-L31

jichenjc:
code good , but we are adding more and more features into the repo
I think we should think about enable such in the CI @dulek @mdbooth , e.g OVN or manage-security-groups=true ?

dulek:
> code good , but we are adding more and more features into the repo I think we should think about enable such in the CI @dulek @mdbooth , e.g OVN or manage-security-groups=true ?

Yeah, that makes perfect sense. I'll see what I can do to extend the testing.

PR #2293, opened 2023-07-12
[occm] Optimize `applyNodeSecurityGroupIDForLB()`

dulek:
Matt noticed several optimizations that can be done in `applyNodeSecurityGroupIDForLB()` function. I added a few more refactorings. In particular:

* We don't need to lookup servers by name as we have `.spec.providerID` in the nodes passed to the function.
* We don't need to lookup servers at all when we have the ID, we can fetch ports by `deviceID`.
* This means we don't need a Nova client in the LB instance anymore.
* `slices.Contains()` from K8s utils can be used to check if SG is in the list.
* It's better to tag the port with SG before we actually apply it. This will protect us from situations when the SG is applied, but tagging the port failed and now when deleting the SG CPO would be unable to find the port and therefore delete the SG.

`[LoadBalancer]manage-security-groups` has to be on.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -40,13 +40,13 @@ import (
 	secgroups "github.com/gophercloud/utils/openstack/networking/v2/extensions/security/groups"
 	"gopkg.in/godo.v2/glob"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes"
 	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/klog/v2"
 	netutils "k8s.io/utils/net"
+	"k8s.io/utils/strings/slices"
 
 	"k8s.io/cloud-provider-openstack/pkg/metrics"
 	cpoutil "k8s.io/cloud-provider-openstack/pkg/util"
@@ -721,15 +721,13 @@ func getSubnetIDForLB(network *gophercloud.ServiceClient, node corev1.Node, pref
 }
 
 // applyNodeSecurityGroupIDForLB associates the security group with all the ports on the nodes.
-func applyNodeSecurityGroupIDForLB(compute *gophercloud.ServiceClient, network *gophercloud.ServiceClient, nodes []*corev1.Node, sg string) error {
+func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, nodes []*corev1.Node, sg string) error {
 	for _, node := range nodes {
-		nodeName := types.NodeName(node.Name)
-		srv, err := getServerByName(compute, nodeName)
+		serverID, _, err := instanceIDFromProviderID(node.Spec.ProviderID)
 		if err != nil {
-			return err
+			return fmt.Errorf("error getting server ID from the node: %w", err)
 		}
-
-		listOpts := neutronports.ListOpts{DeviceID: srv.ID}
+		listOpts := neutronports.ListOpts{DeviceID: serverID}
 		allPorts, err := openstackutil.GetPorts(network, listOpts)
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


pkg/openstack/loadbalancer.go
```diff
@@ -712,48 +712,40 @@ func getSubnetIDForLB(network *gophercloud.ServiceClient, node corev1.Node, pref
...
+func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, nodes []*corev1.Node, sg string) error {
 	for _, node := range nodes {
-		nodeName := types.NodeName(node.Name)
-		srv, err := getServerByName(compute, nodeName)
-		if err != nil {
-			return err
-		}
-
-		listOpts := neutronports.ListOpts{DeviceID: srv.ID}
+		listOpts := neutronports.ListOpts{DeviceID: node.Spec.ProviderID}
```
MaysaMacedo, line 717: shouldn't we take into account that the ProviderID can have the following format `openstack:///a0e45d62-af9d-4453-a569-fe7094204ec6`  and extract the UUID from that?

jichenjc: +1 to the question

I think it's `openstack:///` for node.Spec.ProviderID

dulek: Whoops, yes, I fixed it locally but uploaded the wrong version. Thanks for a careful review!

jichenjc: maybe we need think about how to add those securty group test into CI so we can catch 
but that's a different topic :)

dulek: Being able to mock Gophercloud in the tests would be really cool. I'm not sure if we should wait for some Gophercloud utility or just apply CAPO's approach.

mdbooth: Aside: I would also very much like to be able to unit test code which makes OpenStack calls in CPO. The CAPO approach is one possibility, but I've also started looking into ways we could do this in Gophercloud, up to and including a full simulator. This would not be a small amount of work, though.

mdbooth (commented):
Looks good!

pkg/openstack/loadbalancer.go
```diff
@@ -712,48 +712,44 @@ func getSubnetIDForLB(network *gophercloud.ServiceClient, node corev1.Node, pref
...
-			// slices.Contains.
-			if func() bool {
-				for _, currentSG := range port.SecurityGroups {
-					if currentSG == sg {
-						return true
-					}
-				}
-				return false
-			}() {
+			if slices.Contains(port.SecurityGroups, sg) {
```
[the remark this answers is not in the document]
dulek: It's a `k8s.io/utils/strings/slices` `slices.Contains()`. It's based on strings. The generic `slices` module seems to still be experimental: https://pkg.go.dev/golang.org/x/exp/slices#Contains

pkg/openstack/loadbalancer.go
```diff
@@ -712,48 +712,44 @@ func getSubnetIDForLB(network *gophercloud.ServiceClient, node corev1.Node, pref
...
+			mc := metrics.NewMetricContext("port_tag", "add")
+			err := neutrontags.Add(network, "ports", port.ID, sg).ExtractErr()
+			if mc.ObserveRequest(err) != nil {
+				return fmt.Errorf("failed to add tag %s to port %s: %v", sg, port.ID, err)
+			}
+
+			// Add the SG to the port
+			// TODO(dulek): This isn't an atomic operation. In order to protect from lost update issues we should use
+			//              `revision_number` handling to make sure our update to `security_groups` field wasn't preceded
+			//              by a different one. Same applies to a removal of the SG.
```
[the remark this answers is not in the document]
dulek: https://github.com/kubernetes/cloud-provider-openstack/blob/78a3fa6089c28d48fffea9a29334407c76785cc4/pkg/openstack/loadbalancer.go#L755

This says that "latest state" is based on current state. Imagine 2 actors `A` and `B` doing this in parallel (not even 2 CPO threads, but e.g. CPO and user). If stuff is timed badly:

1. `A` reads SG list with "foo,bar".
2. `B` reads SG list with "foo,bar".
3. `A` updates SG list with "baz" to be "foo,bar,baz".
4. `B` updates SG list with "bazinga" to be "foo,bar,bazinga"
5. Update done by `A` in point 3 is lost now.

pkg/openstack/loadbalancer.go
```diff
@@ -712,48 +712,44 @@ func getSubnetIDForLB(network *gophercloud.ServiceClient, node corev1.Node, pref
 }
 
 // applyNodeSecurityGroupIDForLB associates the security group with all the ports on the nodes.
-func applyNodeSecurityGroupIDForLB(compute *gophercloud.ServiceClient, network *gophercloud.ServiceClient, nodes []*corev1.Node, sg string) error {
+func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, nodes []*corev1.Node, sg string) error {
 	for _, node := range nodes {
-		nodeName := types.NodeName(node.Name)
-		srv, err := getServerByName(compute, nodeName)
+		serverID, _, err := instanceIDFromProviderID(node.Spec.ProviderID)
```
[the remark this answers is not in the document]
dulek: We already do the same to get the subnet to create the LB in [^1]. This happens even before the SG is touched: https://github.com/kubernetes/cloud-provider-openstack/blob/78a3fa6089c28d48fffea9a29334407c76785cc4/pkg/openstack/loadbalancer.go#L702-L705

[^1]: Also that code is not using the function designed to get the server ID from the string that and I'm fixing it in https://github.com/kubernetes/cloud-provider-openstack/pull/2302

pkg/openstack/loadbalancer.go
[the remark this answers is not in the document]
mdbooth: Does neutron provide a revision number that can be used in this way?

mdbooth (commented):
/lgtm

pkg/openstack/loadbalancer.go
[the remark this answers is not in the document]
dulek: Yes [^1]. I wouldn't mention it if it wasn't a thing. ;) We're using it successfully in Kuryr and semi-successfully in OpenShift's EgressIP controller. The latter needs additional lock in place due to Neutron bugs and I'll probably copy that approach here [^2]. And yes, we could create some helpers in Gophercloud.

[^1]: https://docs.openstack.org/api-ref/network/v2/index.html#revisions
[^2]: https://github.com/openshift/cloud-network-config-controller/blob/8376880c790739d3fb62e7faeb94e902b0a83488/pkg/cloudprovider/openstack.go#L758-L809

PR #2302, opened 2023-07-17
[occm] Use `instanceIDFromProviderID()` function

dulek:
This commit makes sure that in all instances of reading Nova instance ID from the `.Spec.ProviderID` field, we're using
`instanceIDFromProviderID()` function which is designed for that purpose.

fixes #

**Release note**:

```release-note
NONE
```


mdbooth (commented):
/lgtm

PR #2304, opened 2023-07-17
[occm] Remove filtering by device_owner.

### LINKED ISSUE #2303: [occm] Cannot expose deployment while on different Availability Zone

>
/kind bug
> /kind feature

**What happened**:
When using different availability zone than default (`nova`), there is no way to expose deployment via service.

The service will be created

**How to reproduce it**:
Deploy k8s nodes on different AZ, create deployment, expose using service. Service will hang on pending state with an error:
```
Error syncing load balancer: failed to ensure load balancer: failed to get subnet to create load balancer for service temp/foo: failed to find object
```

**Environment**:
- OpenStack version: RH OSP 16.2
- Others:

gryf:
In OpenStack default AZ is called 'nova', thus device_owner is filled with 'compute:nova', and that's ok for deployments within one (default) AZ.

Filtering by device_owner however, will make function getAttachedPorts to return zero ports for provided node. In this patch, filtering by device_owner is simply removed.

fixes #2303

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/instances.go
+++ b/pkg/openstack/instances.go
@@ -703,8 +703,7 @@ func getAddressesByName(client *gophercloud.ServiceClient, name types.NodeName, 
 // getAttachedPorts returns a list of ports attached to a server.
 func getAttachedPorts(client *gophercloud.ServiceClient, serverID string) ([]ports.Port, error) {
 	listOpts := ports.ListOpts{
-		DeviceID:    serverID,
-		DeviceOwner: "compute:nova",
+		DeviceID: serverID,
 	}
 
 	return openstack.GetPorts(client, listOpts)


zetaab:
interesting, we do have multiple AZs (any of them are not called `nova`) and we are not facing issues

zetaab (approved):
Anyways the query against openstack api should be still quite fast. I do not see reason to not remove this.

/approve

@dulek opinions?

gryf:
@zetaab this issue manifests itself after transitions from nova os-interfaces to neutron calls - if you on 1.27, you'll not see this.

dulek:
> Anyways the query against openstack api should be still quite fast. I do not see reason to not remove this.

It was caused by kubernetes/cloud-provider-openstack#2250, we've just discovered this today when trying master on a cloud that has AZs.

And queries to OpenStack APIs are expensive when you compare them to K8s API calls. We've seen a lot of issues with Neutron API scalability in the past, so to me anything limiting the number of API calls is worth it.

dulek:
@mdbooth FYI

PR #2305, opened 2023-07-19
[occm] Delete sgs on reconfiguration

dulek:
If we reconfigure CPO to set `manage-security-groups=true`, it will
correctly reconcile LBs to make sure SGs are created and associated with the
nodes. This commit makes sure that SGs are properly deleted when CPO is
reconfigured in the opposite direction, with `manage-security-groups=false`.

Also a small refactoring of the SG deletion code is added in a separate commit.

fixes #

`[LoadBalancer]manage-security-groups` option needs to be manipulated.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -805,21 +805,6 @@ func disassociateSecurityGroupForLB(network *gophercloud.ServiceClient, sg strin
 	return nil
 }
 
-// isSecurityGroupNotFound return true while 'err' is object of gophercloud.ErrResourceNotFound
-func isSecurityGroupNotFound(err error) bool {
-	errType := reflect.TypeOf(err).String()
-	errTypeSlice := strings.Split(errType, ".")
-	errTypeValue := ""
-	if len(errTypeSlice) != 0 {
-		errTypeValue = errTypeSlice[len(errTypeSlice)-1]
-	}
-	if errTypeValue == "ErrResourceNotFound" {
-		return true
-	}
-
-	return false
-}
-
 // deleteListeners deletes listeners and its default pool.
 func (lbaas *LbaasV2) deleteListeners(lbID string, listenerList []listeners.Listener) error {
 	for _, listener := range listenerList {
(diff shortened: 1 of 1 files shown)


pkg/openstack/loadbalancer.go
```diff
@@ -2542,7 +2532,7 @@ func (lbaas *LbaasV2) EnsureSecurityGroupDeleted(_ string, service *corev1.Servi
 	lbSecGroupName := getSecurityGroupName(service)
 	lbSecGroupID, err := secgroups.IDFromName(lbaas.network, lbSecGroupName)
 	if err != nil {
-		if isSecurityGroupNotFound(err) {
+		if cpoerrors.IsNotFound(err) {
```
mdbooth, line 2535: Random thought for @pierreprinetti @EmilienM @kayrus: I wonder if functions like this should be in gophercloud utils. We have basically the same in CAPO, which seems wasteful.

mdbooth (commented):
/lgtm

pkg/openstack/loadbalancer.go
EmilienM: Yes I think https://github.com/dulek/cloud-provider-openstack/blob/48a75a320f8ef9c9fe5d8c5bcd325ee47601ccb6/pkg/util/errors/errors.go#L45-L65 could definitely go in gophercloud/utils and would be welcomed.

mdbooth: @EmilienM https://github.com/gophercloud/utils/pull/193

dulek:
/hold cancel

Okay, glad I've tested, it's `EnsureLoadBalancer()` that's called on startup, not `UpdateLoadBalancer()`.

mdbooth:
> /hold cancel
> 
> Okay, glad I've tested, it's `EnsureLoadBalancer()` that's called on startup, not `UpdateLoadBalancer()`.

Good catch! TBH it's a bit smelly that these are separate code paths. I guess it's safe to only have it in the path which is called on controller initialisation, because the value can't be changed without a controller restart.

/lgtm

dulek:
> > /hold cancel
> > Okay, glad I've tested, it's `EnsureLoadBalancer()` that's called on startup, not `UpdateLoadBalancer()`.
> 
> Good catch! TBH it's a bit smelly that these are separate code paths. I guess it's safe to only have it in the path which is called on controller initialisation, because the value can't be changed without a controller restart.
> 
> /lgtm

Yeah, true, it's smelly. `UpdateLoadBalancer()` is called independently from the `nodeSyncLoop()` and we do need to run the `ensureSecurityGroup()` there to make sure new nodes get the SG applied. We shouldn't need to delete the SG there though, I think I'm going to remove that part there.

We probably should try to detach the SG from the nodes that are removed from the LB too, but that's not trivial as we don't get the list of Nodes being removed from the LB. Currently this is worked around by the fact that when we're removing the SG, we have its ID tagged on the ports, so we can find them easily.

Maybe we can remove them by listing them by ID on the `applyNodeSecurityGroupIDForLB()`? I'll give it a bit of thinking. It doesn't affect this patch, I think.

PR #2306, opened 2023-07-24
[occm] Get IP addresses of neutron subports v3

jingczhang:
Those VMs are plugged into flat networks and use neutron trunk:

Currenlty, nodeAddresses() only returns neutron ports directly attached to VM. For vlan-aware VMs, IP addresses are assigned on neutron subports. Subports are attached to the trunk; they are not attached to the VM directly. This pull request changes nodeAddresses() to return IP addresses of neutron subports when they exist. Without this change, Kubernetes is unable to IP addresses of vlan-aware VMs. This change is transparent to VMs not using neturon trunk.

This is a rebase of the following PR:
https://github.com/kubernetes/cloud-provider-openstack/pull/2270

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/instances.go
+++ b/pkg/openstack/instances.go
@@ -29,7 +29,8 @@ import (
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
-	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
+	neutronports "github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
 	"github.com/gophercloud/gophercloud/pagination"
 	"github.com/mitchellh/mapstructure"
 	v1 "k8s.io/api/core/v1"
@@ -43,7 +44,6 @@ import (
 	"k8s.io/cloud-provider-openstack/pkg/util"
 	"k8s.io/cloud-provider-openstack/pkg/util/errors"
 	"k8s.io/cloud-provider-openstack/pkg/util/metadata"
-	"k8s.io/cloud-provider-openstack/pkg/util/openstack"
 )
 
 // Instances encapsulates an implementation of Instances for OpenStack.
@@ -240,7 +240,7 @@ func (i *Instances) NodeAddressesByProviderID(ctx context.Context, providerID st
 		return []v1.NodeAddress{}, err
 	}
 
-	addresses, err := nodeAddresses(server, ports, i.networkingOpts)
+	addresses, err := nodeAddresses(server, ports, i.network, i.networkingOpts)
 	if err != nil {
 		return []v1.NodeAddress{}, err
 	}
@@ -345,7 +345,7 @@ func (i *Instances) InstanceMetadata(ctx context.Context, node *v1.Node) (*cloud
 	if err != nil {
 		return nil, err
 	}
-	addresses, err := nodeAddresses(srv, ports, i.networkingOpts)
+	addresses, err := nodeAddresses(srv, ports, i.network, i.networkingOpts)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


jingczhang:
Rebase caused the creation of this new PR.
I will update the PR for last comment from https://github.com/kubernetes/cloud-provider-openstack/pull/2270.

dulek:
@mdbooth, you probably want to take another look here.

jingczhang:
Hi @mdbooth, I updated the PR per your comment, pls review. I am testing with k8s 1.23, I am unable to retrieve the trunk_details as you expect, I used continue to skip the code changes for now.

mdbooth:
> Hi @mdbooth, I updated the PR per your comment, pls review. I am testing with k8s 1.23, I am unable to retrieve the trunk_details as you expect, I used continue to skip the code changes for now.

Thanks! This is high on my todo list but unfortunately it doesn't look like I'm going to get to it this week. Hopefully Monday.

jingczhang:
Hi @mdbooth, I managed to test with k8s 1.23, all good, works as expected. It is amazing the master branch works with k8s 1.23.

pkg/openstack/instances.go
```diff
@@ -633,6 +633,44 @@ func nodeAddresses(srv *servers.Server, ports []ports.Port, networkingOpts Netwo
...
 	}
 
+	// Add the addresses assigned on subports via trunk
+	// This exposes the vlan networks to which subports are attached
+	subportAddresses := make(map[string][]Address)
+
+	for _, port := range ports {
+		for _, subport := range port.TrunkDetails.SubPorts {
+			p, err := neutronports.Get(client, subport.PortID).Extract()
+			if err != nil {
```
mdbooth, line 643: Nice! This only has a cost if there are trunk ports. It is free on systems which are not using trunk ports.

pkg/openstack/instances.go
```diff
@@ -633,6 +633,44 @@ func nodeAddresses(srv *servers.Server, ports []ports.Port, networkingOpts Netwo
...
+				klog.Errorf("Failed to get subport %s network details: %v", subport.PortID, err)
+				continue
+			}
+			for _, fixedIP := range p.FixedIPs {
+				isIPv6 := net.ParseIP(fixedIP.IPAddress).To4() == nil
+				if !(isIPv6 && networkingOpts.IPv6SupportDisabled) {
+					addr := Address{IPType: "fixed", Addr: fixedIP.IPAddress}
+					subportAddresses[n.Name] = append(subportAddresses[n.Name], addr)
+					klog.V(5).Infof("Node '%s' is added subport '%s' address '%s/%s'", srv.Name, p.Name, n.Name, addr)
```
mdbooth, line 657: nit: This should use structured logging, but as none of this module uses structured logging that's probably best dealt with consistently in another PR.

pkg/openstack/instances.go
```diff
@@ -633,6 +633,44 @@ func nodeAddresses(srv *servers.Server, ports []ports.Port, networkingOpts Netwo
...
+		srvAddresses, ok := addresses[nw]
+		if !ok {
+			addresses[nw] = ips
+		} else {
+			// this is to take care the corner case
+			// where the same network is attached to the node both directly and via trunk
+			addresses[nw] = append(srvAddresses, ips...)
+		}
+	}
+
```
mdbooth, line 673: Suggestion: we're already looping over this above (in the FixedIPs loop). You could move this to there and eliminate subportAddresses.

mdbooth (commented):
/lgtm

mdbooth:
@jingczhang Those test failures were weird. I'm currently assuming they don't relate to the current version of your PR.

Please could you squash the commits before merging?

mdbooth:
Oh, nm. I had some stale state locally. Yeah, you need to fix the unit tests. If you do that and also squash the commits this still lgtm.

pkg/openstack/instances.go
jingczhang: will check

dulek:
Yup, looks good, just needs the unit tests to be fixed. Thanks!

jingczhang:
Hi @mdbooth, your last comment has been taken care of, please review.

jichenjc:
/approve

I am ok with current code but looks the UT only modify existing cases
can we follow up create more UT To cover, e.g multiple VLAN address returned ?

mdbooth:
> /approve
> 
> I am ok with current code but looks the UT only modify existing cases can we follow up create more UT To cover, e.g multiple VLAN address returned ?

The last time we looked at this the problem was that it's extremely hard to do without being able to mock OpenStack API calls. We can do this in CAPO now, but the solution was quite intrusive and it's not easy to share. It's on my longer-term todo list to find a way to do this, ideally directly in gophercloud.

Issue #2271, opened 2023-06-09
[occm] Support for OpenStack LBaaS providers that do not implement the "fully-populated" loadbalancer API call


m-bull:
/kind feature

**What happened**:
Some Octavia plugins (specifically [`vmwareedge`/`vmware-nsx`](https://opendev.org/x/vmware-nsx) in this case) do not provide an API endpoint for creating "fully-populated" loadbalancers, so calling [this function](https://github.com/kubernetes/cloud-provider-openstack/blob/83fe6be0d9f557a2fed6dfad55c22aab317eff5a/pkg/openstack/loadbalancer.go#L445) results in a http 503. As a result, loadbalancers aren't created. These plugins are deployed on VMWare Integrated OpenStack (VIO) platforms.

While the `vmwareedge` Octavia provider isn't explicitly supported here (just Amphora/OVN), recycling some old code to use the non-fully-populated loadbalancer API calls and placing it behind an opt-in switch would allow it to be used. I had a go a making these changes [here](https://github.com/stackhpc/cloud-provider-openstack/pull/3), and this approach works to create loadbalancers in a VMWare VIO7 environment, although I haven't tested ingress resources in anger.

The patch is a bit old now (and probably not implemented brilliantly), but if the project would be interested in supporting a "legacy-api-call" config option that re-enabled the non-fully-populated loadbalancer creation approach from OCCM <1.21 alongside the existing approach, I'd happily contribute.

**What you expected to happen**:
Loadbalancers are created.

**How to reproduce it**:
On a VMWare VIO7 cloud (OpenStack Train and `vmware-nsx` Neutron/Octavia plugins), using any version of OCCM after 1.20 results in a failure to provision a loadbalancer.

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: >1.21
- OpenStack version:  Train
- Others: VMware Integrated OpenStack 7.0

mdbooth:
@dulek Thoughts?

dulek:
@m-bull: Thanks for raising this! So the code you're attempting to add doesn't feel too cluttered, so I think it's worth considering. I wonder however how do you deal with the fact that after each call CCM needs to wait for the LB to become ACTIVE again before making the next call. In bigger environments (60-100 nodes) we've seen that this causes LB creation to take up to an hour and I believe this is the reason to use the bulk APIs in CCM.

dulek:
[...]

PR #2314, opened 2023-07-27
[occm] Support for LB providers that do not implement the bulk update/create API call

m-bull:
Add support for LB providers that do not implement the bulk create/update API call.

fixes https://github.com/kubernetes/cloud-provider-openstack/issues/2271

As this PR adds new configuration options for both Octavia ingress and Loadbalancers, additional configuration is required to test:

```
[LoadBalancer]
provider-requires-serial-api-calls=true
```
and
```bash
cat <<EOF > /etc/kubernetes/octavia-ingress-controller/config.yaml
---
kind: ConfigMap
apiVersion: v1
metadata:
  name: octavia-ingress-controller-config
  namespace: kube-system
data:
  config: |
    cluster-name: ${cluster_name}
    openstack:
      auth-url: ${auth_url}
      domain-name: ${domain-name}
      username: ${user_name}
      # user-id: ${user_id}
      password: ${password}
      project-id: ${project_id}
      region: ${region}
    octavia:
      subnet-id: ${subnet_id}
      floating-network-id: ${public_net_id}
      provider-requires-serial-api-calls: true
EOF
kubectl apply -f /etc/kubernetes/octavia-ingress-controller/config.yaml
```

**Release note**:

```release-note
[openstack-cloud-controller-manager] Adds provider-requires-serial-api-calls config option to support Octavia providers that do not implement the bulk update/create API call. Set provider-requires-serial-api-calls to true to use serial calls to Octavia to create and populate loadbalancers.
[octavia-ingress-controller] Adds provider-requires-serial-api-calls config option to support Octavia providers that do not implement the bulk update/create API call. Set provider-requires-serial-api-calls to true to use serial calls to Octavia to create and populate loadbalancers.
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -451,7 +451,7 @@ func getListenerProtocol(protocol corev1.Protocol, svcConf *serviceConfig) liste
 	}
 }
 
-func (lbaas *LbaasV2) createFullyPopulatedOctaviaLoadBalancer(name, clusterName string, service *corev1.Service, nodes []*corev1.Node, svcConf *serviceConfig) (*loadbalancers.LoadBalancer, error) {
+func (lbaas *LbaasV2) createOctaviaLoadBalancer(name, clusterName string, service *corev1.Service, nodes []*corev1.Node, svcConf *serviceConfig) (*loadbalancers.LoadBalancer, error) {
 	createOpts := loadbalancers.CreateOpts{
 		Name:        name,
 		Description: fmt.Sprintf("Kubernetes external service %s/%s from cluster %s", service.Namespace, service.Name, clusterName),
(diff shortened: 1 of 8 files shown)


pkg/util/util.go
```diff
@@ -16,6 +16,15 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// CutString makes sure the string length doesn't exceed 255, which is usually the maximum string length in OpenStack.
+func CutString(original string) string {
```
dulek, line 20: If this lands on utils you got to either name this something like `CutStringTo255()` or allow passing `255` as a parameter.

pkg/openstack/loadbalancer.go
```diff
@@ -451,6 +451,57 @@ func getListenerProtocol(protocol corev1.Protocol, svcConf *serviceConfig) liste
...
+			createOpts.VipNetworkID = svcConf.lbNetworkID
+		} else {
+			klog.V(4).Infof("network-id parameter not passed, it will be inferred from subnet-id")
+		}
+	}
+	// For external load balancer, the LoadBalancerIP is a public IP address.
+	loadBalancerIP := service.Spec.LoadBalancerIP
+	if loadBalancerIP != "" && svcConf.internal {
+		createOpts.VipAddress = loadBalancerIP
+	}
```
dulek, line 495: Looks like this whole fragment is duplicated in lines 506-549. Could you extract it to another function and use it in both places? Something like `getOctaviaLoadBalancerCreateOpts()`.

dulek (changes requested):
This looks nice and clean! Couple of remarks inline. Please also make sure to add the information about the new configuration options to the docs: https://github.com/kubernetes/cloud-provider-openstack/blob/dd53a6e535056322548361cbf3345d9590bcc304/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md?plain=1#L268-L282

pkg/util/util.go
m-bull: Renamed to `CutStringTo255()`.

pkg/openstack/loadbalancer.go
m-bull: I ended up switching this around a bit to reuse the original function and just add the "fully-populated" createOpts only when the provider doesn't require the serial-style requests.

Happy to do as you suggest here if this approach isn't as good in your opinion.

dulek: Oh no, this is nice and clean, great idea!

m-bull:
@dulek - thanks for the reviews.

I'm leaving this as [WIP] just for a minute while I do some more testing on the system that I originally made this patch for, will update the title as soon as I can confirm that everything still works as intended with the new options set.

m-bull:
@dulek - I made a final change to correct a log message and tested this patch successfully on the cloud that I wrote the original one for. `[WIP]` removed.

dulek:
/lgtm

I'd still like to have @mdbooth opinion on option naming here.

mdbooth:
> provider-requires-serial-api-calls

I think it's specific enough to not be worth bike-shedding further.

PR #2317, opened 2023-08-07
Add imagePullSecret support

carlotardl:
This PR implements `imagePullSecrets` for the `openstack-cloud-controller-manager` helm chart.
fixes #

**Release note**:

```release-note
NONE
```


dulek:
/ok-to-test

Looks good! Seems like the only chart that'd missing this would be Manila CSI. Could you add the option there too?

Lirt:
> /ok-to-test
> 
> Looks good! Seems like the only chart that'd missing this would be Manila CSI. Could you add the option there too?

I opened a new PR for Manila to synchronize this functionality across all 3 helm charts.
Please see https://github.com/kubernetes/cloud-provider-openstack/pull/2323

PR #1755, opened 2022-01-21
[occm] Fix the service account name

lingxiankong:
**What this PR does / why we need it**:
We were using incorrect service account name which causes service creation failure.

**Which issue this PR fixes(if applicable)**:
fixes #1722

**Special notes for reviewers**:

**Release note**:

```release-note
NONE
```

PR #2332, opened 2023-08-22
[occm] Use standard service account name in OCCM helm chart

### LINKED ISSUE #2049: [occm] (helm chart) `cloud-controller-manager` serviceaccount has no clusterrolebinding
/kind bug

**What happened**:
Using `--use-service-account-crredentials` (as is default in the chart), I'm seeing the following in the logs:

```
I1202 01:58:51.386134       1 event.go:294] "Event occurred" object="test/nginx-ingress-nginx-ingress" fieldPath="" kind="Service" apiVersion="v1" type="Warning" reason="SyncLoadBalancerFailed" message="Error syncing load balancer: failed to ensure load balancer: failed to patch service object test/nginx-ingress-nginx-ingress: services \"nginx-ingress-nginx-ingress\" is forbidden: User \"system:serviceaccount:kube-system:cloud-controller-manager\" cannot patch resource \"services\" in API group \"\" in the namespace \"test\""
```
Should the roles stuff in the [manifests](https://github.com/kubernetes/cloud-provider-openstack/tree/master/manifests/controller-manager) folder be part of the chart, or am I misunderstanding how this works?

When I set `--use-service-account-credentials=false` via `controllerExtraArgs`, things seem to work as expected.

No permission errors in the logs

**How to reproduce it**:
Install the helm chart with values:
```
    image:
      tag: v1.24.5
```

wwentland:

This removes the `openstack-` prefix from the service account name used by the cloud-controller-manager.

The change is motivated by the following:

- Create suitable service accounts, cluster roles and cluster role bindings for use with `--use-service-accounts-credentials=true`
- Normalise service account names in the helm chart and plain manifests
- Adhere to naming conventions across external cloud controller managers for different clouds (e.g. AWS, GCP, ...)

Specifically the first point deserves further details. Prior to this change, users who install the cloud controller manager with helm, would run into the following error when creating load balancers:

```
E0818 08:27:33.802407      11 controller.go:291] error processing service default/hello-bug (will retry): failed to ensure load balancer: failed to patch service object default/hello-bug: services "hello-bug" is forbidden: User "system:serviceaccount:kube-system:cloud-controller-manager" cannot patch resource "services" in API group "" in the namespace "default"
```

Which is due to the fact that the controller is running with the `cloud-controller-manager` service account because `--use-service-account-credentials` is set to `true` by default and the client is initialised with:

```
clientset := clientBuilder.ClientOrDie("cloud-controller-manager")
```

Whilst users can work around this by passing
`--use-service-account-credentials=false`, the desired behaviour would be to install suitable RBAC in the first place.

See:

- https://kubernetes.io/docs/concepts/architecture/cloud-controller/
- https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/
- https://github.com/kubernetes/cloud-provider-openstack/issues/2049
- https://github.com/kubernetes/cloud-provider-openstack/issues/1722
- https://github.com/kubernetes/cloud-provider-openstack/pull/1755

fixes #2049

One slightly tricky aspect of this change is, that the controller creates the `cloud-controller-manager` service account if users installed the helm chart in the default configuration, i.e. without explicitly setting:

```
controllerExtraArgs: |
  - --use-service-account-credentials=false
```

Any service account created by `getOrCreateServiceAccount` would not be managed by Helm and a helm upgrade would fail with:

```
Error: UPGRADE FAILED: rendered manifests contain a resource that already exists. Unable to continue with update: ServiceAccount "cloud-controller-manager" in namespace "kube-system" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "foo"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "kube-system"
```

**Release note**:

```release-note
ACTION REQUIRED: Please note that you might have to delete the `cloud-controller-manager` service account in the `kube-system` namespace if it exists, as upgrading with helm would fail otherwise.
```

### DIFF
--- a/charts/openstack-cloud-controller-manager/templates/clusterrole.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/clusterrole.yaml
@@ -1,7 +1,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: system:openstack-cloud-controller-manager
+  name: {{ .Values.clusterRoleName }}
   annotations:
     {{- with .Values.commonAnnotations }}
     {{- toYaml . | nindent 4 }}
--- a/charts/openstack-cloud-controller-manager/templates/clusterrolebinding.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/clusterrolebinding.yaml
@@ -1,7 +1,8 @@
+---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: system:openstack-cloud-controller-manager
+  name: {{ .Values.clusterRoleName }}
   annotations:
     {{- with .Values.commonAnnotations }}
     {{- toYaml . | nindent 4 }}
@@ -9,8 +10,8 @@ metadata:
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: system:openstack-cloud-controller-manager
+  name: {{ .Values.clusterRoleName }}
 subjects:
 - kind: ServiceAccount
-  name: openstack-cloud-controller-manager
+  name: {{ .Values.serviceAccountName }}
   namespace: {{ .Release.Namespace | quote }}
--- a/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
@@ -37,7 +37,7 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      serviceAccountName: openstack-cloud-controller-manager
+      serviceAccountName: {{ .Values.serviceAccountName }}
       containers:
         - name: openstack-cloud-controller-manager
           image: "{{ .Values.image.repository }}:{{ default .Chart.AppVersion .Values.image.tag }}"
--- a/charts/openstack-cloud-controller-manager/templates/serviceaccount.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/serviceaccount.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: openstack-cloud-controller-manager
+  name: {{ .Values.serviceAccountName }}
   namespace: {{ .Release.Namespace }}
   annotations:
     {{- with .Values.commonAnnotations }}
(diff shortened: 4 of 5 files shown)


wwentland:
There is also a bigger topic of using a more fine-granular separation for the different functions of the cloud-controller-manager, along its functions (cf. [0]):

- Node controller
- Service controller
...

The following are probably of interest in that regard:

- https://github.com/kubernetes/kubernetes/pull/115813/
- https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cloud-provider/names/controller_names.go

Which is, however, outwith the scope of this PR, but something to keep in mind.

[0]: https://kubernetes.io/docs/concepts/architecture/cloud-controller/#functions-of-the-ccm

wwentland:
Thank you for your comments, @dulek!

> Should we consider making it configurable as AWS provider does?

I'm happy to allow users to configure this, but the service account name is closely related to the client name which users won't be able to configure (as that's set in code).

Personally, I prefer to move most settings in a helm chart into the values file as doing so does not prevent users from implementing use cases you might not have foreseen when writing the chart. Happy to update that.

> Also it seems like it skips the "system:" prefix?

Which is in line with the way the service account is named here, whilst the prefix is included for roles and bindings.

FWIW, I am currently wondering if the cluster roles should use the system:controller prefix (cf. [0](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#controller-roles)). I am leaning towards it, but I haven't been able to confirm that yet. Do you, by chance, know if that's desirable these days?

wwentland:
@dulek I've moved both the service account and cluster role name into the helm values files, akin to what's being done in the AWS controller.

What are your thoughts on the upgrade procedure and corresponding release notes?

dulek:
> @dulek I've moved both the service account and cluster role name into the helm values files, akin to what's being done in the AWS controller.
> 
> What are your thoughts on the upgrade procedure and corresponding release notes?

IIUC the only manual step is to delete the old SA and CR? That works for me if mentioned in the release notes.

wwentland:
> IIUC the only manual step is to delete the old SA and CR? That works for me if mentioned in the release notes.

Yeah, the SA would have to be deleted if it exists. Is the release note section above enough for the text to be automatically included or does it require any manual steps?

dulek:
I think it's enough to get release notes generated.
/lgtm

jichenjc:
looks like the manifest use `cloud-controller-manager` , so we should be good

```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: cloud-controller-manager
  namespace: kube-system
---
```

PR #2338, opened 2023-08-25
[cinder-csi-plugin] Fix multiarch builds of Cinder CSI

dulek:
The libs copied into cinder-csi-plugin image are named and placed differently across platforms. This commit attempts to solve this by differentiating where we take them from.

fixes #

**Release note**:

```release-note
NONE
```


zetaab (approved):
/approve
/lgtm

PR #2341, opened 2023-08-25
[occm] Delete LB that gone into ERROR state

dulek:
This commit implements deleting the LB when it goes into ERROR state on creation. This will allow us to be less prone to transient Octavia issues.

A possible improvement could be to extend this behavior to all cases when we wait for the LB to become ACTIVE. The problem with that is that attempt to implement that complicates the code a lot, so let's start with this first step.

fixes #

**Release note**:

```release-note
[openstack-cloud-controller-manager] OCCM will attempt to retry creation of load balancers that have gone into ERROR state.
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -539,6 +539,14 @@ func (lbaas *LbaasV2) createOctaviaLoadBalancer(name, clusterName string, servic
 	}
 
 	if loadbalancer, err = openstackutil.WaitActiveAndGetLoadBalancer(lbaas.lb, loadbalancer.ID); err != nil {
+		if loadbalancer.ProvisioningStatus == errorStatus {
+			// If LB landed in ERROR state we should delete it and retry the creation later.
+			if err = lbaas.deleteLoadBalancer(loadbalancer, service, svcConf, true); err != nil {
+				return nil, fmt.Errorf("loadbalancer %s is in ERROR state and there was an error when removing it: %v", loadbalancer.ID, err)
+			}
+			return nil, fmt.Errorf("loadbalancer %s has gone into ERROR state, please check Octavia for details. Load balancer was "+
+				"deleted and its creation will be retried", loadbalancer.ID)
+		}
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


dulek:
/hold 

Got to run some manual tests here before this is allowed to be merged.

mdbooth (commented):
/lgtm

dulek:
@kayrus, @jichenjc, maybe you can take a look here?

pkg/openstack/loadbalancer.go
```diff
@@ -539,6 +539,14 @@ func (lbaas *LbaasV2) createOctaviaLoadBalancer(name, clusterName string, servic
 	}
 
 	if loadbalancer, err = openstackutil.WaitActiveAndGetLoadBalancer(lbaas.lb, loadbalancer.ID); err != nil {
+		if loadbalancer.ProvisioningStatus == errorStatus {
+			// If LB landed in ERROR state we should delete it and retry the creation later.
+			if err = lbaas.deleteLoadBalancer(loadbalancer, service, svcConf, true); err != nil {
+				return nil, fmt.Errorf("loadbalancer %s is in ERROR state and there was an error when removing it: %v", loadbalancer.ID, err)
```
jichenjc, line 545: not sure this case we should re-create as we failed to delete it, will it create a dupcate later?

jichenjc:
/approve
/hold

@dulek please check my question above, feel free to unhold to merge if it's reasonable flow

pkg/openstack/loadbalancer.go
mdbooth: I don't think so because we check for an existing LB before creation: https://github.com/kubernetes/cloud-provider-openstack/blob/5dc5f0ad12ae325a4c8372f91c8e570ba2f976cd/pkg/openstack/loadbalancer.go#L1919-L1934

@dulek However, that does raise an interesting point. Perhaps we should set the `ServiceAnnotationLoadBalancerID` annotation as soon as we know what it is and not wait until it's ready. We would remove it here only if the deletion was successful.

dulek: Yeah, we always lookup by name, so if the creation fails on ERROR and we won't be able to delete the LB - stuff will work as it was. I was initially thinking about changing the place we put the LB ID on the Service, but I realized that it cannot be bulletproof - updating a Service might fail at any time. The only source of truth that we have is the Octavia API.

I can change it so that we annotate the LB ID immediately and remove it here only after successful deletion, but to me it doesn't change much.

Opinions?

mdbooth: > Yeah, we always lookup by name, so if the creation fails on ERROR and we won't be able to delete the LB - stuff will work as it was. I was initially thinking about changing the place we put the LB ID on the Service, but I realized that it cannot be bulletproof - updating a Service might fail at any time. The only source of truth that we have is the Octavia API.
> 
> I can change it so that we annotate the LB ID immediately and remove it here only after successful deletion, but to me it doesn't change much.
> 
> Opinions?

I agree it doesn't change much in practise. The primary benefit is in naming conflicts, which is really a separate problem.

PR #2345, opened 2023-08-28
[docs] :seedling: Update image repo in the release procedure

dulek:
Seems like we no longer need to modify k8s.gcr.io, it's registry.k8s.io instead.

**Release note**:

```release-note
NONE
```


zetaab (approved):
/approve
/lgtm

PR #2346, opened 2023-08-29
[occm] Set node selector and tolerations by default in OCCM helm chart

wwentland:

This defines suitable default values for `nodeSelector` and `tolerations` to ensure a nicer out-of-the-box experience for users.

Without this change all users would have to maintain these in their respective values file.

See:
- https://kubernetes.io/docs/reference/labels-annotations-taints/

Please note that this (somewhat) matches the default settings used in [cloud-provider-aws](https://github.com/kubernetes/cloud-provider-aws/blob/master/charts/aws-cloud-controller-manager/values.yaml), with the notable difference of me having removed the toleration for the following deprecated taint:

```
tolerations:
- key: node-role.kubernetes.io/master
  effect: NoSchedule
```


wwentland:
The lint error has been addressed in https://github.com/kubernetes/cloud-provider-openstack/pull/2347

I'll rebase this once that PR has been merged.

charts/openstack-cloud-controller-manager/values.yaml
```diff
@@ -37,22 +37,16 @@ livenessProbe: {}
 readinessProbe: {}
 
 # Set nodeSelector where the controller shut run, i.e. controlplane nodes
```
dulek, line 39: Can you fix the description too? I guess "shut" is "should".

charts/openstack-cloud-controller-manager/values.yaml
```diff
@@ -37,22 +37,16 @@ livenessProbe: {}
...
-nodeSelector: {}
-# nodeSelector:
-#   node-role.kubernetes.io/controlplane: "true"
+nodeSelector:
+  node-role.kubernetes.io/control-plane: ""
```
dulek, line 41: Why the change from the comment version?

charts/openstack-cloud-controller-manager/values.yaml
wwentland: Sure, happy to.

charts/openstack-cloud-controller-manager/values.yaml
wwentland: Mostly because the label/taint `node-role.kubernetes.io/controlplane` is not set by kubeadm, nor is its value `true`.

- https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-control-plane
- https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-control-plane-taint

After CNI installation, a node will have the following labels and taints:

```
~/src/localhost/capo/flatcar/live
❯ k get node foo-control-plane-jq5jq -o json | jq '.metadata.labels'
{
  "beta.kubernetes.io/arch": "amd64",
  "beta.kubernetes.io/os": "linux",
  "kubernetes.io/arch": "amd64",
  "kubernetes.io/hostname": "k8s-ww-mgmt-muc5-0-control-plane-jq5jq",
  "kubernetes.io/os": "linux",
  "node-role.kubernetes.io/control-plane": "",
  "node.kubernetes.io/exclude-from-external-load-balancers": ""
}

❯ k get node foo-control-plane-jq5jq -o json|jq '.spec.taints'
[
  {
    "effect": "NoSchedule",
    "key": "node-role.kubernetes.io/control-plane"
  },
  {
    "effect": "NoSchedule",
    "key": "node.cloudprovider.kubernetes.io/uninitialized",
    "value": "true"
  }
]
```

Please note that missing value for `node-role.kubernetes.io/control-plane` as opposed to `node.cloudprovider.kubernetes.io/uninitialized`.

FWIW, you can see similar values in both the AWS and GCP providers:

- [cloud-provider-aws - nodeSelector](https://github.com/kubernetes/cloud-provider-aws/blob/master/charts/aws-cloud-controller-manager/values.yaml#L16)
- [cloud-provider-aws - tolerations](https://github.com/kubernetes/cloud-provider-aws/blob/master/charts/aws-cloud-controller-manager/values.yaml#L125)
- [cloud-provider-gcp - nodeSelector](https://github.com/kubernetes/cloud-provider-gcp/blob/master/deploy/packages/default/manifest.yaml#L28)
- [cloud-provider-gcp - tolerations](https://github.com/kubernetes/cloud-provider-gcp/blob/master/deploy/packages/default/manifest.yaml#L41)

I actually find the solution of the GCP controller interesting in that they use `nodeAffinity` with `requiredDuringSchedulingIgnoredDuringExecution`. I guess [set-based requirements](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement) allow for easier/more distinct differentiation between `Exists` and `In`.

wwentland: Mostly because the label/taint `node-role.kubernetes.io/controlplane` is not set by kubeadm, nor is its value `true`.

- https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-control-plane
- https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-control-plane-taint

After CNI installation, a node will have the following labels and taints:

```
❯ k get node foo-control-plane-jq5jq -o json | jq '.metadata.labels'
{
  "beta.kubernetes.io/arch": "amd64",
  "beta.kubernetes.io/os": "linux",
  "kubernetes.io/arch": "amd64",
  "kubernetes.io/hostname": "k8s-ww-mgmt-muc5-0-control-plane-jq5jq",
  "kubernetes.io/os": "linux",
  "node-role.kubernetes.io/control-plane": "",
  "node.kubernetes.io/exclude-from-external-load-balancers": ""
}

❯ k get node foo-control-plane-jq5jq -o json | jq '.spec.taints'
[
  {
    "effect": "NoSchedule",
    "key": "node-role.kubernetes.io/control-plane"
  },
  {
    "effect": "NoSchedule",
    "key": "node.cloudprovider.kubernetes.io/uninitialized",
    "value": "true"
  }
]
```

Please note the missing value for `node-role.kubernetes.io/control-plane` as opposed to `node.cloudprovider.kubernetes.io/uninitialized`.

FWIW, you can see similar values in both the AWS and GCP providers:

- [cloud-provider-aws - nodeSelector](https://github.com/kubernetes/cloud-provider-aws/blob/master/charts/aws-cloud-controller-manager/values.yaml#L16)
- [cloud-provider-aws - tolerations](https://github.com/kubernetes/cloud-provider-aws/blob/master/charts/aws-cloud-controller-manager/values.yaml#L125)
- [cloud-provider-gcp - nodeSelector](https://github.com/kubernetes/cloud-provider-gcp/blob/master/deploy/packages/default/manifest.yaml#L28)
- [cloud-provider-gcp - tolerations](https://github.com/kubernetes/cloud-provider-gcp/blob/master/deploy/packages/default/manifest.yaml#L41)

I actually find the solution of the GCP controller interesting in that they use `nodeAffinity` with `requiredDuringSchedulingIgnoredDuringExecution`. I guess the `matchExpression` allows them to explicitly use `Exists` rather than havint to define an empty value / `""`.

jichenjc: um... node.cloudprovider.kubernetes.io/uninitialized this seems means OCCM not finished yet
should be rmeoved when the OCCM complete its job?

wwentland: > um... node.cloudprovider.kubernetes.io/uninitialized this seems means OCCM not finished yet
should be rmeoved when the OCCM complete its job?

The output above is the state *before* OCCM has been installed, but *after* I've installed Cilium, so OCCM did not yet have a chance to initialise the node and remove the taint (cf. [node.cloudprovider.kubernetes.io/uninitialized](https://kubernetes.io/docs/reference/labels-annotations-taints/#node-cloudprovider-kubernetes-io-uninitialized)):

> Sets this taint on a Node to mark it as unusable, when kubelet is started with the "external" cloud provider, until a controller from the cloud-controller-manager initializes this Node, and then removes the taint.

We need the toleration, so that new nodes can be initialised.

wwentland:
> I'll rebase this once that PR has been merged.

I've rebased this now and the linter is happy.

wwentland:
What are your thoughts on adding `node-role.kubernetes.io/master:NoSchedule`? The taint is currently [deprecated](https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-master-taint), but we could add it for a better default experience on clusters that used an older version that still sets it.

Documentation states:

> Taint that kubeadm previously applied on control plane nodes to allow only critical workloads to schedule on them. Replaced by the [node-role.kubernetes.io/control-plane](https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-control-plane-taint) taint. kubeadm no longer sets or uses this deprecated taint.

wwentland:
What are your thoughts on adding `node-role.kubernetes.io/master:NoSchedule`? The taint is currently [deprecated](https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-master-taint), but we could add it for a better default experience on clusters that used an older version that still sets it.

Documentation states:

> Taint that kubeadm previously applied on control plane nodes to allow only critical workloads to schedule on them. Replaced by the [node-role.kubernetes.io/control-plane](https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-control-plane-taint) taint. kubeadm no longer sets or uses this deprecated taint.

dulek:
> What are your thoughts on adding `node-role.kubernetes.io/master:NoSchedule`? The taint is currently [deprecated](https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-master-taint), but we could add it for a better default experience on clusters that used an older version that still sets it.
> 
> Documentation states:
> 
> > Taint that kubeadm previously applied on control plane nodes to allow only critical workloads to schedule on them. Replaced by the [node-role.kubernetes.io/control-plane](https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-control-plane-taint) taint. kubeadm no longer sets or uses this deprecated taint.

To me an additional toleration doesn't hurt.

wwentland:
> > To me an additional toleration doesn't hurt.
> 
> That was my thinking too. I'll add it back in, and it can be removed eventually.
> 
> /hold

Done.

/hold cancel

charts/openstack-cloud-controller-manager/values.yaml
```diff
@@ -36,23 +36,20 @@ livenessProbe: {}
...
+
+# Set tolerations for nodes where the controller should run, i.e. node
+# should be uninitialized, controlplane...
+tolerations:
+  - key: node.cloudprovider.kubernetes.io/uninitialized
+    value: "true"
+    effect: NoSchedule
+  - key: node-role.kubernetes.io/master
+    effect: NoSchedule
```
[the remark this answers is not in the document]
wwentland: Yeah, that was my initial reasoning for not including that taint.

I'm happy either way, but lean slightly more towards removing it as it cuts down on technical debt.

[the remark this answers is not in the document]
mdbooth: Don't consider this a downvote from me, btw. Just my 2c.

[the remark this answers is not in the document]
wwentland: It was the way I had initially implemented it. I still lean towards leaving the toleration for deprecated taints out, but have no strong feeling either way. 

That's why I asked for comments. If there really is only a very slim chance of users ever needing this, then we should leave it out and they can always override the default tolerations themselves.

I think I'll stick to tolerations for non-deprecated taints and will remove the one for `node-role.kubernetes.io/master` again.

Thank you for your thoughts!

PR #2347, opened 2023-08-29
[occm] Remove spurious space in OCCM values file

wwentland:

The spurious space character was sadly introduced in an earlier PR and had not been picked up by the helm linter, presumably because the GHA did not run.

**Release note**:

```release-note
NONE
```


wwentland:
This error was introduced in https://github.com/kubernetes/cloud-provider-openstack/pull/2332/

PR #2355, opened 2023-08-30
[occm] Lookup ports by SG and not tags

dulek:
When creating an SG for an LB we tag the ports we attach the SG to in order to be able to find them easily later on when the SG will be getting deleted. We do that to simulate an API allowing us to filter ports by attached SGs, but turns out - such an API already exists in Neutron and Gophercloud v1.6.0 allows to use it.

This commit makes sure that when deleting an SG we're discovering ports to remove it from by querying the Neutron filtering by SG. Moreover tagging the ports with the SGs is removed to as not being necessary anymore.

I've left the code removing the SG tags from ports for backward compatibility.

fixes #

**Release note**:

```release-note
[openstack-cloud-controller-manager] Ports of the nodes will not anymore be tagged with the IDs of SGs OCCM creates for the load balancers.
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -732,22 +732,13 @@ func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, nodes []*
 				continue
 			}
 
-			// Add the security group ID as a tag to the port in order to find all these ports when removing the security group.
-			// We're doing that before actually applying the SG as if tagging would fail we wouldn't be able to find the port
-			// when deleting the SG and operation would be stuck forever. It's better to find more ports than not all of them.
-			mc := metrics.NewMetricContext("port_tag", "add")
-			err := neutrontags.Add(network, "ports", port.ID, sg).ExtractErr()
-			if mc.ObserveRequest(err) != nil {
-				return fmt.Errorf("failed to add tag %s to port %s: %v", sg, port.ID, err)
-			}
-
 			// Add the SG to the port
 			// TODO(dulek): This isn't an atomic operation. In order to protect from lost update issues we should use
 			//              `revision_number` handling to make sure our update to `security_groups` field wasn't preceded
 			//              by a different one. Same applies to a removal of the SG.
 			newSGs := append(port.SecurityGroups, sg)
 			updateOpts := neutronports.UpdateOpts{SecurityGroups: &newSGs}
-			mc = metrics.NewMetricContext("port", "update")
+			mc := metrics.NewMetricContext("port", "update")
 			res := neutronports.Update(network, port.ID, updateOpts)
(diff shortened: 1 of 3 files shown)


pkg/openstack/loadbalancer.go
```diff
@@ -777,17 +768,23 @@ func disassociateSecurityGroupForLB(network *gophercloud.ServiceClient, sg strin
...
 			return fmt.Errorf("failed to update security group for port %s: %v", port.ID, res.Err)
 		}
-		// Remove the security group ID tag from the port.
-		mc = metrics.NewMetricContext("port_tag", "delete")
-		err := neutrontags.Delete(network, "ports", port.ID, sg).ExtractErr()
-		if mc.ObserveRequest(err) != nil {
-			return fmt.Errorf("failed to remove tag %s to port %s: %v", sg, port.ID, res.Err)
+
+		// Remove the security group ID tag from the port. Please note we don't tag ports with SG IDs anymore,
+		// so this stays for backward compatibility. It's reasonable to delete it in the future. 404s are ignored.
```
mdbooth, line 781: What's the backward compatibility? In case the user rolls back an upgrade, perhaps?

pkg/openstack/loadbalancer.go
```diff
@@ -761,7 +752,7 @@ func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, nodes []*
 // disassociateSecurityGroupForLB removes the given security group from the ports
 func disassociateSecurityGroupForLB(network *gophercloud.ServiceClient, sg string) error {
 	// Find all the ports that have the security group associated.
-	listOpts := neutronports.ListOpts{TagsAny: sg}
+	listOpts := neutronports.ListOpts{SecurityGroups: []string{sg}}
```
mdbooth, line 755: Shiny.

mdbooth (commented):
/lgtm

pkg/openstack/loadbalancer.go
dulek: Uhm, well, it will be a problem. The old code won't tag the ports again and deletion of SGs created with new version of the code will fail with the old code.

What's your advice here? Leave the tagging intact? I was really tempted to remove it as it saves us a Neutron call per port, but I guess we can leave it there for one more release?

jichenjc:
/approve
/hold

in case @mdbooth you still want to take another look or feel free to unhold

pkg/openstack/loadbalancer.go
mdbooth: I was thinking just leave the legacy tags there and ignore them, but this is cleaner.

mdbooth (commented):
/lgtm

dulek:
Alright, I'm assuming I can unhold this.

/hold cancel

PR #2358, opened 2023-09-05
[occm] Support get OS instance ID from instance metadata

### LINKED ISSUE #2357: [occm] Support get OS instance ID from instance metadata

/kind feature

**What happened**:

Now, occm support two ways to get the info of OS instance which associated with k8s node:
* First, get instance ID from `node.Spec.ProviderID`, and then query instance info from OS use the instance ID.
* If `node.Spec.ProviderID` is not set, occm try to query OS instance according to the k8s node's name.

But, In some deploy cases the k8s node's name not equal to the node's hostname and then it's not equal to the OS instance's name. In these cases, the occm will report error: `instance not found`.

When `node.Spec.ProviderID` is not set, occm  try to get OS instance id from OS instance's metadata. And use this id to get OS instance info.

**How to reproduce it**:

Deploy k8s cluster in openstack VM and set the kubelet node name not equal to the instance's name. And then startup occm, occm will report `instance not found` in it's log message.

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: latest
- OpenStack version: ussuri
- Others:

jeffyjf:

Now, occm support two ways to get OS instance which associated with k8s node:1. First, get instance ID from `node.Spec.ProviderID`, and then query instance info from OS use the instance ID. 2. If `node.Spec.ProviderID` is not set, occm try to query OS instance according to the k8s node's name.

But, In some deploy cases the k8s node's name not equal to the node's hostname and then it's not equal to the OS instance's name. In these cases, the occm will report error: `instance not found`.

This patch introduce an additonal mechanism to ensure the occm can get the OS instance info correctly.

fixes #2357

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/instancesv2.go
+++ b/pkg/openstack/instancesv2.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/cloud-provider-openstack/pkg/client"
 	"k8s.io/cloud-provider-openstack/pkg/metrics"
 	"k8s.io/cloud-provider-openstack/pkg/util/errors"
+	"k8s.io/cloud-provider-openstack/pkg/util/metadata"
 	"k8s.io/klog/v2"
 )
 
@@ -38,6 +39,7 @@ type InstancesV2 struct {
 	region           string
 	regionProviderID bool
 	networkingOpts   NetworkingOpts
+	metadataServer   metadata.IMetadata
 }
 
 // InstancesV2 returns an implementation of InstancesV2 for OpenStack.
@@ -74,6 +76,7 @@ func (os *OpenStack) instancesv2() (*InstancesV2, bool) {
 		region:           os.epOpts.Region,
 		regionProviderID: regionalProviderID,
 		networkingOpts:   os.networkingOpts,
+		metadataServer:   metadata.GetMetadataProvider(os.metadataOpts.SearchOrder),
 	}, true
 }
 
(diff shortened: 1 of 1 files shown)


dulek:
I don't think this makes sense. My understanding is that [`i.metadataServer.GetInstanceID()`](https://github.com/kubernetes/cloud-provider-openstack/pull/2358/files#diff-adebdcd5abc8ef6cc28c5b140a2dc78cc3d6ca4ae28f2624eadeed570aba5c12R157) will get us the Nova ID of the VM the code is executed from. As we don't know which node the occm is running on, we'll get a wrong ID there most of the time.

We should probably make a step back here and ask - why doesn't your node name match its hostname?

pkg/openstack/instancesv2.go
```diff
@@ -151,6 +154,22 @@ func (i *InstancesV2) makeInstanceID(srv *servers.Server) string {
 
 func (i *InstancesV2) getInstance(ctx context.Context, node *v1.Node) (*ServerAttributesExt, error) {
 	if node.Spec.ProviderID == "" {
+		instanceID, err := i.metadataServer.GetInstanceID()
```
zetaab, line 157: in our case OCCM is executed in control-planes. If we execute this in control-plane it will return the current instance metadata and instance id. NOT for instance the node x which can be parameter for this function. As I see it, this will not work like this

jeffyjf:
@dulek @zetaab Thank you very much for your comments. You are right, I commit a immature propose and patch.

```
We should probably make a step back here and ask - why doesn't your node name match its hostname?
```

Such as, User rename the VM's name after k8s cluster was deployed. Or, Some k8s newbie operators (like me :disappointed:) prefer use ip address  instead of  hostname to deploy k8s cluster.

dulek:
> @dulek @zetaab Thank you very much for your comments. You are right, I commit a immature propose and patch.
> 
> ```
> We should probably make a step back here and ask - why doesn't your node name match its hostname?
> ```
> 
> Such as, User rename the VM's name after k8s cluster was deployed. Or, Some k8s newbie operators (like me 😞) prefer use ip address instead of hostname to deploy k8s cluster.

Oh well, renaming VMs is like shooting yourself in the foot. Nevertheless if you'll find a viable way of identifying nodes using something different than name, we can definitely discuss it. Right now this doesn't seem to do the job.

jeffyjf:
@dulek You are right. I will close the pr.

PR #2360, opened 2023-09-06
[occm] Make sure we don't mask LB tests failures and fix what was failing

dulek:
In `test-lb-service.sh` we do `trap "delete_resources" EXIT` to make sure we cleanup resources on a test failure. In there, we only fetched the `$?` after making a check for `${AUTO_CLEAN_UP}`, which itself alters the code to 0, so function always returns success. This means tests can never really fail.

This commit fixes it by making sure `$ERROR_CODE` is fetched at the very beginning of the cleanup function.

Some additional fixes needed to be made to make tests passing again. In particular:
* Deleting listeners when the secondary shared Service is deleted is fixed now by making sure we're comparing the correct tag.
* Tests no longer create Services as internal in order to make shared LB tests work after #2190.
* FIP is added to the user LB in the sharing test in order for CPO to not complain about not being able to create it for a secondary Service.

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -2390,7 +2390,7 @@ func (lbaas *LbaasV2) deleteLoadBalancer(loadbalancer *loadbalancers.LoadBalance
 					Protocol: proto,
 					Port:     int(port.Port),
 				}]
-				if isPresent && cpoutil.Contains(listener.Tags, loadbalancer.Name) {
+				if isPresent && cpoutil.Contains(listener.Tags, svcConf.lbName) {
 					listenersToDelete = append(listenersToDelete, *listener)
 				}
 			}
@@ -2451,6 +2451,7 @@ func (lbaas *LbaasV2) ensureLoadBalancerDeleted(ctx context.Context, clusterName
 	if err := lbaas.checkServiceDelete(service, svcConf); err != nil {
 		return err
 	}
+	svcConf.lbName = lbName
 
 	if svcConf.lbID != "" {
 		loadbalancer, err = openstackutil.GetLoadbalancerByID(lbaas.lb, svcConf.lbID)
--- a/tests/e2e/cloudprovider/test-lb-service.sh
+++ b/tests/e2e/cloudprovider/test-lb-service.sh
@@ -20,11 +20,11 @@ LB_SUBNET_NAME=${LB_SUBNET_NAME:-"private-subnet"}
 AUTO_CLEAN_UP=${AUTO_CLEAN_UP:-"true"}
 
 function delete_resources() {
+  ERROR_CODE="$?"
+
   if [[ ${AUTO_CLEAN_UP} != "true" ]]; then
     exit ${ERROR_CODE}
   fi
-
-  ERROR_CODE="$?"
 
   printf "\n>>>>>>> Deleting k8s services\n"
   kubectl -n ${NAMESPACE} get svc -o name | xargs -r kubectl -n $NAMESPACE delete
--- a/tests/e2e/cloudprovider/test-lb-service.sh
+++ b/tests/e2e/cloudprovider/test-lb-service.sh
@@ -34,6 +34,12 @@ function delete_resources() {
   printf "\n>>>>>>> Deleting openstack load balancer \n"
   openstack loadbalancer delete test_shared_user_lb --cascade
 
+  printf "\n>>>>>>> Deleting openstack FIPs \n"
+  fips=$(openstack floating ip list --tag occm-test -f value -c ID)
+  for fip in $fips; do
+      openstack floating ip delete ${fip}
+  done
+
   if [[ "$ERROR_CODE" != "0" ]]; then
     printf "\n>>>>>>> Dump openstack-cloud-controller-manager logs \n"
     pod_name=$(kubectl -n kube-system get pod -l k8s-app=openstack-cloud-controller-manager -o name | awk 'NR==1 {print}')
@@ -438,7 +444,6 @@ metadata:
   name: ${service1}
   namespace: $NAMESPACE
   annotations:
-    service.beta.kubernetes.io/openstack-internal-load-balancer: "true"
     loadbalancer.openstack.org/enable-health-monitor: "false"
 spec:
   type: LoadBalancer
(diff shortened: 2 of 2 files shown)


dulek:
`>>>>>>> FAIL: Timeout when waiting for the Service test-shared-2 created`

Well, now we finally start to see failures.

/retest

dulek:
```
I0906 16:55:52.107153      11 event.go:307] "Event occurred" object="octavia-lb-test/test-shared-2" fieldPath="" kind="Service" apiVersion="v1" type="Warning" reason="SyncLoadBalancerFailed" message="Error syncing load balancer: failed to ensure load balancer: internal Service cannot share a load balancer"
```

Okay, so this got broken when I started to limit LB sharing, but I never noticed because tests were busted.

mdbooth:
@dulek Is it worth reverting the previous change to land this quickly? I'm guessing we wouldn't have merged it if we knew it broke a test?

dulek:
> @dulek Is it worth reverting the previous change to land this quickly? I'm guessing we wouldn't have merged it if we knew it broke a test?

```
mdulko:cloud-provider-openstack/ (provider-testing) $ git revert 788ea08fcf5714af8be3b0010bc6d6e5eb1d3a59
Auto-merging pkg/openstack/loadbalancer.go
CONFLICT (content): Merge conflict in pkg/openstack/loadbalancer.go
error: could not revert 788ea08fc... Don't allow internal Services to share an LB (#2190)
```

It's not that trivial anymore. Anyway I plan to look more closely at these tests, my current understanding is that adding `service.beta.kubernetes.io/openstack-internal-load-balancer: "true"` to the LB is just cargo cult from the previous test that did so to avoid unnecessary FIP allocation. I'm pretty sure this is all fixable.

dulek:
@lingxiankong, do you know why we have [this](https://github.com/kubernetes/cloud-provider-openstack/blob/2944f157f0185a712d487ad65e81a19a52097c41/tests/e2e/cloudprovider/test-lb-service.sh#L446) in place in tests?

mdbooth:
> > @dulek Is it worth reverting the previous change to land this quickly? I'm guessing we wouldn't have merged it if we knew it broke a test?
> 
> ```
> mdulko:cloud-provider-openstack/ (provider-testing) $ git revert 788ea08fcf5714af8be3b0010bc6d6e5eb1d3a59
> Auto-merging pkg/openstack/loadbalancer.go
> CONFLICT (content): Merge conflict in pkg/openstack/loadbalancer.go
> error: could not revert 788ea08fc... Don't allow internal Services to share an LB (#2190)
> ```
> 
> It's not that trivial anymore. Anyway I plan to look more closely at these tests, my current understanding is that adding `service.beta.kubernetes.io/openstack-internal-load-balancer: "true"` to the LB is just cargo cult from the previous test that did so to avoid unnecessary FIP allocation. I'm pretty sure this is all fixable.

Probably only worth it if the fix drags on a bit, then. In the meantime we probably shouldn't merge LB-related changes

dulek:
> /retest
> 
> I can see in the logs that listener was removed and LB was ACTIVE again. It couldn't be the ACTIVE-PENDING_UPDATE dance, as we only allow Service to be deleted once it's processed and `kubectl` waits for that.
> 
> Let's try it once again.

Okay, there was a bug, fix is added.

dulek:
```
>>>>>>> Waiting for the Service test-shared-user-lb creation finished

>>>>>>> FAIL: Timeout when waiting for the Service test-shared-user-lb created
```

That's another one.

dulek:
Okay, this one was simple, I left some internal annotations by mistake.

dulek:
Tests are actually passing again! :partying_face: 

@jichenjc, @kayrus, @mdbooth, @zetaab, this is ready for reviews now.

jichenjc:
/lgtm


@dulek given you actually create additional commit for other purpose (e.g compare the LB name) ,can you help 
put that into https://github.com/kubernetes/cloud-provider-openstack/pull/2360#issue-1884352337
as well? Thanks

pkg/openstack/loadbalancer.go
```diff
@@ -2390,7 +2390,7 @@ func (lbaas *LbaasV2) deleteLoadBalancer(loadbalancer *loadbalancers.LoadBalance
 					Protocol: proto,
 					Port:     int(port.Port),
 				}]
-				if isPresent && cpoutil.Contains(listener.Tags, loadbalancer.Name) {
+				if isPresent && cpoutil.Contains(listener.Tags, svcConf.lbName) {
```
mdbooth, line 2393: Can you explain why this is different to loadbalancer.Name? I've read this function and the calling function, in which you're now setting svcConf.lbName to the return value of `lbaas.GetLoadBalancerName`. Does this imply that the loadbalancer name (retrieved by id?) is different to `GetLoadBalancerName`.

mdbooth (commented):
Second and third commits lgtm.

I don't understand the first commit yet.

pkg/openstack/loadbalancer.go
dulek: Yes, so this all boils down to the support for shared LBs that I believe is implemented in a way that is very problematic codewise.

How does it work? You add an annotation to the Service that features the LB ID you want that Service use. The assumption is that the LB exists already and CPO will just add listener, pool and members (members should be the same really[^1]). The _secondary_ LB resources will be tagged with the LB name **that would be used for the secondary Service** so that we can distinguish them later on.

The bug here is that we looked up listeners comparing their tags with the name of the LB taken from OpenStack, which is based on the **primary** Service name. In that case we should use the secondary service name and this is the fix.

[^1]: Now this makes me think that we might have a bug that 2 Services sharing an LB can fight for their view of LB members. Fixed on reconciliation, but not great.

mdbooth: This feels very fragile, but fixing that is definitely outside of the scope of this PR.

/lgtm

dulek:
> /lgtm
> 
> @dulek given you actually create additional commit for other purpose (e.g compare the LB name) ,can you help put that into [#2360 (comment)](https://github.com/kubernetes/cloud-provider-openstack/pull/2360#issue-1884352337) as well? Thanks

Done!

pkg/openstack/loadbalancer.go
dulek: It totally is and you'll probably be happy to see it disabled by default in OpenShift: openshift/cluster-cloud-controller-manager-operator#263.

dulek:
@jichenjc: Anything else to move this forward?

jichenjc:
no, all good ~

/approve

dulek:
/cherry-pick release-1.28

For some reason this haven't made it to 1.28 branch, which is pretty bad. Trying to fix that now.

PR #2362, opened 2023-09-17
[occm] Add octavia listener xForwardedProto option

sungwon-ahn:
Octavia provides the insert_header option as shown below. However, in OCCM, only X-Forwarded-For is available. As a result, we have implemented code to enable you to utilize this functionality in relation to XFP.

(You can find more information about this in the [documentation](https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=create-listener-detail#header-insertions))

fixes #

**Release note**:

```release-note
[openstack-cloud-controller-manager]  The x-forwarded-protocol of the Octavia listener is supported.
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -64,6 +64,7 @@ const (
 	activeStatus                        = "ACTIVE"
 	errorStatus                         = "ERROR"
 	annotationXForwardedFor             = "X-Forwarded-For"
+	annotationXForwardedProtocol        = "X-Forwarded-Proto"
 
 	ServiceAnnotationLoadBalancerInternal             = "service.beta.kubernetes.io/openstack-internal-load-balancer"
 	ServiceAnnotationLoadBalancerConnLimit            = "loadbalancer.openstack.org/connection-limit"
@@ -83,6 +84,7 @@ const (
 	ServiceAnnotationLoadBalancerTimeoutMemberData    = "loadbalancer.openstack.org/timeout-member-data"
 	ServiceAnnotationLoadBalancerTimeoutTCPInspect    = "loadbalancer.openstack.org/timeout-tcp-inspect"
 	ServiceAnnotationLoadBalancerXForwardedFor        = "loadbalancer.openstack.org/x-forwarded-for"
+	ServiceAnnotationLoadBalancerXForwardedProto      = "loadbalancer.openstack.org/x-forwarded-proto"
 	ServiceAnnotationLoadBalancerFlavorID             = "loadbalancer.openstack.org/flavor-id"
 	ServiceAnnotationLoadBalancerAvailabilityZone     = "loadbalancer.openstack.org/availability-zone"
 	// ServiceAnnotationLoadBalancerEnableHealthMonitor defines whether to create health monitor for the load balancer
@@ -333,6 +335,7 @@ type serviceConfig struct {
 	lbPublicNetworkID       string
 	lbPublicSubnetSpec      *floatingSubnetSpec
 	keepClientIP            bool
+	keepClientProtocol      bool
 	enableProxyProtocol     bool
 	timeoutClientData       int
 	timeoutMemberConnect    int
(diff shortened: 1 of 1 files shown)


mdbooth:
Can you describe your full end-to-end use case for this feature?

pkg/openstack/loadbalancer.go
```diff
@@ -1322,6 +1327,19 @@ func (lbaas *LbaasV2) ensureOctaviaListener(lbID string, name string, curListene
...
+			if svcConf.keepClientProtocol {
+				if *updateOpts.InsertHeaders == nil {
+					*updateOpts.InsertHeaders = make(map[string]string)
+				}
+				(*updateOpts.InsertHeaders)[annotationXForwardedProtocol] = "true"
+			} else {
+				delete(*updateOpts.InsertHeaders, annotationXForwardedProtocol)
+			}
+			listenerChanged = true
+		}
```
dulek, line 1342: This completely duplicates the code above, please extract it to a function so that we can avoid repeating it.

pkg/openstack/loadbalancer.go
```diff
@@ -1399,6 +1420,9 @@ func (lbaas *LbaasV2) buildListenerCreateOpt(port corev1.ServicePort, svcConf *s
 	} else if svcConf.keepClientIP && listenerCreateOpt.Protocol != listeners.ProtocolHTTP {
 		klog.V(4).Infof("Forcing to use %q protocol for listener because %q annotation is set", listeners.ProtocolHTTP, ServiceAnnotationLoadBalancerXForwardedFor)
 		listenerCreateOpt.Protocol = listeners.ProtocolHTTP
+	} else if svcConf.keepClientProtocol && listenerCreateOpt.Protocol != listeners.ProtocolHTTP {
+		klog.V(4).Infof("Forcing to use %q protocol for listener because %q annotation is set", listeners.ProtocolHTTP, ServiceAnnotationLoadBalancerXForwardedProto)
+		listenerCreateOpt.Protocol = listeners.ProtocolHTTP
 	}
```
dulek, line 1426: This if does exactly the same thing as the previous one. You should join them.

pkg/openstack/loadbalancer.go
```diff
@@ -438,6 +441,8 @@ func getListenerProtocol(protocol corev1.Protocol, svcConf *serviceConfig) liste
 			return listeners.ProtocolTerminatedHTTPS
 		} else if svcConf.keepClientIP {
 			return listeners.ProtocolHTTP
+		} else if svcConf.keepClientProtocol {
+			return listeners.ProtocolHTTP
```
dulek, line 445: This if does exactly the same thing as the previous one. You should join them.

dulek (changes requested):

dulek:
There are more header insertions supported by Octavia. I think instead of implementing them one-by-one using annotations we may want to aim at creating a single annotation that will take a list of them and then OCCM will read that and populate InsertHeaders accordingly.

sungwon-ahn:
> Can you describe your full end-to-end use case for this feature?

Hi mdbooth! thx you for asking

I think it is necessary in the SSL offloading use case.

Octavia(HTTPS) < - > Nginx (HTTP)  <-> Pod (HTTP)
The k8s external type load balancer uses Octavia to use TERMINED_HTTP. In this case, the API server in the pod generates a response to the request in http.

mdbooth:
> > Can you describe your full end-to-end use case for this feature?
> 
> Hi mdbooth! thx you for asking
> 
> I think it is necessary in the SSL offloading use case.
> 
> Octavia(HTTPS) < - > Nginx (HTTP) <-> Pod (HTTP)
> 
> You can enable the HTTPS protocol when using a load balancer of the load balancer type in Octavia.
> 
> In this use case, HTTPS requests passing through Octavia must be translated to HTTP and forwarded to the Pod (backend API server).
> 
> You can identify the protocol being used in this request by checking the 'x-forwarded-protocol' header.

Would you be able to use the Octavia ingress controller instead for this use case? I worry in general about continuing to add annotations to service objects.

PR #2367, opened 2023-09-19
[occm] Make sure we don't mask LB tests failures and fix what was failing

dulek:
In `test-lb-service.sh` we do `trap "delete_resources" EXIT` to make sure we cleanup resources on a test failure. In there, we only fetched the `$?` after making a check for `${AUTO_CLEAN_UP}`, which itself alters the code to 0, so function always returns success. This means tests can never really fail.

This commit fixes it by making sure `$ERROR_CODE` is fetched at the very beginning of the cleanup function.

Some additional fixes needed to be made to make tests passing again. In particular:
* Tests no longer create Services as internal in order to make shared LB tests work after #2190.
* FIP is added to the user LB in the sharing test in order for CPO to not complain about not being able to create it for a secondary Service.

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/e2e/cloudprovider/test-lb-service.sh
+++ b/tests/e2e/cloudprovider/test-lb-service.sh
@@ -20,11 +20,11 @@ LB_SUBNET_NAME=${LB_SUBNET_NAME:-"private-subnet"}
 AUTO_CLEAN_UP=${AUTO_CLEAN_UP:-"true"}
 
 function delete_resources() {
+  ERROR_CODE="$?"
+
   if [[ ${AUTO_CLEAN_UP} != "true" ]]; then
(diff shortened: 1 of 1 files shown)


dulek:
This is a backport of testing fixes into release-1.27. I'll need to add it to earlier releases too.

dulek:
I forced-rebased this, let's see.

dulek:
Yup, looks good now!

dulek:
@jichenjc, @kayrus, @zetaab, care for a `/lgtm` here?

dulek:
@jichenjc, @kayrus, @zetaab, care for a /lgtm here?

zetaab (approved):
/lgtm

PR #2368, opened 2023-09-25
fix(occm/loadbalancer): enable the proxy-protocol only for supported listeners

Lucasgranet:
This PR fix an issue when trying to activate the ProxyProtocol on a K8S LoadBalancer with, at least, one UDP or SCTP port.

The CCM try, when the proxy-protocol's annotation is set, to enable the proxy-protocols on all pools, whatever the protocol used by the listener.

Octavia doesn't support  the proxy protocol with a UDP or SCTP listener: <https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=create-pool-detail#protocol-combinations-listener-pool>

N/A

ATM, this service __cannot__ be deployed. This kind of service will be more common due to "new" protocol HTTP/3/QUIC (TCP and UDP usage on the same service).
```yaml
apiVersion: v1
kind: Service
metadata:
  name: octavia-udp-tcp-with-proxyprotocol
  annotations:
    loadbalancer.openstack.org/proxy-protocol: "true"
spec:
  ports:
  - name: client
    port: 443
    protocol: TCP
    targetPort: 443
  - name: client-udp
    port: 443
    protocol: UDP
    targetPort: 443
  selector:
    app: nginx
  type: LoadBalancer
```
Without this patch, this will result to a 409 from the Octavia's API.

**Release note**:

```release-note
the proxy-protocol is ignored when the listener protocol is not supported (UDP/SCTP)
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	expslices "golang.org/x/exp/slices"
 	"reflect"
 	"regexp"
 	"strconv"
@@ -129,7 +130,11 @@ type servicePatcher struct {
 	updated *corev1.Service
 }
 
-var _ cloudprovider.LoadBalancer = &LbaasV2{}
+var (
+	// List based on Octavia documentation https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=create-pool-detail#protocol-combinations-listener-pool
+	proxyProtocolSupportedListeners                            = []listeners.Protocol{listeners.ProtocolHTTP, listeners.ProtocolHTTPS, listeners.ProtocolTCP, listeners.ProtocolTerminatedHTTPS}
+	_                               cloudprovider.LoadBalancer = &LbaasV2{}
+)
 
 // negate returns a negated matches for a given one
 func negate(f matcher) matcher { return func(s *subnets.Subnet) bool { return !f(s) } }
@@ -1133,7 +1138,7 @@ func (lbaas *LbaasV2) ensureOctaviaPool(lbID string, name string, listener *list
 
 	// By default, use the protocol of the listener
 	poolProto := v2pools.Protocol(listener.Protocol)
-	if svcConf.enableProxyProtocol {
+	if svcConf.enableProxyProtocol && isProtocolProxyProtocolSupported(listener.Protocol) {
 		poolProto = v2pools.ProtocolPROXY
 	} else if (svcConf.keepClientIP || svcConf.tlsContainerRef != "") && poolProto != v2pools.ProtocolHTTP {
 		poolProto = v2pools.ProtocolHTTP
@@ -1201,7 +1206,7 @@ func (lbaas *LbaasV2) ensureOctaviaPool(lbID string, name string, listener *list
 func (lbaas *LbaasV2) buildPoolCreateOpt(listenerProtocol string, service *corev1.Service, svcConf *serviceConfig) v2pools.CreateOpts {
 	// By default, use the protocol of the listener
 	poolProto := v2pools.Protocol(listenerProtocol)
-	if svcConf.enableProxyProtocol {
+	if svcConf.enableProxyProtocol && isProtocolProxyProtocolSupported(listenerProtocol) {
 		poolProto = v2pools.ProtocolPROXY
 	} else if (svcConf.keepClientIP || svcConf.tlsContainerRef != "") && poolProto != v2pools.ProtocolHTTP {
 		if svcConf.keepClientIP && svcConf.tlsContainerRef != "" {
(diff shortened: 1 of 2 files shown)


kayrus:
@Lucasgranet for me this fix looks like a dirty hack. How does QUIC server behave when TCP does contain a proxy protocol header, but UDP doesn't?

Lucasgranet:
Hello @kayrus,

With the current version of Octavia, the QUIC protocol cannot be enable with the ProxyProtocol indeed. (because he relies on UDP).

Whatever is the protocol, Octavia, atm, doesn't support the ProxyProtocol with the UDP (and SCTP) listeners.
You can find [here](https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=create-pool-detail#protocol-combinations-listener-pool) a compatibility matrix for ProxyProtocol support.

This PR lead to fix LoadBalancer Service when:
- the ProxyProtocol annotation is set
- An unsupported protocol is used on a port (UDP or SCTP)

With the current upstream code, the CCM will get this error from the Octavia API:
```json
{
    "faultcode": "Client", 
    "faultstring": "Validation failure: The pool protocol 'PROXY' is invalid while the listener protocol is 'UDP'.", 
    "debuginfo": null
}
```

Lucas,

pkg/openstack/loadbalancer.go
```diff
@@ -129,7 +130,11 @@ type servicePatcher struct {
 	updated *corev1.Service
 }
 
-var _ cloudprovider.LoadBalancer = &LbaasV2{}
+var (
+	// List based on Octavia documentation https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=create-pool-detail#protocol-combinations-listener-pool
+	proxyProtocolSupportedListeners                            = []listeners.Protocol{listeners.ProtocolHTTP, listeners.ProtocolHTTPS, listeners.ProtocolTCP, listeners.ProtocolTerminatedHTTPS}
+	_                               cloudprovider.LoadBalancer = &LbaasV2{}
```
kayrus, line 136: what do you want to achieve with this string?

pkg/openstack/loadbalancer.go
```diff
@@ -129,7 +130,11 @@ type servicePatcher struct {
 	updated *corev1.Service
 }
 
-var _ cloudprovider.LoadBalancer = &LbaasV2{}
+var (
+	// List based on Octavia documentation https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=create-pool-detail#protocol-combinations-listener-pool
+	proxyProtocolSupportedListeners                            = []listeners.Protocol{listeners.ProtocolHTTP, listeners.ProtocolHTTPS, listeners.ProtocolTCP, listeners.ProtocolTerminatedHTTPS}
```
kayrus, line 135: it's better to define each supported protocols on a new line. further list updates will look more readable.

pkg/openstack/loadbalancer.go
```diff
@@ -2604,6 +2609,14 @@ func GetLoadBalancerSourceRanges(service *corev1.Service, preferredIPFamily core
 	return ipnets, nil
 }
 
+// isProtocolProxyProtocolSupported checks if the ListenProtocol can be used with the ProxyProtocol
+// for the pool configuration. In the case of UDP listener, Octavia doesn't support the ProxyProtocol
+// and should not be enabled to avoid 403 from Octavia API.
+// https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=create-pool-detail#protocol-combinations-listener-pool
+func isProtocolProxyProtocolSupported(protocol string) bool {
```
kayrus, line 2616: we already have https://github.com/kubernetes/cloud-provider-openstack/blob/5b36a539124160a070a98e1a23eee9357fc07f2b/pkg/util/util.go#L65-L73 which can be converted into:

```
func Contains[T comparable](list []T, strToSearch T) bool {
        for _, item := range list {
                if item == strToSearch {
                        return true
                }
        }
        return false
}
```

dulek:
I think I understand the idea here. However instead of relying on a static list of supported protocols I think I'd prefer to catch the creation error and ignore it if message indicates that the problem is with the combination. Moreover we should only mask the error if the `Service` has other ports. If not, we should fail creation of the LB for that service and inform the user about this.

kayrus:
> catch the creation error and ignore it if message indicates that the problem is with the combination

is there a way to view these warnings using `kubectl describe svc`? I'm asking because such events must be visible for end users, who don't always have an access to OCCM logs.

Lucasgranet:
> I think I understand the idea here. However instead of relying on a static list of supported protocols I think I'd prefer to catch the creation error and ignore it if message indicates that the problem is with the combination. 

Indeed, this behavior seems better, especially if Octavia will support this combination one day.

> Moreover we should only mask the error if the `Service` has other ports. If not, we should fail creation of the LB for that service and inform the user about this.

I'm agree with this too. We should not allow the ProxyProtocol's annotation set to `true` when **all** used protocols are not compatible.

> is there a way to view these warnings using `kubectl describe svc`?

An event is returned to the customer.

```console
  Warning  SyncLoadBalancerFailed  18s (x5 over 18m)   service-controller  Error syncing load balancer: failed to ensure load balancer: Bad request with: [POST https://load-balancer.gra9.cloud.ovh.net/v2.0/lbaas/pools], error message: {"faultcode": "Client", "faultstring": "Validation failure: The pool protocol 'PROXY' is invalid while the listener protocol is 'UDP'.", "debuginfo": null}
```

dulek:
> > catch the creation error and ignore it if message indicates that the problem is with the combination
> 
> is there a way to view these warnings using `kubectl describe svc`? I'm asking because such events must be visible for end users, who don't always have an access to OCCM logs.

Fair point, only errors are logged as Events and I was proposing to *not* error. I'll try to draft some interface that would allow us to create events, based on [how GCP provider does this](https://github.com/kubernetes/cloud-provider-gcp/blob/b55897dfbbc4abf4ea7709eec56ea1f4ef30313c/providers/gce/gce_loadbalancer.go#L155).

dulek:
Please see #2383. It's yet to be tested, but I don't see why it wouldn't work.

dulek:
@Lucasgranet https://github.com/kubernetes/cloud-provider-openstack/pull/2383 is just merged, so great timing. ;)

dulek:
I rebased this on #2549. Closing this one.

PR #2369, opened 2023-09-26
[cinder-csi-plugin] Add finalizer to VolumeSnapshot when create PVC with the VolumeSnapshot

### LINKED ISSUE #2294: [cinder-csi-plugin:bug] Volume Snapshot deletion is not detecting errors
> /kind bug

**What happened**:
I am using Delete policy for both PV class and VS(VolumeSnapshot) class.
We use TrilioVault for Kubernetes for K8s backup tool and run [TVK preflight plugin](https://github.com/trilioData/tvk-plugins/blob/main/docs/preflight/README.md) as a part of preflight check of the cluster provisioning.
This TVK preflight check plugin creates a PVC and a VS from this source PVC, then create a restore PVC from the snapshot.
Once all checks are finished, it deletes all test resources.

In openstack , if a volume was created from a volume snapshot, this snapshot cannot be deleted before its child volume is deleted.
So the first attempt to delete VS is failed because it takes a bit time to delete the restored volume. Ofc, the source volume deletion is also failed because it can be deleted once the VS and restored volume are deleted.

But openstack snapshot api is async and it means it responds 202 (return value 0) to DELETE snapshot request every time even if the deletion failed. i.e. this `cs.Cloud.DeleteSnapshot` func will never fail in our scenario and k8s VS and VSC(VolumeSnapshotContent) objects are deleted without any issue and it will not be requeued even though openstack resources are there.
https://github.com/kubernetes/cloud-provider-openstack/blob/9ed6d961c6ee5a4f51533877ae981aa6d9753f2d/pkg/csi/cinder/controllerserver.go#L415

It results garbage resources(volumes of the source PVC, and volumeSnapshots of the VS) on Openstack cloud.

I think this could be considered as Openstack API issue more or less. But i think we can make a workaround at this plugin level.
So csi plugin will check the volume snapshot status after requesting the deletion. If the volume snapshot is removed, then it will return OK. If the volume snapshot status is changed to ERROR or its status remains Available (we can set timeout for this check), it will return Error so it can be requeued.

**How to reproduce it**:
It can be reproduced by using these example resources https://github.com/kubernetes/cloud-provider-openstack/tree/master/examples/cinder-csi-plugin/snapshot
- 1. Create a PVC.
- 2. Create a VS from it.
- 3. Create a PVC from VS
- 4. Delete VS first.

- Volume deletion is requeued if it is failed correctly because Openstack cinder api returns fail as sync mode.
Only volume snapshot deletion is the problem.
- The VS object on k8s has been deleted but Openstack volume snapshot exists. So the original PV cannot be deleted. So this issue produces garbage volumes as well as gargage volume snapshot.

**Environment**:
We just use cinder-csi-plugin. The version list is
csiplugin: registry.k8s.io/provider-os/cinder-csi-plugin:v1.25.6
snapshotter: k8s.gcr.io/sig-storage/csi-snapshotter:v6.2.1
provisioner: k8s.gcr.io/sig-storage/csi-provisioner:v3.1.1
atttacher: k8s.gcr.io/sig-storage/csi-attacher:v3.5.1

- OpenStack version: ocata, wallaby (i think all version, even antelope will has the same issue)
- Others:
K8s: 1.25.9

jeffyjf:

This PR implement a PVC watcher. When the watcher receive a PVC ADDED event and PVC with VolumeSnapshot kind dataSource, the VolumeSnapshot will be added a finalizer. And the finalizer will be removed when the watcher receive the PVC's DELETED event. It can used to protect VolumeSnapshot, avoid that the VS be deleted while there are any PersistentVolume on it.

fixes #2294

Asynchronously add finalizer by watcher seems like unsafety, might result in some potential issues. After the PR [1] merged and released, we need to refactor it and implement a synchronous way to add finalizer.

[1] https://github.com/kubernetes-csi/external-provisioner/pull/1070

**Release note**:

```release-note
Add  a command flag `--enable-vs-deletion-protection`. If set, The VS will be added a finalizer while a PVC created base on it, ensure the VS couldn't be deleted before the PVC be deleted.
```


mdbooth (commented):
As mentioned in the issue, I do like the concept of adding a finalizer to the snapshot for each volume which requires it. However, I can't think of another instance where finalizers are used this way, so it might be worth some effort to discover if there are any problems with potentially very big lists of finalizers. Imagine, for example, a some CI system which uses a 'golden image' extremely heavily and has 1,000 volumes using it. Would that be ok?

However, I don't think we can safely add the finalizer asynchronously. A finalizer would typically be added synchronously *before* resource creation. It's also worth noting that the resource in this case would be the PV rather than the PVC.

Could we add code which adds *and persists* the snapshot finalizer immediately before creating the PV?

We would also need to add code to remove the finalizer in the deletion flow of the PV *strictly before* removing the PV finalizer.

mdbooth:
Looks very WIP right now so probably not worth actually running tests, but

/ok-to-test

Note that tests won't run automatically while the PR is marked WIP, but you can run them manually on demand when you're ready with `/test all`.

pkg/csi/cinder/controllerserver.go
```diff
@@ -653,6 +661,62 @@ func (cs *controllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 	}, nil
 }
 
+func (cs *controllerServer) StartWathcer() error {
```
kayrus, line 664:
```suggestion
func (cs *controllerServer) StartWatcher() error {
```

pkg/csi/cinder/controllerserver.go
```diff
@@ -653,6 +661,62 @@ func (cs *controllerServer) ControllerExpandVolume(ctx context.Context, req *csi
...
+	if err != nil {
+		klog.ErrorS(err, "Error building watcher")
+		return err
+	}
+	watchInterface, err := watcher.Watch(context.TODO(), &corev1.PersistentVolumeClaimList{})
+	if err != nil {
+		klog.ErrorS(err, "Error watching PersistentVolumeClaim")
+		return err
+	}
+	go func() {
```
kayrus, line 680: Can we define this function statically and call it here? e.g. `go watcher(watchInterface)`?

jeffyjf:
Hi @mdbooth 

> However, I don't think we can safely add the finalizer asynchronously. A finalizer would typically be added synchronously _before_ resource creation. It's also worth noting that the resource in this case would be the PV rather than the PVC.

I am a newbie. In my opinion, this cannot be implemented just by watcher. We need to add a webhook to do this. right? If it's not right, do you have any other suggestions?

jeffyjf:
> As mentioned in the issue, I do like the concept of adding a finalizer to the snapshot for each volume which requires it. However, I can't think of another instance where finalizers are used this way, so it might be worth some effort to discover if there are any problems with potentially very big lists of finalizers. Imagine, for example, a some CI system which uses a 'golden image' extremely heavily and has 1,000 volumes using it. Would that be ok?
> 
> However, I don't think we can safely add the finalizer asynchronously. A finalizer would typically be added synchronously _before_ resource creation. It's also worth noting that the resource in this case would be the PV rather than the PVC.
> 
> Could we add code which adds _and persists_ the snapshot finalizer immediately before creating the PV?
> 
> We would also need to add code to remove the finalizer in the deletion flow of the PV _strictly before_ removing the PV finalizer.

In the past few days, I tried a few methods to implement add finalizer before creating the PV synchronously:
1. I tried implement this in [CreateVolume](https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/controllerserver.go#L47) function, but I found I can't get enough context informations by graceful way. Such as, the k8s snapshot's name and namespace.
2. I tried add a webhook to intercept the PV's creation, this can add finalizer before creating the PV. But, this will increase the difficulty of deployment and maintenance. Moreover, If a cluster have multiple webhook about PV and another webhook reject the PV's creation after this webhook, I don't konw how to resolve this situation. So, I don't like this way.

I think the first method is more graceful and appropriate. But, the first methon require [external-provisioner sidecar](https://kubernetes-csi.github.io/docs/external-provisioner.html) provide more necessary informations. So I think we should submit proposal to external-provisioner firstly in order to we can resolve this issue appropriately.

jeffyjf:
> > As mentioned in the issue, I do like the concept of adding a finalizer to the snapshot for each volume which requires it. However, I can't think of another instance where finalizers are used this way, so it might be worth some effort to discover if there are any problems with potentially very big lists of finalizers. Imagine, for example, a some CI system which uses a 'golden image' extremely heavily and has 1,000 volumes using it. Would that be ok?
> > However, I don't think we can safely add the finalizer asynchronously. A finalizer would typically be added synchronously _before_ resource creation. It's also worth noting that the resource in this case would be the PV rather than the PVC.
> > Could we add code which adds _and persists_ the snapshot finalizer immediately before creating the PV?
> > We would also need to add code to remove the finalizer in the deletion flow of the PV _strictly before_ removing the PV finalizer.
> 
> In the past few days, I tried a few methods to implement add finalizer before creating the PV synchronously:
> 
> 1. I tried implement this in [CreateVolume](https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/controllerserver.go#L47) function, but I found I can't get enough context informations by graceful way. Such as, I can't obtain the k8s VolumeSnapshot's name and namespace, this result in I can't add finalizer  to it.
> 2. I tried add a webhook to intercept the PV's creation, this can add finalizer before creating the PV. But, this will increase the difficulty of deployment and maintenance. Moreover, If a cluster have multiple webhook about PV and another webhook reject the PV's creation after this webhook, I don't konw how to resolve this situation. So, I don't like this way.
> 
> I think the NO.1 method is more graceful and appropriate. But, the first methon require [external-provisioner sidecar](https://kubernetes-csi.github.io/docs/external-provisioner.html) provide more necessary informations. So I think we should submit a KEP to CSI spec in order to we can get enough necessary informations about VolumeSnatshot in CreateVolume function to resolve this issue appropriately.

I've already commit a [PR](https://github.com/kubernetes-csi/external-provisioner/pull/1070) to [external-provisioner](https://github.com/kubernetes-csi/external-provisioner/pull/1070), in order to we can add finalizer to VolumeSnapshot synchronously in future. But now, let's continue to complete this imperfect solution to fix the issue #2294 and hold back resource leak in production environment  asap.

dulek:
> Hi @mdbooth, thanks very much for your comments. But I have a bit question.
> 
> > However, I don't think we can safely add the finalizer asynchronously. A finalizer would typically be added synchronously _before_ resource creation. It's also worth noting that the resource in this case would be the PV rather than the PVC.
> 
> I am a newbie. In my opinion, this cannot be implemented just by watcher. We need to add a webhook to do this. right? If it's not right, do you have any other suggestions?

Hey, what is the problem with adding a finalizer asynchronously? We've did that all the time in Kuryr and it worked totally well when we made sure we're doing it by patching the resource properly.

jeffyjf:
Hi CPO Co-Leads @dulek @jichenjc @kayrus @zetaab 

This PR  is ready for review. Please take a look this PR in your spare time. I'm eager to get your advices.

jichenjc:
>>>Asynchronously add finalizer by watcher seems like unsafety, might result in some potential issues. After the PR [1] merged and released, we need to refactor it and implement a synchronous way to add finalizer.

thanks for the PR, I thought our PR rely on https://github.com/kubernetes-csi/external-provisioner/pull/1070
but read again seems you are proposing a long term solution in external-provisioner then after that's done
we will do another round of fix to current one ? so this PR is a temp solution as stop gap ?

jeffyjf:
> > > > Asynchronously add finalizer by watcher seems like unsafety, might result in some potential issues. After the PR [1] merged and released, we need to refactor it and implement a synchronous way to add finalizer.
> 
> thanks for the PR, I thought our PR rely on [kubernetes-csi/external-provisioner#1070](https://github.com/kubernetes-csi/external-provisioner/pull/1070) but read again seems you are proposing a long term solution in external-provisioner then after that's done we will do another round of fix to current one ? so this PR is a temp solution as stop gap ?

Yep, I am not sure when the [kubernetes-csi/external-provisioner#1070](https://github.com/kubernetes-csi/external-provisioner/pull/1070) can be merged and released. This PR can be temporarily used to hold back resources leak.

jeffyjf:
> > Hi @mdbooth, thanks very much for your comments. But I have a bit question.
> > > However, I don't think we can safely add the finalizer asynchronously. A finalizer would typically be added synchronously _before_ resource creation. It's also worth noting that the resource in this case would be the PV rather than the PVC.
> > 
> > 
> > I am a newbie. In my opinion, this cannot be implemented just by watcher. We need to add a webhook to do this. right? If it's not right, do you have any other suggestions?
> 
> Hey, what is the problem with adding a finalizer asynchronously? We've did that all the time in Kuryr and it worked totally well when we made sure we're doing it by patching the resource properly.

There may just be some potential problems, but I have no concrete example. Maybe in an extremely busy environment, the resource may be deleted before the finalizer be successfully added.

kayrus:
@jeffyjf can you rebase on latest master?

jeffyjf:
> @jeffyjf can you rebase on latest master?

Of course I will rebase it late on.

kayrus:
@jeffyjf can you fix unit tests?

jeffyjf:
> @jeffyjf can you fix unit tests?

Of course, I'll fix it asap.

pkg/csi/cinder/watcher.go
```diff
@@ -0,0 +1,106 @@
...
+
+	snap "github.com/kubernetes-csi/external-snapshotter/client/v6/clientset/versioned"
+	"golang.org/x/net/context"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
```
dulek, line 30: Looks like this is mixing K8s and non-K8s imports. No biggie.

pkg/csi/cinder/watcher.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+		update := false
+		finalizer := fmt.Sprintf("%s/pvc-%s", driverName, pvc.UID)
+		// TODO(JeffYang): Asynchronously add finalizer seems like unsafety, might have some potential issues.
+		// Move it into controllerServer CreateVolume after the PR https://github.com/kubernetes-csi/external-provisioner/pull/1070 merged and released.
+		// We can synchronously add finalizer there
+		if event.Type == watch.Added {
+			klog.V(5).InfoS("Receive ADDED event try to add finalizer to VolumeSnapshot", "PersistentVolumeClaim", pvc.Name, "VolumeSnapshot", snapshotObj.Name)
+			update = controllerutil.AddFinalizer(snapshotObj, finalizer)
+		}
```
dulek, line 70: Uhm, what if we missed an ADDED event somehow, but then get the MODIFIED one? I believe we should rather look for ADDED/MODIFIED and check if finalizer is there and if it's not add it.

pkg/csi/cinder/watcher.go
```diff
@@ -0,0 +1,106 @@
...
+		if event.Type == watch.Deleted {
+			klog.V(5).InfoS("Receive DELETED event try to remove finalizer from VolumeSnapshot", "PersistentVolumeClaim", pvc.Name, "VolumeSnapshot", snapshotObj.Name)
+			update = controllerutil.RemoveFinalizer(snapshotObj, finalizer)
+		}
```
dulek, line 74: What if you've missed a deletion event here? Are we stuck with a snapshot?

pkg/csi/cinder/watcher.go
```diff
@@ -0,0 +1,106 @@
...
+		if update {
+			_, err := cw.snapClient.SnapshotV1().VolumeSnapshots(pvc.Namespace).Update(context.Background(), snapshotObj, metav1.UpdateOptions{})
```
dulek, line 76: This should be a `Patch`, otherwise we might end up replacing changes made by some other controller.

dulek (changes requested):
This feels like it should be a proper K8s controller and not just a simplified version of a watcher.

cmd/cinder-csi-plugin/main.go
```diff
@@ -65,6 +69,7 @@ func main() {
 
 	cmd.PersistentFlags().StringVar(&cluster, "cluster", "", "The identifier of the cluster that the plugin is running in.")
 	cmd.PersistentFlags().StringVar(&httpEndpoint, "http-endpoint", "", "The TCP network address where the HTTP server for providing metrics for diagnostics, will listen (example: `:8080`). The default is empty string, which means the server is disabled.")
+	cmd.PersistentFlags().BoolVar(&vsDeletionProtectionEnabled, "enable-vs-deletion-protection", false, "If set, The VS will be added a finalizer while a PVC created base on it, ensure the VS couldn't be deleted before the PVC be deleted.")
```
jichenjc, line 72: `VS` is not widely used I think you mean virtual server ?

pkg/csi/cinder/watcher.go
```diff
@@ -0,0 +1,106 @@
...
+	annStorageProvisioner = "volume.kubernetes.io/storage-provisioner"
+	snapshotKind          = "VolumeSnapshot"
+)
+
+type controllerWatcher struct {
+	snapClient snap.Interface
+	watcher    watch.Interface
+}
+
+func (cw *controllerWatcher) run() {
```
jichenjc, line 43: I am not sure but this to me is general feature instead of CPO ? e.g every cloud has this problem?
should this be part of snapshot general thing e.g external snapshot controller to handle ?

PR #2383, opened 2023-10-03
[occm] Add ability to add Events by OCCM

dulek:
We often find situations when it's useful to use events to warn user about implicit choices that the OCCM makes, e.g. when a provided option is ignored because Octavia does not support it.

This commit adds `eventRecorder` to the `LbaasV2` interface that can be used to create events in these cases. In addition a few places when we just logged a warning are now extended to create events on Services that users will be able to read.

fixes #

**Release note**:

```release-note
[openstack-cloud-controller-manager] OCCM will now emit events on Services regarding various implicit decisions it makes.
```

### DIFF
--- a/pkg/openstack/events.go
+++ b/pkg/openstack/events.go
@@ -1,0 +1,24 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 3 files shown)


dulek:
Alright, this is now tested and works fine. We had an issue in OpenShift but it turned out to be on our side.

@zetaab, @jichenjc, @kayrus, can you take a look?

zetaab (approved):
/approve

jichenjc:
this is great and helpful ,we may extend this to other places in the CPO later

/hold cancel

PR #2386, opened 2023-10-04
test : getSecurityGroupName [load balancer]

KingDaemonX:

fixes #

**Release note**:

```release-note
NONE
```
implemented a test case for the getSecurity group name to check if the function is implemented correctly and also to check the length of name generated if input is too long

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -416,7 +416,7 @@ func popListener(existingListeners []listeners.Listener, id string) []listeners.
 
 func getSecurityGroupName(service *corev1.Service) string {
 	securityGroupName := fmt.Sprintf("lb-sg-%s-%s-%s", service.UID, service.Namespace, service.Name)
-	//OpenStack requires that the name of a security group is shorter than 255 bytes.
+	// OpenStack requires that the name of a security group is shorter than 255 bytes.
 	if len(securityGroupName) > 255 {
 		securityGroupName = securityGroupName[:255]
 	}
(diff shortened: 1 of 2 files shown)


pkg/openstack/loadbalancer.go
```diff
@@ -416,7 +416,7 @@ func popListener(existingListeners []listeners.Listener, id string) []listeners.
 
 func getSecurityGroupName(service *corev1.Service) string {
 	securityGroupName := fmt.Sprintf("lb-sg-%s-%s-%s", service.UID, service.Namespace, service.Name)
-	//OpenStack requires that the name of a security group is shorter than 255 bytes.
+	// OpenStack requires that the name of a security group is shorter than 255 bytes.
```
dulek, line 419: Looks like your IDE is making some automatic formatting changes that aren't related to this PR. It's better to disable that and not mix unrelated changes into PRs as it makes them harder to be reviewed. Please make sure only the test is added into the PR.

pkg/openstack/loadbalancer_test.go
```diff
@@ -1,13 +1,15 @@
...
 import (
+	"fmt"
 	"sort"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
-
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/listeners"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/rules"
+	"github.com/stretchr/testify/assert"
```
dulek, line 10: Feels like we tried to keep this one separated for a reason, I guess because it's a testing-only import? No big deal here though.

dulek (changes requested):
Thanks for the PR, it looks good! I've proposed a bunch of improvements in the inline comments.

pkg/openstack/loadbalancer.go
KingDaemonX: okay I will fix this immiediately

pkg/openstack/loadbalancer_test.go
```diff
@@ -460,3 +462,45 @@ func TestGetRulesToCreateAndDelete(t *testing.T) {
...
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			expectedSecurityGroupName := fmt.Sprintf("lb-sg-%s-%s-%s", test.service.UID, test.service.Namespace, test.service.Name)
+
+			if len(expectedSecurityGroupName) > 255 {
+				expectedSecurityGroupName = expectedSecurityGroupName[:255]
+			}
```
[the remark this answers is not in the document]
KingDaemonX: to maintain the master integrity on my end , i will be making the changes in a new branch, if that is fine with you 
i will drop the link to the new PR if it needed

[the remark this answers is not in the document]
KingDaemonX: here is the new and improved version of this on a different branch 

https://github.com/kubernetes/cloud-provider-openstack/pull/2389

PR #2390, opened 2023-10-04
test : GetLoadBalancerName method

KingDaemonX:

fixes #

**Release note**:

```release-note
NONE
```
implement the test to GetLoadBalanceName method and test all logical part of the code including edge cases

i found out the if all the input are empty, it returns `kube_service___` ; i think this should be looked into since it failed during test

// i think there should be a check for "kube_service___ and an error should be returned if the important input are empty"

### DIFF
--- a/pkg/openstack/loadbalancer_test.go
+++ b/pkg/openstack/loadbalancer_test.go
@@ -1,6 +1,7 @@
 package openstack
 
 import (
+	"context"
 	"reflect"
 	"sort"
 	"testing"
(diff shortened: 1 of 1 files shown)


pkg/openstack/loadbalancer_test.go
```diff
@@ -460,3 +463,64 @@ func TestGetRulesToCreateAndDelete(t *testing.T) {
...
+				ctx:         ctx,
+				clusterName: "a_longer_valid_cluster",
+				service: &corev1.Service{
+					ObjectMeta: v1.ObjectMeta{
+						Namespace: "a_longer_valid_cluster_namespace",
+						Name:      "a_longer_valid_name_for_the_load_balance_name_to_test_if_the_length_of_value_is_longer_than_required_maximum_length_random_addition_hardcode_number_to_make_it_above_length_255_at_the_end_yeah_so_the_rest_is_additional_input",
+					},
+				},
+			},
+			expect: "kube_service_a_longer_valid_cluster_a_longer_valid_cluster_namespace_a_longer_valid_name_for_the_load_balance_name_to_test_if_the_length_of_value_is_longer_than_required_maximum_length_random_addition_hardcode_number_to_make_it_above_length_255_at_the_end",
```
dulek, line 507: Same thing as in SG patch, I bet it should be truncated to 255 characters here.

dulek (commented):
/ok-to-test

pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +537,63 @@ func Test_getListenerProtocol(t *testing.T) {
...
+
+	testArg := &testArgs{
+		ctx:         ctx,
+		clusterName: "my_valid_cluster",
+		service: &corev1.Service{
+			ObjectMeta: v1.ObjectMeta{
+				Namespace: "valid_cluster_namespace",
+				Name:      "valid_name",
+			},
+		},
```
dulek, line 559: It's a minor thing, but Kubernetes will not allow names containing "_": https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +537,63 @@ func Test_getListenerProtocol(t *testing.T) {
...
+	}
+
+	expected := "kube_service_my_valid_cluster_valid_cluster_namespace_valid_name"
+	t.Run("check valid input with short name", func(t *testing.T) {
+		if got := lbaas.GetLoadBalancerName(testArg.ctx, testArg.clusterName, testArg.service); got != expected {
+			t.Errorf("LbaasV2.GetLoadBalancerName() = %v, want %v", got, expected)
+		}
```
dulek, line 566: This should use `assert.Equal()` instead of a manual check.

dulek (changes requested):
Can you do it in a loop with multiple cases like in one of the previous versions of `getSecurityGroupName()` tests?

KingDaemonX:
> Can you do it in a loop with multiple cases like in one of the previous versions of `getSecurityGroupName()` tests?

Okay I will do that

pkg/openstack/loadbalancer_test.go
KingDaemonX: Thank you for the review  :) 
I looked at the url for the naming convention and also switched it to the readable loop test function 

I also used the assert to check for the cases results as specified

pkg/openstack/loadbalancer_test.go
```diff
@@ -593,3 +594,63 @@ func Test_getIntFromServiceAnnotation(t *testing.T) {
...
+		clusterName string
+		service     *corev1.Service
+	}
+	tests := []struct {
+		name     string
+		testArgs testArgs
+		expected string
+	}{
+		{
+			name: "check valid input with short name",
```
pierreprinetti, line 613: nit: these are tests, so there's probably no need to say "check" in the test case name each time :)

KingDaemonX: made all the changes you suggested and push them already
thanks!

pierreprinetti:
Thank you!
/lgtm

PR #2394, opened 2023-10-05
Added Test for buildBatchUpdateMemberOpts Function [loadbalancer_test]

FavourEva:

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer_test.go
+++ b/pkg/openstack/loadbalancer_test.go
@@ -1,13 +1,14 @@
 package openstack
 
 import (
+	"reflect"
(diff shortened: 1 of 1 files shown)


dulek:
/ok-to-test

This is a great start and a nice test of a happy path, but we should have a bunch of more cases here:
1. A case where `ErrNoAddressFound` happens and a member is ignored.
2. Case with a `NodePort` == 0.
3. Case with `healthCheckNodePort` when it's both `canUseHTTPMonitor()=true` and `canUseHTTPMonitor()=false`.

PR #2397, opened 2023-10-06
test-getboolfromserviceannotation [cloud provider openstack]

FavourEva:

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer_test.go
+++ b/pkg/openstack/loadbalancer_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/rules"
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 type testPopListener struct {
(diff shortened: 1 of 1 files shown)


pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +535,149 @@ func Test_getListenerProtocol(t *testing.T) {
...
+			name: "Valid 'true' annotation with default 'false'",
+			service: &corev1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{"annotation7": "true"},
+				},
+			},
+			annotationKey:  "annotation7",
+			defaultSetting: false,
+			expect:         true,
+		},
```
dulek, line 635: This is the same as the `Valid 'true' annotation` case, isn't it?

pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +535,149 @@ func Test_getListenerProtocol(t *testing.T) {
...
+			name: "Valid 'false' annotation with default 'true'",
+			service: &corev1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{"annotation8": "false"},
+				},
+			},
+			annotationKey:  "annotation8",
+			defaultSetting: true,
+			expect:         false,
+		},
```
dulek, line 646: And this one seems the same as `Valid 'false' annotation` if I see correctly?

dulek (changes requested):
Seems like some cases are duplicated if I read this correctly. Can you confirm and remove the duplicates?

dulek:
CI also tells us that code is not formatted with `gofmt`: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/cloud-provider-openstack/2397/pull-cloud-provider-openstack-check/1710300693216104448#1:build-log.txt%3A180

pkg/openstack/loadbalancer_test.go
FavourEva: They cover similar grounds but the default settings differs. I thought of keeping them separate to maintain district test for each combination of annotation and default setting.

I will immediately merge them into single cases that tests different default settings for the same annotation value to reduce redundancy in the test cases

Is this ok?

dulek: `Valid 'true' annotation` has:
```
Annotations: map[string]string{"annotation2": "false"},
defaultSetting: false,
expect:         true,
```

`"Valid 'true' annotation with default 'false'"` has:
Annotations: map[string]string{"annotation7": "true"},
defaultSetting: false,
expect:         true,

The `defaultSetting` is `false` in both. They still feel redundant to me, so yeah, let's make sure we don't have two tests for the exactly the same case. Thank you!

FavourEva: Sorry about. Confirmed and resolved it.
Please help review to confirm if its okay now

PR #2398, opened 2023-10-06
test : getMemberSubnetID

KingDaemonX:

fixes #

**Release note**:

```release-note
NONE
```
implemented the unit test for the complex  `getMemberSubnetID`

test for possible edge cases in the program lbaasv2 method

### DIFF
--- a/pkg/openstack/loadbalancer_test.go
+++ b/pkg/openstack/loadbalancer_test.go
@@ -1543,3 +1543,89 @@ func Test_nodeAddressForLB(t *testing.T) {
 		})
 	}
 }
+
+func TestLbaasV2_getMemberSubnetID(t *testing.T) {
+	lbaasOpts := LoadBalancerOpts{
+		LBClasses: map[string]*LBClass{
+			"lbclassKey": {
+				MemberSubnetID: "lb-class-member-subnet-id-5678",
+			},
+		},
+		MemberSubnetID: "default-memberSubnetId",
+	}
+
+	tests := []struct {
+		name    string
+		opts    LoadBalancerOpts
+		service *corev1.Service
+		want    string
+		wantErr string
+	}{
+		{
+			name: "get member subnet id from service annotation",
+			opts: LoadBalancerOpts{},
+			service: &corev1.Service{
+				ObjectMeta: v1.ObjectMeta{
+					Annotations: map[string]string{
+						ServiceAnnotationLoadBalancerMemberSubnetID: "member-subnet-id",
+						ServiceAnnotationLoadBalancerClass:          "svc-annotation-loadbalance-class",
+					},
+				},
+			},
+			want:    "member-subnet-id",
+			wantErr: "",
+		},
+		{
+			name: "get member subnet id from config class",
+			opts: lbaasOpts,
+			service: &corev1.Service{
+				ObjectMeta: v1.ObjectMeta{
+					Annotations: map[string]string{
+						ServiceAnnotationLoadBalancerClass: "lbclassKey",
+					},
+				},
+			},
+			want:    "lb-class-member-subnet-id-5678",
+			wantErr: "",
+		},
+		{
+			name:    "get member subnet id from default config",
+			opts:    lbaasOpts,
+			service: &corev1.Service{},
+			want:    "default-memberSubnetId",
+			wantErr: "",
+		},
+		{
+			name: "error when loadbalancer class not found",
+			opts: LoadBalancerOpts{},
+			service: &corev1.Service{
+				ObjectMeta: v1.ObjectMeta{
+					Annotations: map[string]string{
+						ServiceAnnotationLoadBalancerClass: "invalid-lb-class",
+					},
+				},
+			},
+			want:    "",
+			wantErr: "invalid loadbalancer class \"invalid-lb-class\"",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			lbaas := LbaasV2{
+				LoadBalancer: LoadBalancer{
+					opts: tt.opts,
+				},
+			}
+
+			got, err := lbaas.getMemberSubnetID(tt.service, &serviceConfig{})
+			if tt.wantErr != "" {
+				assert.EqualError(t, err, tt.wantErr)
+			} else {
+				assert.NoError(t, err)
+			}
+
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}


pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +536,104 @@ func Test_getListenerProtocol(t *testing.T) {
...
+			lbaas := tt.fields.lbaas
+
+			got, err := lbaas.getMemberSubnetID(tt.args.service, tt.args.svcConf)
+			if (err != nil) != (tt.wantErr != nil) || (err != nil && err.Error() != tt.wantErr.Error()) {
+				t.Errorf("LbaasV2.getMemberSubnetID() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if got != tt.want {
+				t.Errorf("LbaasV2.getMemberSubnetID() = %v, want %v", got, tt.want)
+			}
```
dulek, line 636: Why not `assert.Equal()`?

pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +536,104 @@ func Test_getListenerProtocol(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			lbaas := tt.fields.lbaas
+
+			got, err := lbaas.getMemberSubnetID(tt.args.service, tt.args.svcConf)
+			if (err != nil) != (tt.wantErr != nil) || (err != nil && err.Error() != tt.wantErr.Error()) {
+				t.Errorf("LbaasV2.getMemberSubnetID() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
```
dulek, line 633: You should rather do something like this:

```
if tt.wantErr != nil {
    assert.ErrorIs(...)
} else {
    assert.NoError(...)
}
```
Maybe even `assert.ErrorIs` can be used with a `nil` target to make it one line?

pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +536,104 @@ func Test_getListenerProtocol(t *testing.T) {
...
+			wantErr: nil,
+		},
+		{
+			name:   "get member subnet id from default config",
+			fields: fields{},
+			args: args{
+				service: &corev1.Service{},
+				svcConf: &serviceConfig{},
+			},
+			want:    "",
```
dulek, line 605: Can you populate the config to make this have a value? It's `opts.MemberSubnetID`.

dulek (changes requested):

pkg/openstack/loadbalancer_test.go
KingDaemonX: i want to bring you attention to the import in this method, i am testing 

the `svcConfig *serviceConfig` function parameter was never used, i don't know if it is intentional but i tried finding a use case for it in the method and did not see any

dulek:
/retest

Timeout in downloading go packages.

pkg/openstack/loadbalancer_test.go
```diff
@@ -945,3 +946,106 @@ func Test_getBoolFromServiceAnnotation(t *testing.T) {
...
+			opts: LoadBalancerOpts{
+				LBClasses:      map[string]*LBClass{"lbclassKey": {MemberSubnetID: "lb-class-member-subnet-id-5678"}},
+				MemberSubnetID: "lb-class-memberSubnetId",
+			},
+		},
+	}
+
+	type fields struct {
+		lbaas LbaasV2
+	}
```
dulek, line 962: Okay, I think I now understand why this is put here, there are two cases where we need the LbaasV2 struct and use the one defined above. It's quite a bit hard to read. Could we instead extend testcase structure with `opts LoadBalancerOpts` and define the options required in each case? Then we can initialize the LbaasV2 object with these options in the loop itself. I think it'll be clearer.

dulek (changes requested):

pierreprinetti (changes requested):

pkg/openstack/loadbalancer_test.go
```diff
@@ -973,3 +974,89 @@ func TestLbaasV2_updateServiceAnnotations(t *testing.T) {
...
+	lbaasOpts := LoadBalancerOpts{
+		LBClasses:      map[string]*LBClass{"lbclassKey": {MemberSubnetID: "lb-class-member-subnet-id-5678"}},
+		MemberSubnetID: "lb-class-memberSubnetId",
+	}
+
+	tests := []struct {
+		name    string
+		opts    LoadBalancerOpts
+		service *corev1.Service
+		svcConf *serviceConfig
```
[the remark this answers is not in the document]
KingDaemonX: yeah but i need it to complete the argument except i made a changes to the code, removing the argument

pkg/openstack/loadbalancer_test.go
```diff
@@ -973,3 +974,89 @@ func TestLbaasV2_updateServiceAnnotations(t *testing.T) {
...
+		want    string
+		wantErr error
```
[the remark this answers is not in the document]
KingDaemonX: i can't check a nil error if it was string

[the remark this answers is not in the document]
KingDaemonX: @pierreprinetti I just made some changes you suggested and made some of the modification in a PR here 
https://github.com/kubernetes/cloud-provider-openstack/pull/2438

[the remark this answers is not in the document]
pierreprinetti: > i can't check a nil error if it was string the program returns nil sometimes, hence the error check

If `wantErr` is a string, you can say that when it's the empty string, then we don't expect an error. This works well because `assert.EqualError` accepts a string as the expected error.

if wantErr != "" { assert.EqualError(...) } else { assert.NoError(...) }

pkg/openstack/loadbalancer_test.go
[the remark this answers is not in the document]
pierreprinetti: You can pass an empty (or even nil) `*serviceConfig` in the `t.Run` loop.

pkg/openstack/loadbalancer_test.go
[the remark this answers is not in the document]
KingDaemonX: yeah you are perfectly right, thank you
i should be checking the state of what the test wanted instead of the return value; that was my mistake ;)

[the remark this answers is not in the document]
KingDaemonX: @pierreprinetti  i took all your advice and made the neccesary changes already
can you re-review please

pkg/openstack/loadbalancer_test.go
```diff
@@ -1543,3 +1543,89 @@ func Test_nodeAddressForLB(t *testing.T) {
...
+				LoadBalancer: LoadBalancer{
+					opts: tt.opts,
+				},
+			}
+
+			got, err := lbaas.getMemberSubnetID(tt.service, &serviceConfig{})
+			if tt.wantErr != "" {
+				assert.EqualError(t, err, tt.wantErr)
+			} else {
+				if assert.NoError(t, err) {
```
pierreprinetti, line 1625: Can you explain to me what this `if` is doing? I am not sure I understand.

pierreprinetti: Oh I see. You only want to check `got` if `err` is nil.

I think it's fine to check `got` unconditionally, outside the main `if` block.

KingDaemonX: Okay no problem 
I’d just do this instead

KingDaemonX: @pierreprinetti fixed it

pierreprinetti: Thank you, in the name of all the easily-distracted reviewers out there!

pierreprinetti (approved):
Nice work, thanks!
/lgtm
/assign jichenjc

PR #2399, opened 2023-10-06
 test-buildbatchupdatememberopts [cloud provider openstack]

FavourEva:
Add unit test for building update member opts

To run the test, on your terminal run ```make unit```
**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer_test.go
+++ b/pkg/openstack/loadbalancer_test.go
@@ -945,3 +945,31 @@ func Test_getBoolFromServiceAnnotation(t *testing.T) {
 		})
 	}
 }
+
+func TestLbaasV2_updateServiceAnnotations(t *testing.T) {
+	service := &corev1.Service{
+		ObjectMeta: v1.ObjectMeta{
+			Annotations: nil,
+		},
+	}
+
+	annotations := map[string]string{
+		"key1": "value1",
+		"key2": "value2",
+	}
+
+	lbaas := LbaasV2{}
+	lbaas.updateServiceAnnotations(service, annotations)
+
+	serviceAnnotations := make([]map[string]string, 0)
+	for key, value := range service.ObjectMeta.Annotations {
+		serviceAnnotations = append(serviceAnnotations, map[string]string{key: value})
+	}
+
+	expectedAnnotations := []map[string]string{
+		{"key1": "value1"},
+		{"key2": "value2"},
+	}
+
+	assert.ElementsMatch(t, expectedAnnotations, serviceAnnotations)
+}


pkg/openstack/loadbalancer.go
```diff
@@ -352,6 +352,7 @@ type serviceConfig struct {
 	healthMonitorTimeout        int
 	healthMonitorMaxRetries     int
 	healthMonitorMaxRetriesDown int
+	canUseHTTPMonitor			bool
```
dulek, line 355: Why is this necessary? It's odd that a change to the code needs to be made to allow unit testing.

pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +535,160 @@ func Test_getListenerProtocol(t *testing.T) {
...
+        
+	}
+	//Instantiate an object of your LbaasV2 struct
+	lbaas := &LbaasV2{}
+
+	// Functions being tested
+	_, _, err := lbaas.buildBatchUpdateMemberOpts(port, []*corev1.Node{node1, node2}, svcConf)
+
+	// Assertions
+	assert.NoError(t, err, "Expected no error")
```
dulek, line 590: Could that be one of the cases tested in loop instead?

pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +535,160 @@ func Test_getListenerProtocol(t *testing.T) {
...
+
+	testCases := []struct {
+		name        string
+		nodes       []*corev1.Node
+		port        corev1.ServicePort
+		svcConf     *serviceConfig
+		expectedLen int
```
dulek, line 597: This seems to be unused.

pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +535,160 @@ func Test_getListenerProtocol(t *testing.T) {
...
+				lbMemberSubnetID:    "subnet-12345-test",
+				healthCheckNodePort: 8081,
+			},
+			expectedLen: 2,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			_, _, err := lbaas.buildBatchUpdateMemberOpts(tc.port, tc.nodes, tc.svcConf)
```
dulek, line 690: The code seems to ignore the result of the function (besides error) and as a test it should check if what the function returns matches our expectation.

dulek (changes requested):

pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +535,160 @@ func Test_getListenerProtocol(t *testing.T) {
 		})
 	}
 }
+
+
+func TestBuildBatchUpdateMemberOpts(t *testing.T) {
+	t.Parallel()
```
sakshi-1505, line 541: Let's avoid using t.Parallel() since it doesn't help much in table driven tests

pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +535,160 @@ func Test_getListenerProtocol(t *testing.T) {
...
+	testCases := []struct {
+		name        string
+		nodes       []*corev1.Node
+		port        corev1.ServicePort
+		svcConf     *serviceConfig
+		expectedLen int
+		
+	}{
+		{
+			name: "NodePort == 0",
```
sakshi-1505, line 601: Can we only use alphabets while naming a test case as a best practice?

pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +535,173 @@ func Test_getListenerProtocol(t *testing.T) {
...
+				}
+			} else {
+				// No error expected
+				if assert.NoError(t, err, "Expected no error in test case: "+tc.name) {
+					assert.Len(t, members, tc.expectedLen, "Unexpected number of members in test case: "+tc.name)
+
+					// Check for empty newMembers when the flag is set
+					if tc.checkNewMembersEmpty {
+						if assert.Len(t, newMembers, 0, "NewMembers should be empty when there's no error in test case: "+tc.name) {
+							assert.True(t, newMembers.Has(fmt.Sprintf("%s-%s-%d-%d", node1.Name, "192.168.1.1", 8080, 8081)))
```
pierreprinetti, line 698: `fmt` is not imported in this file

pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +535,173 @@ func Test_getListenerProtocol(t *testing.T) {
...
+					assert.Nil(t, newMembers, "NewMembers should be nil when there's an error in test case: "+tc.name)
+				}
+			} else {
+				// No error expected
+				if assert.NoError(t, err, "Expected no error in test case: "+tc.name) {
+					assert.Len(t, members, tc.expectedLen, "Unexpected number of members in test case: "+tc.name)
+
+					// Check for empty newMembers when the flag is set
+					if tc.checkNewMembersEmpty {
+						if assert.Len(t, newMembers, 0, "NewMembers should be empty when there's no error in test case: "+tc.name) {
```
pierreprinetti, line 697: this can be `assert.Empty`

pkg/openstack/loadbalancer_test.go
pierreprinetti: Also, I'd try something like this:

```suggestion
							assert.Contains(t, newMembers, fmt.Sprintf("%s-%s-%d-%d", node1.Name, "192.168.1.1", 8080, 8081))
```

Since the underlying type for `Set` [is a map](https://github.com/kubernetes/apimachinery/blob/1ca6c696c8e470d0745788dad2f69de813b2237d/pkg/util/sets/set.go#L24), this has a chance of working.

pkg/openstack/loadbalancer_test.go
```diff
@@ -694,7 +695,7 @@ func TestBuildBatchUpdateMemberOpts(t *testing.T) {
 
 					// Check for empty newMembers when the flag is set
 					if tc.checkNewMembersEmpty {
-						if assert.Len(t, newMembers, 0, "NewMembers should be empty when there's no error in test case: "+tc.name) {
+						if assert.Empty(t, newMembers, 0, "NewMembers should be empty when there's no error in test case: "+tc.name) {
```
pierreprinetti, line 698:
```suggestion
						if assert.Empty(t, newMembers, "NewMembers should be empty when there's no error in test case: "+tc.name) {
```

pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +536,173 @@ func Test_getListenerProtocol(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			members, newMembers, err := lbaas.buildBatchUpdateMemberOpts(tc.port, tc.nodes, tc.svcConf)
+
+			if tc.expectError {
+				// Expect an error
+				if assert.EqualError(t, err, "Expected an error in test case: "+tc.name) {
```
pierreprinetti, line 687: I understand that `if assert` is documented in the library, however I find it very confusing. What does it do exactly? I suppose that this code would only check that `members` and `newMembers` are nil IF the error matches with the expected error.

Provided that you want to check that `members` and `newMembers` are nil, why wouldn't you want check that in case the error is not the same as expected?

Honestly I would remove the `if` block and put the three assertions one after the other without conditionals. Unless I am missing something?

pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +536,173 @@ func Test_getListenerProtocol(t *testing.T) {
...
+					assert.Nil(t, members, "Members should be nil when there's an error in test case: "+tc.name)
```
pierreprinetti, line 688: The test case name is already listed in the logs. Consider:

```suggestion
					assert.Nil(t, members, "Members should be nil when the function returns an error")
```

pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +536,173 @@ func Test_getListenerProtocol(t *testing.T) {
...
+					assert.Nil(t, newMembers, "NewMembers should be nil when there's an error in test case: "+tc.name)
```
pierreprinetti, line 689: same as above: the test name will already appear in the logs

Issue #2400, opened 2023-10-06
[tests] Add unit tests for the loadbalancer.go functions


sakshi-1505:
/kind feature

**What happened**:

The cloud-provider-openstack code most related to this project lives in [loadbalancer.go](https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/openstack/loadbalancer.go) which is lacking almost all unit tests. This task is to write tests for all of the functions there. Please note that it is untrivial to create tests for methods using gophercloud library, so calling the OpenStack APIs.

**List of Tests**:
- [x] getSecurityGroupName
- [x] getListenerProtocol
- [x] GetLoadBalancerName
- [x] nodeAddressForLB
- [x] getStringFromServiceAnnotation
- [x] getIntFromServiceAnnotation
- [x] getBoolFromServiceAnnotation
- [x] buildMonitorCreateOpts
- [x] buildPoolCreateOpt
- [x] buildBatchUpdateMemberOpts
- [x] buildListenerCreateOpt
- [x] getMemberSubnetID
- [x] getSubnetID
- [x] getNetworkID
- [x] checkListenerPorts
- [x] updateServiceAnnotations
- [x] createLoadBalancerStatus
- [ ] GetLoadBalancerSourceRanges

FavourEva:
I am currently working on these functions: 
getBoolFromServiceAnnotation
buildBatchUpdateMemberOpts
buildListenerCreateOpt
updateServiceAnnotations

sakshi-1505:
Working on getSecurityGroupName,GetLoadBalancerSourceRanges,getSubnetId

dulek:
@sakshi-1505: I think `getSecurityGroupName()` is being done here: https://github.com/kubernetes/cloud-provider-openstack/pull/2389. A bunch of PRs that are already opened: https://github.com/kubernetes/cloud-provider-openstack/pulls?q=is%3Apr+is%3Aopen+test

sakshi-1505:
Ugh, I had created a PR already for that https://github.com/kubernetes/cloud-provider-openstack/pull/2401. I will close the PR since I guess there is one already created but needs some more changes.

sakshi-1505:
@dulek I guess there are multiple PRs created, out of which none is merged but those Prs occupy 50% of the function listed above. Do you think there is some other level 2 task I can pick? I have already built devstack on my local machine as was suggested in the Outreachy guide before I jumped on picking these issues.

[...]

PR #2389, opened 2023-10-04
test : getSecurityGroupName()

KingDaemonX:
**What this PR does / why we need it**:

**Which issue this PR fixes(if applicable)**:
fixes #

**Special notes for reviewers**:

**Release note**:

```release-note
NONE
```
implement a test case that checks for edge case in the `getSecurityGroupName` function to testing all logic and hardcode to check for changes in the function code

PR #2401, opened 2023-10-06
enh: add unit test for get SG name

sakshi-1505:
Adds unit test for function in load balancer.go

refers https://github.com/kubernetes/cloud-provider-openstack/issues/2400

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer_test.go
+++ b/pkg/openstack/loadbalancer_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/rules"
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 type testPopListener struct {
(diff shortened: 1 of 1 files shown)


pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +535,48 @@ func Test_getListenerProtocol(t *testing.T) {
...
+				},
+			},
+			want: `lb-sg-test_uid_with_more_than_255_characters-test_namespace_with_more_than_255_characters_LoremipsumdolorsitametconsecteturadipiscingelitMaurissemperveldiamaposuereVivamustempusliberonecmaximuspellentesque-test_name_with_more_than_255_characters_Lo`,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := getSecurityGroupName(tt.args.service); got != tt.want {
+				t.Errorf("getSecurityGroupName() = %v, want %v", got, tt.want)
+			}
```
dulek, line 579: You should use `assert.Equal` instead here.

dulek (changes requested):
Seems like the test is failing. You can try them locally with `make unit`. Besides that and the inline comment this looks exactly like it should.

To be completely fair with the process #2389 was here first, so I'd rather merge the earlier one once it's finalized. Can you help with reviews on #2389 instead? Thanks!

sakshi-1505:
>Seems like the test is failing. You can try them locally with make unit. Besides that and the inline comment this looks exactly like it should.

I actually did make unit before the pushing the PR through, looks like the spaces are creating issue.

sakshi-1505:
@dulek I think it's pretty fair that we continue the discussion/changes on the PR opened by @KingDaemonX , I will be closing this PR in lieu of the existing PR https://github.com/kubernetes/cloud-provider-openstack/pull/2389

PR #2403, opened 2023-10-06
test : getSecurityGroupName ()

KingDaemonX:

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer_test.go
+++ b/pkg/openstack/loadbalancer_test.go
@@ -824,3 +824,49 @@ func Test_buildPoolCreateOpt(t *testing.T) {
 		})
 	}
 }
+
+func Test_getSecurityGroupName(t *testing.T) {
+	tests := []struct {
+		name     string
+		service  *corev1.Service
+		expected string
+	}{
+		{
+			name: "regular test security group name and length",
+			service: &corev1.Service{
+				ObjectMeta: v1.ObjectMeta{
+					UID:       "12345",
+					Namespace: "security-group-namespace",
+					Name:      "security-group-name",
+				},
+			},
+			expected: "lb-sg-12345-security-group-namespace-security-group-name",
+		},
+		{
+			name: "security group name longer than 255 byte",
+			service: &corev1.Service{
+				ObjectMeta: v1.ObjectMeta{
+					UID:       "12345678-90ab-cdef-0123-456789abcdef",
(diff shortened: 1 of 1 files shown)


KingDaemonX:
@dulek I think i closed my previous PR but mistake when i was trying to push changes 

so i reopened it using one

pkg/openstack/loadbalancer_test.go
```diff
@@ -5,10 +5,12 @@ import (
 	"sort"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
```
sakshi-1505, line 8: Let's not place the declaration of assert module, ideally all third party deps should be listed together like they had been earlier.

pkg/openstack/loadbalancer_test.go
```diff
@@ -534,3 +536,49 @@ func Test_getListenerProtocol(t *testing.T) {
...
+				ObjectMeta: v1.ObjectMeta{
+					UID:       "12345",
+					Namespace: "security-group-namespace",
+					Name:      "security-group-name",
+				},
+			},
+			expected: "lb-sg-12345-security-group-namespace-security-group-name",
+		},
+		{
+			name: "security group name longer than 255 byte",
```
sakshi-1505, line 558: Can we name test more verbose? It's an ideal practice that a test name should also mention the ideal condition it had expected to be in.
 
```
security group name longer than 255 byte should get trimmed to 255 byte
```

KingDaemonX: Yeah I understand you want to help but what the test does is check if it actually trimmed to 255 byte 

so i think it’s verbose enough that I am testing a security group name longer than 255 byte 

hence the name

pkg/openstack/loadbalancer_test.go
dulek: +1, I don't understand why this import is moved.

pkg/openstack/loadbalancer_test.go
dulek: I agree, current name seems self-explanatory enough.

dulek (approved):
Looks good, but if you'll be resolving the conflict please move the import to back where it was.

dulek:
Looks good, just needs another rebase after #2390 got merged.

PR #2406, opened 2023-10-08
enh: add unit test for instanceID from providerID

sakshi-1505:

Adds unit test for fun instanceIDFromProviderID in instances.go file

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/instances_test.go
+++ b/pkg/openstack/instances_test.go
@@ -22,6 +22,7 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	v1 "k8s.io/api/core/v1"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/openstack/instances_test.go
```diff
@@ -189,3 +190,56 @@ func TestSortNodeAddressesWithMultipleCIDRs(t *testing.T) {
...
+		{
+			name: "it appends openstack:// in case of missing protocal in providerID",
+			args: args{
+				providerID: "us-east-1/testInstanceID",
+			},
+			wantInstanceID: "testInstanceID",
+			wantRegion:     "us-east-1",
+			wantErr:        false,
+		},
+	}
```
dulek, line 232: The most common case in OpenStack is like this: `openstack:///testInstanceID` Let's add it too.

dulek (changes requested):

pkg/openstack/instances_test.go
sakshi-1505: Acknowledged.

sakshi-1505:
@dulek please check

sakshi-1505:
bump @dulek @pierreprinetti

pkg/openstack/instances_test.go
```diff
@@ -189,3 +190,65 @@ func TestSortNodeAddressesWithMultipleCIDRs(t *testing.T) {
...
+			name: "it returns error in case of invalid providerID format with no region",
+			args: args{
+				providerID: "openstack://us-east-1-testInstanceID",
+			},
+			wantInstanceID: "",
+			wantRegion:     "",
+			wantErr:        true,
+		},
+		{
+			name: "it parses correct instanceID in case of extra slash protocal",
```
pierreprinetti, line 224: "extra slash" makes it appear as a mistake. However, I _think_ that in our case the default region (which may be what we call "availability zone" in OpenStack) is the empty string (`""`). So the "region" is actually there; it's just that the region name has length zero.

If my speculation is correct (@dulek please keep me honest), I would change the test name to reflect that.

```suggestion
{
			name: "it parses correct instanceID in case the region name is the empty string",
}```

File: pkg/openstack/instances_test.go (hunk 1 above)
Comment (pierreprinetti) [reply pkg/openstack/instances_test.go:241]: I'd also test protocol injection in case the instance as an empty region. Can you please add a case like this?

```
providerID: "/testInstanceID",
```

File: pkg/openstack/instances_test.go (hunk 2 above)
Comment (dulek) [reply pkg/openstack/instances_test.go:224]: My understanding is that region is not an AZ, region has whole new set of API services and only Keystone is shared. I don't think that matters too much here, your rename request seems valid, especially with the typo that's currently there. ;)

**Comment by dulek** [comment]
/approve

@pierreprinetti can mark this with `/lgtm` once it's ready. Approve will stay.

File: pkg/openstack/instances_test.go (hunk 1 above)
Comment (sakshi-1505) [reply pkg/openstack/instances_test.go:241]: I have added the test-case

File: pkg/openstack/instances_test.go (hunk 2 above)
Comment (sakshi-1505) [reply pkg/openstack/instances_test.go:224]: Made requested changes

**Review by pierreprinetti — COMMENTED** [review]
Please fix the last typo and squash: this is good to go! Thank you.

File: pkg/openstack/instances_test.go [hunk 3]
```diff
@@ -189,3 +190,74 @@ func TestSortNodeAddressesWithMultipleCIDRs(t *testing.T) {
...
+			name: "it parses correct instanceID in case the region name is the empty string",
+			args: args{
+				providerID: "openstack:///testInstanceID",
+			},
+			wantInstanceID: "testInstanceID",
+			wantRegion:     "",
+			wantErr:        false,
+		},
+		{
+			name: "it appends openstack:// in case of missing protocal in providerID",
```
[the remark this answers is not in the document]
sakshi-1505: done

sakshi-1505:
Please check again @pierreprinetti

pierreprinetti:
Just a last-minute doubt: https://github.com/kubernetes/cloud-provider-openstack/pull/2406#discussion_r1361824758

pkg/openstack/instances_test.go
sakshi-1505: I just checked, & I had named that test-case incorrectly. Here it is

sakshi-1505: https://github.com/kubernetes/cloud-provider-openstack/blob/c4b34e1459587b652c305c6788dd1380e64c0dc2/pkg/openstack/instances_test.go#L215

pierreprinetti:
Excellent, thanks.
/lgtm

PR #2408, opened 2023-10-09
feat: add unit test for createLBStatus

sakshi-1505:

refers #2400

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer_test.go
+++ b/pkg/openstack/loadbalancer_test.go
@@ -538,6 +538,115 @@ func Test_getListenerProtocol(t *testing.T) {
 	}
 }
 
+func TestLbaasV2_createLoadBalancerStatus(t *testing.T) {
+	type fields struct {
+		LoadBalancer LoadBalancer
+	}
+	type result struct {
+		HostName  string
+		IPAddress string
+	}
+	type args struct {
+		service *corev1.Service
+		svcConf *serviceConfig
+		addr    string
+	}
+	tests := []struct {
+		name   string
+		fields fields
+		args   args
+		want   result
+	}{
+		{
+			name: "it should return hostname from service annotation",
+			fields: fields{
+				LoadBalancer: LoadBalancer{
+					opts: LoadBalancerOpts{
+						EnableIngressHostname: false,
+						IngressHostnameSuffix: "test",
+					},
+				},
+			},
+			args: args{
+				service: &corev1.Service{
+					ObjectMeta: v1.ObjectMeta{
+						Annotations: map[string]string{"loadbalancer.openstack.org/hostname": "testHostName"},
+					},
+				},
+				svcConf: &serviceConfig{
+					enableProxyProtocol: false,
+				},
+				addr: "10.10.0.6",
+			},
+			want: result{
+				HostName: "testHostName",
+			},
+		},
+		{
+			name: "it should return fakehostname if proxyProtocol & IngressHostName is enabled without svc annotation",
+			fields: fields{
+				LoadBalancer: LoadBalancer{
+					opts: LoadBalancerOpts{
+						EnableIngressHostname: true,
+						IngressHostnameSuffix: "ingress-suffix",
+					},
+				},
+			},
+			args: args{
+				service: &corev1.Service{
+					ObjectMeta: v1.ObjectMeta{
+						Annotations: map[string]string{"test": "key"},
+					},
+				},
+				svcConf: &serviceConfig{
+					enableProxyProtocol: true,
+				},
+				addr: "10.10.0.6",
+			},
+			want: result{
+				HostName: "10.10.0.6.ingress-suffix",
+			},
+		},
+		{
+			name: "it should default to ip address if not hostname can be found from svc or proxyProtocol",
+			fields: fields{
+				LoadBalancer: LoadBalancer{
+					opts: LoadBalancerOpts{
+						EnableIngressHostname: false,
+						IngressHostnameSuffix: "ingress-suffix",
+					},
+				},
+			},
+			args: args{
+				service: &corev1.Service{
+					ObjectMeta: v1.ObjectMeta{
+						Annotations: map[string]string{"test": "key"},
+					},
+				},
+				svcConf: &serviceConfig{
+					enableProxyProtocol: false,
+				},
+				addr: "10.10.0.6",
+			},
+			want: result{
+				IPAddress: "10.10.0.6",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			lbaas := &LbaasV2{
+				LoadBalancer: tt.fields.LoadBalancer,
+			}
+
+			result := lbaas.createLoadBalancerStatus(tt.args.service, tt.args.svcConf, tt.args.addr)
+			assert.Equal(t, tt.want.HostName, result.Ingress[0].Hostname)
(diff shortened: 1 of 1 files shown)


dulek:
Looks good I think, just needs a rebase.

sakshi-1505:
@dulek please take a look now

sakshi-1505:
bump @dulek @pierreprinetti

dulek:
Please squash the commits before this can be merged.

dulek:
/approve

I'm leaving approval, will lgtm once it's squashed.

sakshi-1505:
@pierreprinetti @dulek Can you please approve this now? I did the rebase, there is addition of another test in this PR because the authors are adding test in b/w of the file rather than appending things at the end which is the major cause of all this conflicts & errors.

pkg/openstack/loadbalancer_test.go
```diff
@@ -538,6 +538,119 @@ func Test_getListenerProtocol(t *testing.T) {
...
+			want: &corev1.LoadBalancerStatus{
+				Ingress: []corev1.LoadBalancerIngress{
+					{
+						Hostname: "testHostName",
+					},
+				},
+			},
+		},
+		{
+			name: "it should return fakehostname if proxyProtocal & IngressHostName is enabled without svc annotation",
```
pierreprinetti, line 586: little typo:

```suggestion
			name: "it should return fakehostname if proxyProtocol & IngressHostName is enabled without svc annotation",
```

pkg/openstack/loadbalancer_test.go
```diff
@@ -538,6 +538,119 @@ func Test_getListenerProtocol(t *testing.T) {
...
+		{
+			name: "it should return hostname from service annotation",
+			fields: fields{
+				LoadBalancer: LoadBalancer{
+					opts: LoadBalancerOpts{
+						EnableIngressHostname: false,
+						IngressHostnameSuffix: "test",
+					},
+				},
+			},
```
pierreprinetti, line 565: Is this setup required for the test case? Why did you choose to include it?

pierreprinetti (changes requested):
Your test is correct, as in: it validates the logic that is currently in place.

However, I think you could walk an extra mile and make it more robust.

Essentially, your test validates two fields of `Status`: `Hostname` and `IP`. Now, imagine that tomorrow we need to add another field: for example (and this is totally made up), `maximumConcurrentConnections`, that defaults to 1000.

As soon as we add that new field, your test fails, because you are comparing for equality an entire `corev1.LoadBalancerStatus` object.

---

One way to make your test future-proof is to only check the properties you are actually testing for. Instead of comparing the entire object, you could check that the single `Hostname` and `IP` properties match what you need them to. Do you want to try that?

pkg/openstack/loadbalancer_test.go
sakshi-1505: Actually, this is included to make sure that if at some later time we change the way of getHostFromSVC annotation the function doesn't go into this code block https://github.com/kubernetes/cloud-provider-openstack/blob/60ad70c0b53d545f323df71b4603bfc8a98dd2df/pkg/openstack/loadbalancer.go#L1881C1-L1882C1. Moreover this is to ensure the declarative behaviour of the unit test as they should be.

pkg/openstack/loadbalancer_test.go
```diff
@@ -824,7 +937,6 @@ func Test_buildPoolCreateOpt(t *testing.T) {
 		})
 	}
 }
-
```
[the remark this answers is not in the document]
sakshi-1505: Done

pkg/openstack/loadbalancer_test.go
sakshi-1505: Fixed

pkg/openstack/loadbalancer_test.go
```diff
@@ -538,6 +538,116 @@ func Test_getListenerProtocol(t *testing.T) {
...
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			lbaas := &LbaasV2{
+				LoadBalancer: tt.fields.LoadBalancer,
+			}
+			if tt.want.HostName != "" {
+				assert.Equal(t, tt.want.HostName, lbaas.createLoadBalancerStatus(tt.args.service, tt.args.svcConf, tt.args.addr).Ingress[0].Hostname)
```
pierreprinetti, line 642: I'd call the function under test just once, save the result in a variable, and assert the variable. I know it's not a big deal in this case, but calling it once per assertion  can be a problematic pattern when the function under test is more expensive.

pkg/openstack/loadbalancer_test.go
```diff
@@ -538,6 +538,116 @@ func Test_getListenerProtocol(t *testing.T) {
...
+				IPAddress: "10.10.0.6",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			lbaas := &LbaasV2{
+				LoadBalancer: tt.fields.LoadBalancer,
+			}
+			if tt.want.HostName != "" {
```
pierreprinetti, line 641: What happens if you remove this conditional?

pkg/openstack/loadbalancer_test.go
```diff
@@ -538,6 +538,116 @@ func Test_getListenerProtocol(t *testing.T) {
...
+				assert.Equal(t, tt.want.HostName, lbaas.createLoadBalancerStatus(tt.args.service, tt.args.svcConf, tt.args.addr).Ingress[0].Hostname)
+			}
+			if tt.want.IPAddress != "" {
```
pierreprinetti, line 644: What happens if you remove this conditional?

pkg/openstack/loadbalancer_test.go
sakshi-1505: I don't know why I believed that the want.result would be a pointer where it's a string type itself. I removed both conditional

pkg/openstack/loadbalancer_test.go
sakshi-1505: https://github.com/kubernetes/cloud-provider-openstack/pull/2408#discussion_r1361051543

pkg/openstack/loadbalancer_test.go
sakshi-1505: done

pkg/openstack/loadbalancer_test.go
```diff
@@ -538,6 +538,115 @@ func Test_getListenerProtocol(t *testing.T) {
...
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			lbaas := &LbaasV2{
+				LoadBalancer: tt.fields.LoadBalancer,
+			}
+
+			expectedResult := lbaas.createLoadBalancerStatus(tt.args.service, tt.args.svcConf, tt.args.addr)
```
pierreprinetti, line 642: nit: This is not the "expected" result (which we would define in the test case), but rather the actual result. Consider renaming this variable to just "result" to avoid confusion in the matchers.

```suggestion
			result := lbaas.createLoadBalancerStatus(tt.args.service, tt.args.svcConf, tt.args.addr)
```

pierreprinetti (commented):
Looks good to me! Please apply the little fix in the inline comments, then squash, and we're ready to merge. Thank you!

sakshi-1505:
PLease review again @pierreprinetti , I have made all requested changes

PR #2411, opened 2023-10-09
Test get subnet

majorchork:
this pr adds unit test for getSubnetID LbaasV2 method in loadbalancer.go
refers #2400 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer_test.go
+++ b/pkg/openstack/loadbalancer_test.go
@@ -2,6 +2,7 @@ package openstack
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 	"sort"
 	"testing"
@@ -767,11 +768,10 @@ func TestLbaasV2_GetLoadBalancerName(t *testing.T) {
 
 func Test_buildPoolCreateOpt(t *testing.T) {
 	type args struct {
-		protocol    string
-		svcConf     *serviceConfig
-		service     *corev1.Service
-		lbaasV2     *LbaasV2
-		lbAlgorithm string
+		protocol string
+		svcConf  *serviceConfig
+		service  *corev1.Service
+		lbaasV2  *LbaasV2
 	}
 	tests := []struct {
 		name string
@@ -1752,6 +1752,132 @@ func TestBuildBatchUpdateMemberOpts(t *testing.T) {
 			} else {
 				assert.Len(t, newMembers, tc.expectedNewMembersCount)
 			}
+		})
+	}
+}
+
+func Test_getSubnetID(t *testing.T) {
+	type args struct {
+		svcConf *serviceConfig
+		service *corev1.Service
+		lbaasV2 *LbaasV2
+	}
+	tests := []struct {
+		name        string
+		args        args
+		want        string
+		expectedErr string
+	}{
+		{
+			name: "test get subnet from service annotation",
+			args: args{
+				svcConf: &serviceConfig{},
+				lbaasV2: &LbaasV2{
+					LoadBalancer{
+						opts: LoadBalancerOpts{
+							LBClasses: map[string]*LBClass{
+								"test-class": {
+									SubnetID: "test-class-subnet-id",
+								},
+							},
+						},
+					},
+				},
+				service: &corev1.Service{
+					ObjectMeta: v1.ObjectMeta{
+						Annotations: map[string]string{
+							"loadbalancer.openstack.org/subnet-id": "annotation-test-id",
+							"loadbalancer.openstack.org/class":     "test-class",
+						},
+					},
+				},
+			},
+			want: "annotation-test-id",
+		},
+		{
+			name: "test get subnet from config class",
+			args: args{
+				svcConf: &serviceConfig{},
+				lbaasV2: &LbaasV2{
+					LoadBalancer{
+						opts: LoadBalancerOpts{
+							LBClasses: map[string]*LBClass{
+								"test-class": {
+									SubnetID: "test-class-subnet-id",
+								},
+							},
+						},
+					},
+				},
+				service: &corev1.Service{
+					ObjectMeta: v1.ObjectMeta{
+						Annotations: map[string]string{
+							"loadbalancer.openstack.org/class": "test-class",
+						},
+					},
+				},
+			},
+			want: "test-class-subnet-id",
+		},
+		{
+			name: "test get subnet from config class with invalid loadbalancer class",
+			args: args{
+				svcConf: &serviceConfig{},
+				lbaasV2: &LbaasV2{
+					LoadBalancer{
+						opts: LoadBalancerOpts{
+							LBClasses: map[string]*LBClass{
+								"decoy-class": {
+									SubnetID: "test-id",
+								},
+							},
+							SubnetID: "test-subnet-id",
+						},
+					},
+				},
+				service: &corev1.Service{
+					ObjectMeta: v1.ObjectMeta{
+						Annotations: map[string]string{
+							"loadbalancer.openstack.org/class": "test-class",
+						},
+					},
+				},
+			},
+			want:        "",
+			expectedErr: fmt.Sprintf("invalid loadbalancer class %q", "test-class"),
+		},
+		{
+			name: "test get subnet from default config",
+			args: args{
+				svcConf: &serviceConfig{},
+				lbaasV2: &LbaasV2{
+					LoadBalancer{
+						opts: LoadBalancerOpts{
+							LBClasses: map[string]*LBClass{
+								"test-config-class-subnet-id": {
+									SubnetID: "test-id",
+								},
+							},
+							SubnetID: "test-default-subnet-id",
+						},
+					},
+				},
+				service: &corev1.Service{},
+			},
+			want: "test-default-subnet-id",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := tt.args.lbaasV2.getSubnetID(tt.args.service, tt.args.svcConf)
+			if tt.expectedErr != "" {
+				assert.EqualError(t, err, tt.expectedErr)
+			}
+			if tt.expectedErr == "" {
+				assert.NoError(t, err)
+			}
+			assert.Equal(t, tt.want, got)
 		})
 	}
 }
--- a/pkg/openstack/loadbalancer_test.go
+++ b/pkg/openstack/loadbalancer_test.go
@@ -767,10 +767,11 @@ func TestLbaasV2_GetLoadBalancerName(t *testing.T) {
 
 func Test_buildPoolCreateOpt(t *testing.T) {
 	type args struct {
-		protocol string
-		svcConf  *serviceConfig
-		service  *corev1.Service
-		lbaasV2  *LbaasV2
+		protocol    string
+		svcConf     *serviceConfig
+		service     *corev1.Service
+		lbaasV2     *LbaasV2
+		lbAlgorithm string
 	}
 	tests := []struct {
 		name string


pkg/openstack/loadbalancer_test.go
```diff
@@ -787,3 +788,149 @@ func Test_buildPoolCreateOpt(t *testing.T) {
...
+							"loadbalancer.openstack.org/class": "test-class",
+						},
+					},
+				},
+			},
+			want:        "test-class-subnet-id",
+			expectedErr: nil,
+		},
+		{
+			name: "test get subnet from config class",
```
dulek, line 877: This name is the same as in line 841, we should have them unique.

pkg/openstack/loadbalancer_test.go
```diff
@@ -787,3 +788,149 @@ func Test_buildPoolCreateOpt(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := tt.args.lbaasV2.getSubnetID(tt.args.service, tt.args.svcConf)
+			if tt.expectedErr != nil {
+				assert.EqualError(t, err, tt.expectedErr.Error())
+			}
+
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("getSubnetID() = %v, expected %v", got, tt.want)
+			}
```
dulek, line 933: Why not `assert.Equal()`?

dulek (commented):
Seems like a case testing [this codepath](https://github.com/kubernetes/cloud-provider-openstack/blob/8af230a0d5b5443ef405017a3c541a9556ee0158/pkg/openstack/loadbalancer.go#L1467-L1470) is missing?

pkg/openstack/loadbalancer_test.go
dulek: I think this one still stands.

majorchork: @dulek updated

majorchork:
> Seems like a case testing [this codepath](https://github.com/kubernetes/cloud-provider-openstack/blob/8af230a0d5b5443ef405017a3c541a9556ee0158/pkg/openstack/loadbalancer.go#L1467-L1470) is missing?

yea! that was the one with the wrong name, I've updated it now

pkg/openstack/loadbalancer_test.go
```diff
@@ -593,3 +595,296 @@ func Test_getIntFromServiceAnnotation(t *testing.T) {
...
+		{
+			name: "test get subnet from service annotation",
+			args: args{
+				svcConf: &serviceConfig{},
+				lbaasV2: &LbaasV2{
+					LoadBalancer{
+						opts: LoadBalancerOpts{
+							LBClasses: map[string]*LBClass{
+								"test-class": {
+									SubnetID:           "test-id",
```
pierreprinetti, line 785: What do you think about putting something like `test-class-subnet-id` here, to make sure that the subnet ID is actually taken from the annotation?

pkg/openstack/loadbalancer_test.go
```diff
@@ -593,3 +595,296 @@ func Test_getIntFromServiceAnnotation(t *testing.T) {
...
+			want:        "test-class-subnet-id",
+			expectedErr: nil,
+		},
+		{
+			name: "test get subnet from config class with invalid loadbalancer class",
+			args: args{
+				svcConf: &serviceConfig{},
+				lbaasV2: &LbaasV2{
+					LoadBalancer{
+						opts: LoadBalancerOpts{
```
pierreprinetti, line 846: What do you think about adding one LBClass here that is NOT `test-class`?

pkg/openstack/loadbalancer_test.go
```diff
@@ -593,3 +595,296 @@ func Test_getIntFromServiceAnnotation(t *testing.T) {
...
+			expectedErr: fmt.Errorf("invalid loadbalancer class %q", "test-class"),
+		},
+		{
+			name: "test get subnet from default config",
+			args: args{
+				svcConf: &serviceConfig{},
+				lbaasV2: &LbaasV2{
+					LoadBalancer{
+						opts: LoadBalancerOpts{
+							SubnetID: "test-default-subnet-id",
```
pierreprinetti, line 869: I'd add a LBClass here as well, just to make sure it's ignored in case the class is not set below. What do you think?

pkg/openstack/loadbalancer_test.go
```diff
@@ -593,3 +595,296 @@ func Test_getIntFromServiceAnnotation(t *testing.T) {
...
+			},
+			want:        "test-default-subnet-id",
+			expectedErr: nil,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := tt.args.lbaasV2.getSubnetID(tt.args.service, tt.args.svcConf)
+			if tt.expectedErr != nil {
```
pierreprinetti, line 883: Watch out! You are not checking the case where `tt.expectedErr` is nil and `err` is not nil

pierreprinetti (commented):
Reviewing `Test_getSubnetID`

pkg/openstack/loadbalancer_test.go
majorchork: That sounds great, should I add an extra LBClass or completely replace the other LBClass?

pkg/openstack/loadbalancer_test.go
majorchork: sounds nice but looking at the function the default config is only selected when the config class is empty, so if I add the extra class, it's supposed to pick it up and not ignore it

pkg/openstack/loadbalancer_test.go
majorchork: yea!! that's what I'm referring to as well, checking it is repetitive since the assert checks it, as shown in this screenshot
Screenshot 2023-10-10 at 13 17 47

pkg/openstack/loadbalancer_test.go
majorchork: sound great! i've added it in

pkg/openstack/loadbalancer_test.go
majorchork: i've added an extra LBClass, if i completely remove it the function will set LBClass to "" and pick the default class instead

pkg/openstack/loadbalancer_test.go
majorchork: if you look at the screenshot, if the configClassName is not set to ""(empty string) the default config class is not picked, if I add a LbClass here the default config class won't be ignored, and we're testing for the default config class, if you'd like I can add an extra test case with LbClass set and the default  config class set, it should ignore the default config class and pick the LBClass
Screenshot 2023-10-10 at 13 27 20

pkg/openstack/loadbalancer_test.go
pierreprinetti: Am I blind? I think your code only checks err if tt.expectedErr is not nil.

pkg/openstack/loadbalancer_test.go
majorchork: i misunderstood you before, I've added it in now

pkg/openstack/loadbalancer_test.go
```diff
@@ -824,3 +825,151 @@ func Test_buildPoolCreateOpt(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := tt.args.lbaasV2.getSubnetID(tt.args.service, tt.args.svcConf)
+			if tt.expectedErr != "" {
+				assert.EqualError(t, err, tt.expectedErr)
+			}
+
+			assert.Equal(t, got, tt.want)
```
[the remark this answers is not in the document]
majorchork: Sure! I've added in the check. Please review again

pkg/openstack/loadbalancer_test.go
```diff
@@ -870,3 +871,151 @@ func Test_getSecurityGroupName(t *testing.T) {
...
+				service: &corev1.Service{},
+			},
+			want: "test-default-subnet-id",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := tt.args.lbaasV2.getSubnetID(tt.args.service, tt.args.svcConf)
+			if tt.expectedErr != "" || err != nil {
```
pierreprinetti, line 1014: This would indeed work in all likely cases. However, it is not quite correct.

`assert.EqualError` is designed for cases where an error is expected, and only match its message against a string.

Now consider this edge case:
```Go
func TestAssertEqualError(t *testing.T) {
	t.Run("edge case: non-nil error with empty message", func(t *testing.T) {
		var (
			err         error  = fmt.Errorf("")
			expectedErr string = ""
		)
		assert.EqualError(t, err, expectedErr) // PASS: the error message matches the expected string
	})
}
```
In this example the error is not nil, however the assertion would not trigger any test failure.

Were the function under test to return a non-nil error with an empty string as its message, this test would not catch the issue.

Clearly, we are unlikely to hit this particular case. However, by the law of large numbers, an issue like this WILL happen, and those are precisely the hardest to debug. Let's be a little defensive in our tests, at the price of a couple more lines of code.

Here you probably should write a separate block for when `tt.expectedErr` is the empty string, because in that case we really want to `assert.NoError(t, err)`.

dulek (changes requested):

pkg/openstack/loadbalancer_test.go
majorchork: wow!! nice catch i really didn't consider that edge case! thanks a lot for the insight, I'll fix it

dulek:
/approve

I'll leave `/lgtm` for @pierreprinetti, I'm not 100% sure this is what he meant on the error checking front.

majorchork:
> /approve
> 
> I'll leave `/lgtm` for @pierreprinetti, I'm not 100% sure this is what he meant on the error checking front.

Thanks a lot! Noted

pkg/openstack/loadbalancer_test.go
```diff
@@ -973,3 +974,129 @@ func TestLbaasV2_updateServiceAnnotations(t *testing.T) {
...
```
pierreprinetti, line 1093:
```suggestion
			if tt.expectedErr != "" {
```

pierreprinetti (changes requested):

pkg/openstack/loadbalancer_test.go
```diff
@@ -1629,3 +1630,129 @@ func TestLbaasV2_getMemberSubnetID(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := tt.args.lbaasV2.getSubnetID(tt.args.service, tt.args.svcConf)
+			if tt.expectedErr != "" {
+				assert.EqualError(t, err, tt.expectedErr)
+			}
+			if tt.expectedErr == "" {
+				assert.NoError(t, err)
+			}
+			assert.Equal(t, got, tt.want)
```
pierreprinetti, line 1755: I know it's very confusing, but the argument order in `assert.Equal`'s [is the other way around](https://github.com/stretchr/testify/blob/4ae48e988c8c84c138d688eba40bfc869f4be907/assert/assertions.go#L415C18-L415C18). This is important to get a clear error message in case of mismatch.

```suggestion
			assert.Equal(t, tt.want, got)
```

pierreprinetti (changes requested):

pkg/openstack/loadbalancer_test.go
majorchork: I agree, nice catch! very true! thanks for always being thorough! 🙏🏾

pierreprinetti: thorough is my middle name

pierreprinetti (approved):
/lgtm

pkg/openstack/loadbalancer_test.go
pierreprinetti: shameless

majorchork: 😂😂😂 perks of the job!

pierreprinetti:
Please rebase to enable the merge :pray:

majorchork:
> Please rebase to enable the merge 🙏

I already have seems like the bot forgot to remove the label, I'll just force push again and see if the bot picks it this time
Screenshot 2023-10-19 at 06 43 21

majorchork:
@k8s-ci-robot  verify that this pr need rebase

pierreprinetti:
> > Please rebase to enable the merge 🙏
> 
> I already have seems like the bot forgot to remove the label, I'll just force push again and see if the bot picks it this time Screenshot 2023-10-19 at 06 43 21

What do you mean, "Your branch is ahead of 'origin/master' by 17 commits"? This is not expected. Your local `master` branch is not aligned with origin.

To rebase on origin's master, try:

```sh
git checkout test-getSubnetID
git fetch origin
git rebase origin/master
```

---

Additionally, if you want to get rid of those 17 commits on your local `master` branch, here's how to do it:
```sh
git checkout master
git fetch origin
git reset --hard origin/master
git clean -fdx
```

majorchork:
> ```shell
> git reset --hard origin/master
> ```
Screenshot 2023-10-19 at 07 13 47


😅 my bad, it's ahead cause I've not published my local to my origin, I always just fetch upstream and merge then checkout to the branch I want to rebase before pushing, I never actually push the changes on my local origin
`git fetch upstream (upstream is  https://github.com/kubernetes/cloud-provider-openstack/master)
 git merge upstream/master
 git checkout test-getSubnetId
 git rebase master`
so my local master is ahead by all the changes I've fetched from the upstream, 
I've updated, gotten rid of the commits, and rebased now

PR #2428, opened 2023-10-12
[1.26.4] 1.26.4 release

kayrus:

This PR bumps a release version for manifests and docs using a new `hack/bump-release.sh` script

**Release note**:

```release-note
NONE
```


go.mod
```diff
@@ -147,36 +147,3 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
-
-replace (
```
zetaab, line 151: removing entire section is little bit scary

kayrus: I'm looking into a solution to avoid this. Replace was a hack when there were dependency issues. Now it should be ok, especially for k8s.io deps. I'd like to play a bit with these dependencies before continuing.

dulek:
Can we squash the commit doing the revert of the k3s version?

PR #2429, opened 2023-10-13
Test buid monitor create opts

majorchork:
this pr adds unit test for buildMonitorCreateOpts LbaasV2 method in loadbalancer.go
refers #2400

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer_test.go
+++ b/pkg/openstack/loadbalancer_test.go
@@ -7,7 +7,9 @@ import (
 	"sort"
 	"testing"
 
+	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/listeners"
+	v2monitors "github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/monitors"
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/pools"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/rules"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


pkg/openstack/loadbalancer_test.go
```diff
@@ -973,3 +976,281 @@ func TestLbaasV2_updateServiceAnnotations(t *testing.T) {
...
+				MaxRetriesDown: 1,
+			},
+		},
+		{
+			name: "using tcp protocol with not-ovn provider",
+			testArg: testArg{
+				lbaas: &LbaasV2{
+					LoadBalancer{
+						opts: LoadBalancerOpts{
+							LBProvider: "not-ovn",
```
dulek, line 1216: You can use "amphora" here to match what actually supports HTTP monitors.

pkg/openstack/loadbalancer_test.go
```diff
@@ -973,3 +976,281 @@ func TestLbaasV2_updateServiceAnnotations(t *testing.T) {
...
+						},
+						lb: &gophercloud.ServiceClient{
+							Endpoint: "/endpoint/",
+						},
```
dulek, line 1220: Do we even need these? I know `canUseHTTPMonitor()` uses it, but we cannot test that codepath anyway as there's no easy way to mock gophercloud in CPO.

pkg/openstack/loadbalancer_test.go
```diff
@@ -973,3 +976,281 @@ func TestLbaasV2_updateServiceAnnotations(t *testing.T) {
...
+					},
+				},
+				svcConf: &serviceConfig{
+					healthMonitorDelay:          1,
+					healthMonitorTimeout:        1,
+					healthMonitorMaxRetries:     1,
+					healthMonitorMaxRetriesDown: 1,
```
dulek, line 1134: It would be better to have these values different to also check if function doesn't mix them.

pkg/openstack/loadbalancer_test.go
```diff
@@ -973,3 +976,281 @@ func TestLbaasV2_updateServiceAnnotations(t *testing.T) {
...
+					healthCheckNodePort:         2,
```
dulek, line 1135: This should be in the NodePort range: https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport

dulek (changes requested):

majorchork:
hello @pierreprinetti, kindly take a look at this test and please let me know if you have any suggestions 😃

pkg/openstack/loadbalancer_test.go
```diff
@@ -1629,3 +1632,277 @@ func TestLbaasV2_getMemberSubnetID(t *testing.T) {
...
+			},
+			want: v2monitors.CreateOpts{
+				Name:           "using udp with ovn provider",
+				Type:           "UDP-CONNECT",
+				Delay:          8,
+				Timeout:        4,
+				MaxRetries:     3,
+				MaxRetriesDown: 2,
+			},
+		},
```
pierreprinetti, line 1863: How is this different from the first test case?

pierreprinetti (changes requested):

pkg/openstack/loadbalancer_test.go
```diff
@@ -1629,3 +1632,277 @@ func TestLbaasV2_getMemberSubnetID(t *testing.T) {
...
+			if tt.expectedErr != "" {
+				assert.EqualError(t, err, tt.expectedErr)
+			}
+			if tt.expectedErr == "" {
+				assert.NoError(t, err)
+			}
+			assert.Equal(t, got, tt.want)
+		})
+	}
+}
```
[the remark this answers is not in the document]
majorchork: yes, i authored that, i just checked out from that branch, once #2411 is merged and I rebase it should balance out

pkg/openstack/loadbalancer_test.go
majorchork: True! nice catch, i was thinking of way to test the gophercloud part of the function, but it was tricky to mock so I go removed it and forgot both test cases were thesame without that

pkg/openstack/loadbalancer_test.go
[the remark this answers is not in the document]
pierreprinetti: didn't balance out :/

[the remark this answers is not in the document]
pierreprinetti: oh MERGED you say. hmm OK, but then turn this PR as a WIP if it isn't ready to be reviewed until another merges.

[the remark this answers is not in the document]
majorchork: oh! noted! 😃

majorchork:
> Thanks. Please squash and we should be ready to merge!

thanks a lot, squashed!😅

dulek:
Seems like this is missing a case when `healthCheckNodePort=0`?

majorchork:
> Seems like this is missing a case when `healthCheckNodePort=0`?

Wow! thanks a lot. Nice catch! I've added it now

PR #2431, opened 2023-10-13
test-buildbatchupdatememberopts

FavourEva:
Add unit test to build batch update member opts

To run the test, on your terminal run ```make unit```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer_test.go
+++ b/pkg/openstack/loadbalancer_test.go
@@ -1629,3 +1629,128 @@ func TestLbaasV2_getMemberSubnetID(t *testing.T) {
 		})
 	}
 }
+
+func TestBuildBatchUpdateMemberOpts(t *testing.T) {
+
+	// Sample Nodes
+	node1 := &corev1.Node{
+		ObjectMeta: v1.ObjectMeta{
+			Name: "node-1",
+		},
+		Status: corev1.NodeStatus{
+			Addresses: []corev1.NodeAddress{
+				{
+					Type:    corev1.NodeInternalIP,
+					Address: "192.168.1.1",
+				},
+			},
+		},
+	}
+	node2 := &corev1.Node{
+		ObjectMeta: v1.ObjectMeta{
+			Name: "node-2",
+		},
+		Status: corev1.NodeStatus{
+			Addresses: []corev1.NodeAddress{
+				{
+					Type:    corev1.NodeInternalIP,
+					Address: "192.168.1.2",
+				},
+			},
+		},
+	}
+	testCases := []struct {
+		name                    string
+		nodes                   []*corev1.Node
+		port                    corev1.ServicePort
+		svcConf                 *serviceConfig
+		expectedLen             int
+		expectedNewMembersCount int
+	}{
+		{
+			name:  "NodePortequalszero",
+			nodes: []*corev1.Node{node1, node2},
+			port:  corev1.ServicePort{NodePort: 0},
+			svcConf: &serviceConfig{
+				preferredIPFamily:   corev1.IPv4Protocol,
+				lbMemberSubnetID:    "subnet-12345-test",
+				healthCheckNodePort: 8081,
+			},
+			expectedLen:             0,
+			expectedNewMembersCount: 0,
+		},
+		{
+			name:  "Valid nodes, canUseHTTPMonitor=false",
+			nodes: []*corev1.Node{node1, node2},
+			port:  corev1.ServicePort{NodePort: 8080},
+			svcConf: &serviceConfig{
+				preferredIPFamily:   corev1.IPv4Protocol,
+				lbMemberSubnetID:    "subnet-12345-test",
+				healthCheckNodePort: 8081,
+				enableMonitor:       false,
+			},
+			expectedLen:             2,
+			expectedNewMembersCount: 2,
+		},
+		{
+			name:  "Valid nodes, canUseHTTPMonitor=true",
+			nodes: []*corev1.Node{node1, node2},
+			port:  corev1.ServicePort{NodePort: 8080},
+			svcConf: &serviceConfig{
+				preferredIPFamily:   corev1.IPv4Protocol,
+				lbMemberSubnetID:    "subnet-12345-test",
+				healthCheckNodePort: 8081,
+				enableMonitor:       true,
+			},
+			expectedLen:             2,
+			expectedNewMembersCount: 2,
+		},
+		{
+			name:  "Invalid preferred IP family, fallback to default",
+			nodes: []*corev1.Node{node1, node2},
+			port:  corev1.ServicePort{NodePort: 0},
+			svcConf: &serviceConfig{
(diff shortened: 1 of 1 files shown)


pkg/openstack/loadbalancer_test.go
```diff
@@ -973,3 +973,175 @@ func TestLbaasV2_updateServiceAnnotations(t *testing.T) {
...
+				assert.Len(t, members, tc.expectedLen, "Unexpected number of members")
+			case invalidPreferredIPFamily:
+				members, _, err := lbaas.buildBatchUpdateMemberOpts(tc.port, tc.nodes, tc.svcConf)
+				assert.NoError(t, err, "Expected no error")
+				assert.Len(t, members, tc.expectedLen, "Unexpected number of members")
+			case errNoAddressFoundMemberIgnored:
+				members, _, err := lbaas.buildBatchUpdateMemberOpts(tc.port, tc.nodes, tc.svcConf)
+				assert.NoError(t, err, "Expected no error")
+				assert.Len(t, members, tc.expectedLen, "Unexpected number of members")
+			}
```
pierreprinetti, line 1143: In this PR, you are defining a slice of cases and then you `range` over it. Which means that you want to apply the same testing logic multiple times. But then you use a combination of `iota` constants and `switch` to define a separate behaviour for each test case. Which means that you haven't found that common testing logic.

If I read this correctly, each `case` here do exactly the same, except `case errNoAddressFound` where you `assert.Nil`. Are you sure that `assert.Len(t, members, 0)` doesn't do what you want? Because if that works, then you can do the same for all cases:

```Go
members, _, err := lbaas.buildBatchUpdateMemberOpts(tc.port, tc.nodes, tc.svcConf)
assert.NoError(t, err)
assert.Len(t, members, tc.expectedLen)
```

---

Also a nit: the error messages you pass as arguments to the assertions don't add any context specific to the case: you can remove them and let the library provide a generic error message.

FavourEva: Fixed. Please review @pierreprinetti

pkg/openstack/loadbalancer_test.go
```diff
@@ -1108,3 +1108,162 @@ func Test_getStringFromServiceAnnotation(t *testing.T) {
...
+			testType:             errNoAddressFoundMemberIgnored,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			members, _, err := lbaas.buildBatchUpdateMemberOpts(tc.port, tc.nodes, tc.svcConf)
+			assert.NoError(t, err, "Expected no error")
+
+			if tc.testType == nodePortEquals0 {
```
pierreprinetti, line 1256: I don't understand this "testType" pattern. What are you looking to achieve? Is there a way you can have the same assertions for all test cases?

FavourEva: I usde the "typeType" to distinguish the different test cases and apply specific assertions based on the nature of each scenario as it allows different expectations for each type of test case.

I would go ahead to resolve this using same assertion as you said.

pierreprinetti: My point is: I don't think you should do that. These scenarios all require the same assertions.

pkg/openstack/loadbalancer_test.go
```diff
@@ -1108,3 +1108,135 @@ func Test_getStringFromServiceAnnotation(t *testing.T) {
...
+			nodes:       []*corev1.Node{},
+			port:        port,
+			svcConf:     svcConf,
+			expectedLen: 0,
+			expectError: false,
+		},
+		{
+			name:  "Valid nodes, canUseHTTPMonitor=false",
+			nodes: []*corev1.Node{node1, node2},
+			port:  corev1.ServicePort{NodePort: 8080},
```
pierreprinetti, line 1184: If I'm not mistaken, this port is identical to `port` (line 1114). Please always use the variable, or always use the literal.

pkg/openstack/loadbalancer_test.go
```diff
@@ -1108,3 +1108,135 @@ func Test_getStringFromServiceAnnotation(t *testing.T) {
...
+				lbMemberSubnetID:    "subnet-12345-test",
+				healthCheckNodePort: 8081,
+			},
+			expectedLen: 2,
+			expectError: false,
+		},
+		{
+			name:  "ErrNoAddressFound happens and a member is ignored",
+			nodes: []*corev1.Node{},
+			port:  corev1.ServicePort{NodePort: 8080},
```
pierreprinetti, line 1222: If I'm not mistaken, this port is identical to `port` (line 1114). Please always use the variable, or always use the literal.

pierreprinetti (changes requested):

pkg/openstack/loadbalancer_test.go
FavourEva: Resolved. Please review @pierreprinetti

pkg/openstack/loadbalancer_test.go
```diff
@@ -1108,3 +1108,130 @@ func Test_getStringFromServiceAnnotation(t *testing.T) {
...
+				{
+					Type:    corev1.NodeInternalIP,
+					Address: "192.168.1.2",
+				},
+			},
+		},
+	}
+
+	// Instantiate an object of your LbaasV2 struct
+	lbaas := &LbaasV2{}
```
pierreprinetti, line 1143: Move this declaration in the `t.Run` loop, so that use use a fresh `LbaasV2` object each time

pkg/openstack/loadbalancer_test.go
```diff
@@ -1108,3 +1108,130 @@ func Test_getStringFromServiceAnnotation(t *testing.T) {
...
+				enableMonitor:       false,
+			},
+			expectedLen: 0,
+			expectError: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			members, _, err := lbaas.buildBatchUpdateMemberOpts(tc.port, tc.nodes, tc.svcConf)
```
pierreprinetti, line 1232: Do you want to also write some assertions regarding the second returned value (`newMembers`)?

pierreprinetti (changes requested):
I think you are missing the case where the function returns an error. This happens when one node exists and its `node.Status.Addresses` is empty

pkg/openstack/loadbalancer_test.go
FavourEva: Yes. I would do that right away. Thanks

FavourEva: @pierreprinetti Please you can check it out again

pierreprinetti:
I was wrong: there can never be a non-nil error. The only error that `nodeAddressForLB` can return is `cpoerrors.ErrNoAddressFound`, and `buildBatchUpdateMemberOpts` ignores it.

jichenjc:
/approve

thanks for the additional UT ~

Issue #1980, opened 2022-08-31
[cinder-csi-plugin] Allow StorageClass parameters to be set in Helm chart


mkjpryor:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

It would be good to be able to set parameters for the storage classes created by the CSI Cinder Helm chart. In particular, I would like to be able to set the availability zone and volume type.

jichenjc:
@mkjpryor do you want to provide such PR? Thanks

mkjpryor:
@jichenjc I'm definitely up for doing it when I get time.

mkjpryor:
/remove-lifecycle stale

I still want to look at this at some point

sakshi-1505:
@mkjpryor If you don't mind, I would be taking over this issue

PR #2434, opened 2023-10-15
feat: add support for parameters for storageClass

sakshi-1505:
fixes https://github.com/kubernetes/cloud-provider-openstack/issues/1980

 ```release-note 
- adds support for parameter in storageClass for csi_chart
 ```

### DIFF
--- a/charts/cinder-csi-plugin/templates/storageclass.yaml
+++ b/charts/cinder-csi-plugin/templates/storageclass.yaml
@@ -9,6 +9,18 @@ metadata:
   {{- end }}
 provisioner: cinder.csi.openstack.org
 reclaimPolicy: Delete
+{{ if .Values.storageClass.parameters.enabled }}
+parameters:
+  {{ if .Values.storageClass.parameters.type }}
+  type: {{ .Values.storageClass.parameters.type }}
+  {{ end }}
+  {{ if .Values.storageClass.parameters.availabilityZone }}
+  availabilityZone: {{ .Values.storageClass.parameters.availabilityZone }}
+  {{ end }}
+  {{ range .Values.storageClass.parameters.additionalParameters }}
+  {{ .key }}: {{ .value }}
+  {{ end }}
+{{ end }}
 allowVolumeExpansion: {{ .Values.storageClass.delete.allowVolumeExpansion }}
 ---
 apiVersion: storage.k8s.io/v1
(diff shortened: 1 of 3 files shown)


sakshi-1505:
@zetaab @pierreprinetti @jichenjc Please take a look.

charts/cinder-csi-plugin/values.yaml
```diff
@@ -160,6 +160,12 @@ secret:
 
 storageClass:
   enabled: true
+  parameters:
```
jichenjc, line 163: is this sample ? so should be comment out?

sakshi-1505: I don't really believe commenting our things in helm's values but always like to put each & every change behind a toggle to maintain backward compatibility. It's not a sample but actual values file, & I have kept all changed behind the `storageClass.parameters.Enabled` toggle, so this is a backward compatible change where folks if want can enable & by default it won't take effect.

charts/cinder-csi-plugin/values.yaml
```diff
@@ -160,6 +160,12 @@ secret:
...
+    enabled: false
+    type: gp2
+    availabilityZone: us-east-1a
```
jichenjc, line 166: I am talking about here, e.g apparently we don't want any chart default to this AWS region?

sakshi-1505: I got your point, let me fix this real quick

sakshi-1505:
Please check again @jichenjc

sakshi-1505:
@jichenjc @pierreprinetti can I please get a review here?

dulek:
Oh, you're missing the release note box too. Grab it from the template: https://github.com/kubernetes/cloud-provider-openstack/blob/66624e34b1884781dca75f2ae6dc9736c0de9f28/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L31-L33

sakshi-1505:
@dulek please recheck

charts/cinder-csi-plugin/templates/storageclass.yaml
```diff
@@ -21,5 +33,17 @@ metadata:
...
+{{ if .Values.storageClass.parameters.enabled }}
+parameters:
+  {{ if .Values.storageClass.parameters.type }}
+  type: {{ .Values.storageClass.parameters.type }}
+  {{ end }}
+  {{ if .Values.storageClass.parameters.availabilityZone }}
+  availabilityZone: {{ .Values.storageClass.parameters.availabilityZone }}
+  {{ end }}
+  {{ range .Values.storageClass.parameters.additionalParameters }}
+  {{ .key }}: {{ .value }}
```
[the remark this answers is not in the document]
dulek: @sakshi-1505: Seems like you've missed this one comment.

dulek (changes requested):
Can you check one of my previous comments?

Besides that let's improve the release note a bit. How about:

```
[cinder-csi-plugin] Added support for modifying StorageClass parameters in the Helm chart.
```

Issue #2243, opened 2023-05-12
[Octavia-ingress-controller] Add configurable listener timeouts through annotations


gmauleon:
/kind feature

Hey,

In some circonstances for long lived connections the default timeout is not enough.
Could we make the load balancer listener timeouts configurable through ingress annotations?

Similar to the existing `octavia.ingress.kubernetes.io/whitelist-source-range`, we could add some `octavia.ingress.kubernetes.io/timeout_member_data` etc... or similar to change the default listener timeouts.

For my specific use case we would only need `timeout_member_data` and `timeout_client_data` but I guess all timeouts could be exposed.

Thanks in advance.
Gael.

jichenjc:
not sure I follow the question here, we have some const definition here 
https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/util/openstack/loadbalancer.go#L49

do you want to update those or other param?

gmauleon:
Hey @jichenjc,

No I meant the timeout on the listener itself, those: https://github.com/gophercloud/gophercloud/blob/master/openstack/loadbalancer/v2/listeners/requests.go#L152

A bit like we already do for the [AllowedCIDRs](https://github.com/gophercloud/gophercloud/blob/master/openstack/loadbalancer/v2/listeners/requests.go#L168) through [this](https://github.com/kubernetes/cloud-provider-openstack/blob/8872f001944426e599fe6ea2b6a53e533523a6d2/pkg/ingress/controller/controller.go#L730) annotation.

So one can configured the listener timeouts by adding annotations on the ingress resource.

Thanks!

dulek:
Oh, you mean for the ingress controller. Please update the PR title with the correct tag then, this got me confused. And yeah, it's a fair request, happy to review once the PR is up.

jichenjc:
updated the title directly :)

gmauleon:
Indeed sorry for the confusing title 😅

dulek:
This is still missing a case when listener timeouts are updated and listener has to be updated too. I'm reopening this.

kush-work:
Oops thats my bad that I forgot to take listener updation and only took creating in mind. I will get this fixed.

PR #2435, opened 2023-10-15
[octavia-ingress-controller]feat: add configurable listener timeout via ingress annotation

sakshi-1505:

This PR basically adds the ability to configure timeout via ingress annotations in ingress-controller.

fixes https://github.com/kubernetes/cloud-provider-openstack/issues/2243

**Release note**:

```release-note
- Added support for configurable timeouts via ingress annotation in octavia-ingress-controller
```

### DIFF
--- a/pkg/ingress/controller/controller.go
+++ b/pkg/ingress/controller/controller.go
@@ -108,6 +108,26 @@ const (
 	// IngressControllerTag is added to the related resources.
 	IngressControllerTag = "octavia.ingress.kubernetes.io"
 
+	// IngressAnnotationTimeoutClientData is the timeout for frontend client inactivity.
+	// If not set, this value defaults to the Octavia configuration key `timeout_client_data`.
+	// Refer to https://docs.openstack.org/octavia/latest/configuration/configref.html#haproxy_amphora.timeout_client_data
+	IngressAnnotationTimeoutClientData = "octavia.ingress.kubernetes.io/timeout-client-data"
+
+	// IngressAnnotationTimeoutMemberData is the timeout for backend member inactivity.
+	// If not set, this value defaults to the Octavia configuration key `timeout_member_data`.
+	// Refer to https://docs.openstack.org/octavia/latest/configuration/configref.html#haproxy_amphora.timeout_member_data
+	IngressAnnotationTimeoutMemberData = "octavia.ingress.kubernetes.io/timeout-member-data"
+
+	// IngressAnnotationTimeoutMemberConnect is the timeout for backend member connection.
+	// If not set, this value defaults to the Octavia configuration key `timeout_member_connect`.
+	// Refer to https://docs.openstack.org/octavia/latest/configuration/configref.html#haproxy_amphora.timeout_member_connect
+	IngressAnnotationTimeoutMemberConnect = "octavia.ingress.kubernetes.io/timeout-member-connect"
+
+	// IngressAnnotationTimeoutTCPInspect is the time to wait for TCP packets for content inspection.
+	// If not set, this value defaults to the Octavia configuration key `timeout_tcp_inspect`.
+	// Refer to https://docs.openstack.org/octavia/latest/configuration/configref.html#haproxy_amphora.timeout_tcp_inspect
+	IngressAnnotationTimeoutTCPInspect = "octavia.ingress.kubernetes.io/timeout-tcp-inspect"
+
 	// IngressSecretCertName is certificate key name defined in the secret data.
 	IngressSecretCertName = "tls.crt"
 	// IngressSecretKeyName is private key name defined in the secret data.
@@ -728,8 +748,13 @@ func (c *Controller) ensureIngress(ing *nwv1.Ingress) error {
 
 	// Create listener
 	sourceRanges := getStringFromIngressAnnotation(ing, IngressAnnotationSourceRangesKey, "0.0.0.0/0")
+	timeoutClientData := maybeGetIntFromIngressAnnotation(ing, IngressAnnotationTimeoutClientData)
+	timeoutMemberConnect := maybeGetIntFromIngressAnnotation(ing, IngressAnnotationTimeoutMemberConnect)
+	timeoutMemberData := maybeGetIntFromIngressAnnotation(ing, IngressAnnotationTimeoutMemberData)
+	timeoutTCPInspect := maybeGetIntFromIngressAnnotation(ing, IngressAnnotationTimeoutTCPInspect)
+
 	listenerAllowedCIDRs := strings.Split(sourceRanges, ",")
-	listener, err := c.osClient.EnsureListener(resName, lb.ID, secretRefs, listenerAllowedCIDRs)
+	listener, err := c.osClient.EnsureListener(resName, lb.ID, secretRefs, listenerAllowedCIDRs, timeoutClientData, timeoutMemberData, timeoutTCPInspect, timeoutMemberConnect)
 	if err != nil {
 		return err
 	}
@@ -1015,6 +1040,23 @@ func getStringFromIngressAnnotation(ingress *nwv1.Ingress, annotationKey string,
 	}
 
 	return defaultValue
+}
+
+// maybeGetIntFromIngressAnnotation searches a given Ingress for a specific annotationKey and either returns the
+// annotation's value
+func maybeGetIntFromIngressAnnotation(ingress *nwv1.Ingress, annotationKey string) *int {
+	klog.V(4).Infof("maybeGetIntFromIngressAnnotation(%s/%s, %v33)", ingress.Namespace, ingress.Name, annotationKey)
+	if annotationValue, ok := ingress.Annotations[annotationKey]; ok {
+		klog.V(4).Infof("Found a Service Annotation for key: %v", annotationKey)
+		returnValue, err := strconv.Atoi(annotationValue)
+		if err != nil {
+			klog.V(4).Infof("Invalid integer found on Service Annotation: %v = %v", annotationKey, annotationValue)
+			return nil
+		}
+		return &returnValue
+	}
+	klog.V(4).Infof("Could not find a Service Annotation; falling back to default setting for annotation %v", annotationKey)
+	return nil
 }
 
 // privateKeyFromPEM converts a PEM block into a crypto.PrivateKey.
--- a/pkg/ingress/controller/openstack/octavia.go
+++ b/pkg/ingress/controller/openstack/octavia.go
@@ -330,7 +330,7 @@ func (os *OpenStack) UpdateLoadBalancerDescription(lbID string, newDescription s
 }
 
 // EnsureListener creates a loadbalancer listener in octavia if it does not exist, wait for the loadbalancer to be ACTIVE.
-func (os *OpenStack) EnsureListener(name string, lbID string, secretRefs []string, listenerAllowedCIDRs []string) (*listeners.Listener, error) {
+func (os *OpenStack) EnsureListener(name string, lbID string, secretRefs []string, listenerAllowedCIDRs []string, timeoutClientData, timeoutMemberData, timeoutTCPInspect, timeoutMemberConnect *int) (*listeners.Listener, error) {
 	listener, err := openstackutil.GetListenerByName(os.Octavia, name, lbID)
 	if err != nil {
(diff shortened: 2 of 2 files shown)


sakshi-1505:
@pierreprinetti @dulek @jichenjc Request you to please review the above PR & let me know if I thought in the right redirection or not. I would say it was pretty much fun to navigate through the code of complete open stack & then finding all the different binary it produces and then investigating the octavia-controller's code.

pkg/ingress/controller/controller.go
```diff
@@ -728,8 +740,13 @@ func (c *Controller) ensureIngress(ing *nwv1.Ingress) error {
 
 	// Create listener
 	sourceRanges := getStringFromIngressAnnotation(ing, IngressAnnotationSourceRangesKey, "0.0.0.0/0")
+	timeoutClientData := getIntFromIngressAnnotation(ing, IngressAnnotationTimeoutClientData, 50000)
+	timeoutMemberConnect := getIntFromIngressAnnotation(ing, IngressAnnotationTimeoutMemberConnect, 5000)
+	timeoutMemberData := getIntFromIngressAnnotation(ing, IngressAnnotationTimeoutMemberData, 50000)
+	timeoutTCPInspect := getIntFromIngressAnnotation(ing, IngressAnnotationTimeoutTCPInspect, 0)
```
pierreprinetti, line 746: I think that one requirement for this change should be: **if the new annotations are not set in config, the Ingress behaves exactly as it was before this patch**.

I am worried that setting an explicit default timeout changes the default behaviour. If no annotation is set, `TimeoutXxx` fields in `listeners.CreateOpts` should be left nil as they were previously.

sakshi-1505: Understandable, thanks @pierreprinetti I have made the requested changes

pierreprinetti (commented):
I am worried that in case the annotation input is malformed, it's very hard to diagnose an anomalous behaviour.

I think there is at least one case we should take into consideration: the case where the annotation is not properly formatted as an int (for example: a rogue character, maybe a space, slipped in). `strconv.Atoi` would fail parsing and the Ingress would be created with the default Octavia setting, without any notice to the user.

At a minimum, I think that we should log if the parsing of an annotation failed, with a log line that states exactly which annotation failed parsing.

What do you think?

pkg/ingress/controller/controller.go
```diff
@@ -108,6 +108,18 @@ const (
...
+	// IngressAnnotationTimeoutClientData is the timeout for client data in ms
+	IngressAnnotationTimeoutClientData = "octavia.ingress.kubernetes.io/timeout-client-data"
+
+	// IngressAnnotationTimeoutClientData is the timeout for member data in ms
+	IngressAnnotationTimeoutMemberData = "octavia.ingress.kubernetes.io/timeout-member-data"
+
+	// IngressAnnotationTimeoutMemberConnet is the timeout for memer connect in ms
+	IngressAnnotationTimeoutMemberConnect = "octavia.ingress.kubernetes.io/timeout-member-connect"
+
+	// IngressAnnotationTimeoutMemberConnet is the timeout for memer connect in ms
```
pierreprinetti, line 120: Maybe for completeness you could mention what's the default value (I suppose it's an Octavia setting?)

sakshi-1505: Should I just document default values here?

pkg/ingress/controller/controller.go
```diff
@@ -1017,6 +1034,19 @@ func getStringFromIngressAnnotation(ingress *nwv1.Ingress, annotationKey string,
 	return defaultValue
 }
 
+// maybeGetIntFromIngressAnnotation searches a given Ingress for a specific annotationKey and either returns the
+// annotation's value or a specified defaultSetting
+func maybeGetIntFromIngressAnnotation(ingress *nwv1.Ingress, annotationKey string) *int {
+	if annotationValue, ok := ingress.Annotations[annotationKey]; ok {
+		returnValue, err := strconv.Atoi(annotationValue)
+		if err != nil {
+			return nil
```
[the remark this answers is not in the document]
pierreprinetti: WRT the logging level: I'd keep it consistent with this: https://github.com/kubernetes/cloud-provider-openstack/blob/60ad70c0b53d545f323df71b4603bfc8a98dd2df/pkg/openstack/loadbalancer.go#L693

Note that this getBool function also has a logging issue, and I think @FavourEva is on it.

pkg/ingress/controller/controller.go
sakshi-1505: It's not an octavia config argument but it's available for LB, when the controller starts to create listener it doesn't pass this value post so that LoadBalancer default values can get applied, so basically we are passing this in CreateOpts to let override the defaults available in LB. I will still add the docs of the CLI configs.

pkg/ingress/controller/controller.go
```diff
@@ -1017,6 +1034,19 @@ func getStringFromIngressAnnotation(ingress *nwv1.Ingress, annotationKey string,
 	return defaultValue
 }
 
+// maybeGetIntFromIngressAnnotation searches a given Ingress for a specific annotationKey and either returns the
+// annotation's value or a specified defaultSetting
```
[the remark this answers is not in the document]
sakshi-1505: fixed

pkg/ingress/controller/controller.go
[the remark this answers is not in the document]
sakshi-1505: fixed

pkg/ingress/controller/controller.go
sakshi-1505: fixed

sakshi-1505:
@pierreprinetti please re-review

pkg/ingress/controller/controller.go
```diff
@@ -108,6 +108,22 @@ const (
 	// IngressControllerTag is added to the related resources.
 	IngressControllerTag = "octavia.ingress.kubernetes.io"
 
+	// IngressAnnotationTimeoutClientData is the timeout for client data in ms. Defaults to 50000
+	// Refer https://docs.openstack.org/octavia/latest/configuration/configref.html
```
pierreprinetti, line 112: Adding a link is a nice touch!
I would also mention the exact configuration switch though.

```suggestion
	// IngressAnnotationTimeoutClientData is the timeout for frontend client inactivity.
	// If not set, this value defaults to the Octavia configuration key `timeout_client_data`.  
	// Refer to https://docs.openstack.org/octavia/latest/configuration/configref.html#haproxy_amphora.timeout_client_data
```
timeout_client_data

pkg/ingress/controller/controller.go
```diff
@@ -108,6 +108,22 @@ const (
...
+	IngressAnnotationTimeoutClientData = "octavia.ingress.kubernetes.io/timeout-client-data"
+
+	// IngressAnnotationTimeoutClientData is the timeout for member data in ms. Defaults to 50000
+	// Refer https://docs.openstack.org/octavia/latest/configuration/configref.html
```
pierreprinetti, line 116:
```suggestion
	// IngressAnnotationTimeoutMemberData is the timeout for backend member inactivity.
	// If not set, this value defaults to the Octavia configuration key `timeout_member_data`.  
	// Refer to https://docs.openstack.org/octavia/latest/configuration/configref.html#haproxy_amphora.timeout_member_data
```

pkg/ingress/controller/controller.go
```diff
@@ -108,6 +108,22 @@ const (
...
+	IngressAnnotationTimeoutMemberData = "octavia.ingress.kubernetes.io/timeout-member-data"
+
+	// IngressAnnotationTimeoutMemberConnect is the timeout for memer connect in ms. Defaults to 50000
+	// Refer https://docs.openstack.org/octavia/latest/configuration/configref.html
```
pierreprinetti, line 120:
```suggestion
	// IngressAnnotationTimeoutMemberConnect is the timeout for backend member connection.
	// If not set, this value defaults to the Octavia configuration key `timeout_member_connect`.  
	// Refer to https://docs.openstack.org/octavia/latest/configuration/configref.html#haproxy_amphora.timeout_member_connect
```

pkg/ingress/controller/controller.go
```diff
@@ -108,6 +108,22 @@ const (
...
+	IngressAnnotationTimeoutMemberConnect = "octavia.ingress.kubernetes.io/timeout-member-connect"
+
+	// IngressAnnotationTimeoutMemberConnect is the timeout for memer connect in ms. Defaults to 0
+	// Refer https://docs.openstack.org/octavia/latest/configuration/configref.html
```
pierreprinetti, line 124:
```suggestion
	// IngressAnnotationTimeoutTCPInspect is the time to wait for TCP packets for content inspection.
	// If not set, this value defaults to the Octavia configuration key `timeout_tcp_inspect`.  
	// Refer to https://docs.openstack.org/octavia/latest/configuration/configref.html#haproxy_amphora.timeout_tcp_inspect
```

pkg/ingress/controller/controller.go
```diff
@@ -1017,6 +1038,23 @@ func getStringFromIngressAnnotation(ingress *nwv1.Ingress, annotationKey string,
...
+	if annotationValue, ok := ingress.Annotations[annotationKey]; ok {
+		klog.V(4).Infof("Found a Service Annotation for key: %v", annotationKey)
+		returnValue, err := strconv.Atoi(annotationValue)
+		if err != nil {
+			klog.V(4).Infof("Invalid integer found on Service Annotation: %v = %v", annotationKey, annotationValue)
+			return nil
+		}
+		return &returnValue
+	}
+	klog.V(4).Infof("Could not find a Service Annotation; falling back to default setting for annotation %v", annotationKey)
```
pierreprinetti, line 1054: :+1:

pierreprinetti (changes requested):

pkg/ingress/controller/controller.go
```diff
@@ -108,6 +108,22 @@ const (
...
+
+	// IngressAnnotationTimeoutClientData is the timeout for member data in ms. Defaults to 50000
+	// Refer https://docs.openstack.org/octavia/latest/configuration/configref.html
+	IngressAnnotationTimeoutMemberData = "octavia.ingress.kubernetes.io/timeout-member-data"
+
+	// IngressAnnotationTimeoutMemberConnect is the timeout for memer connect in ms. Defaults to 50000
+	// Refer https://docs.openstack.org/octavia/latest/configuration/configref.html
+	IngressAnnotationTimeoutMemberConnect = "octavia.ingress.kubernetes.io/timeout-member-connect"
+
+	// IngressAnnotationTimeoutMemberConnect is the timeout for memer connect in ms. Defaults to 0
```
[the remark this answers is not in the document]
sakshi-1505: Yes, thanks for pointing out the typo. I will fix this right away.

sakshi-1505:
@pierreprinetti Please review again

pierreprinetti (commented):
Looks good to me.
Please run `go fmt ./...` before we ping a maintainer again for review

sakshi-1505:
> Looks good to me.
> Please run `go fmt ./...` before we ping a maintainer again for review

Done, it was just a matter of few seconds before I squashed it 🙈

sakshi-1505:
please do check again @pierreprinetti

jichenjc:
/approve

thanks for the update~

pkg/ingress/controller/openstack/octavia.go
```diff
@@ -340,10 +340,14 @@ func (os *OpenStack) EnsureListener(name string, lbID string, secretRefs []strin
...
-			ProtocolPort:   80, // Ingress Controller only supports http/https for now
-			LoadbalancerID: lbID,
+			Name:                 name,
+			Protocol:             "HTTP",
+			ProtocolPort:         80, // Ingress Controller only supports http/https for now
+			LoadbalancerID:       lbID,
+			TimeoutClientData:    timeoutClientData,
+			TimeoutMemberData:    timeoutMemberData,
+			TimeoutMemberConnect: timeoutMemberConnect,
+			TimeoutTCPInspect:    timeoutTCPInspect,
```
dulek, line 350: @sakshi-1505: This is missing the update case, these properties can be updated, just like the `AllowedCIDRs` option is in lines 368-371. Can you update that too?

PR #2437, opened 2023-10-16
Improve logging for boolean Service Annotations

FavourEva:
Improve logging for bool service annotation

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -677,6 +677,7 @@ func getIntFromServiceAnnotation(service *corev1.Service, annotationKey string, 
 }
 
 // getBoolFromServiceAnnotation searches a given v1.Service for a specific annotationKey and either returns the annotation's boolean value or a specified defaultSetting
+// If the annotation is not found or is not a valid boolean ("true" or "false"), it falls back to the defaultSetting and logs a message accordingly.
 func getBoolFromServiceAnnotation(service *corev1.Service, annotationKey string, defaultSetting bool) bool {
 	klog.V(4).Infof("getBoolFromServiceAnnotation(%s/%s, %v, %v)", service.Namespace, service.Name, annotationKey, defaultSetting)
 	if annotationValue, ok := service.Annotations[annotationKey]; ok {
@@ -687,7 +688,8 @@ func getBoolFromServiceAnnotation(service *corev1.Service, annotationKey string,
 		case "false":
 			returnValue = false
 		default:
-			returnValue = defaultSetting
+			klog.Infof("Found a non-boolean Service Annotation: %v = %v (falling back to default setting: %v)", annotationKey, annotationValue, defaultSetting)
+			return defaultSetting
 		}
 
 		klog.V(4).Infof("Found a Service Annotation: %v = %v", annotationKey, returnValue)


pkg/openstack/loadbalancer.go
```diff
@@ -687,13 +688,14 @@ func getBoolFromServiceAnnotation(service *corev1.Service, annotationKey string,
...
 		default:
+			klog.Error("Invalid Service Annotation: %v = %v (falling back to default setting: %v)", annotationKey, annotationValue, defaultSetting)
 			returnValue = defaultSetting
 		}
 
 		klog.V(4).Infof("Found a Service Annotation: %v = %v", annotationKey, returnValue)
 		return returnValue
 	}
-	klog.V(4).Infof("Could not find a Service Annotation; falling back to default setting: %v = %v", annotationKey, defaultSetting)
+	klog.V(4).Infof("Found a non-boolean Service Annotation; falling back to default setting: %v = %v", annotationKey, defaultSetting)
```
sakshi-1505, line 698: I don't think it's the right place for non-boolean, if the func execution ever reaches here. It means that we couldn't find any service annotation for the following key. The log message currently indicates that if found an annotation when it didn't.

pkg/openstack/loadbalancer.go
```diff
@@ -687,13 +688,14 @@ func getBoolFromServiceAnnotation(service *corev1.Service, annotationKey string,
 		case "false":
 			returnValue = false
 		default:
+			klog.Error("Invalid Service Annotation: %v = %v (falling back to default setting: %v)", annotationKey, annotationValue, defaultSetting)
```
sakshi-1505, line 691: Here you should log that, `found a non-boolean annotation`

sakshi-1505 (changes requested):

pkg/openstack/loadbalancer.go
```diff
@@ -687,6 +688,7 @@ func getBoolFromServiceAnnotation(service *corev1.Service, annotationKey string,
 		case "false":
 			returnValue = false
 		default:
+			klog.Error("Found a non-boolean Service Annotation: %v = %v (falling back to default setting: %v)", annotationKey, annotationValue, defaultSetting)
```
pierreprinetti, line 691: Good! Now return `defaultSetting` after this line.

FavourEva: > Good! Now return `defaultSetting` after this line.

Done! Is it  cool now? @pierreprinetti

pierreprinetti: Perfect! Thanks.

pierreprinetti (approved):
/lgtm

pierreprinetti:
/lgtm cancel

Wait. I am not sure that `klog.Error` is what we want here. @jichenjc @kayrus what would be the appropriate logging level here for consistency with the rest of the logic?

FavourEva:
> /lgtm cancel
> 
> Wait. I am not sure that `klog.Error` is what we want here. @jichenjc @kayrus what would be the appropriate logging level here for consistency with the rest of the logic?

`klog.warningf`?

jichenjc:
are we good enough to go or need fix the log warning suggestion?

FavourEva:
> are we good enough to go or need fix the log warning suggestion?

We actually need a suggestion on what the appropriate logging level should be or is `klog.Error` what we want there? @jichenjc

jichenjc:
I can't find where's orignial info but my understanding is 

Log.info => general info for reference
Log.warning => warning, potentially lead to error in the future or something worthy attention
Log.error => definitely something wrong and already cause issues

FavourEva:
> I can't find where's orignial guideline from golang/k8s cmmunity but my understanding is
> 
> Log.info => general info for reference Log.warning => warning, potentially lead to error in the future or something worthy attention Log.error => definitely something wrong and already cause issues

I guess the logging level should align with the requirements and severity of the situation. Since falling back to default setting is an expected behaviour, I think using a lower log level `klog.Info` would be better. What do you think @jichenjc

jichenjc:
>>>Since falling back to default setting is an expected behaviour

if it is fall back to default (normal way) then yes, Info is better

FavourEva:
> /lgtm cancel
> 
> Wait. I am not sure that `klog.Error` is what we want here. @jichenjc @kayrus what would be the appropriate logging level here for consistency with the rest of the logic?

Resolved! @pierreprinetti Please review

pierreprinetti (approved):
/lgtm

pierreprinetti:
/assign jichenjc
when you have time :pray:

FavourEva:
@dulek @jichenjc Could you please consider merging my PR when you have a moment? I believe its ready. Thanks a lot

jichenjc:
/approve 

sorry for the delay :)

dulek:
This is a neat improvement, thank you!

PR #2438, opened 2023-10-18
Get member subnet id method mod

KingDaemonX:
mod : removed the unused service config argument and modified the test to fit new function modification

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1433,7 +1433,7 @@ func (lbaas *LbaasV2) buildListenerCreateOpt(port corev1.ServicePort, svcConf *s
 }
 
 // getMemberSubnetID gets the configured member-subnet-id from the different possible sources.
-func (lbaas *LbaasV2) getMemberSubnetID(service *corev1.Service, svcConf *serviceConfig) (string, error) {
+func (lbaas *LbaasV2) getMemberSubnetID(service *corev1.Service) (string, error) {
 	// Get Member Subnet from Service Annotation
 	memberSubnetIDAnnotation := getStringFromServiceAnnotation(service, ServiceAnnotationLoadBalancerMemberSubnetID, "")
 	if memberSubnetIDAnnotation != "" {
@@ -1532,7 +1532,7 @@ func (lbaas *LbaasV2) checkServiceUpdate(service *corev1.Service, nodes []*corev
 	svcConf.supportLBTags = openstackutil.IsOctaviaFeatureSupported(lbaas.lb, openstackutil.OctaviaFeatureTags, lbaas.opts.LBProvider)
 
 	// Find subnet ID for creating members
-	memberSubnetID, err := lbaas.getMemberSubnetID(service, svcConf)
+	memberSubnetID, err := lbaas.getMemberSubnetID(service)
 	if err != nil {
 		return fmt.Errorf("unable to get member-subnet-id, %w", err)
 	}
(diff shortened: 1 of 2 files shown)


pkg/openstack/loadbalancer_test.go
```diff
@@ -1217,3 +1216,89 @@ func Test_getStringFromServiceAnnotation(t *testing.T) {
...
+			}
+
+			got, err := lbaas.getMemberSubnetID(tt.service)
+			if err != nil {
+				assert.EqualError(t, err, tt.wantErr)
+			} else {
+				if assert.NoError(t, err) {
+					assert.Equal(t, tt.want, got)
+				}
+			}
```
pierreprinetti, line 1301: Take a step back and look at what happens here. Your assertions are inside an `if` block. The `if` block determines which assertions will be run based on the content of **`err`**. Sometimes you need to have different assertions based on the expected values... but based on the _actual_ values? That looks suspicious.

Now look closer at what happens line by line, while you keep the variable values in your head. Imagine you expect `(tt.want == "", tt.wantErr == io.EOF)`, and your function returns `(got == "", err == nil)`. What would happen in this case?

pkg/openstack/loadbalancer.go
```diff
@@ -1433,7 +1433,7 @@ func (lbaas *LbaasV2) buildListenerCreateOpt(port corev1.ServicePort, svcConf *s
 }
 
 // getMemberSubnetID gets the configured member-subnet-id from the different possible sources.
-func (lbaas *LbaasV2) getMemberSubnetID(service *corev1.Service, svcConf *serviceConfig) (string, error) {
+func (lbaas *LbaasV2) getMemberSubnetID(service *corev1.Service) (string, error) {
```
pierreprinetti, line 1436: I'll defer to the maintainers to chime in on this change; `svcConf` is indeed unused, but I don't know why it was there in the first place.

CC @jichenjc @kayrus

pierreprinetti (changes requested):

pkg/openstack/loadbalancer_test.go
KingDaemonX: Don’t mind that part 
I will change it, I don’t know if to tag the PR a draft or suggestion to modify the function 
I know it can’t be merged yet because i will need to rebase and restructure it if the modification

KingDaemonX: I made the push before your correction in that other PR and I was waiting for a review to know if the function modification looks good 
I’d have to close the PR if the removal of the `serviceConfig` isn’t authorised since there is an existing test already for the function and complete the entire modification when it is authorised 

so I apologize for the repetition, it wasn’t intentional as I have learnt the best usecase already from you and will not intentionally repeat same mistake twice

pierreprinetti: A good workflow may be:
1. Submit a PR
2. Make the required edits until you get a LGTM
3. Rebase
4. you'll get a new LGTM
5. wait for approval
6. merge

Conflicts are extraordinarily frequent these days, because several people are submitting PRs on the same test files. That's not a normal situation, and we have to bear with that for now.

pierreprinetti: In general, you can defer rebasing until the very last moment.

pierreprinetti: You don't have to turn your PR into WIP, unless you realise that it is NOT ready for review and you want to avoid wasting the reviewer's time. This is NOT the case here: your PR is tidy and definitely ready for reviews.

I have asked you to turn another PR of yours to WIP, just because you mentioned that it is not ready for review.

pkg/openstack/loadbalancer.go
dulek: I've took a quick look at the `git blame`, looks like it was there by mistake:

PR #2439, opened 2023-10-18
[occm] feat : add load balancer listener tag using service annotation

### LINKED ISSUE #2327: [occm] - add annotation for custom octavia listener tags
/kind feature

Hi team,
sometimes we need to enable custom configuration for loadbalancer listener on the loadbalancing infrastructure backend,
like for example special handling of UDP datagrams for Octobus, or special settings for HW acceleration for Hana Cloud.

We implement these special settings in Octavia using listener tags. Customer sipmly adds a tag to the listener and the provider driver will configure the necessary settings on the loadbalancing backend.

Could we make such custom load balancer listener tags configurable through loadbalancer service annotations?
This would greatly improve customer's experience, as they would not have to go through the extra step to add the tag after the listener is created.

The tags are documented here:
https://documentation.global.cloud.sap/docs/customer/networking/load-balancers/#enhanced-service-definition---esd-ccloud-specific

FYI @kayrus

yorubad-dev:

/kind feat

this PR solves the [ add annotation for custom octavia listener tags problem](https://github.com/kubernetes/cloud-provider-openstack/issues/2327)

this is currently a draft PR that needs review, also i am a bit confuse on where to plug the method 😢

fixes #2327

### DIFF
--- a/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
+++ b/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
@@ -236,6 +236,11 @@ Request Body:
   This annotation is automatically added and it contains the floating ip address of the load balancer service.
   When using `loadbalancer.openstack.org/hostname` annotation it is the only place to see the real address of the load balancer.
 
+- `loadbalancer.openstack.org/custom-tags`
+
+  Allows to specify custom tags that all load balancer resources for that Service will be tagged with. 
+  Tags are arbitrary strings, to specify multiple tags separate them using a comma `,` in the annotation.
+
 ### Switching between Floating Subnets by using preconfigured Classes
 
 If you have multiple `FloatingIPPools` and/or `FloatingIPSubnets` it might be desirable to offer the user logical meanings for `LoadBalancers` like `internetFacing` or `DMZ` instead of requiring the user to select a dedicated network or subnet ID at the service object level as an annotation.
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -86,6 +86,7 @@ const (
 	ServiceAnnotationLoadBalancerXForwardedFor        = "loadbalancer.openstack.org/x-forwarded-for"
 	ServiceAnnotationLoadBalancerFlavorID             = "loadbalancer.openstack.org/flavor-id"
 	ServiceAnnotationLoadBalancerAvailabilityZone     = "loadbalancer.openstack.org/availability-zone"
+	ServiceAnnotationLoadBalancerCustomTags           = "loadbalancer.openstack.org/custom-tags"
 	// ServiceAnnotationLoadBalancerEnableHealthMonitor defines whether to create health monitor for the load balancer
 	// pool, if not specified, use 'create-monitor' config. The health monitor can be created or deleted dynamically.
 	ServiceAnnotationLoadBalancerEnableHealthMonitor         = "loadbalancer.openstack.org/enable-health-monitor"
@@ -469,6 +470,7 @@ func (lbaas *LbaasV2) createOctaviaLoadBalancer(name, clusterName string, servic
 
 	if svcConf.supportLBTags {
 		createOpts.Tags = []string{svcConf.lbName}
+		createOpts.Tags = append(createOpts.Tags, getCustomLoadBalancerTags(service, svcConf)...)
 	}
 
 	if svcConf.flavorID != "" {
@@ -508,8 +510,8 @@ func (lbaas *LbaasV2) createOctaviaLoadBalancer(name, clusterName string, servic
 
 	if !lbaas.opts.ProviderRequiresSerialAPICalls {
 		for portIndex, port := range service.Spec.Ports {
-			listenerCreateOpt := lbaas.buildListenerCreateOpt(port, svcConf, cpoutil.Sprintf255(listenerFormat, portIndex, name))
-			members, newMembers, err := lbaas.buildBatchUpdateMemberOpts(port, nodes, svcConf)
+			listenerCreateOpt := lbaas.buildListenerCreateOpt(port, svcConf, service, cpoutil.Sprintf255(listenerFormat, portIndex, name))
+			members, newMembers, err := lbaas.buildBatchUpdateMemberOpts(port, nodes, svcConf, service)
 			if err != nil {
 				return nil, err
 			}
@@ -604,6 +606,23 @@ func (lbaas *LbaasV2) GetLoadBalancerName(_ context.Context, clusterName string,
 // getLoadBalancerLegacyName returns the legacy load balancer name for backward compatibility.
 func (lbaas *LbaasV2) getLoadBalancerLegacyName(_ context.Context, _ string, service *corev1.Service) string {
 	return cloudprovider.DefaultLoadBalancerName(service)
+}
+
+// Returns a list of custom loadbalancer tags for the supported service resources.
+func getCustomLoadBalancerTags(service *corev1.Service, svcConf *serviceConfig) []string {
+	if !svcConf.supportLBTags {
+		return nil
+	}
+
+	annotationVal := getStringFromServiceAnnotation(service, ServiceAnnotationLoadBalancerCustomTags, "")
+
+	tags := strings.Split(annotationVal, ",")
+
+	for i, tag := range tags {
+		tags[i] = strings.TrimSpace(tag)
+	}
+
+	return tags
 }
 
 // The LB needs to be configured with instance addresses on the same
@@ -739,7 +758,7 @@ func isPortMember(port PortWithPortSecurity, IP string, subnetID string) bool {
 }
 
 // applyNodeSecurityGroupIDForLB associates the security group with the ports being members of the LB on the nodes.
-func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, svcConf *serviceConfig, nodes []*corev1.Node, sg string) error {
+func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, service *corev1.Service, svcConf *serviceConfig, nodes []*corev1.Node, sg string) error {
 	for _, node := range nodes {
 		serverID, _, err := instanceIDFromProviderID(node.Spec.ProviderID)
 		if err != nil {
@@ -784,6 +803,20 @@ func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, svcConf *
 			res := neutronports.Update(network, port.ID, updateOpts)
 			if mc.ObserveRequest(res.Err) != nil {
 				return fmt.Errorf("failed to update security group for port %s: %v", port.ID, res.Err)
+			}
+
+			tags := getCustomLoadBalancerTags(service, svcConf)
+
+			mc = metrics.NewMetricContext("security_group_tag", "replace")
+			_, err := neutrontags.ReplaceAll(network, "security_groups", port.ID, neutrontags.ReplaceAllOpts{Tags: tags}).Extract()
+			if mc.ObserveRequest(err) != nil {
+				return fmt.Errorf("failed to add tag %s to port %s: %v", tags, port.ID, err)
+			}
+
+			mc = metrics.NewMetricContext("floating_ip_tag", "replace")
+			_, err = neutrontags.ReplaceAll(network, "floatingips", port.ID, neutrontags.ReplaceAllOpts{Tags: tags}).Extract()
+			if mc.ObserveRequest(err) != nil {
+				return fmt.Errorf("failed to add tag %s to port %s of floating_ips: %v", tags, port.ID, err)
 			}
 		}
 	}
@@ -896,10 +929,16 @@ func (lbaas *LbaasV2) deleteOctaviaListeners(lbID string, listenerList []listene
 	return nil
 }
 
-func (lbaas *LbaasV2) createFloatingIP(msg string, floatIPOpts floatingips.CreateOpts) (*floatingips.FloatingIP, error) {
+func (lbaas *LbaasV2) createFloatingIP(msg string, floatIPOpts floatingips.CreateOpts, service *corev1.Service, svcConf *serviceConfig) (*floatingips.FloatingIP, error) {
 	klog.V(4).Infof("%s floating ip with opts %+v", msg, floatIPOpts)
 	mc := metrics.NewMetricContext("floating_ip", "create")
 	floatIP, err := floatingips.Create(lbaas.network, floatIPOpts).Extract()
+
+	tags := getCustomLoadBalancerTags(service, svcConf)
+
+	if _, err := neutrontags.ReplaceAll(lbaas.network, "floatingips", floatIP.ID, neutrontags.ReplaceAllOpts{Tags: tags}).Extract(); err != nil {
+		return nil, fmt.Errorf("failed to add custom tags %s to floatingIPs %s with a projectID (%s)", tags, floatIP.ID, floatIP.ProjectID)
+	}
 	err = PreserveGopherError(err)
 	if mc.ObserveRequest(err) != nil {
 		return floatIP, fmt.Errorf("error creating LB floatingip: %s", err)
@@ -1037,7 +1076,7 @@ func (lbaas *LbaasV2) ensureFloatingIP(clusterName string, service *corev1.Servi
 					svcConf.lbPublicSubnetSpec, svcConf.lbPublicNetworkID)
 				for _, subnet := range foundSubnets {
 					floatIPOpts.SubnetID = subnet.ID
-					floatIP, err = lbaas.createFloatingIP(fmt.Sprintf("Trying subnet %s for creating", subnet.Name), floatIPOpts)
+					floatIP, err = lbaas.createFloatingIP(fmt.Sprintf("Trying subnet %s for creating", subnet.Name), floatIPOpts, service, svcConf)
 					if err == nil {
 						foundSubnet = subnet
 						break
@@ -1054,7 +1093,7 @@ func (lbaas *LbaasV2) ensureFloatingIP(clusterName string, service *corev1.Servi
 					floatIPOpts.SubnetID = svcConf.lbPublicSubnetSpec.subnetID
 				}
 				floatIPOpts.FloatingIP = loadBalancerIP
-				floatIP, err = lbaas.createFloatingIP("Creating", floatIPOpts)
+				floatIP, err = lbaas.createFloatingIP("Creating", floatIPOpts, service, svcConf)
 				if err != nil {
 					return "", err
 				}
@@ -1237,7 +1276,7 @@ func (lbaas *LbaasV2) ensureOctaviaPool(lbID string, name string, listener *list
 		curMembers.Insert(fmt.Sprintf("%s-%s-%d-%d", m.Name, m.Address, m.ProtocolPort, m.MonitorPort))
 	}
 
-	members, newMembers, err := lbaas.buildBatchUpdateMemberOpts(port, nodes, svcConf)
+	members, newMembers, err := lbaas.buildBatchUpdateMemberOpts(port, nodes, svcConf, service)
 	if err != nil {
 		return nil, err
 	}
@@ -1254,6 +1293,8 @@ func (lbaas *LbaasV2) ensureOctaviaPool(lbID string, name string, listener *list
 }
 
 func (lbaas *LbaasV2) buildPoolCreateOpt(listenerProtocol string, service *corev1.Service, svcConf *serviceConfig, name string) v2pools.CreateOpts {
+	customTags := getCustomLoadBalancerTags(service, svcConf)
+
 	// By default, use the protocol of the listener
 	poolProto := v2pools.Protocol(listenerProtocol)
 	if svcConf.enableProxyProtocol {
@@ -1284,13 +1325,16 @@ func (lbaas *LbaasV2) buildPoolCreateOpt(listenerProtocol string, service *corev
 		Protocol:    poolProto,
 		LBMethod:    lbmethod,
 		Persistence: persistence,
+		Tags:        customTags,
 	}
 }
 
 // buildBatchUpdateMemberOpts returns v2pools.BatchUpdateMemberOpts array for Services and Nodes alongside a list of member names
-func (lbaas *LbaasV2) buildBatchUpdateMemberOpts(port corev1.ServicePort, nodes []*corev1.Node, svcConf *serviceConfig) ([]v2pools.BatchUpdateMemberOpts, sets.Set[string], error) {
+func (lbaas *LbaasV2) buildBatchUpdateMemberOpts(port corev1.ServicePort, nodes []*corev1.Node, svcConf *serviceConfig, service *corev1.Service) ([]v2pools.BatchUpdateMemberOpts, sets.Set[string], error) {
 	var members []v2pools.BatchUpdateMemberOpts
 	newMembers := sets.New[string]()
+
+	customTags := getCustomLoadBalancerTags(service, svcConf)
 
 	for _, node := range nodes {
 		addr, err := nodeAddressForLB(node, svcConf.preferredIPFamily)
@@ -1315,6 +1359,7 @@ func (lbaas *LbaasV2) buildBatchUpdateMemberOpts(port corev1.ServicePort, nodes 
 				ProtocolPort: int(port.NodePort),
 				Name:         &node.Name,
 				SubnetID:     memberSubnetID,
+				Tags:         customTags,
 			}
 			if svcConf.healthCheckNodePort > 0 && lbaas.canUseHTTPMonitor(port) {
 				member.MonitorPort = &svcConf.healthCheckNodePort
@@ -1327,13 +1372,13 @@ func (lbaas *LbaasV2) buildBatchUpdateMemberOpts(port corev1.ServicePort, nodes 
 }
 
 // Make sure the listener is created for Service
-func (lbaas *LbaasV2) ensureOctaviaListener(lbID string, name string, curListenerMapping map[listenerKey]*listeners.Listener, port corev1.ServicePort, svcConf *serviceConfig, _ *corev1.Service) (*listeners.Listener, error) {
+func (lbaas *LbaasV2) ensureOctaviaListener(lbID string, name string, curListenerMapping map[listenerKey]*listeners.Listener, port corev1.ServicePort, svcConf *serviceConfig, service *corev1.Service) (*listeners.Listener, error) {
 	listener, isPresent := curListenerMapping[listenerKey{
 		Protocol: getListenerProtocol(port.Protocol, svcConf),
 		Port:     int(port.Port),
 	}]
 	if !isPresent {
-		listenerCreateOpt := lbaas.buildListenerCreateOpt(port, svcConf, name)
+		listenerCreateOpt := lbaas.buildListenerCreateOpt(port, svcConf, service, name)
 		listenerCreateOpt.LoadbalancerID = lbID
 
 		klog.V(2).Infof("Creating listener for port %d using protocol %s", int(port.Port), listenerCreateOpt.Protocol)
@@ -1419,7 +1464,7 @@ func (lbaas *LbaasV2) ensureOctaviaListener(lbID string, name string, curListene
 }
 
 // buildListenerCreateOpt returns listeners.CreateOpts for a specific Service port and configuration
-func (lbaas *LbaasV2) buildListenerCreateOpt(port corev1.ServicePort, svcConf *serviceConfig, name string) listeners.CreateOpts {
+func (lbaas *LbaasV2) buildListenerCreateOpt(port corev1.ServicePort, svcConf *serviceConfig, service *corev1.Service, name string) listeners.CreateOpts {
 	listenerCreateOpt := listeners.CreateOpts{
 		Name:         name,
 		Protocol:     listeners.Protocol(port.Protocol),
@@ -1429,6 +1474,8 @@ func (lbaas *LbaasV2) buildListenerCreateOpt(port corev1.ServicePort, svcConf *s
 
 	if svcConf.supportLBTags {
 		listenerCreateOpt.Tags = []string{svcConf.lbName}
+		// add custom tags to LB listener
+		listenerCreateOpt.Tags = append(listenerCreateOpt.Tags, getCustomLoadBalancerTags(service, svcConf)...)
 	}
 
 	if openstackutil.IsOctaviaFeatureSupported(lbaas.lb, openstackutil.OctaviaFeatureTimeout, lbaas.opts.LBProvider) {
@@ -2317,6 +2364,12 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
 			return fmt.Errorf("failed to create Security Group for loadbalancer service %s/%s: %v", apiService.Namespace, apiService.Name, err)
 		}
 		lbSecGroupID = lbSecGroup.ID
+
+		tags := getCustomLoadBalancerTags(apiService, svcConf)
+
+		if _, err := neutrontags.ReplaceAll(lbaas.network, "security-groups", lbSecGroupID, neutrontags.ReplaceAllOpts{Tags: tags}).Extract(); err != nil {
+			return fmt.Errorf("failed to add custom tags %s to security group %s (%s)", tags, lbSecGroupID, lbSecGroupName)
+		}
 	}
 
 	mc := metrics.NewMetricContext("subnet", "get")
@@ -2409,7 +2462,7 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
 		}
 	}
 
-	if err := applyNodeSecurityGroupIDForLB(lbaas.network, svcConf, nodes, lbSecGroupID); err != nil {
+	if err := applyNodeSecurityGroupIDForLB(lbaas.network, apiService, svcConf, nodes, lbSecGroupID); err != nil {
 		return err
 	}
 	return nil
--- a/pkg/openstack/loadbalancer_test.go
+++ b/pkg/openstack/loadbalancer_test.go
@@ -708,6 +708,7 @@ func TestLbaasV2_checkListenerPorts(t *testing.T) {
 		})
 	}
 }
+
 func TestLbaasV2_createLoadBalancerStatus(t *testing.T) {
 	type fields struct {
 		LoadBalancer LoadBalancer
@@ -1831,6 +1832,7 @@ func TestBuildBatchUpdateMemberOpts(t *testing.T) {
 		nodes                   []*corev1.Node
 		port                    corev1.ServicePort
 		svcConf                 *serviceConfig
+		service                 *corev1.Service
 		expectedLen             int
 		expectedNewMembersCount int
 	}{
@@ -1909,7 +1911,7 @@ func TestBuildBatchUpdateMemberOpts(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			lbaas := &LbaasV2{}
-			members, newMembers, err := lbaas.buildBatchUpdateMemberOpts(tc.port, tc.nodes, tc.svcConf)
+			members, newMembers, err := lbaas.buildBatchUpdateMemberOpts(tc.port, tc.nodes, tc.svcConf, tc.service)
 			assert.Len(t, members, tc.expectedLen)
 			assert.NoError(t, err)
 
@@ -2298,6 +2300,7 @@ func TestBuildListenerCreateOpt(t *testing.T) {
 		name              string
 		port              corev1.ServicePort
 		svcConf           *serviceConfig
+		service           *corev1.Service
 		expectedCreateOpt listeners.CreateOpts
 	}{
 		{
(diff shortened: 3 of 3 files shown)


KingDaemonX:
@pierreprinetti can you take a look at this ??

pkg/openstack/loadbalancer.go
```diff
@@ -85,6 +85,7 @@ const (
 	ServiceAnnotationLoadBalancerXForwardedFor        = "loadbalancer.openstack.org/x-forwarded-for"
 	ServiceAnnotationLoadBalancerFlavorID             = "loadbalancer.openstack.org/flavor-id"
 	ServiceAnnotationLoadBalancerAvailabilityZone     = "loadbalancer.openstack.org/availability-zone"
+	ServiceAnnotationLoadBalancerListenerTags         = "loadbalancer.openstack.org/listener-Tags"
```
kayrus, line 88:
```suggestion
	ServiceAnnotationLoadBalancerListenerTags         = "loadbalancer.openstack.org/listener-tags"
```

pkg/openstack/loadbalancer.go
```diff
@@ -605,6 +606,32 @@ func (lbaas *LbaasV2) getLoadBalancerLegacyName(_ context.Context, _ string, ser
 	return cloudprovider.DefaultLoadBalancerName(service)
 }
 
+// this function gets the listeners tags from the service annotation and   
+func (lbaas *LbaasV2) customLoadBalanceListenerTag(service *corev1.Service) ([]string, error) {
+	var listener listeners.Listener
+	annotationVal := getStringFromServiceAnnotation(service, ServiceAnnotationLoadBalancerListenerTags, "")
+
+	listOfTags := strings.Split(annotationVal, "=")
```
kayrus, line 614:
```suggestion
	listOfTags := strings.Split(annotationVal, ",")
```

pkg/openstack/loadbalancer.go
```diff
@@ -605,6 +606,32 @@ func (lbaas *LbaasV2) getLoadBalancerLegacyName(_ context.Context, _ string, ser
...
+
+	if annotationVal == "" {
+		return listener.Tags, fmt.Errorf("annotation is missing or empty for %q", service.Name)
+	}
+
+	if annotationVal != "" {
```
kayrus, line 620: this condition doesn't make sense, please remove it.

kayrus (commented):
thanks for the PR. Is it still WIP? If so, please add documentation and tests. See also some major findings

pkg/openstack/loadbalancer.go
```diff
@@ -605,6 +606,32 @@ func (lbaas *LbaasV2) getLoadBalancerLegacyName(_ context.Context, _ string, ser
...
+		for _, tags := range listOfTags {
+			tag := strings.Split(strings.TrimSpace(tags), "=")
+
+			if len(tag) == 2 {
+				listener.Tags = append(listener.Tags, tags)
+			} else {
+				return listener.Tags, fmt.Errorf("invalid tag format: %s", tags)
+			}
```
pierreprinetti, line 628: This code doesn't allow for a tag to contain `=` in the value. Is that expected? Is this restriction necessary?

If not, and a `=` is an acceptable character in this context, then consider using [`strings.Cut`](https://pkg.go.dev/strings#Cut) instead (see [this example](https://go.dev/play/p/DG3gj7BShcE)).

pierreprinetti:
Also: please don't forget to add tests! At a minimum, a unit test to exercise tag splitting (with various configurations of spaces in it) would probably be a good idea

pkg/openstack/loadbalancer.go
```diff
@@ -605,6 +606,32 @@ func (lbaas *LbaasV2) getLoadBalancerLegacyName(_ context.Context, _ string, ser
 	return cloudprovider.DefaultLoadBalancerName(service)
 }
 
+// this function gets the listeners tags from the service annotation and   
+func (lbaas *LbaasV2) customLoadBalanceListenerTag(service *corev1.Service) ([]string, error) {
```
sakshi-1505, line 610:
```suggestion
func (lbaas *LbaasV2) customLoadBalancerListenerTag(service *corev1.Service) ([]string, error) {
```

This belongs to an LB object

pkg/openstack/loadbalancer.go
sakshi-1505: Why is this marked as resolved?

pkg/openstack/loadbalancer.go
sakshi-1505, line 610: I don't see this function being called anywhere, can we add WIP in the title since it's not yet complete?

KingDaemonX: Did you read the previous comment or my text on the commit before adding yours ?

KingDaemonX: I’m fully aware and thank you for pointing out the almost invisible to see typo

pkg/openstack/loadbalancer.go
KingDaemonX: because i fixed it in the code 
is there an issue about the resolving of the comment after fix ??

pkg/openstack/loadbalancer.go
KingDaemonX: i understand you want to help but i think it's also great to read through the messages and have a grasp of the whole idea and phase of the current work, look into previous reviewer comment, so as not to repeat thesame thing they commented instead of looking only at the code

pkg/openstack/loadbalancer.go
sakshi-1505: I don't see this fixed in the code, you would want to recheck if you even pushed the commit or now https://github.com/kubernetes/cloud-provider-openstack/blob/97a664a539ff6cfc9414068da04464f540c9c7a0/pkg/openstack/loadbalancer.go#L620

KingDaemonX: you should see it's a WIP and i haven't made recent push of the changes i made
thank you helping 
i appreciate you

pkg/openstack/loadbalancer.go
KingDaemonX: i actually took your comment in good faith and only pointed that it'd be great to not repeat others comment because of bulky messages in the PR is kinda confusing and repetitive ones are overwhelming 

also, The typo is almost invisible and even the previous two reviewer  @pierreprinetti and @kayrus missed it
these two are almost impossible to notice the difference `customLoadBalanceListenerTag` and `customLoadBalancerListenerTag`

so i am good with you and you probably took my reviews out of context
i'm sorry if you feel different

KingDaemonX:
@pierreprinetti @kayrus  please re-review

i am working currently working on the test case, hence the WIP

pkg/openstack/loadbalancer.go
```diff
@@ -467,7 +468,12 @@ func (lbaas *LbaasV2) createOctaviaLoadBalancer(name, clusterName string, servic
 	}
 
 	if svcConf.supportLBTags {
-		createOpts.Tags = []string{svcConf.lbName}
+		customTags, err := lbaas.customLoadBalancerListenerTag(service, svcConf)
+		if err != nil {
+			return nil, err
```
dulek, line 473: Should this be fatal? Honestly I'm not 100% sure. Probably it doesn't hurt it currently is.

pkg/openstack/loadbalancer.go
```diff
@@ -605,6 +611,35 @@ func (lbaas *LbaasV2) getLoadBalancerLegacyName(_ context.Context, _ string, ser
...
+	for _, tags := range listOfTags {
+		tags = strings.TrimSpace(tags)
+		tagKey, tagValue, ok := strings.Cut(tags, "=")
+		if !ok {
+			continue
+		}
+
+		if tagKey == "" || tagValue == "" {
+			return listenerTags, fmt.Errorf("invalid tagging format: %q", tags)
+		}
```
dulek, line 635: Tags [are just arbitrary strings in Octavia](https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=create-listener-detail#id3). What's the point of forcing them to have `key=value` format?

pkg/openstack/loadbalancer.go
```diff
@@ -646,13 +681,13 @@ func nodeAddressForLB(node *corev1.Node, preferredIPFamily corev1.IPFamily) (str
...
 	if annotationValue, ok := service.Annotations[annotationKey]; ok {
-		//if there is an annotation for this setting, set the "setting" var to it
+		// if there is an annotation for this setting, set the "setting" var to it
 		// annotationValue can be empty, it is working as designed
 		// it makes possible for instance provisioning loadbalancer without floatingip
 		klog.V(4).Infof("Found a Service Annotation: %v = %v", annotationKey, annotationValue)
 		return annotationValue
 	}
-	//if there is no annotation, set "settings" var to the value from cloud config
+	// if there is no annotation, set "settings" var to the value from cloud config
```
dulek, line 690: These changes are unrelated.

pkg/openstack/loadbalancer.go
```diff
@@ -2067,7 +2102,7 @@ func (lbaas *LbaasV2) EnsureLoadBalancer(ctx context.Context, clusterName string
 }
 
 func (lbaas *LbaasV2) listSubnetsForNetwork(networkID string, tweak ...TweakSubNetListOpsFunction) ([]subnets.Subnet, error) {
-	var opts = subnets.ListOpts{NetworkID: networkID}
+	opts := subnets.ListOpts{NetworkID: networkID}
```
dulek, line 2105: Unrelated change, please remove.

dulek (changes requested):
This isn't adding tags to load balancer listeners, only to the load balancer itself.

In my opinion we should allow tagging of all the LB resources being created, not only load balancers, so also: listeners, pools, members, health monitors and security groups. Search for "CreateOpts" in loadbalancer.go to find creations of the other resources.

We also need to account for updates - it should be possible to update the set of tags in the Service annotation and CPO should then update all the LB resources to reflect that. A tricky part here is we might want to keep the tags created by the user intact, which triggers a lost update problem for us, but let's deal with that later.

KingDaemonX:
> This isn't adding tags to load balancer listeners, only to the load balancer itself.
> 
> In my opinion we should allow tagging of all the LB resources being created, not only load balancers, so also: listeners, pools, members, health monitors and security groups. Search for "CreateOpts" in loadbalancer.go to find creations of the other resources.
> 
> We also need to account for updates - it should be possible to update the set of tags in the Service annotation and CPO should then update all the LB resources to reflect that. A tricky part here is we might want to keep the tags created by the user intact, which triggers a lost update problem for us, but let's deal with that later.

it just got really wide and clear to me now, thank you

pkg/openstack/loadbalancer.go
```diff
@@ -467,7 +468,12 @@ func (lbaas *LbaasV2) createOctaviaLoadBalancer(name, clusterName string, servic
 	}
 
 	if svcConf.supportLBTags {
-		createOpts.Tags = []string{svcConf.lbName}
+		customTags, err := lbaas.customLoadBalancerListenerTag(service, svcConf)
```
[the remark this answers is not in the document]
KingDaemonX: renamed :+1:

pkg/openstack/loadbalancer.go
```diff
@@ -605,6 +611,35 @@ func (lbaas *LbaasV2) getLoadBalancerLegacyName(_ context.Context, _ string, ser
...
 
+// Returns a list of custom load balancer listener tags for the specified  service.
+func (lbaas *LbaasV2) customLoadBalancerListenerTag(service *corev1.Service, svcConf *serviceConfig) ([]string, error) {
+	var listenerTags []string
+	annotationVal := getStringFromServiceAnnotation(service, ServiceAnnotationLoadBalancerListenerTags, "")
+
+	if annotationVal == "" {
+		listenerTags = append(listenerTags, svcConf.lbName)
+		return listenerTags, nil
+	}
```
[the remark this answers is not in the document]
KingDaemonX: removed :+1:

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
```diff
@@ -208,6 +208,12 @@ Request Body:
 
   The name of the loadbalancer availability zone to use. It is ignored if the Octavia version doesn't support availability zones yet.
 
+- `loadbalancer.openstack.org/listener-tags`
+
+  Allows customable loadbalancer tag configurable added during initial stage of loadbalancer creation.
+
+  If this custom tags are created with a `=` to denote key and value pair, the creation of the listener tags are setup in the initial stage easing the process of creation after listener is created
+
```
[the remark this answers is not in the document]
KingDaemonX: yeah, i just saw it 
changes made :+1:

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
```diff
@@ -208,6 +208,12 @@ Request Body:
 
   The name of the loadbalancer availability zone to use. It is ignored if the Octavia version doesn't support availability zones yet.
 
+- `loadbalancer.openstack.org/listener-tags`
```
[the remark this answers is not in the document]
KingDaemonX: done :+1:

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
```diff
@@ -208,6 +208,12 @@ Request Body:
...
+
+  Allows customable loadbalancer tag configurable added during initial stage of loadbalancer creation.
+
+  If this custom tags are created with a `=` to denote key and value pair, the creation of the listener tags are setup in the initial stage easing the process of creation after listener is created
```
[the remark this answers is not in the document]
KingDaemonX: okay, thank you for highlighting that part

pkg/openstack/loadbalancer.go
KingDaemonX: fixed and removed fatality

pkg/openstack/loadbalancer.go
KingDaemonX: i though it will be key-value pair after reading through the opened issue 

image

pkg/openstack/loadbalancer.go
KingDaemonX: fixed :+1:

pkg/openstack/loadbalancer.go
KingDaemonX: fixed :+1:

pkg/openstack/loadbalancer.go
```diff
@@ -422,7 +423,7 @@ func popListener(existingListeners []listeners.Listener, id string) []listeners.
 
 func getSecurityGroupName(service *corev1.Service) string {
 	securityGroupName := fmt.Sprintf("lb-sg-%s-%s-%s", service.UID, service.Namespace, service.Name)
-	//OpenStack requires that the name of a security group is shorter than 255 bytes.
+	// OpenStack requires that the name of a security group is shorter than 255 bytes.
```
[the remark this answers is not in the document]
KingDaemonX: yeah i will stop using the linter on the CPO codebase

KingDaemonX:
> This isn't adding tags to load balancer listeners, only to the load balancer itself.
> 
> In my opinion we should allow tagging of all the LB resources being created, not only load balancers, so also: listeners, pools, members, health monitors and security groups. Search for "CreateOpts" in loadbalancer.go to find creations of the other resources.
> 
> We also need to account for updates - it should be possible to update the set of tags in the Service annotation and CPO should then update all the LB resources to reflect that. A tricky part here is we might want to keep the tags created by the user intact, which triggers a lost update problem for us, but let's deal with that later.

can i create tags field for other resources,for example the security group doesn't have a tag

pkg/openstack/loadbalancer.go
dulek: It's not exactly true, it's just the custom format that might be used by the env of the person raising the issue, but tags are free-form really in Octavia.

pkg/openstack/loadbalancer.go
```diff
@@ -652,7 +669,7 @@ func getStringFromServiceAnnotation(service *corev1.Service, annotationKey strin
 		klog.V(4).Infof("Found a Service Annotation: %v = %v", annotationKey, annotationValue)
 		return annotationValue
 	}
-	//if there is no annotation, set "settings" var to the value from cloud config
+	// if there is no annotation, set "settings" var to the value from cloud config
```
dulek, line 672: Looks like unrelated change.

pkg/openstack/loadbalancer.go
```diff
@@ -1180,6 +1197,9 @@ func (lbaas *LbaasV2) ensureOctaviaPool(lbID string, name string, listener *list
 	if pool == nil {
 		createOpt := lbaas.buildPoolCreateOpt(listener.Protocol, service, svcConf, name)
 		createOpt.ListenerID = listener.ID
+		if svcConf.supportLBTags {
+			createOpt.Tags = append(createOpt.Tags, lbaas.getCustomLoadBalancerTags(service)...)
```
dulek, line 1201: I think this is empty by default, no need to use `append`, you can just set it. Now I wonder - could we just put this inside `buildPoolCreateOpt` instead?

pkg/openstack/loadbalancer.go
```diff
@@ -1400,6 +1420,8 @@ func (lbaas *LbaasV2) buildListenerCreateOpt(port corev1.ServicePort, svcConf *s
 
 	if svcConf.supportLBTags {
 		listenerCreateOpt.Tags = []string{svcConf.lbName}
+		// add custom tags to LB listener
+		listenerCreateOpt.Tags = append(listenerCreateOpt.Tags, lbaas.getCustomLoadBalancerTags(service)...)
```
dulek, line 1424: Nice!

pkg/openstack/loadbalancer.go
```diff
@@ -467,7 +468,7 @@ func (lbaas *LbaasV2) createOctaviaLoadBalancer(name, clusterName string, servic
 	}
 
 	if svcConf.supportLBTags {
-		createOpts.Tags = []string{svcConf.lbName}
+		createOpts.Tags = append(createOpts.Tags, lbaas.getCustomLoadBalancerTags(service)...)
```
dulek, line 471: I think `.Tags` is a nil here, you don't need to use `append()`, just assign directly.

dulek:
Nice, this looks better!

So according to Octavia API reference, the health monitors include tags [1]. I guess gophercloud doesn't support that yet, I'll look into that.

This is still missing tagging of the LB members here [2] and here [3].

The SGs are tagged in a different way - they're part of Neutron service, not Octavia. Something like this should work here [4]:

```
		err := neutrontags.Add(lbaas.network, "securitygroups", lbSecGroupID, "foo").ExtractErr()
```

And we should still think about how do we update resources with tags if annotation changes.

[1] https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=create-listener-detail#create-health-monitor
[2] https://github.com/kubernetes/cloud-provider-openstack/blob/4ff1458c9a10dbd50bca90c1c33b2088745ba5fe/pkg/util/openstack/loadbalancer_serial.go#L71
[3] https://github.com/kubernetes/cloud-provider-openstack/blob/23291097decc2258985e4139e0de9d159f76a79d/pkg/openstack/loadbalancer.go#L1284-L1289
[4] https://github.com/kubernetes/cloud-provider-openstack/blob/23291097decc2258985e4139e0de9d159f76a79d/pkg/openstack/loadbalancer.go#L2270

dulek (changes requested):

KingDaemonX:
> Nice, this looks better!
> 
> So according to Octavia API reference, the health monitors include tags [1]. I guess gophercloud doesn't support that yet, I'll look into that.
> 
> This is still missing tagging of the LB members here [2] and here [3].
> 
> The SGs are tagged in a different way - they're part of Neutron service, not Octavia. Something like this should work here [4]:
> 
> ```
> 		err := neutrontags.Add(lbaas.network, "securitygroups", lbSecGroupID, "foo").ExtractErr()
> ```
> 
> And we should still think about how do we update resources with tags if annotation changes.
> 
> [1] https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=create-listener-detail#create-health-monitor [2]
> 
> https://github.com/kubernetes/cloud-provider-openstack/blob/4ff1458c9a10dbd50bca90c1c33b2088745ba5fe/pkg/util/openstack/loadbalancer_serial.go#L71
> 
> 
> [3]
> https://github.com/kubernetes/cloud-provider-openstack/blob/23291097decc2258985e4139e0de9d159f76a79d/pkg/openstack/loadbalancer.go#L1284-L1289
> 
> 
> [4]
> https://github.com/kubernetes/cloud-provider-openstack/blob/23291097decc2258985e4139e0de9d159f76a79d/pkg/openstack/loadbalancer.go#L2270

for the `[4]`, i'm i suppose to use the custom tags as the `lbSecGroupID`

pkg/openstack/loadbalancer.go
```diff
@@ -606,6 +607,22 @@ func (lbaas *LbaasV2) getLoadBalancerLegacyName(_ context.Context, _ string, ser
...
+
+	listOfTags := strings.Split(annotationVal, ",")
+
+	for _, tags := range listOfTags {
+		tags = strings.TrimSpace(tags)
+
+		listenerTags = append(listenerTags, tags)
+	}
+
+	return listenerTags
```
jeffyjf, line 623: Are there just listener tags? According to your [description](https://github.com/kubernetes/cloud-provider-openstack/pull/2439/files#diff-7f585750e84ff958bbc414045cc178a9fa7f91c86be09356ff549142e1430a24R242). It should also contain loadbalancer tag and pool tag.

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
```diff
@@ -236,6 +236,13 @@ Request Body:
   This annotation is automatically added and it contains the floating ip address of the load balancer service.
   When using `loadbalancer.openstack.org/hostname` annotation it is the only place to see the real address of the load balancer.
 
+- `loadbalancer.openstack.org/custom-tags`
+
+  Allows customable loadbalancer tag configurable added during initial stage of loadbalancer creation.
+  Tags are abitrary strings that can me added to loadbalancer and using this annotation allows adding of one or more custom tag to the `LoadBalancer`, `Listener` and `Pool`
```
jeffyjf, line 242: I noticed that you add all of the tags to all of kinds resources of the LB associate with the Service. Why not split it to different annotation?  Just like: `loadbalancer.openstack.org/loadbalancer-tags`, `loadbalancer.openstack.org/listener-tags` and `loadbalancer.openstack.org/pool-tags`.

dulek: I specifically asked to not do that, I do not think we should end up with 10 more annotations supported for each of the resource we create in CPO.

pkg/openstack/loadbalancer.go
```diff
@@ -606,6 +607,22 @@ func (lbaas *LbaasV2) getLoadBalancerLegacyName(_ context.Context, _ string, ser
...
 
+// Returns a list of custom load balancer listener tags for the specified  service.
+func (lbaas *LbaasV2) getCustomLoadBalancerTags(service *corev1.Service) []string {
+	var listenerTags []string
+	annotationVal := getStringFromServiceAnnotation(service, ServiceAnnotationLoadBalancerCustomTags, "")
+
+	listOfTags := strings.Split(annotationVal, ",")
+
+	for _, tags := range listOfTags {
+		tags = strings.TrimSpace(tags)
```
dulek, line 618: I think you can save some lines and a var by doing it like this:

```
    tags := strings.Split(annotationVal, ",")
	for i, tag := range tags {
		listOfTags[i] = strings.TrimSpace(tag)
	}
    return tags
```

pkg/openstack/loadbalancer.go
```diff
@@ -1226,6 +1243,12 @@ func (lbaas *LbaasV2) ensureOctaviaPool(lbID string, name string, listener *list
 }
 
 func (lbaas *LbaasV2) buildPoolCreateOpt(listenerProtocol string, service *corev1.Service, svcConf *serviceConfig, name string) v2pools.CreateOpts {
+	var customTag []string
```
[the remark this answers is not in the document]
dulek: +1

pkg/openstack/loadbalancer.go
```diff
@@ -1226,6 +1243,12 @@ func (lbaas *LbaasV2) ensureOctaviaPool(lbID string, name string, listener *list
...
+
+	if svcConf.supportLBTags {
+		customTag = lbaas.getCustomLoadBalancerTags(service)
+	}
```
dulek, line 1250: Why don't we check that inside `getCustomLoadBalancerTags()` and if it's `false`, the function will return `nil`? That way you can just directly use `getCustomLoadBalancerTags()` in line 1282.

pkg/openstack/loadbalancer.go
```diff
@@ -1282,11 +1306,18 @@ func (lbaas *LbaasV2) buildBatchUpdateMemberOpts(port corev1.ServicePort, nodes
 		}
 
 		if port.NodePort != 0 { // It's 0 when AllocateLoadBalancerNodePorts=False
+			var customTag []string
+
+			if svcConf.supportLBTags {
+				customTag = lbaas.getCustomLoadBalancerTags(service)
```
dulek, line 1312: Same here.

pkg/openstack/loadbalancer.go
```diff
@@ -2283,6 +2316,10 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
 			return fmt.Errorf("failed to create Security Group for loadbalancer service %s/%s: %v", apiService.Namespace, apiService.Name, err)
 		}
 		lbSecGroupID = lbSecGroup.ID
+		if svcConf.supportLBTags {
+			lbSecGroup.Tags = lbaas.getCustomLoadBalancerTags(apiService)
+		}
```
dulek, line 2321: In this case you don't need to check `supportLBTags`, SGs are part of Neutron, not Octavia and we can assume they always support tagging. Also this is not the way to do it. In Neutron tagging is an operation separate from resource creation. I've provided an example of how to do it in this comment: https://github.com/kubernetes/cloud-provider-openstack/pull/2439#issuecomment-1785728438

dulek (changes requested):
This still needs a bunch of improvements.

pkg/openstack/loadbalancer.go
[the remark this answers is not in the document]
KingDaemonX: noted !

pkg/openstack/loadbalancer.go
```diff
@@ -468,7 +469,7 @@ func (lbaas *LbaasV2) createOctaviaLoadBalancer(name, clusterName string, servic
 	}
 
 	if svcConf.supportLBTags {
-		createOpts.Tags = []string{svcConf.lbName}
+		createOpts.Tags = lbaas.getCustomLoadBalancerTags(service)
```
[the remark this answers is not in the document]
KingDaemonX: fixed

pkg/openstack/loadbalancer.go
KingDaemonX: modified already ! but this check will be double checked in a couple of method logic

pkg/openstack/loadbalancer.go
KingDaemonX: https://go.dev/play/p/fiUy3r7Arg9

i came up with this following your suggestion

pkg/openstack/loadbalancer.go
KingDaemonX: noted and fixed !

thank you for the clarification

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
jeffyjf: > I specifically asked to not do that, I do not think we should end up with 10 more annotations supported for each of the resource we create in CPO.

Why shouldn't we do that.

According to the [issue descriptoin](https://github.com/kubernetes/cloud-provider-openstack/issues/2327#issue-1856603420), these tags maybe used to config some special functions. So, I really don't think we can config them arbitrary. We couldn't ensure what will happen if we add a tag that originally used to config listener's function to pool. Because of we don't know Octavia driver how to use it,  especially when cloud provider has a custom Octavia driver.

Another possible solution could be to add a special prefix to classify tags. Such as: 
```
loadbalancer.openstack.org/custom-tags: "listener-<tag1>,pool-<tag2>"
```

How do you think?

jeffyjf:
Hi @KingDaemonX

If you addressed something, you can resolve the related comments.

pkg/openstack/loadbalancer.go
```diff
@@ -1310,11 +1332,14 @@ func (lbaas *LbaasV2) buildBatchUpdateMemberOpts(port corev1.ServicePort, nodes
 		}
 
 		if port.NodePort != 0 { // It's 0 when AllocateLoadBalancerNodePorts=False
+			customTag := lbaas.getCustomLoadBalancerTags(service, svcConf)
```
dulek, line 1335: I think this is pretty much constant, so we can do this just once, outside of the loop over `nodes`.

pkg/openstack/loadbalancer.go
```diff
@@ -2317,6 +2344,14 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
 			return fmt.Errorf("failed to create Security Group for loadbalancer service %s/%s: %v", apiService.Namespace, apiService.Name, err)
 		}
 		lbSecGroupID = lbSecGroup.ID
+
+		tags := lbaas.getCustomLoadBalancerTags(apiService, svcConf)
+		for _, tag := range tags {
+			if err := neutrontags.Add(lbaas.network, "securitygroups", lbSecGroupID, tag).ExtractErr(); err != nil {
```
dulek, line 2350: Ouch, true. That might be a lot of calls, let's try to use `neutrontags.ReplaceAll()` instead. Something like this should do. Note that I've improved the error message a bit and replaced "securitygroups" with "security-groups" (fixing a previous mistake).

```
		tags := lbaas.getCustomLoadBalancerTags(apiService, svcConf)
		if _, err := neutrontags.ReplaceAll(lbaas.network, "security-groups", lbSecGroupID, neutrontags.ReplaceAllOpts{Tags: tags}).Extract(); err != nil {
			return fmt.Errorf("failed to add custom tags %s to security group %s (%s)", tags, lbSecGroupID, lbSecGroupName)
		}
```

dulek (commented):
I realized we're missing tagging of floating IPs here. You should be able to add it here: https://github.com/kubernetes/cloud-provider-openstack/blob/03c525d5e3ede43c313014a8b073992cc492a7af/pkg/openstack/loadbalancer.go#L918-L927

Please note it's a Neutron resource again, so the tagging is done the same way as in case of the security groups (the string should be "floatingips", I check that in the API reference).

I'm trying to build this and run a bit of tests, might get back with more comments.

dulek:
> I'm trying to build this and run a bit of tests, might get back with more comments.

Okay, nevermind, after changing `"securitygroups"` to `"security-groups` it just works.

KingDaemonX:
> Unrelated change, please remove.

yeah i get you
i will do that

pkg/openstack/loadbalancer.go
KingDaemonX: ** An Opinion

i was looking into the codebase,this update all existing tags :
```
		tags := lbaas.getCustomLoadBalancerTags(apiService, svcConf)
		if _, err := neutrontags.ReplaceAll(lbaas.network, "security-groups", lbSecGroupID, neutrontags.ReplaceAllOpts{Tags: tags}).Extract(); err != nil {
			return fmt.Errorf("failed to add custom tags %s to security group %s (%s)", tags, lbSecGroupID, lbSecGroupName)
		}
```

this create a new tag to the resource and i think using the creation instead of the update is kinda valid because of it isb an O(n) and the little time difference won't be noticed I think but i will make the changes as requested
```
for _, tag := range tags {
		if err := neutrontags.Add(lbaas.network, "securitygroups", lbSecGroupID, tag).ExtractErr(); err != nil {
				return fmt.Errorf("failed to add custom tag %q to securitygroups neutrontags ", tag)
		}
}
```

dulek: The time difference will be noticed - each of these `neutrontags.Add()` is a call to OpenStack Neutron API. In practice it can be really slow [^1] so saving calls to Neutron is really important.

[^1]: I've even created a meme once: https://dulek.github.io/kuryr-kubernetization/#18

KingDaemonX:
i have made all the required changes to the PR @dulek

pkg/openstack/loadbalancer.go
```diff
@@ -2317,6 +2350,12 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
 			return fmt.Errorf("failed to create Security Group for loadbalancer service %s/%s: %v", apiService.Namespace, apiService.Name, err)
 		}
 		lbSecGroupID = lbSecGroup.ID
+
+		tags := lbaas.getCustomLoadBalancerTags(apiService, svcConf)
+
+		if _, err := neutrontags.ReplaceAll(lbaas.network, "security-group", lbSecGroupID, neutrontags.ReplaceAllOpts{Tags: tags}).Extract(); err != nil {
```
dulek, line 2356: Had you tested this and it works? I've only tried with `security-groups` and can confirm it was working as commented here: https://github.com/kubernetes/cloud-provider-openstack/pull/2439#issuecomment-1804183413

dulek (changes requested):
Seems like you've used `security-group` instead of `security-groups`. I'm fairly sure only the latter works, but please prove me wrong if you've tested it and it works.

Also seems like my remarks from [this comment about metrics](https://github.com/kubernetes/cloud-provider-openstack/pull/2439#issuecomment-1804190858) are not implemented.

KingDaemonX:
> Seems like you've used `security-group` instead of `security-groups`. I'm fairly sure only the latter works, but please prove me wrong if you've tested it and it works.
> 
> Also seems like my remarks from [this comment about metrics](https://github.com/kubernetes/cloud-provider-openstack/pull/2439#issuecomment-1804190858) are not implemented.

i totally miss the remark on metric until now that you pointed it again 
i will fix them now

kayrus:
Hm, I also don't like an idea that we need to tag all the resources at once. Neutron is modular and there is a possibility that it doesn't have tags module enabled, while octavia can have this module. If tags in neutron are not enabled, the OCCM will fail the reconciliation with an error.
Though, cascade tagging of the octavia resources seems fine with me.

Also security groups and especially FIPs can have their own set of tags, e.g. FIP can be preallocated in advance and a special tag can be assigned on it. Once OCCM starts to manage this FIP, it will remove the tags assigned manually. This should not happen.

pkg/openstack/loadbalancer.go
```diff
@@ -785,6 +804,20 @@ func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, svcConf *
...
+			_, err := neutrontags.ReplaceAll(network, "security_groups", port.ID, neutrontags.ReplaceAllOpts{Tags: tags}).Extract()
+			if mc.ObserveRequest(err) != nil {
+				return fmt.Errorf("failed to add tag %s to port %s: %v", tags, port.ID, err)
+			}
+
+			mc = metrics.NewMetricContext("floating_ip_tag", "replace")
+			_, err = neutrontags.ReplaceAll(network, "floatingips", port.ID, neutrontags.ReplaceAllOpts{Tags: tags}).Extract()
+			if mc.ObserveRequest(err) != nil {
+				return fmt.Errorf("failed to add tag %s to port %s of floating_ips: %v", tags, port.ID, err)
+			}
```
dulek, line 820: This shouldn't be here, this function doesn't have anything to do with floating IPs.

Also this won't ever work, you specify `port.ID` which will not match a floating IP.

pkg/openstack/loadbalancer.go
```diff
@@ -785,6 +804,20 @@ func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, svcConf *
...
 				return fmt.Errorf("failed to update security group for port %s: %v", port.ID, res.Err)
 			}
+
+			tags := getCustomLoadBalancerTags(service, svcConf)
+
+			mc = metrics.NewMetricContext("security_group_tag", "replace")
+			_, err := neutrontags.ReplaceAll(network, "security_groups", port.ID, neutrontags.ReplaceAllOpts{Tags: tags}).Extract()
+			if mc.ObserveRequest(err) != nil {
+				return fmt.Errorf("failed to add tag %s to port %s: %v", tags, port.ID, err)
+			}
```
dulek, line 814: I think this is wrong placement, this method applies the security group to a certain port. For now we should limit the SG tagging to the place we're creating it.

Also this won't ever work, you specify `port.ID` which will not match a security group.

pkg/openstack/loadbalancer.go
```diff
@@ -896,10 +929,16 @@ func (lbaas *LbaasV2) deleteOctaviaListeners(lbID string, listenerList []listene
...
+func (lbaas *LbaasV2) createFloatingIP(msg string, floatIPOpts floatingips.CreateOpts, service *corev1.Service, svcConf *serviceConfig) (*floatingips.FloatingIP, error) {
 	klog.V(4).Infof("%s floating ip with opts %+v", msg, floatIPOpts)
 	mc := metrics.NewMetricContext("floating_ip", "create")
 	floatIP, err := floatingips.Create(lbaas.network, floatIPOpts).Extract()
+
+	tags := getCustomLoadBalancerTags(service, svcConf)
+
+	if _, err := neutrontags.ReplaceAll(lbaas.network, "floatingips", floatIP.ID, neutrontags.ReplaceAllOpts{Tags: tags}).Extract(); err != nil {
+		return nil, fmt.Errorf("failed to add custom tags %s to floatingIPs %s with a projectID (%s)", tags, floatIP.ID, floatIP.ProjectID)
+	}
```
dulek, line 941: This is missing the metrics addition. This is where you should add `mc = metrics.NewMetricContext("floating_ip_tag", "replace")` and all the other stuff.

pkg/openstack/loadbalancer.go
```diff
@@ -2317,6 +2364,12 @@ func (lbaas *LbaasV2) ensureAndUpdateOctaviaSecurityGroup(clusterName string, ap
 			return fmt.Errorf("failed to create Security Group for loadbalancer service %s/%s: %v", apiService.Namespace, apiService.Name, err)
 		}
 		lbSecGroupID = lbSecGroup.ID
+
+		tags := getCustomLoadBalancerTags(apiService, svcConf)
+
+		if _, err := neutrontags.ReplaceAll(lbaas.network, "security-groups", lbSecGroupID, neutrontags.ReplaceAllOpts{Tags: tags}).Extract(); err != nil {
+			return fmt.Errorf("failed to add custom tags %s to security group %s (%s)", tags, lbSecGroupID, lbSecGroupName)
+		}
```
dulek, line 2372: This is missing the `mc = metrics.NewMetricContext("security_group_tag", "replace")` and then the subsequent `ObserveRequest()`.

dulek (changes requested):
This version adds the metrics in a wrong place and using wrong IDs. I've pointed to that codeplace as an example where we did some tagging and metrics for another purpose, but the actual tagging should happen just after the resource is created.

pkg/openstack/loadbalancer.go
dulek: I also realized we should only call this code when `len(tags) > 0`.

pkg/openstack/loadbalancer.go
dulek: Same here should only do tagging when `len(tags) > 0`.

dulek (changes requested):

dulek:
> Hm, I also don't like an idea that we need to tag all the resources at once. Neutron is modular and there is a possibility that it doesn't have tags module enabled, while octavia can have this module. If tags in neutron are not enabled, the OCCM will fail the reconciliation with an error. Though, cascade tagging of the octavia resources seems fine with me.

True, we can do that. Octavia is not modular, so it's just a matter of API version, but we should check for tags extension in Neutron case too.

@KingDaemonX - we need to add a function checking this and only attempt to tag FIPs and SGs when it exists. This is an example of how to do it: https://github.com/kubernetes/cloud-provider-openstack/blob/fdba36babb2c4b46e759c99cca50ac7eba2ee06f/pkg/openstack/openstack.go#L475-L484

We need to check for `standard-attr-tag` extension [as explained in Neutron API reference](https://docs.openstack.org/api-ref/network/v2/index.html#tag-extension).

> Also security groups and especially FIPs can have their own set of tags, e.g. FIP can be preallocated in advance and a special tag can be assigned on it. Once OCCM starts to manage this FIP, it will remove the tags assigned manually. This should not happen.

Current code (after my remarks will be addressed) is only tagging FIPs on creation. Same happens with SGs, we assume it's always us creating it, if `manage-security-groups=false` that code won't be executed. Would that work for you? I also suggested to drop the tags replace call if user haven't defined any tags in the config.

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
vladorf: FYI: https://github.com/kubernetes/cloud-provider-openstack/issues/2327#issuecomment-1947301740

I understand it would mean complete rework, but IMHO:
1. it would solve this issue in more elegant fashion
2. there is more room for accomodating future use-cases (e.g. adding more options / adding new tag types)

dulek:
Ah wait, this is indeed stale. I'll just close this, we have other PRs looking at this.

PR #2446, opened 2023-10-20
[charts/occm] Implement imagePullSecret support for master branch

carlotardl:
This PR implements `imagePullSecrets` for the `openstack-cloud-controller-manager` helm chart on master branch.

None applicable

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
@@ -25,6 +25,10 @@ spec:
       labels:
         {{- include "occm.controllermanager.labels" . | nindent 8 }}
     spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
--- a/charts/openstack-cloud-controller-manager/values.yaml
+++ b/charts/openstack-cloud-controller-manager/values.yaml
@@ -9,6 +9,11 @@ commonAnnotations: {}
 #   "helm.sh/hook-weight": "-100"
 #   "helm.sh/hook-delete-policy": before-hook-creation
 
+# List of secrets to use as image pull secret
+imagePullSecrets: []
+# - pull-secret-1
+# - pull-secret-2
+
 # Image repository name and tag
 image:
   repository: registry.k8s.io/provider-os/openstack-cloud-controller-manager
--- a/charts/openstack-cloud-controller-manager/Chart.yaml
+++ b/charts/openstack-cloud-controller-manager/Chart.yaml
@@ -4,7 +4,7 @@ description: Openstack Cloud Controller Manager Helm Chart
 icon: https://object-storage-ca-ymq-1.vexxhost.net/swift/v1/6e4619c416ff4bd19e1c087f27a43eea/www-images-prod/openstack-logo/OpenStack-Logo-Vertical.png
 home: https://github.com/kubernetes/cloud-provider-openstack
 name: openstack-cloud-controller-manager
-version: 2.29.0-alpha.4
+version: 2.29.0-alpha.5
 maintainers:
   - name: eumel8
     email: f.kloeker@telekom.de


wwentland:
Thank you! 

Please note that the linter is currently failing because you did not bump the chart version in [Chart.yaml](https://github.com/kubernetes/cloud-provider-openstack/blob/master/charts/openstack-cloud-controller-manager/Chart.yaml#L7). It would be fantastic if you could incorporate that here and squash/fixup your commits into a single one.

charts/manila-csi-plugin/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v1.28.0
 description: Manila CSI Chart for OpenStack
 name: openstack-manila-csi
-version: 2.29.0-alpha.1
+version: 2.29.0-alpha.4
```
dulek, line 5: Why do we jump 3 versions here?

dulek (changes requested):

charts/manila-csi-plugin/Chart.yaml
carlotardl: I updated the wrong file, I corrected the mistake on the latest commit

jichenjc:
/lgtm

please squash the commits before merge ..

carlotardl:
It's been rebased, I had different sources and there was a mixup but I updated it on the last commit

charts/openstack-cloud-controller-manager/templates/daemonset.yaml
```diff
@@ -25,6 +25,8 @@ spec:
       labels:
         {{- include "occm.controllermanager.labels" . | nindent 8 }}
     spec:
+      imagePullSecrets:
+        {{- toYaml .Values.imagePullSecrets | nindent 8 }}
```
wwentland, line 29: It might be an idea to enclose this in a `with`  block, similar to other settings in the chart.

This would look similar to:

```
{{- with .Values.imagePullSecrets }}
imagePullSecrets:
  {{- toYaml . | nindent 8 }}
{{- end }}
```

wwentland (commented):
Look much better, thanks. I was wondering if we'd like to treat this setting similar to others in that template.

charts/openstack-cloud-controller-manager/templates/daemonset.yaml
dulek: +1, makes sense.

dulek (changes requested):

charts/openstack-cloud-controller-manager/values.yaml
```diff
@@ -10,6 +10,7 @@ commonAnnotations: {}
 #   "helm.sh/hook-delete-policy": before-hook-creation
 
 # Image repository name and tag
+imagePullSecrets: []
```
wwentland, line 13: This should probably be on the line above, to ensure that the comment `# Image repository name and tag` remains in the right place.

Furthermore, it would be great to add a similar comment that details how to use `imagePullSecrets` with an example.

charts/openstack-cloud-controller-manager/templates/daemonset.yaml
```diff
@@ -25,6 +25,10 @@ spec:
       labels:
         {{- include "occm.controllermanager.labels" . | nindent 8 }}
     spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
```
wwentland, line 31: I am not entirely sure this works with the default of `imagePullSecrets: []` that is currently being set in the values file.

We have to make sure that this works in the following cases:

1. Default case (user does not set `imagePullSecrets` and they are `imagePullSecrets: []`
2. Single element in `imagePullSecrets`
3. Multiple elements in `imagePullSecrets`

This might require a combination of `if`, `range` and/or `with` to achive the desired outcome.

wwentland (changes requested):
Thanks, I am sure that the functionality to reference `imagePullSecrets` will come in handy in a variety of contexts.

I've added a couple of inline comments and just want to make sure that this works as desired in the variety of use-cases users would want to implement.

Furthermore, I am curious if we want to consider adding this to the `ServiceAccount` instead or let users choose one approach over the other.

The former ties closer in how access through workload identity is typically implemented and might rather be what users expect.

See https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/  for details.

charts/openstack-cloud-controller-manager/templates/daemonset.yaml
carlotardl: With the values file, the output should be empty, but once we overwrite the values.yml with
```
imagePullSecrets:
- one
```
The output will be:
```
imagePullSecrets:
        - one
```
It also works with more values, it can be tested with 
`helm template --values values.yaml .`

I updated the last commit with a few comments.

charts/openstack-cloud-controller-manager/values.yaml
Lirt: This was fixed, please re-check.

charts/openstack-cloud-controller-manager/templates/daemonset.yaml
Lirt: lgtm, tested with helm template and all possibilities render correctly.

charts/openstack-cloud-controller-manager/templates/daemonset.yaml
Lirt: This was fixed, please re-check.

PR #2447, opened 2023-10-20
[release-1.28] [occm] Use standard service account name in OCCM helm chart

### LINKED ISSUE #2440: [occm] service account `cloud-controller-manager` cannot patch service object

/kind bug

**What happened**:

Installed the OCCM helm chart and wanted to provision an Octavia load  balancer (via the haproxy-ingress helm chart) but it's stuck in pending.

The error is:

```
Error syncing load balancer: failed to ensure load balancer: failed to patch service object haproxy-ingress/haproxy-ingress: services "haproxy-ingress" is forbidden: User "system:serviceaccount:kube-system:cloud-controller-manager" cannot patch resource "services" in API group "" in the namespace "haproxy-ingress"
```

Probably related: #1722

A load balancer.

**How to reproduce it**:

Install the helm chart: `openstack-cloud-controller-manager`, version: `2.28.0`

The helm chart creates a SA which is also used by the `openstack-cloud-controller` pod. This SA seems to have the correct CRB as far as I can tell.

When the pod starts, I see another SA created, `cloud-controller-manager`. This SA has no CRB to be able to patch the service object. I am not sure why this is created or when. It doesn't seem to come with the helm chart (see below) and I am not creating it anywhere else as far as I see.

Somewhat a side effect, the created SA (`cloud-controller-manager`) also blocks helm from doing updates to the OCCM:

> UPGRADE FAILED: rendered manifests contain a resource that already exists. Unable to continue with update: ServiceAccount "cloud-controller-manager" in namespace "kube-system" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "openstack-cloud-controller-manager"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "kube-system"

wwentland:

This is a cherry-pick of:

- #2332
- #2347

fixes #2440

At the time #2332 was merged the CI was sadly somewhat broken and missed a spurious space. This has been fixed in #2347 which is also included here to ensure that CI is passing.

As discussed, I bumped the helm chart version to `2.28.2`.

I've also tested this helm chart version in a new 1.28 cluster and was unable to reproduce the bug fixed in #2332.

**Release note**:

```release-note
ACTION REQUIRED: Please note that you might have to delete the `cloud-controller-manager` service account in the `kube-system` namespace if it exists, as upgrading with helm would fail otherwise.
```

### DIFF
--- a/charts/openstack-cloud-controller-manager/values.yaml
+++ b/charts/openstack-cloud-controller-manager/values.yaml
@@ -130,3 +130,7 @@ extraVolumeMounts:
 # cluster name that used for created cluster
 cluster:
   name: kubernetes
+
+clusterRoleName : system:cloud-controller-manager
+
+serviceAccountName: cloud-controller-manager
--- a/charts/openstack-cloud-controller-manager/values.yaml
+++ b/charts/openstack-cloud-controller-manager/values.yaml
@@ -131,6 +131,6 @@ extraVolumeMounts:
 cluster:
   name: kubernetes
 
-clusterRoleName : system:cloud-controller-manager
+clusterRoleName: system:cloud-controller-manager
 
 serviceAccountName: cloud-controller-manager
--- a/charts/openstack-cloud-controller-manager/Chart.yaml
+++ b/charts/openstack-cloud-controller-manager/Chart.yaml
@@ -4,7 +4,7 @@ description: Openstack Cloud Controller Manager Helm Chart
 icon: https://object-storage-ca-ymq-1.vexxhost.net/swift/v1/6e4619c416ff4bd19e1c087f27a43eea/www-images-prod/openstack-logo/OpenStack-Logo-Vertical.png
 home: https://github.com/kubernetes/cloud-provider-openstack
 name: openstack-cloud-controller-manager
-version: 2.28.1
+version: 2.28.2
 maintainers:
   - name: eumel8
     email: f.kloeker@telekom.de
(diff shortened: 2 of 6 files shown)


wwentland:
Please let me know if your preference would be for me to squash both commits into a single, correct, one.

till:
@wwentland could you push a 2.28-pre-something while we wait? Then I will test it as well. :)

charts/openstack-cloud-controller-manager/values.yaml
```diff
@@ -130,3 +130,7 @@ extraVolumeMounts:
 # cluster name that used for created cluster
 cluster:
   name: kubernetes
+
+clusterRoleName: system:cloud-controller-manager
+
+serviceAccountName: cloud-controller-manager
```
kayrus, line 136: hm. I'm a bit skeptical about removing the `openstack-` suffix, since it may cause inconsistencies during an upgrade. But if this was already merged into master, I'm fine with this. @zetaab @dulek what do you think?

wwentland: Yes, that is why it is important to include the release notes. 

Once users install this the RBAC should be like it should have been in the first place.

wwentland: Just to give a bit more context from #2332: The client is initialised as `cloud-controller-manager` which is why a service account of the same name is created in the `kube-system` system.

The helm chart currently creates RBAC for an `openstack-cloud-controller-manager` one, which results in broken permissions for every installation that runs with `--use-service-account-credentials=true`.

You can find additional details in the original PR.

till:
@wwentland do you think this can land today?

wwentland:
> @wwentland do you think this can land today?

The PR still requires an approval by, for example, @zetaab or @dulek.

PR #2377, opened 2023-09-28
[occm] improve LB logic maintainability

kayrus:
**What this PR does / why we need it**:

This PR improves the loadbalancer code readability and maintainability.

**Which issue this PR fixes(if applicable)**:

fixes #2374

**Special notes for reviewers**:

**Release note**:

```release-note
NONE
```

PR #2451, opened 2023-10-24
[octavia-lb] Add the logic to reattach the VIP to the load balancer if the IP has changed

### LINKED ISSUE #2443: [occm] loadbalancer doesn't honor the `spec.loadBalancerIP` on service update

/kind bug

**What happened**:

Documentation [says](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md#creating-service-by-specifying-a-floating-ip):

> Sometimes it's useful to use an existing available floating IP rather than creating a new one, especially in the automation scenario. In the example below, 122.112.219.229 is an available floating IP created in the OpenStack Networking service.

>> NOTE: If 122.112.219.229 is not available, a new floating IP will be created automatically from the configured public network. If 122.112.219.229 is already associated with another port, the Service creation will fail.

However once I create a svc with no `spec.loadBalancerIP` defined, OCCM assigns the random FIP to a LB. If I want to change the FIP on the existing service by setting the `spec.loadBalancerIP`, the setting persists, but the actual FIP is not assigned.

I expect OCCM to assign a new FIP once `spec.loadBalancerIP` is changed.

**How to reproduce it**:

* create a service without a FIP
* assign a  `spec.loadBalancerIP` to a desired value
* FIP is not changed

https://github.com/kubernetes/cloud-provider-openstack/blob/ed517e1f07057cd439542c73bff12102d8849dde/pkg/openstack/loadbalancer.go#L1039-L1041

see also #2377, where this bug is seen more clearly.

**Environment**:
- OpenStack version: ???
- Others: ???

yang-wang11:

fixes #2443

**Release note**:

```release-note
Enable OCCM to reassign the new Floating IP from service.spec when spec.loadBalancerIP is changed.
```

### DIFF
--- a/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
+++ b/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
@@ -303,13 +303,17 @@ spec:
 
 Sometimes it's useful to use an existing available floating IP rather than creating a new one, especially in the automation scenario. In the example below, 122.112.219.229 is an available floating IP created in the OpenStack Networking service.
 
-> NOTE: If 122.112.219.229 is not available, a new floating IP will be created automatically from the configured public network. If 122.112.219.229 is already associated with another port, the Service creation will fail.
+The service’s annotation loadbalancer.openstack.org/floating-ip can also be utilized to specify a desired floating IP. If this annotation exists and is not empty, the value contained in it will be used as the floating IP. Otherwise, the spec.loadBalancerIP field will be used as shown in the example below.
+
+> NOTE: If the specified floating-ip is not available, a new floating IP will be created automatically from the configured public network. If the specified floating-ip is already associated with another port, the Service creation will fail. 
 
 ```yaml
 apiVersion: v1
 kind: Service
 metadata:
   name: nginx-internet
+  annotations:
+    loadbalancer.openstack.org/floating-ip: "122.112.219.220"  
 spec:
   type: LoadBalancer
   selector:
--- a/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
+++ b/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
@@ -134,6 +134,12 @@ Request Body:
 
   If 'true', the floating IP will **NOT** be deleted. Default is 'false'.
 
+- `loadbalancer.openstack.org/floating-ip`
+
+  This annotation is utilized to specify the floating IP for the load balancer. 
+
+  If set, both newly created and pre-existing load balancers will utilize it. Furthermore, its precedence is higher than that of `Spec.LoadBalanccerIP`.
+
 - `loadbalancer.openstack.org/proxy-protocol`
 
   If 'true', the loadbalancer pool protocol will be set as `PROXY`. Default is 'false'.
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -73,6 +73,7 @@ const (
 	ServiceAnnotationLoadBalancerFloatingSubnetTags   = "loadbalancer.openstack.org/floating-subnet-tags"
 	ServiceAnnotationLoadBalancerClass                = "loadbalancer.openstack.org/class"
 	ServiceAnnotationLoadBalancerKeepFloatingIP       = "loadbalancer.openstack.org/keep-floatingip"
+	ServiceAnnotationLoadBalancerFloatingIP           = "loadbalancer.openstack.org/floating-ip"
 	ServiceAnnotationLoadBalancerPortID               = "loadbalancer.openstack.org/port-id"
 	ServiceAnnotationLoadBalancerProxyEnabled         = "loadbalancer.openstack.org/proxy-protocol"
 	ServiceAnnotationLoadBalancerSubnetID             = "loadbalancer.openstack.org/subnet-id"
@@ -431,7 +432,13 @@ func getSecurityGroupName(service *corev1.Service) string {
 }
 
 func getLoadBalancerIP(service *corev1.Service) string {
-	return service.Spec.LoadBalancerIP
+	loadBalancerIP := service.Spec.LoadBalancerIP
+	if annotations := service.Annotations; annotations != nil {
+		if val, ok := annotations[ServiceAnnotationLoadBalancerFloatingIP]; ok && val != "" {
+			return val
+		}
+	}
+	return loadBalancerIP
 }
 
 func getFullServiceName(service *corev1.Service) string {
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -431,7 +431,6 @@ func getSecurityGroupName(service *corev1.Service) string {
 }
 
 func getLoadBalancerIP(service *corev1.Service) string {
-
 	return service.Spec.LoadBalancerIP
 }
 
@@ -559,7 +558,7 @@ func (lbaas *LbaasV2) createOctaviaLoadBalancer(name, clusterName string, servic
 				return nil, fmt.Errorf("loadbalancer %s is in ERROR state and there was an error when removing it: %v", loadbalancer.ID, err)
 			}
 			return nil, fmt.Errorf("loadbalancer %s has gone into ERROR state, please check Octavia for details. Load balancer was "+
-					"deleted and its creation will be retried", loadbalancer.ID)
+				"deleted and its creation will be retried", loadbalancer.ID)
 		}
 		return nil, err
 	}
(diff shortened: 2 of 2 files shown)


dulek (changes requested):
I think this enables users of the cluster to set `spec.loadBalancerIP=""`, wait for a new FIP to get attached and repeat the process until all FIPs are exhausted. We should allow that. At minimum old FIP has to be deleted if it was created by the CPO.

dulek:
Seems like this also breaks shared LBs:

```
E1024 12:12:45.460335      11 controller.go:298] error processing service octavia-lb-test/test-shared-2 (retrying with exponential backoff): failed to ensure load balancer: floating IP  is not available
I1024 12:12:45.460700      11 event.go:376] "Event occurred" object="octavia-lb-test/test-shared-2" fieldPath="" kind="Service" apiVersion="v1" type="Warning" reason="SyncLoadBalancerFailed" message="Error syncing load balancer: failed to ensure load balancer: floating IP  is not available"
```

Not that I like the feature, but we cannot just break it.

yang-wang11:
> I think this enables users of the cluster to set `spec.loadBalancerIP=""`, wait for a new FIP to get attached and repeat the process until all FIPs are exhausted. We shouldn't allow that. At minimum old FIP has to be deleted if it was created by the CPO.

@dulek  Great observation, I will take some time to refactor this function to lessen the if-else.

pkg/openstack/loadbalancer.go
```diff
@@ -430,6 +430,14 @@ func getSecurityGroupName(service *corev1.Service) string {
 	return securityGroupName
 }
 
+func getLoadBalancerIP(service *corev1.Service) string {
+	return service.Spec.LoadBalancerIP
```
yang-wang11, line 434: Should further annotation support be provided, enhancement can be realized here.

yang-wang11: enhanced in the commit https://github.com/kubernetes/cloud-provider-openstack/pull/2451/commits/58812808fa44205510af62d654c6e4e55c25bf9d.

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
```diff
@@ -303,13 +303,17 @@ spec:
 
 Sometimes it's useful to use an existing available floating IP rather than creating a new one, especially in the automation scenario. In the example below, 122.112.219.229 is an available floating IP created in the OpenStack Networking service.
 
-> NOTE: If 122.112.219.229 is not available, a new floating IP will be created automatically from the configured public network. If 122.112.219.229 is already associated with another port, the Service creation will fail.
+The service’s annotation loadbalancer.openstack.org/floating-ip can also be utilized to specify a desired floating IP. If this annotation exists and is not empty, the value contained in it will be used as the floating IP. Otherwise, the spec.loadBalancerIP field will be used as shown in the example below.
+
```
jeffyjf, line 307: The new annotation should be listed [here](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md?plain=1#L91)

jeffyjf (changes requested):

pkg/openstack/loadbalancer.go
```diff
@@ -944,103 +962,131 @@ func (lbaas *LbaasV2) ensureFloatingIP(clusterName string, service *corev1.Servi
...
+}
+
+func (lbaas *LbaasV2) processWithFloatIP(portID string, floatIP *floatingips.FloatingIP, service *corev1.Service) (string, error) {
+
+	klog.V(4).Infof("Found floating ip %v by loadbalancer port id %q", floatIP, portID)
+
+	loadBalancerIP := getLoadBalancerIP(service)
+
+	if floatIP.FloatingIP != loadBalancerIP && loadBalancerIP != "" {
+		if err := lbaas.associateFloatingIP(loadBalancerIP, portID); err != nil {
```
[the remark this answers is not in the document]
yang-wang11: yes, the limitation is needed, only the LB-owned can be updated.

yang-wang11:
@kayrus I'm not trying to rush you, but kindly look forward to your code review.

pkg/openstack/loadbalancer.go
```diff
@@ -74,6 +74,7 @@ const (
 	ServiceAnnotationLoadBalancerFloatingSubnetTags   = "loadbalancer.openstack.org/floating-subnet-tags"
 	ServiceAnnotationLoadBalancerClass                = "loadbalancer.openstack.org/class"
 	ServiceAnnotationLoadBalancerKeepFloatingIP       = "loadbalancer.openstack.org/keep-floatingip"
+	ServiceAnnotationLoadBalancerFloatingIP           = "loadbalancer.openstack.org/floating-ip"
```
kayrus, line 77: in order to comply with the previous naming, I'd suggest to use `loadbalancer.openstack.org/floatingip`
in future for dual stack we can use `loadbalancer.openstack.org/floatingip-v6`

pkg/openstack/loadbalancer.go
```diff
@@ -431,6 +432,19 @@ func getSecurityGroupName(service *corev1.Service) string {
 	return securityGroupName
 }
 
+func getLoadBalancerIP(service *corev1.Service) string {
+	if annotations := service.Annotations; annotations != nil {
```
kayrus, line 436: we can skip this check. `if val, ok := annotations[ServiceAnnotationLoadBalancerFloatingIP]; ok && val != "" {` is enough.

kayrus (commented):
@yang-wang11 apologies that you had to wait. See my comments. As for the new `ensureFloatingIP` logic, give me a bit more time.

kayrus:
@yang-wang11 Unfortunately I have doubts about the proposed `ensureFloatingIP` logic. I'm afraid the reconciliation port of the code is too fragile, and it requires a wider refactoring.

A couple of examples:

https://github.com/kubernetes/cloud-provider-openstack/blob/d6f294d0a0bdb53e2dd71023be007ef7b26d4834/pkg/openstack/loadbalancer.go#L965-L968

why would we fail in this case? What if the loadbalancer has a `service.beta.kubernetes.io/openstack-internal-load-balancer` flag?

There are also a bunch of minor concerns, which combination in result raise other questions.

Basically this is the reason why the #2377 PR is still in draft.

IMO the first step that we need to perform is to extend the mock API with Octavia support and add test cases which include a collection of different annotation combinations for services.

@yang-wang11 how critical is the issue for you? Does it occur only on svc update only?

pkg/openstack/loadbalancer.go
yang-wang11: Yes, it is used to avoid panic if there is no annotation. adopt your suggestion, it can cover my concern. 
https://go.dev/play/p/RDerITo3dqp?v=goprev

yang-wang11:
> @yang-wang11 Unfortunately I have doubts about the proposed `ensureFloatingIP` logic. I'm afraid the reconciliation port of the code is too fragile, and it requires a wider refactoring.
> 
> An example:
> 
> https://github.com/kubernetes/cloud-provider-openstack/blob/d6f294d0a0bdb53e2dd71023be007ef7b26d4834/pkg/openstack/loadbalancer.go#L965-L968
> 
> why would we fail in this case? What if the loadbalancer has a `service.beta.kubernetes.io/openstack-internal-load-balancer` flag? There is no indication that this loadbalancer is going to be updated, therefore there is no reason to return an error.
> 
> There are also a bunch of minor concerns, which combination in result raises other questions.
> 
> Basically this is the reason why the #2377 PR is still in draft.
> 
> IMO the first step that we need to perform is to extend the mock API with Octavia support and add test cases which include a collection of different annotation combinations for services.
> 
> @yang-wang11 how critical is the issue for you? Does it occur only on svc update only?

@kayrus 
I'm unsure if this change belongs to a refactor or not, the thing I did is combine similar logic. It aims to eliminate the if-else logic.
As we discussed in the Slack channel, the issue that occurred from the cluster was the update case, but not sure if there are others. we hope we can fix this issue at the earliest convenience.

kayrus:
@yang-wang11 can you resolve the conflicts?

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
```diff
@@ -134,6 +134,12 @@ Request Body:
 
   If 'true', the floating IP will **NOT** be deleted. Default is 'false'.
 
+- `loadbalancer.openstack.org/floating-ip`
+
+  This annotation is utilized to specify the floating IP for the load balancer. 
+
+  If set, both newly created and pre-existing load balancers will utilize it. Furthermore, its precedence is higher than that of `Spec.LoadBalanccerIP`.
```
kayrus, line 141:
```suggestion
  If set, both newly created and pre-existing load balancers will utilize it. Furthermore, its precedence is higher than that of `Spec.LoadBalancerIP`.
```

PR #2455, opened 2023-10-25
[occm] manage-security-groups: Only add SGs to LB members

### LINKED ISSUE #2454: [occm] LB failing to be created with `manage-security-groups=true` and `port_security_enabled=false`
/kind bug

**What happened**:
LB is failing to get created when an additional port with `port_security_enabled=false` is added to the node and `manage-security-groups=true` in the CPO configuration. This is because an SG is tried to be added to it and it fails, as without port security you cannot assign SGs.

LB gets created normally.

**How to reproduce it**:
All is above.

**Environment**:
- OpenStack version: N/A
- Others:

dulek:
Previous code for manage-security-groups were adding the SGs to all the ports attached to the Node. That doesn't make a lot of sense, we only need this traffic opened to the LB members and we only add one port of the Node as the LB member. Moreover we've even tried to add SGs to the ports that had `port_security_enabled=false` which obviously failed.

This patch makes sure that we only add the SG opening NodePort traffic to the port that is actually plugged into the LB. Moreover it adds a check for `port_security_enabled` to make sure that ports with it disabled are ignored.

fixes #2454

**Release note**:
```release-note
[openstack-cloud-controller-manager] `manage-security-groups=true` will only attach the SG to the port that is a member of the load balancer and not to all ports attached to the node as previously.
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -727,22 +727,49 @@ func getSubnetIDForLB(network *gophercloud.ServiceClient, node corev1.Node, pref
 	return "", cpoerrors.ErrNotFound
 }
 
-// applyNodeSecurityGroupIDForLB associates the security group with all the ports on the nodes.
-func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, nodes []*corev1.Node, sg string) error {
+// isPortMember returns true if IP and subnetID are one of the FixedIPs on the port
+func isPortMember(port PortWithPortSecurity, IP string, subnetID string) bool {
+	for _, fixedIP := range port.FixedIPs {
+		if (subnetID == "" || subnetID == fixedIP.SubnetID) && IP == fixedIP.IPAddress {
+			return true
+		}
+	}
+	return false
+}
+
+// applyNodeSecurityGroupIDForLB associates the security group with the ports being members of the LB on the nodes.
+func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, svcConf *serviceConfig, nodes []*corev1.Node, sg string) error {
 	for _, node := range nodes {
 		serverID, _, err := instanceIDFromProviderID(node.Spec.ProviderID)
 		if err != nil {
 			return fmt.Errorf("error getting server ID from the node: %w", err)
 		}
+
+		addr, _ := nodeAddressForLB(node, svcConf.preferredIPFamily)
+		if addr == "" {
+			// If node has no viable address let's ignore it.
+			continue
+		}
+
 		listOpts := neutronports.ListOpts{DeviceID: serverID}
-		allPorts, err := openstackutil.GetPorts(network, listOpts)
+		allPorts, err := openstackutil.GetPorts[PortWithPortSecurity](network, listOpts)
 		if err != nil {
 			return err
 		}
 
 		for _, port := range allPorts {
+			// You can't assign an SG to a port with port_security_enabled=false, skip them.
+			if !port.PortSecurityEnabled {
+				continue
+			}
+
 			// If the Security Group is already present on the port, skip it.
 			if slices.Contains(port.SecurityGroups, sg) {
+				continue
+			}
+
+			// Only add SGs to the port actually attached to the LB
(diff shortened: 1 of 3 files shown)


dulek:
/hold

I need to run some more manual testing on this.

pkg/openstack/loadbalancer.go
```diff
@@ -727,25 +727,52 @@ func getSubnetIDForLB(network *gophercloud.ServiceClient, node corev1.Node, pref
...
+
+		addr, err := nodeAddressForLB(node, svcConf.preferredIPFamily)
+		if err != nil {
+			// If node has no viable address let's ignore it.
+			continue
+		}
+
 		listOpts := neutronports.ListOpts{DeviceID: serverID}
-		allPorts, err := openstackutil.GetPorts(network, listOpts)
+		allPorts, err := openstackutil.GetPorts[PortWithPortSecurity](network, listOpts)
```
jichenjc, line 755: um... where's the usage of this ? I only see it's used but where's this created?

dulek: What do you mean? If where `allPorts` var is used, then in for loop just below. And to have `PortSecurityEnabled` property on ports I need to use `PortWithPortSecurity` struct and that's why I made `GetPorts()` a generic.

dulek:
/hold cancel

Okay, a tiny fix and this works as expected now.

pkg/openstack/loadbalancer.go
jichenjc: sorry, I didn't notice change in
https://github.com/kubernetes/cloud-provider-openstack/pull/2455/files#diff-a89aa055a2316c0db33d258ec1c17fc3948b8f53fbc3a475d5a6602025096b33L143
at first so didn't get the logic here, but read that code more I understand the purpose , thanks for the info

pkg/openstack/loadbalancer.go
```diff
@@ -727,25 +727,52 @@ func getSubnetIDForLB(network *gophercloud.ServiceClient, node corev1.Node, pref
...
+func isPortMember(port PortWithPortSecurity, IP string, subnetID string) bool {
+	for _, fixedIP := range port.FixedIPs {
+		if (subnetID == "" || subnetID == fixedIP.SubnetID) && IP == fixedIP.IPAddress {
+			return true
+		}
+	}
+	return false
+}
+
 // applyNodeSecurityGroupIDForLB associates the security group with all the ports on the nodes.
```
jeffyjf, line 740: I think this comment should be modified, It not all the ports now.

pkg/openstack/loadbalancer.go
```diff
@@ -727,25 +727,52 @@ func getSubnetIDForLB(network *gophercloud.ServiceClient, node corev1.Node, pref
...
-func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, nodes []*corev1.Node, sg string) error {
+func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, svcConf *serviceConfig, nodes []*corev1.Node, sg string) error {
 	for _, node := range nodes {
 		serverID, _, err := instanceIDFromProviderID(node.Spec.ProviderID)
 		if err != nil {
 			return fmt.Errorf("error getting server ID from the node: %w", err)
 		}
+
+		addr, err := nodeAddressForLB(node, svcConf.preferredIPFamily)
+		if err != nil {
```
jeffyjf, line 749: Why not `addr == ""`, I know it is equivalent to `err != nil` here. But, I think it seems like a bit informal. In my opinion, it will confuse someone like me who aren't familiar with the code. I need to view details of `nodeAddressForLB` to  understand this.

pkg/openstack/loadbalancer.go
dulek: Nice catch, fixed!

pkg/openstack/loadbalancer.go
dulek: No big deal, but yeah. Fixed.

jeffyjf:
Look good to me

jichenjc:
/lgtm

we got a lgtm from @EmilienM , my lgtm is on behalf of him :)

PR #2456, opened 2023-10-26
[occm] fix: octavia tlsContainerRef validation for barbican secrets

Nuckal777:
Octavia accepts secrets and containers for `default_tls_container_ref` [ref](https://docs.openstack.org/api-ref/load-balancer/v2/index.html#create-listener). Validation currently checks a secret UUID against the containers endpoint.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/keymanager/v1/containers"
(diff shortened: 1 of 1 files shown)


jichenjc:
I think we can take this as a bug and cherry pick to older release

wwentland:
Looks as if the cherry pick to `release-1.27` didn't work.

dulek:
Can we add an issue here so that we'll track it as a bugfix? Just to keep it formally clean that we're backporting this.

jichenjc:
https://github.com/kubernetes/cloud-provider-openstack/issues/2461 is created to track the issue
we can close that issue when all backport merged

PR #2463, opened 2023-11-02
Add Test to generate listener options [OCP]

FavourEva:

fixes #

**Release note**:

```release-note
NONE
```


dulek:
Nice! It's missing a case when `Protocol` ends up as `listeners.ProtocolHTTP`, but I think this is good enough.
/lgtm
/approve

FavourEva:
> Nice! It's missing a case when `Protocol` ends up as `listeners.ProtocolHTTP`, but I think this is good enough. /lgtm /approve

You're right! I would have to open a new PR including that case to the test

PR #2466, opened 2023-11-03
[release-1.28] manage-security-groups: Only add SGs to LB members (#2455)

### LINKED ISSUE #2454: [occm] LB failing to be created with `manage-security-groups=true` and `port_security_enabled=false`

dulek:
Previous code for manage-security-groups were adding the SGs to all the ports attached to the Node. That doesn't make a lot of sense, we only need this traffic opened to the LB members and we only add one port of the Node as the LB member. Moreover we've even tried to add SGs to the ports that had `port_security_enabled=false` which obviously failed.

This patch makes sure that we only add the SG opening NodePort traffic to the port that is actually plugged into the LB. Moreover it adds a check for `port_security_enabled` to make sure that ports with it disabled are ignored.

fixes #2454

**Release note**:
```release-note
[openstack-cloud-controller-manager] `manage-security-groups=true` will only attach the SG to the port that is a member of the load balancer and not to all ports attached to the node as previously.
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -717,22 +717,49 @@ func getSubnetIDForLB(network *gophercloud.ServiceClient, node corev1.Node, pref
 	return "", cpoerrors.ErrNotFound
 }
 
-// applyNodeSecurityGroupIDForLB associates the security group with all the ports on the nodes.
-func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, nodes []*corev1.Node, sg string) error {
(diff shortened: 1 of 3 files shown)


dulek:
We're missing most of the manage-sg fixes in 1.27, I'll just assume nobody used that anyway and stop the backports here.

PR #2468, opened 2023-11-06
[octavia-ingress-controller] Update listener timeouts

### LINKED ISSUE #2243: [Octavia-ingress-controller] Add configurable listener timeouts through annotations

dulek:
This commit makes sure that listener timeouts are updated when their values are changed on the Ingress resource annotations.

cc @sakshi-1505 

fixes #2243

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/ingress/controller/openstack/octavia.go
+++ b/pkg/ingress/controller/openstack/octavia.go
@@ -367,7 +367,11 @@ func (os *OpenStack) EnsureListener(name string, lbID string, secretRefs []strin
 	} else {
 		if len(listenerAllowedCIDRs) > 0 && !reflect.DeepEqual(listener.AllowedCIDRs, listenerAllowedCIDRs) {
 			_, err := listeners.Update(os.Octavia, listener.ID, listeners.UpdateOpts{
-				AllowedCIDRs: &listenerAllowedCIDRs,
+				AllowedCIDRs:         &listenerAllowedCIDRs,
+				TimeoutClientData:    timeoutClientData,
+				TimeoutMemberData:    timeoutMemberData,
+				TimeoutMemberConnect: timeoutMemberConnect,
+				TimeoutTCPInspect:    timeoutTCPInspect,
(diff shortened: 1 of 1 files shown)


dulek:
This extends #2435 a bit.

dulek:
@pierreprinetti and @jichenjc, you might want to take a look.

pkg/ingress/controller/openstack/octavia.go
```diff
@@ -367,7 +367,11 @@ func (os *OpenStack) EnsureListener(name string, lbID string, secretRefs []strin
 	} else {
 		if len(listenerAllowedCIDRs) > 0 && !reflect.DeepEqual(listener.AllowedCIDRs, listenerAllowedCIDRs) {
 			_, err := listeners.Update(os.Octavia, listener.ID, listeners.UpdateOpts{
-				AllowedCIDRs: &listenerAllowedCIDRs,
+				AllowedCIDRs:         &listenerAllowedCIDRs,
```
MaysaMacedo, line 370: Shouldn't the timeouts be updated when there is a change in the timeouts and not when allowedCIDRs has changed?

zetaab (approved):
/lgtm

pkg/ingress/controller/openstack/octavia.go
dulek: @MaysaMacedo: Whoops, I missed this, obviously valid comment. It's fixed in #2518.

PR #2480, opened 2023-11-25
cinder-csi: Adds support for managing backups (kubernetes#2473)

### LINKED ISSUE #2473: [cinder-csi-plugin] Programatically create cinder backups from Kubernetes via CSI
/kind feature

**What happened**:
Currently, there is no method to programmatically create volume backups from K8S.
This disallows for Kubernetes administrators to easily make sure their data is redundantly stored.

I believe the best option would be to do something similar to what [Longhorn does](https://longhorn.io/docs/1.5.2/snapshots-and-backups/csi-snapshot-support/csi-volume-snapshot-associated-with-longhorn-backup/).
By setting the value on the volumesnapshotclass' parameters.type to `snapshot` or `backup`, you can choose between linking
the volumesnapshot object to a volume snapshot or a backup.

```
kind: VolumeSnapshotClass
apiVersion: snapshot.storage.k8s.io/v1
driver: driver.longhorn.io
deletionPolicy: Delete
parameters:
  type: <type>
```

**How to reproduce it**:
Not applicable

I would love to contribute the necessary changes to the project

**Environment**:
- OpenStack version: openstack 6.3.0
- Others:

Sebastian-RG:
This allows for creating and deleting cinder backups from K8S via VolumeSnapshot objects that have a VolumeSnapshotClass with `parameters.type = "backup"`. 

Backups are different from snapshots in that they are usually stored off-site, for example via S3 or NFS.

fixes #2473 

Create a VolumeSnapshotClass that has parameters.type equal to "backup"
```
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshotClass
metadata:
  name: csi-cinder-snapclass
driver: cinder.csi.openstack.org
deletionPolicy: Delete
parameters:
  force-create: "false"
  type: "snapshot"
```

Create a volumeSnapshot that has a Cinder PVC as the Source. This creates a Backup. Deleting this also deletes the backup.
```
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshot
metadata:
  name: new-snapshot-demo
spec:
  volumeSnapshotClassName: csi-cinder-snapclass
  source:
    persistentVolumeClaimName: pvc-snapshot-demo
```

Create a PVC that has the previouse VolumeSnapshot as the source.
The requested size must be greater than the backup size.
This creates a volume populated with the backup data.
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: snapshot-demo-restore
spec:
  storageClassName: csi-sc-cinderplugin
  dataSource:
    name: new-snapshot-demo
    kind: VolumeSnapshot
    apiGroup: snapshot.storage.k8s.io
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 2Gi
```

**Release note**:
```release-note
[cinder-csi-plugin] Adds support for managing cinder backups via volumeSnapshot objects by adding a parameter to the corresponding volumeSnapshotClass
```

### DIFF
--- a/docs/cinder-csi-plugin/using-cinder-csi-plugin.md
+++ b/docs/cinder-csi-plugin/using-cinder-csi-plugin.md
@@ -267,7 +267,9 @@ helm install --namespace kube-system --name cinder-csi ./charts/cinder-csi-plugi
 | StorageClass `parameters`  | `availability`          | `nova`          | String. Volume Availability Zone |
 | StorageClass `parameters`  | `type`                  | Empty String    | String. Name/ID of Volume type. Corresponding volume type should exist in cinder     |
 | VolumeSnapshotClass `parameters` | `force-create`    | `false`         | Enable to support creating snapshot for a volume in in-use status |
-| Inline Volume `volumeAttributes`   | `capacity`              | `1Gi`       | volume size for creating inline volumes|
+| VolumeSnapshotClass `parameters` | `type`            | Empty String    | `snapshot` creates a VolumeSnapshot object linked to a Cinder volume snapshot. `backup` creates a VolumeSnapshot object linked to a cinder volume backup. Defaults to `snapshot` if not defined |
+| VolumeSnapshotClass `parameters` | `backup-max-duration-seconds-per-gb`  | `20`    | Defines the amount of time to wait for a backup to complete in seconds per GB of volume size |
+| Inline Volume `volumeAttributes`   | `capacity`              | `1Gi`       | volume size for creating inline volumes| 
 | Inline Volume `VolumeAttributes`   | `type`              | Empty String  | Name/ID of Volume type. Corresponding volume type should exist in cinder |
 
 ## Local Development
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -17,9 +17,11 @@ limitations under the License.
 package cinder
 
 import (
+	"fmt"
 	"strconv"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/backups"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/snapshots"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes"
 	"github.com/kubernetes-csi/csi-lib-utils/protosanitizer"
@@ -110,36 +112,71 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 	content := req.GetVolumeContentSource()
 	var snapshotID string
-	var sourcevolID string
+	var sourceVolID string
+	var sourceBackupID string
+	var backupsAreEnabled bool
+	backupsAreEnabled, err = cloud.BackupsAreEnabled()
+	klog.V(4).Infof("Backups enabled: %v", backupsAreEnabled)
+	if err != nil {
+		klog.Errorf("Failed to check if backups are enabled: %v", err)
+	}
 
 	if content != nil && content.GetSnapshot() != nil {
 		snapshotID = content.GetSnapshot().GetSnapshotId()
-		_, err := cloud.GetSnapshotByID(snapshotID)
-		if err != nil {
-			if cpoerrors.IsNotFound(err) {
-				return nil, status.Errorf(codes.NotFound, "VolumeContentSource Snapshot %s not found", snapshotID)
-			}
-			return nil, status.Errorf(codes.Internal, "Failed to retrieve the snapshot %s: %v", snapshotID, err)
+
+		snap, err := cloud.GetSnapshotByID(snapshotID)
+		if err != nil && !cpoerrors.IsNotFound(err) {
+			return nil, err
 		}
+		// If the snapshot exists but is not yet available, fail.
+		if err == nil && snap.Status != "available" {
+			return nil, status.Errorf(codes.Unavailable, "VolumeContentSource Snapshot %s is not yet available. status: %s", snapshotID, snap.Status)
+		}
+
+		// In case a snapshot is not found
+		// check if a Backup with the same ID exists
+		if backupsAreEnabled && cpoerrors.IsNotFound(err) {
+			back, err := cloud.GetBackupByID(snapshotID)
+			if err != nil {
+				//If there is an error getting the backup as well, fail.
+				return nil, status.Errorf(codes.NotFound, "VolumeContentSource Snapshot or Backup with ID %s not found", snapshotID)
+			}
+			if back.Status != "available" {
+				// If the backup exists but is not yet available, fail.
+				return nil, status.Errorf(codes.Unavailable, "VolumeContentSource Backup %s is not yet available. status: %s", snapshotID, back.Status)
+			}
+			// If an available backup is found, create the volume from the backup
+			sourceBackupID = snapshotID
+			snapshotID = ""
+		}
+		// In case GetSnapshotByID has error IsNotFound and backups are not enabled
+		// TODO: Change 'snapshotID == ""' to '!backupsAreEnabled' when cloud.BackupsAreEnabled() is correctly implemented
+		if cpoerrors.IsNotFound(err) && snapshotID == "" {
+			return nil, err
+		}
+
 	}
 
 	if content != nil && content.GetVolume() != nil {
-		sourcevolID = content.GetVolume().GetVolumeId()
-		_, err := cloud.GetVolume(sourcevolID)
+		sourceVolID = content.GetVolume().GetVolumeId()
+		_, err := cloud.GetVolume(sourceVolID)
 		if err != nil {
 			if cpoerrors.IsNotFound(err) {
-				return nil, status.Errorf(codes.NotFound, "Source Volume %s not found", sourcevolID)
+				return nil, status.Errorf(codes.NotFound, "Source Volume %s not found", sourceVolID)
 			}
-			return nil, status.Errorf(codes.Internal, "Failed to retrieve the source volume %s: %v", sourcevolID, err)
+			return nil, status.Errorf(codes.Internal, "Failed to retrieve the source volume %s: %v", sourceVolID, err)
 		}
 	}
 
-	vol, err := cloud.CreateVolume(volName, volSizeGB, volType, volAvailability, snapshotID, sourcevolID, &properties)
+	vol, err := cloud.CreateVolume(volName, volSizeGB, volType, volAvailability, snapshotID, sourceVolID, sourceBackupID, properties)
+	// When creating a volume from a backup, the response does not include the backupID.
+	if sourceBackupID != "" {
+		vol.BackupID = &sourceBackupID
+	}
 
 	if err != nil {
 		klog.Errorf("Failed to CreateVolume: %v", err)
 		return nil, status.Errorf(codes.Internal, "CreateVolume failed with error %v", err)
-
 	}
 
 	klog.V(4).Infof("CreateVolume: Successfully created volume %s in Availability Zone: %s of size %d GiB", vol.ID, vol.AvailabilityZone, vol.Size)
@@ -326,6 +363,25 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 
 	name := req.Name
 	volumeID := req.GetSourceVolumeId()
+	snapshotType := req.Parameters[openstack.SnapshotType]
+	filters := map[string]string{"Name": name}
+	backupMaxDurationSecondsPerGB := openstack.BackupMaxDurationSecondsPerGBDefault
+
+	// Current time, used for CreatedAt
+	var ctime *timestamppb.Timestamp
+	// Size of the created snapshot, used to calculate the amount of time to wait for the backup to finish
+	var snapSize int
+	// If true, skips creating a snapshot because a backup already exists
+	var backupAlreadyExists bool
+	var snap *snapshots.Snapshot
+	var backup *backups.Backup
+	var backups []backups.Backup
+	var err error
+
+	// Set snapshot type to 'snapshot' by default
+	if snapshotType == "" {
+		snapshotType = "snapshot"
+	}
 
 	if name == "" {
 		return nil, status.Error(codes.InvalidArgument, "Snapshot name must be provided in CreateSnapshot request")
@@ -335,73 +391,216 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		return nil, status.Error(codes.InvalidArgument, "VolumeID must be provided in CreateSnapshot request")
 	}
 
-	// Verify a snapshot with the provided name doesn't already exist for this tenant
-	var snap *snapshots.Snapshot
+	// Verify snapshot type has a valid value
+	if snapshotType != "snapshot" && snapshotType != "backup" {
+		return nil, status.Error(codes.InvalidArgument, "Snapshot type must be 'backup', 'snapshot' or not defined")
+	}
+	var backupsAreEnabled bool
+	backupsAreEnabled, err = cs.Cloud.BackupsAreEnabled()
+	klog.V(4).Infof("Backups enabled: %v", backupsAreEnabled)
+	if err != nil {
+		klog.Errorf("Failed to check if backups are enabled: %v", err)
+	}
+
+	// Prechecks in case of a backup
+	if snapshotType == "backup" {
+		if !backupsAreEnabled {
+			return nil, status.Error(codes.FailedPrecondition, "Backups are not enabled in Cinder")
+		}
+		// Get a list of backups with the provided name
+		backups, err = cs.Cloud.ListBackups(filters)
+		if err != nil {
+			klog.Errorf("Failed to query for existing Backup during CreateSnapshot: %v", err)
+			return nil, status.Error(codes.Internal, "Failed to get backups")
+		}
+		// If more than one backup with the provided name exists, fail
+		if len(backups) > 1 {
+			klog.Errorf("found multiple existing backups with selected name (%s) during create", name)
+			return nil, status.Error(codes.Internal, "Multiple backups reported by Cinder with same name")
+		}
+
+		if len(backups) == 1 {
+			backup = &backups[0]
+
+			// Verify the existing backup has the same VolumeID, otherwise it belongs to another volume
+			if backup.VolumeID != volumeID {
+				return nil, status.Error(codes.AlreadyExists, "Backup with given name already exists, with different source volume ID")
+			}
+
+			// If a backup of the volume already exists, skip creating the snapshot
+			backupAlreadyExists = true
+			klog.V(3).Infof("Found existing backup %s from volume with ID: %s", name, volumeID)
+		}
+
+		// Get the max duration to wait in seconds per GB of snapshot and fail if parsing fails
+		if item, ok := (req.Parameters)[openstack.BackupMaxDurationPerGB]; ok {
+			backupMaxDurationSecondsPerGB, err = strconv.Atoi(item)
+			if err != nil {
+				klog.Errorf("Setting backup-max-duration-seconds-per-gb failed due to a parsing error: %v", err)
+				return nil, status.Error(codes.Internal, "Failed to parse backup-max-duration-seconds-per-gb")
+			}
+		}
+	}
+
+	// Create the snapshot if the backup does not already exist and wait for it to be ready
+	if !backupAlreadyExists {
+		snap, err = cs.createSnapshot(name, volumeID, req.Parameters)
+		if err != nil {
+			return nil, err
+		}
+
+		ctime = timestamppb.New(snap.CreatedAt)
+		if err = ctime.CheckValid(); err != nil {
+			klog.Errorf("Error to convert time to timestamp: %v", err)
+		}
+
+		snap.Status, err = cs.Cloud.WaitSnapshotReady(snap.ID)
+		if err != nil {
+			klog.Errorf("Failed to WaitSnapshotReady: %v", err)
+			return nil, status.Errorf(codes.Internal, "CreateSnapshot failed with error: %v. Current snapshot status: %v", err, snap.Status)
+		}
+
+		snapSize = snap.Size
+	}
+
+	if snapshotType == "snapshot" {
+		return &csi.CreateSnapshotResponse{
+			Snapshot: &csi.Snapshot{
+				SnapshotId:     snap.ID,
+				SizeBytes:      int64(snap.Size * 1024 * 1024 * 1024),
+				SourceVolumeId: snap.VolumeID,
+				CreationTime:   ctime,
+				ReadyToUse:     true,
+			},
+		}, nil
+	}
+
+	// If snapshotType is 'backup', create a backup from the snapshot and delete the snapshot.
+	if snapshotType == "backup" {
+
+		if !backupAlreadyExists {
+			backup, err = cs.createBackup(name, volumeID, snap, req.Parameters)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		ctime = timestamppb.New(backup.CreatedAt)
+		if err := ctime.CheckValid(); err != nil {
+			klog.Errorf("Error to convert time to timestamp: %v", err)
+		}
+
+		backup.Status, err = cs.Cloud.WaitBackupReady(backup.ID, snapSize, backupMaxDurationSecondsPerGB)
+		if err != nil {
+			klog.Errorf("Failed to WaitBackupReady: %v", err)
+			return nil, status.Error(codes.Internal, fmt.Sprintf("CreateBackup failed with error %v. Current backups status: %s", err, backup.Status))
+		}
+
+		// Necessary to get all the backup information, including size.
+		backup, err = cs.Cloud.GetBackupByID(backup.ID)
+		if err != nil {
+			klog.Errorf("Failed to GetBackupByID after backup creation: %v", err)
+			return nil, status.Error(codes.Internal, fmt.Sprintf("GetBackupByID failed with error %v", err))
+		}
+
+		err = cs.Cloud.DeleteSnapshot(snap.ID)
+		if err != nil && !cpoerrors.IsNotFound(err) {
+			klog.Errorf("Failed to DeleteSnapshot: %v", err)
+			return nil, status.Error(codes.Internal, fmt.Sprintf("DeleteSnapshot failed with error %v", err))
+		}
+	}
+
+	return &csi.CreateSnapshotResponse{
+		Snapshot: &csi.Snapshot{
+			SnapshotId:     backup.ID,
+			SizeBytes:      int64(backup.Size * 1024 * 1024 * 1024),
+			SourceVolumeId: backup.VolumeID,
+			CreationTime:   ctime,
+			ReadyToUse:     true,
+		},
+	}, nil
+
+}
+
+func (cs *controllerServer) createSnapshot(name string, volumeID string, parameters map[string]string) (snap *snapshots.Snapshot, err error) {
+
 	filters := map[string]string{}
 	filters["Name"] = name
+
+	// List existing snapshots with the same name
 	snapshots, _, err := cs.Cloud.ListSnapshots(filters)
 	if err != nil {
 		klog.Errorf("Failed to query for existing Snapshot during CreateSnapshot: %v", err)
 		return nil, status.Error(codes.Internal, "Failed to get snapshots")
 	}
 
+	// If more than one snapshot with the provided name exists, fail
+	if len(snapshots) > 1 {
+		klog.Errorf("found multiple existing snapshots with selected name (%s) during create", name)
+
+		return nil, status.Error(codes.Internal, "Multiple snapshots reported by Cinder with same name")
+	}
+
+	// Verify a snapshot with the provided name doesn't already exist for this tenant
 	if len(snapshots) == 1 {
 		snap = &snapshots[0]
-
 		if snap.VolumeID != volumeID {
 			return nil, status.Error(codes.AlreadyExists, "Snapshot with given name already exists, with different source volume ID")
 		}
 
+		// If the snapshot for the correct volume already exists, return it
 		klog.V(3).Infof("Found existing snapshot %s from volume with ID: %s", name, volumeID)
-
-	} else if len(snapshots) > 1 {
-		klog.Errorf("found multiple existing snapshots with selected name (%s) during create", name)
-		return nil, status.Error(codes.Internal, "Multiple snapshots reported by Cinder with same name")
-
-	} else {
-		// Add cluster ID to the snapshot metadata
-		properties := map[string]string{cinderCSIClusterIDKey: cs.Driver.cluster}
-
-		// see https://github.com/kubernetes-csi/external-snapshotter/pull/375/
-		// Also, we don't want to tag every param but we still want to send the
-		// 'force-create' flag to openstack layer so that we will honor the
-		// force create functions
-		for _, mKey := range []string{"csi.storage.k8s.io/volumesnapshot/name", "csi.storage.k8s.io/volumesnapshot/namespace", "csi.storage.k8s.io/volumesnapshotcontent/name", openstack.SnapshotForceCreate} {
-			if v, ok := req.Parameters[mKey]; ok {
-				properties[mKey] = v
-			}
-		}
-
-		// TODO: Delegate the check to openstack itself and ignore the conflict
-		snap, err = cs.Cloud.CreateSnapshot(name, volumeID, &properties)
-		if err != nil {
-			klog.Errorf("Failed to Create snapshot: %v", err)
-			return nil, status.Errorf(codes.Internal, "CreateSnapshot failed with error %v", err)
-		}
-
-		klog.V(3).Infof("CreateSnapshot %s from volume with ID: %s", name, volumeID)
+		return snap, nil
 	}
 
-	ctime := timestamppb.New(snap.CreatedAt)
-	if err := ctime.CheckValid(); err != nil {
-		klog.Errorf("Error to convert time to timestamp: %v", err)
+	// Add cluster ID to the snapshot metadata
+	properties := map[string]string{cinderCSIClusterIDKey: cs.Driver.cluster}
+
+	// see https://github.com/kubernetes-csi/external-snapshotter/pull/375/
+	// Also, we don't want to tag every param but we still want to send the
+	// 'force-create' flag to openstack layer so that we will honor the
+	// force create functions
+	for _, mKey := range []string{"csi.storage.k8s.io/volumesnapshot/name", "csi.storage.k8s.io/volumesnapshot/namespace", "csi.storage.k8s.io/volumesnapshotcontent/name", openstack.SnapshotForceCreate} {
+		if v, ok := parameters[mKey]; ok {
+			properties[mKey] = v
+		}
 	}
 
-	err = cs.Cloud.WaitSnapshotReady(snap.ID)
+	// TODO: Delegate the check to openstack itself and ignore the conflict
+	snap, err = cs.Cloud.CreateSnapshot(name, volumeID, properties)
 	if err != nil {
-		klog.Errorf("Failed to WaitSnapshotReady: %v", err)
+		klog.Errorf("Failed to Create snapshot: %v", err)
 		return nil, status.Errorf(codes.Internal, "CreateSnapshot failed with error %v", err)
 	}
 
-	return &csi.CreateSnapshotResponse{
-		Snapshot: &csi.Snapshot{
-			SnapshotId:     snap.ID,
-			SizeBytes:      int64(snap.Size * 1024 * 1024 * 1024),
-			SourceVolumeId: snap.VolumeID,
-			CreationTime:   ctime,
-			ReadyToUse:     true,
-		},
-	}, nil
+	klog.V(3).Infof("CreateSnapshot %s from volume with ID: %s", name, volumeID)
+
+	return snap, nil
+}
+
+func (cs *controllerServer) createBackup(name string, volumeID string, snap *snapshots.Snapshot, parameters map[string]string) (*backups.Backup, error) {
+
+	// Add cluster ID to the snapshot metadata
+	properties := map[string]string{cinderCSIClusterIDKey: cs.Driver.cluster}
+
+	// see https://github.com/kubernetes-csi/external-snapshotter/pull/375/
+	// Also, we don't want to tag every param but we still want to send the
+	// 'force-create' flag to openstack layer so that we will honor the
+	// force create functions
+	for _, mKey := range []string{"csi.storage.k8s.io/volumesnapshot/name", "csi.storage.k8s.io/volumesnapshot/namespace", "csi.storage.k8s.io/volumesnapshotcontent/name", openstack.SnapshotForceCreate, openstack.SnapshotType} {
+		if v, ok := parameters[mKey]; ok {
+			properties[mKey] = v
+		}
+	}
+
+	backup, err := cs.Cloud.CreateBackup(name, volumeID, snap.ID, properties)
+	if err != nil {
+		klog.Errorf("Failed to Create backup: %v", err)
+		return nil, status.Error(codes.Internal, fmt.Sprintf("CreateBackup failed with error %v", err))
+	}
+	klog.V(4).Infof("Backup created: %+v", backup)
+
+	return backup, nil
 }
 
 func (cs *controllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteSnapshotRequest) (*csi.DeleteSnapshotResponse, error) {
@@ -413,8 +612,18 @@ func (cs *controllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
 		return nil, status.Error(codes.InvalidArgument, "Snapshot ID must be provided in DeleteSnapshot request")
 	}
 
+	// If volumeSnapshot object was linked to a cinder backup, delete the backup.
+	back, err := cs.Cloud.GetBackupByID(id)
+	if err == nil && back != nil {
+		err = cs.Cloud.DeleteBackup(id)
+		if err != nil {
+			klog.Errorf("Failed to Delete backup: %v", err)
+			return nil, status.Error(codes.Internal, fmt.Sprintf("DeleteBackup failed with error %v", err))
+		}
+	}
+
 	// Delegate the check to openstack itself
-	err := cs.Cloud.DeleteSnapshot(id)
+	err = cs.Cloud.DeleteSnapshot(id)
 	if err != nil {
 		if cpoerrors.IsNotFound(err) {
 			klog.V(3).Infof("Snapshot %s is already deleted.", id)
@@ -670,6 +879,16 @@ func getCreateVolumeResponse(vol *volumes.Volume, ignoreVolumeAZ bool, accessibl
 			Type: &csi.VolumeContentSource_Volume{
 				Volume: &csi.VolumeContentSource_VolumeSource{
 					VolumeId: vol.SourceVolID,
+				},
+			},
+		}
+	}
+
+	if vol.BackupID != nil && *vol.BackupID != "" {
+		volsrc = &csi.VolumeContentSource{
+			Type: &csi.VolumeContentSource_Snapshot{
+				Snapshot: &csi.VolumeContentSource_SnapshotSource{
+					SnapshotId: *vol.BackupID,
 				},
 			},
 		}
--- a/pkg/csi/cinder/openstack/openstack.go
+++ b/pkg/csi/cinder/openstack/openstack.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
+	"github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/backups"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/snapshots"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
@@ -44,7 +45,7 @@ func AddExtraFlags(fs *pflag.FlagSet) {
 }
 
 type IOpenStack interface {
-	CreateVolume(name string, size int, vtype, availability string, snapshotID string, sourcevolID string, tags *map[string]string) (*volumes.Volume, error)
+	CreateVolume(name string, size int, vtype, availability string, snapshotID string, sourceVolID string, sourceBackupID string, tags map[string]string) (*volumes.Volume, error)
 	DeleteVolume(volumeID string) error
 	AttachVolume(instanceID, volumeID string) (string, error)
 	ListVolumes(limit int, startingToken string) ([]volumes.Volume, string, error)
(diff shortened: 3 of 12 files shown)


jichenjc:
>>>  force-create: "false"
  type: "snapshot"

Sebastian-RG:
It seems the E2E test failed, but I can't tell how it's related to the changes I made :thinking: 
```
 + lib/databases/mysql:configure_database_mysql:107 :   '[' mysql == mariadb ']'
+ lib/databases/mysql:configure_database_mysql:108 :   sudo mysqladmin -u root password password
mysqladmin: connect to server at 'localhost' failed
error: 'Access denied for user 'root'@'localhost' (using password: NO)'
+ lib/databases/mysql:configure_database_mysql:108 :   true
+ lib/databases/mysql:configure_database_mysql:113 :   is_ubuntu 
```

```
Nov 27 18:19:19 devstack neutron-server[63269]: DEBUG dbcounter [-] [63269] Writing DB stats neutron:SELECT=99,neutron:UPDATE=1 {{(pid=63269) stat_writer /usr/local/lib/python3.10/dist-packages/dbcounter.py:114}}
Nov 27 18:19:19 devstack neutron-server[63269]: WARNING neutron.plugins.ml2.drivers.ovn.mech_driver.ovsdb.ovn_client [None req-7f03ad2a-0a07-4a62-9942-c63275e34d16 None None] No hosting information found for port 7a747b1a-aa6d-4d21-b14f-765b9978e33b: RuntimeError: No hosting information found for port 7a747b1a-aa6d-4d21-b14f-765b9978e33b
Nov 27 18:19:19 devstack neutron-server[63268]: DEBUG ovsdbapp.backend.ovs_idl.transaction [-] Running txn n=1 command(idx=0): DbSetCommand(_result=None, table=Logical_Switch_Port, record=03cac3e3-48cb-46dc-95ff-acae066ef47a, col_values=(('external_ids', {'neutron:host_id': 'devstack'}),), if_exists=True) {{(pid=63268) do_commit /usr/local/lib/python3.10/dist-packages/ovsdbapp/backend/ovs_idl/transaction.py:89}} 
```

zetaab (changes requested):

pkg/csi/cinder/controllerserver.go
```diff
@@ -337,6 +373,11 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		return nil, status.Error(codes.InvalidArgument, "VolumeID must be provided in CreateSnapshot request")
 	}
 
+	// Verify snapshot type has a valid value
+	if snapshotType != "snapshot" && snapshotType != "backup" {
```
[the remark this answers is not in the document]
Sebastian-RG: Line 361-366 takes care of this issue
```
	snapshotType, snapshotTypeDeclared := req.Parameters[openstack.SnapshotType]
	// Set snapshot type to 'snapshot' by default
	if !snapshotTypeDeclared {
		snapshotType = "snapshot"
	}
```

pkg/csi/cinder/openstack/openstack_backups.go
```diff
@@ -0,0 +1,197 @@
...
+	mc := metrics.NewMetricContext("backup", "list")
+	err := backups.List(os.blockstorage, opts).EachPage(func(page pagination.Page) (bool, error) {
+		var err error
+
+		backs, err = backups.ExtractBackups(page)
+		if err != nil {
+			return false, err
+		}
+
+		nextPageURL, err := page.NextPageURL()
```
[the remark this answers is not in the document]
Sebastian-RG: I adapted the code from another section that did this. Turns out this is needed when implementing `ListResource` capabilities in CSI. At least i think so based on ListVolumes' and ListSnapshots' List response in controllerserver.go

```
	return &csi.ListVolumesResponse{
		Entries:   ventries,
		NextToken: nextPageToken,
	}
```

```
	return &csi.ListSnapshotsResponse{
		Entries:   sentries,
		NextToken: nextPageToken,
	}, nil
```
But this is not needed in this case. I should have checked before submitting

pkg/csi/cinder/openstack/openstack_backups.go
```diff
@@ -0,0 +1,197 @@
...
+
+func (os *OpenStack) WaitBackupReady(backupID string) error {
+	backoff := wait.Backoff{
+		Duration: backupReadyDuration,
+		Factor:   backupReadyFactor,
+		Steps:    backupReadySteps,
+	}
+
+	// Waits for a maximum of 295 seconds
+	err := wait.ExponentialBackoff(backoff, func() (bool, error) {
```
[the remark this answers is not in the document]
Sebastian-RG: What do you think about this approach? Any ideas for a factor?

pkg/csi/cinder/controllerserver.go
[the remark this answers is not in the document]
Sebastian-RG: Reverted because slices is part of golang 1.21 https://pkg.go.dev/slices?tab=versions :smiling_face_with_tear:

pkg/csi/cinder/openstack/openstack_backups.go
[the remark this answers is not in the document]
jeffyjf: What do you think of introduce a new parameter as the factor, just like: `max-duration-per-gb`.

pkg/csi/cinder/controllerserver.go
```diff
@@ -337,6 +394,11 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		return nil, status.Error(codes.InvalidArgument, "VolumeID must be provided in CreateSnapshot request")
 	}
 
+	// Verify snapshot type has a valid value
+	if snapshotType != "snapshot" && snapshotType != "backup" && snapshotType != "" {
```
jeffyjf, line 398: What if the user set `SnapshotType` as `""`.

IMO,  `&& snapshotType != ""` should be removed.  `if snapshotType != "snapshot" && snapshotType != "backup" { ... }` is enough.

jeffyjf (changes requested):

pkg/csi/cinder/openstack/openstack_backups.go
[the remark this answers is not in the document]
Sebastian-RG: And we get `max-duration-per-gb` from an environment variable to configure it from K8S?

pkg/csi/cinder/controllerserver.go
```diff
@@ -394,6 +476,61 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+		// see https://github.com/kubernetes-csi/external-snapshotter/pull/375/
+		// Also, we don't want to tag every param but we still want to send the
+		// 'force-create' flag to openstack layer so that we will honor the
+		// force create functions
+		for _, mKey := range []string{"csi.storage.k8s.io/volumesnapshot/name", "csi.storage.k8s.io/volumesnapshot/namespace", "csi.storage.k8s.io/volumesnapshotcontent/name", openstack.SnapshotForceCreate, openstack.SnapshotType} {
+			if v, ok := req.Parameters[mKey]; ok {
+				properties[mKey] = v
+			}
+		}
+		backup, err = cs.Cloud.CreateBackup(name, volumeID, snap.ID, &properties)
```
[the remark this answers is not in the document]
Sebastian-RG: My mistake. In case of a snapshot it works because here it will detect the volumeID being the same
```
	if len(snapshots) == 1 {
		snap = &snapshots[0]

		if snap.VolumeID != volumeID {
			return nil, status.Error(codes.AlreadyExists, "Snapshot with given name already exists, with different source volume ID")
		}
		klog.V(3).Infof("Found existing snapshot %s from volume with ID: %s", name, volumeID)
}
```

i will change the logic so that it also does this check before trying to create the backup again.
It will go directly to `waitBackupReady()` if a backup with the same volumeID and name exists.

kayrus:
@Sebastian-RG can you fill in the release notes in the initial PR message?

pkg/csi/cinder/controllerserver.go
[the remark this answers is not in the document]
Sebastian-RG: There should be no issues with cinder-csi restarts now

docs/cinder-csi-plugin/using-cinder-csi-plugin.md
```diff
@@ -254,6 +254,8 @@ helm install --namespace kube-system --name cinder-csi ./charts/cinder-csi-plugi
 | StorageClass `parameters`  | `availability`          | `nova`          | String. Volume Availability Zone |
 | StorageClass `parameters`  | `type`                  | Empty String    | String. Name/ID of Volume type. Corresponding volume type should exist in cinder     |
 | VolumeSnapshotClass `parameters` | `force-create`    | `false`         | Enable to support creating snapshot for a volume in in-use status |
+| VolumeSnapshotClass `parameters` | `type`            | Empty String    | `snapshot` creates a VolumeSnapshot object linked to a Cinder volume snapshot. `backup` creates a VolumeSnapshot object linked to a cinder volume backup. Defaults to `snapshot` if not defined |
+| VolumeSnapshotClass `parameters` | `backup-max-duration-per-gb`  | `5`    | Defines the amount of time to wait for a backup to complete per volume size in GB |
```
zetaab, line 258: duration but what is the unit? seconds? minutes?

zetaab: if that is seconds, it should be quite fast. 200MB/s, well perhaps worth to use something like 50MB/s by default, which means 20 default value?

zetaab (commented):
cc @dulek @jichenjc asking for review

docs/cinder-csi-plugin/using-cinder-csi-plugin.md
Sebastian-RG: Done :)

zetaab:
looks good, waiting for reviews from another contributors/maintainers

pkg/csi/cinder/controllerserver.go
```diff
@@ -345,6 +408,14 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		return nil, status.Error(codes.Internal, "Failed to get snapshots")
 	}
 
+	// Verify a backup with the provided name doesn't already exist for this tenant
+	var backup *backups.Backup
+	backups, err := cs.Cloud.ListBackups(filters)
```
jeffyjf, line 413: Do we still need to list backups even if `snaphotType` is `snapshot`?

pkg/csi/cinder/controllerserver.go
```diff
@@ -392,6 +475,78 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+		// Also, we don't want to tag every param but we still want to send the
+		// 'force-create' flag to openstack layer so that we will honor the
+		// force create functions
+		for _, mKey := range []string{"csi.storage.k8s.io/volumesnapshot/name", "csi.storage.k8s.io/volumesnapshot/namespace", "csi.storage.k8s.io/volumesnapshotcontent/name", openstack.SnapshotForceCreate, openstack.SnapshotType, openstack.BackupMaxDurationPerGB} {
+			if v, ok := req.Parameters[mKey]; ok {
+				properties[mKey] = v
+			}
+		}
+
+		// We repeat the check from the snapshot creation in case Cinder-CSI gets restarted while a backup is in progress.
```
jeffyjf, line 494: I'm not sure I understand here. Why do we need to check it again?

pkg/csi/cinder/openstack/openstack_backups.go
```diff
@@ -0,0 +1,188 @@
...
+const (
+	backupReadyStatus                    = "available"
+	backupDescription                    = "Created by OpenStack Cinder CSI driver"
+	backupReadyDuration                  = 5 * time.Second
+	backupReadyFactor                    = 1.2
+	backupReadySteps                     = 15
+	backupMaxDurationSecondsPerGBDefault = 20
+	BackupMaxDurationPerGB               = "backup-max-duration-seconds-per-gb"
+)
+
```
jeffyjf, line 44: Could you add some comments on top of these public functions? Just like `openstack_instances.go` and `openstack_snapshots.go` did. So that `godoc` can auto-generate docs.

pkg/csi/cinder/controllerserver.go
```diff
@@ -392,6 +475,78 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		klog.Errorf("Failed to WaitSnapshotReady: %v", err)
 		return nil, status.Errorf(codes.Internal, "CreateSnapshot failed with error %v", err)
 	}
+	// If snapshotType is 'backup', create a backup from the snapshot and delete the snapshot.
```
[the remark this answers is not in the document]
zetaab: it makes the backup process robust. Lets say that we have 5TB cinder volume, you will take "cinder backup" from it. It means that it will copy entire 5TB volume from storage system A to storage system B by using external network. It will take a lot of time. During copying the content of the volume can change a lot and it means that in some situations it will corrupt the backup. Usually when taking snapshot from existing volume it means that it can be really snapshot if the storage system supports that (taking snapshot inside same volume backend can support that, but not all support that still..). Which means that the possibility to changes and issues are a lot smaller.

pkg/csi/cinder/controllerserver.go
Sebastian-RG: Basically, the previous check stops the controller from creating a new snapshot if either a snapshot or a backup with the same name exists
If a snapshot with the sane name exists, it skips to `WaitSnapshotReady`.
If a backup with the same name exists it sets `skipSnapshot` to true and skips waiting for the snapshot. It also skips deleting it. (I just added this `skipSnapshot` part)

The logic for the backup starts after we know that the snapshot is ready. We do the same thing where we avoid creating a new backup if a backup with the same name already exists, and skip to `WaitBackupReady`.

We always do these checks because the controller pod might restart after a snapshot or backup is created, but before a `CreateSnapshotResponse` has been returned.

pkg/csi/cinder/controllerserver.go
```diff
@@ -111,15 +113,29 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+		snapshotError, err := GetSnapshot(cloud, snapshotID)
 		if err != nil {
-			if cpoerrors.IsNotFound(err) {
-				return nil, status.Errorf(codes.NotFound, "VolumeContentSource Snapshot %s not found", snapshotID)
-			}
-			return nil, status.Errorf(codes.Internal, "Failed to retrieve the snapshot %s: %v", snapshotID, err)
+			return nil, err
+		}
+		// Look for a Backup with the same ID as the snapshot.
+		BackupError, err := GetBackup(cloud, snapshotID)
```
kayrus, line 125: please try to avoid variables that start with a capital letter. it's a good practice in Go to use lowercase letters for variables that are not intended to be exported.
```suggestion
		backupError, err := GetBackup(cloud, snapshotID)
```

pkg/csi/cinder/controllerserver.go
```diff
@@ -111,15 +113,29 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+
+		// If Backup and snapshot not found
+		if snapshotError != nil && BackupError != nil {
+			return nil, status.Errorf(codes.NotFound, "VolumeContentSource Snapshot or Backup %s not found", snapshotID)
+		}
+
+		// Case: Backup found, Snapshot not found. Create Volume from Backup.
+		if BackupError == nil && snapshotError != nil {
+			sourcebackupID = snapshotID
+			snapshotID = ""
```
kayrus, line 138: this whole block can be simplified:
```golang
        if snapshotError != nil {
                if BackupError != nil {
                        return nil, status.Errorf(codes.NotFound, "VolumeContentSource Snapshot or Backup %s not found", snapshotID)
                }
                sourcebackupID = snapshotID
                snapshotID = ""
        }
```

pkg/csi/cinder/controllerserver.go
```diff
@@ -147,6 +167,38 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	return getCreateVolumeResponse(vol, ignoreVolumeAZ, req.GetAccessibilityRequirements()), nil
 }
 
+func GetSnapshot(cloud openstack.IOpenStack, snapshotID string) (error, error) {
+	snap, errsnap := cloud.GetSnapshotByID(snapshotID)
```
kayrus, line 171: Could you just use `err` instead of `errsnap`. This is not the case, when you need to differentiate between multiple `err` variables. The same applies to the `GetBackup` func.

pkg/csi/cinder/controllerserver.go
```diff
@@ -147,6 +167,38 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	return getCreateVolumeResponse(vol, ignoreVolumeAZ, req.GetAccessibilityRequirements()), nil
 }
 
+func GetSnapshot(cloud openstack.IOpenStack, snapshotID string) (error, error) {
```
kayrus, line 170: if you're calling this func only from this go package, it's doesn't make sense to export the function. Moreover you can convert it to a `controllerServer` private method.

pkg/csi/cinder/controllerserver.go
```diff
@@ -382,15 +477,92 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+		// 'force-create' flag to openstack layer so that we will honor the
+		// force create functions
+		for _, mKey := range []string{"csi.storage.k8s.io/volumesnapshot/name", "csi.storage.k8s.io/volumesnapshot/namespace", "csi.storage.k8s.io/volumesnapshotcontent/name", openstack.SnapshotForceCreate, openstack.SnapshotType, openstack.BackupMaxDurationPerGB} {
+			if v, ok := req.Parameters[mKey]; ok {
+				properties[mKey] = v
+			}
+		}
+
+		// We repeat the check from the snapshot creation in case Cinder-CSI gets restarted while a backup is in progress.
+		if len(backups) == 1 {
```
kayrus, line 511: try to avoid `if/then-else-if/else` blocks.

You can do the following:

```golang
if len(backups) > 1 {
   return err
}

if len(backup) == 0 {
      // create a backup
} else {
    // do smth
}
}
```

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -60,6 +61,11 @@ type IOpenStack interface {
 	DeleteSnapshot(snapID string) error
 	GetSnapshotByID(snapshotID string) (*snapshots.Snapshot, error)
 	WaitSnapshotReady(snapshotID string) error
+	CreateBackup(name, volID string, snapshotID string, tags *map[string]string) (*backups.Backup, error)
```
kayrus, line 64: it there a reason to pass a pointer to the `map[string]string`?

kayrus (changes requested):
@Sebastian-RG thanks for the PR. See my comments.

pkg/csi/cinder/openstack/openstack.go
Sebastian-RG: I couldn't find the reason why they did it this way in `CreateSnapshot()`, so I just did the same. If I undestood the reason I would be more comfortable changing it :sweat_smile:

Sebastian-RG:
Hi! Are all comments resolved?

zetaab (approved):
/approve

hopefully you tested the modified version, as we do not have integration tests that everything works. Anyway, I do not see huge issue merging this as we can do fixes quite fast if we see any issues with it. This PR is already +600 rows so quite difficult to get everything working with the first time.

kayrus:
/hold
I have some more comments on the code. I'll post a new CR soon.

pkg/csi/cinder/controllerserver.go
```diff
@@ -147,6 +164,38 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	return getCreateVolumeResponse(vol, ignoreVolumeAZ, req.GetAccessibilityRequirements()), nil
 }
 
+func (cs *controllerServer) getSnapshot(cloud openstack.IOpenStack, snapshotID string) (error, error) {
+	snap, err := cloud.GetSnapshotByID(snapshotID)
```
kayrus, line 168: I don't see a reason in returning two errors. Can you return a single error? Current code is very complex and hard to read.

pkg/csi/cinder/controllerserver.go
```diff
@@ -147,6 +164,38 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+	if err != nil && !cpoerrors.IsNotFound(err) {
+		return err, status.Errorf(codes.Internal, "Failed to retrieve the snapshot %s: %v", snapshotID, err)
+	}
+
+	return err, nil
+}
+
+func (cs *controllerServer) getBackup(cloud openstack.IOpenStack, snapshotID string) (error, error) {
+	// Backup exists but is not yet available.
+	back, err := cloud.GetBackupByID(snapshotID)
```
kayrus, line 183: you don't need to pass the `cloud openstack.IOpenStack` since it's a part of the `cs *controllerServer` struct.
```suggestion
	back, err := cs.Cloud.GetBackupByID(snapshotID)
```

kayrus: while I read this code more, I am increasingly convinced that it is quite possible to get rid of these two methods: `getSnapshot` and `getBackup`. Probably you can combine the logic from them and `CreateVolume`'s one into one method.

pkg/csi/cinder/controllerserver.go
```diff
@@ -326,6 +375,23 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+	// Default to 1 GB
+	snapSize := 1
+	// If true, skips creating a snapshot because a backup already exists
+	var skipSnapshot bool
+	var snap *snapshots.Snapshot
+	var backup *backups.Backup
+	var backups []backups.Backup
+
+	// Set snapshot type to 'snapshot' by default
+	if !snapshotTypeDeclared || snapshotType == "" {
```
kayrus, line 392: IMO the `snapshotTypeDeclared` is useless, especially if a snapshot type is explicitly set to an empty string. Relates to an already discussed comment below.

pkg/csi/cinder/controllerserver.go
```diff
@@ -345,6 +415,15 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		return nil, status.Error(codes.Internal, "Failed to get snapshots")
 	}
 
+	// Verify a backup with the provided name doesn't already exist for this tenant
```
kayrus, line 418: the below code is extremely complex and hard to read. I'd suggest to split it into two parts: snapshots only, backups only. The mixing of number if/then/else-if/else-if/else is not an option.

pkg/csi/cinder/openstack/openstack_backups.go
```diff
@@ -0,0 +1,202 @@
...
+		case "Marker":
+			opts.Marker = val
+		case "Limit":
+			opts.Limit, _ = strconv.Atoi(val)
+		default:
+			klog.V(3).Infof("Not a valid filter key %s", key)
+		}
+	}
+	mc := metrics.NewMetricContext("backup", "list")
+	err := backups.List(os.blockstorage, opts).EachPage(func(page pagination.Page) (bool, error) {
```
kayrus, line 116: since you're not using the advantage of page iteration (stop pagination once the required object is found), you can just use the `ExtractBackupsInto` call:

```go
        allPages, err := backups.List(client, listOpts).AllPages()
        if err != nil {
                panic(err)
        }

        allBackups, err := backups.ExtractBackups(allPages)
        if err != nil {
                panic(err)
        }
```

pkg/csi/cinder/openstack/openstack_backups.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	return backup, nil
+}
+
+// WaitBackupReady waits until backup is ready. It waits longer depending on
+// the size of the corresponding snapshot.
+func (os *OpenStack) WaitBackupReady(backupID string, snapshotSize int, tags *map[string]string) error {
+	var err error
+
+	backup_max_duration_seconds_per_gb := backupMaxDurationSecondsPerGBDefault
```
kayrus, line 157: please follow a common golang variable name convention: https://google.github.io/styleguide/go/decisions#underscores

zetaab:
@Sebastian-RG do you have time to make those changes that was asked? Kubernetes 1.29.0 got released yesterday and we are soonish making release from the new cloud provider. I would like to include this PR in release

Sebastian-RG:
> @Sebastian-RG do you have time to make those changes that was asked? Kubernetes 1.29.0 got released yesterday and we are soonish making release from the new cloud provider. I would like to include this PR in release

I can update the PR by tommorrow

pkg/csi/cinder/controllerserver.go
Sebastian-RG: I deleted both functions, I hope the code is easier to read now

Sebastian-RG:
@kayrus @jichenjc @zetaab thank you for your comments on the PR. I have updated it as requested. Please re-review it when you have the time.
Thank you for your patience in the process!

Sebastian-RG:
Hi @kayrus! Any additional comments?

zetaab:
@kayrus could you check this

pkg/csi/cinder/openstack/openstack.go
kayrus: I'd change the `tags *map[string]string` to `tags map[string]string` in all methods, since there is no need to pass the pointer. https://play.golang.com/p/-NjjexjOEx5

pkg/csi/cinder/openstack/openstack_backups.go
```diff
@@ -0,0 +1,195 @@
...
+			opts.Limit, _ = strconv.Atoi(val)
+		default:
+			klog.V(3).Infof("Not a valid filter key %s", key)
+		}
+	}
+	mc := metrics.NewMetricContext("backup", "list")
+
+	allPages, err := backups.List(os.blockstorage, opts).AllPages()
+	if err != nil {
+		panic(err)
```
kayrus, line 118: why do you use `panic(err)`?

```suggestion
		return nil, err
```

pkg/csi/cinder/openstack/openstack_backups.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	allBackups, err = backups.ExtractBackups(allPages)
+	if err != nil {
+		panic(err)
```
kayrus, line 122:
```suggestion
		return nil, err
```

pkg/csi/cinder/controllerserver.go
```diff
@@ -335,73 +381,208 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+
+		ctime = timestamppb.New(backup.CreatedAt)
+		if err := ctime.CheckValid(); err != nil {
+			klog.Errorf("Error to convert time to timestamp: %v", err)
+		}
+
+		// Get the max duration to wait in seconds per GB of snapshot
+		backupMaxDurationSecondsPerGB, err = strconv.Atoi(openstack.BackupMaxDurationPerGB)
+		if err != nil {
+			klog.V(5).Infof("Set backup-max-duration-seconds-per-gb to 20 due to: %v", err)
```
kayrus, line 469:
```suggestion
			klog.V(5).Infof("Set backup-max-duration-seconds-per-gb to %d due to parsing error: %v", openstack.BackupMaxDurationSecondsPerGBDefault, err)
```

pkg/csi/cinder/openstack/openstack_backups.go
```diff
@@ -0,0 +1,195 @@
...
+	_, cancel := context.WithTimeout(context.Background(), duration*time.Second)
+	defer cancel()
+	var ok bool
+	var err error
+	for !ok {
+		ok, err = os.backupIsReady(backupID)
+		if err != nil {
+			return err
+		}
+		time.Sleep(7 * time.Second)
```
kayrus, line 181: can you put the `7` to a const, or to a config property?

pkg/csi/cinder/openstack/openstack_backups.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// Supporting function for WaitBackupReady().
+// Allows for a timeout while waiting for the backup to be ready.
+func (os *OpenStack) waitBackupReadyWithContext(ctx context.Context, backupID string, duration time.Duration) error {
+	_, cancel := context.WithTimeout(context.Background(), duration*time.Second)
+	defer cancel()
+	var ok bool
+	var err error
+	for !ok {
```
kayrus, line 176: It's better to use context's `Done` channel, see https://pkg.go.dev/context#WithTimeout

```go
	ctx, cancel := context.WithTimeout(context.Background(), duration)
	defer cancel()

	ticker := time.NewTicker(10 * time.Second)
	defer ticker.Stop()

	for {
		select {
		case <-ticker.C:
			ok, err = os.backupIsReady(backupID)
			if err != nil {
				return err
			}

			if ok {
				return nil
			}
		case <-ctx.Done():
			return ctx.Err()
		}
	}

	return nil
```

pkg/csi/cinder/openstack/openstack_backups.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	return backup, nil
+}
+
+// WaitBackupReady waits until backup is ready. It waits longer depending on
+// the size of the corresponding snapshot.
+func (os *OpenStack) WaitBackupReady(backupID string, snapshotSize int, backupMaxDurationSecondsPerGB int) error {
+	var err error
+
+	duration := time.Duration(backupMaxDurationSecondsPerGB*snapshotSize + 30)
```
kayrus, line 158: can you put the `30` to a const, or to a config property?

pkg/csi/cinder/openstack/openstack_backups.go
```diff
@@ -0,0 +1,195 @@
...
+	"k8s.io/cloud-provider-openstack/pkg/metrics"
+	"k8s.io/klog/v2"
+)
+
+const (
+	backupReadyStatus                    = "available"
+	backupDescription                    = "Created by OpenStack Cinder CSI driver"
+	backupReadyDuration                  = 5 * time.Second
+	backupReadyFactor                    = 1.2
+	backupReadySteps                     = 15
```
kayrus, line 39: these constants are not used anywhere.

kayrus (changes requested):
see my recent comments

pkg/csi/cinder/controllerserver.go
kayrus: Actually I think that it makes sense to parse the `BackupMaxDurationPerGB` during the controller startup and fail with an error, if the value cannot be parsed.

zetaab:
@Sebastian-RG do you have time to update the PR? Hopefully it will be the last one. We are soon releasing 1.29.0 and asking because of that. If you cannot do that, I can help you

pkg/csi/cinder/controllerserver.go
```diff
@@ -335,73 +381,208 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+		}
+
+		// Necessary to get all the backup information, including size.
+		backup, err = cs.Cloud.GetBackupByID(backup.ID)
+		if err != nil {
+			klog.Errorf("Failed to GetBackupByID after backup creation: %v", err)
+			return nil, status.Error(codes.Internal, fmt.Sprintf("GetBackupByID failed with error %v", err))
+		}
+
+		if !backupAlreadyExists {
```
dulek, line 486: What if backup and snapshot got created, but then `GetBackupByID()` in line 480 failed because of some transient reason? That way we orphan the snapshot. I think we should always attempt to delete the snapshot if `snapshotType="backup"`.

pkg/csi/cinder/controllerserver.go
```diff
@@ -335,73 +381,208 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+			err = cs.Cloud.DeleteSnapshot(snap.ID)
+			if err != nil {
```
dulek, line 488: We should ignore NotFound here.

dulek (changes requested):
Quite a few remarks, I don't think we can merge this before they're fixed.

I also might have had commented with issues on the code that you're only moving around, please feel free to inform me about this in the comment and ignore that one.

pkg/csi/cinder/controllerserver.go
Sebastian-RG: here I was following what they do for the force create flag for snapshots:
```
func (os *OpenStack) CreateSnapshot(name, volID string, tags map[string]string) (*snapshots.Snapshot, error) {

	force := false
	// if no flag given, then force will be false by default
	// if flag it given , check it
	if item, ok := (tags)[SnapshotForceCreate]; ok {
		var err error
		force, err = strconv.ParseBool(item)
		if err != nil {
			klog.V(5).Infof("Make force create flag to false due to: %v", err)
		}

		delete(tags, SnapshotForceCreate)
	}
...
```
Where if the parsing of an argument that comes from the `volumeSnapshotClass` fails, it goes to the default value instead of failing.

However now that I think about it, in the cases where you meant to set force create flag to true and it did not parse correctly, the snapshot would have failed anyway because it is necessary to have it set to true to create the snapshot...

I can make it fail at the start of `CreateSnapshot()`, but not at controller startup since you only get this value from `req.parameters`. It's from the `volumeSnapshotClass`.

pkg/csi/cinder/openstack/openstack_backups.go
Sebastian-RG: Thank you, this looks much better than what I made up

pkg/csi/cinder/controllerserver.go
```diff
@@ -110,31 +112,56 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
-		if err != nil {
-			if cpoerrors.IsNotFound(err) {
-				return nil, status.Errorf(codes.NotFound, "VolumeContentSource Snapshot %s not found", snapshotID)
+
+		snap, snapErr := cs.Cloud.GetSnapshotByID(snapshotID)
+		if snapErr != nil && !cpoerrors.IsNotFound(snapErr) {
+			return nil, snapErr
+		}
+		// If the snapshot exists but is not yet available, fail.
+		if snapErr == nil && snap.Status != "available" {
```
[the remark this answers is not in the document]
Sebastian-RG: Since the check is 
`if snapErr == nil && snap.Status != "available" {...`
It first checks for the error to be `nil` (so snap exists) and only then would it check for the snap status, so we are protected :)

pkg/csi/cinder/controllerserver.go
```diff
@@ -110,31 +112,56 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+			return nil, status.Errorf(codes.NotFound, "VolumeContentSource Snapshot %s is not yet available. status: %s", snapshotID, snap.Status)
```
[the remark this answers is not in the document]
Sebastian-RG: ty

pkg/csi/cinder/controllerserver.go
```diff
@@ -110,31 +112,56 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+		}
+
+		// In case a snapshot is not found
+		// check if a Backup with the same ID exists
+		if cpoerrors.IsNotFound(snapErr) {
+			back, err := cs.Cloud.GetBackupByID(snapshotID)
+			if err != nil {
+				//If there is an error getting the backup as well, fail.
+				return nil, status.Errorf(codes.NotFound, "VolumeContentSource Snapshot or Backup with ID %s not found", snapshotID)
+			}
```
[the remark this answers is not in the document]
Sebastian-RG: So, since we have the same interface for looking for an snapshot and looking for a backup, we first check if a snapshot with name N exists, and if it doesn't, we check for a backup with name N. However, some systems will not support cinder backups as they are not required. 

This will lead to this `err` being a "real" (not not found) error many times, but it is not necessarily an issue.

So:

if the system does not support backups, and there is an error: No issue, we should return codes.NotFound due to snapshot not being found
 
if the system supports backups, and this is a NotFound error: No issue, we should return codes.NotFound due to snapshot and backup not being found

if the system supports backups, and this has a real error: Oops!, we should have sent codes.Internal

So we are covering 2/3 cases. Since we can't differentiate cases 1 and 3 (because we can't know if the cinder installation supports backups), we should aim for the case where the users that aren't using this specific feature do not get affected

pkg/csi/cinder/controllerserver.go
```diff
@@ -110,31 +112,56 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+			return nil, status.Errorf(codes.Internal, "Failed to retrieve the source volume %s: %v", sourceVolID, err)
 		}
 	}
 
-	vol, err := cloud.CreateVolume(volName, volSizeGB, volType, volAvailability, snapshotID, sourcevolID, &properties)
+	vol, err := cloud.CreateVolume(volName, volSizeGB, volType, volAvailability, snapshotID, sourceVolID, sourceBackupID, &properties)
+	// When creating a volume from a backup, the response does not include the backupID.
+	if sourceBackupID != "" {
+		vol.BackupID = &sourceBackupID
+	}
```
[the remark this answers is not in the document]
Sebastian-RG: It can't return a backupID since the returned object comes from the actual cinder volume object, which doesn't have a field to store what backup it was created from :sweat_smile:

pkg/csi/cinder/openstack/openstack_backups.go
```diff
@@ -0,0 +1,195 @@
...
+
+// Supporting function for waitBackupReadyWithContext().
+// Returns true when the backup is ready.
+func (os *OpenStack) backupIsReady(backupID string) (bool, error) {
+	backup, err := os.GetBackupByID(backupID)
+	if err != nil {
+		return false, err
+	}
+
+	return backup.Status == backupReadyStatus, nil
```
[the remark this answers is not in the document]
Sebastian-RG: Yup! We do this in the `waitBackupReadyWithContext()` function, which calls this one (`backupIsReady()`). If this function returns an error, the error is passed to `waitBackupReadyWithContext()` and returned all the way up

pkg/csi/cinder/controllerserver.go
```diff
@@ -413,8 +594,18 @@ func (cs *controllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
 		return nil, status.Error(codes.InvalidArgument, "Snapshot ID must be provided in DeleteSnapshot request")
 	}
 
+	// If volumeSnapshot object was linked to a cinder backup, delete the backup.
+	back, err := cs.Cloud.GetBackupByID(id)
```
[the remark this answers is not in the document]
Sebastian-RG: Hmm, how do I match for IsNotFound errors? Do i just search for "not found" in `err.Error()`? My environment for testing is temporarily down so if you could help me with this it would be great :pray:

pkg/csi/cinder/controllerserver.go
Sebastian-RG: Agree! Same with previous comment, could you help me with the method to match IsNotFound errors so I can ignore them?

Sebastian-RG:
Fixed most of the comments. 2 comments from @dulek pending. Since my testing environment is temporarily down I can't figure out how to check for IsNotFound errors. Any help with that would be very much appreciated.

pkg/csi/cinder/controllerserver.go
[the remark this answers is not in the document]
dulek: All clear, thanks!

pkg/csi/cinder/controllerserver.go
dulek: [This should be it, isn't it? ](https://github.com/kubernetes/cloud-provider-openstack/blob/1fdebaa1c3c29fd27772c75c2a0d9b512cdd7983/pkg/util/errors/errors.go#L45-L65)

Your code is using it in a few places already.

pkg/csi/cinder/controllerserver.go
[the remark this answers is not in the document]
Sebastian-RG: Since we can't know if the cinder installation supports backups, I think this is the best method to do it.
Since DeleteBackup() is more error prone than GetBackup().

I think we should first test with GetBackup, and if it does not work, we assume this cinder installation does not support Backups or the backup does not exist. In either case we ignore the error. If it does work but DeleteBackup fails, then we should fail

pkg/csi/cinder/openstack/openstack_backups.go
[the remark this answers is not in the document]
Sebastian-RG: if os.GetBackupByID() returns any error, os.backupIsReady() returns it:
```
func (os *OpenStack) backupIsReady(backupID string) (bool, error) {
	backup, err := os.GetBackupByID(backupID)
	if err != nil {
		return false, err
	}

	return backup.Status == backupReadyStatus, nil
}
```
waitBackupReadyWithContext returns that same error:
```
	for {
		select {
		case <-ticker.C:
			done, err = os.backupIsReady(backupID)
			if err != nil {
				return err
			}

			if done {
				return nil
			}
		case <-ctx.Done():
			return ctx.Err()
		}
	}
```
Then WaitBackupReady receives that error and returns it:
```
	err = os.waitBackupReadyWithContext(backupID, duration)
	if err == context.DeadlineExceeded {
		err = fmt.Errorf("timeout, Backup %s is still not Ready: %v", backupID, err)
	}

	back, _ := os.GetBackupByID(backupID)

	if back != nil {
		return back.Status, err
	} else {
		return "Failed to get backup status", err
	}
```
This goes up CreateSnapshot()
```
		backup.Status, err = cs.Cloud.WaitBackupReady(backup.ID, snapSize, backupMaxDurationSecondsPerGB)
		if err != nil {
			klog.Errorf("Failed to WaitBackupReady: %v", err)
			return nil, status.Error(codes.Internal, fmt.Sprintf("CreateBackup failed with error %v", err))
		}
```

pkg/csi/cinder/controllerserver.go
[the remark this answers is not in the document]
dulek: I don't think API will behave differently if `cinder-backup` is not deployed. In such case API will still get us a 404, as `GET` call doesn't RPC down to `cinder-backup` and just looks the items up in the DB: https://opendev.org/openstack/cinder/src/branch/master/cinder/backup/api.py#L71-L76.

If backups are optional, wouldn't it be better to put all the backup operations behind an `if` based on existence of the backup API extension: https://docs.openstack.org/api-ref/block-storage/v3/#list-known-api-extensions ?

pkg/csi/cinder/controllerserver.go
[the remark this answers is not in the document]
dulek: As I explained above, GET call on backups will not tell us anything about `cinder-backup` being present in the deployment, but now I think it does make some sense to probe for backup existence, because on DELETE call we'll crash here: https://opendev.org/openstack/cinder/src/branch/master/cinder/backup/api.py#L120. Thanks, please mark this one as resolved.

pkg/csi/cinder/openstack/openstack_backups.go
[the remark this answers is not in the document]
dulek: I don't mean API returning an error. I mean `backup.Status` becoming `error`. This can happen and such state is irrecoverable. Here I caused it by specifying a non-existing AZ, but it can happen for various reasons, e.g. storage being full.

```
mdulko:~/ $ openstack volume backup create --availability-zone non-existing --snapshot test test 
+-----------+--------------------------------------+
| Field     | Value                                |
+-----------+--------------------------------------+
| id        | eb01d44f-5b1a-4e9e-b23e-7f979bf38dc1 |
| name      | None                                 |
| volume_id | d35ddf06-be8c-4205-861a-fae31c8d3600 |
+-----------+--------------------------------------+
mdulko:~/ $ openstack volume backup list
+--------------------------------------+------+-------------+--------+------+-------------+
| ID                                   | Name | Description | Status | Size | Incremental |
+--------------------------------------+------+-------------+--------+------+-------------+
| eb01d44f-5b1a-4e9e-b23e-7f979bf38dc1 | None | None        | error  |   10 | False       |
+--------------------------------------+------+-------------+--------+------+-------------+
```

dulek (changes requested):

pkg/csi/cinder/controllerserver.go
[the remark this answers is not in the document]
dulek: Oh, interesting, so it's 404 from delete call as well:

```
REQ: curl -g -i --cacert "/home/mdulko/.config/openstack/standalone-ca.crt" -X DELETE https://10.8.1.135:13776/v3/9e1f1eda803b4dd18ec77b3ed4d572a0/backups/eb01d44f-5b1a-4e9e-b23e-7f979bf38dc1 -H "OpenStack-API-Version: volume 3.64" -H "User-Agent: openstacksdk/2.0.0 keystoneauth1/4.5.0 python-requests/2.28.1 CPython/3.10.13" -H "X-Auth-Token: {SHA256}885c7009b71be65f06ddf13031866d95a73b040d50670d7cf7e928ca6eed4914"
https://10.8.1.135:13776 "DELETE /v3/9e1f1eda803b4dd18ec77b3ed4d572a0/backups/eb01d44f-5b1a-4e9e-b23e-7f979bf38dc1 HTTP/1.1" 404 87
RESP: [404] content-length: 87 content-type: application/json date: Thu, 11 Jan 2024 16:42:08 GMT openstack-api-version: volume 3.64 server: Apache vary: OpenStack-API-Version x-compute-request-id: req-f451ec29-8125-4864-9cbb-3476dda8ee3d x-openstack-request-id: req-f451ec29-8125-4864-9cbb-3476dda8ee3d
RESP BODY: {"itemNotFound": {"code": 404, "message": "Service cinder-backup could not be found."}}
DELETE call to block-storage for https://10.8.1.135:13776/v3/9e1f1eda803b4dd18ec77b3ed4d572a0/backups/eb01d44f-5b1a-4e9e-b23e-7f979bf38dc1 used request id req-f451ec29-8125-4864-9cbb-3476dda8ee3d
Failed to delete backup with name or ID 'eb01d44f-5b1a-4e9e-b23e-7f979bf38dc1': ResourceNotFound: 404: Client Error for url: https://10.8.1.135:13776/v3/9e1f1eda803b4dd18ec77b3ed4d572a0/backups/eb01d44f-5b1a-4e9e-b23e-7f979bf38dc1, Service cinder-backup could not be found.
```

Sebastian-RG:
@dulek @kayrus  Added a check `BackupsAreEnabled()` to see if a cluster supports backups. Please tell me if any issues are pending.

pkg/csi/cinder/openstack/openstack_backups.go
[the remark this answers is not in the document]
dulek: Nice, thanks!

dulek (changes requested):

pkg/csi/cinder/openstack/openstack_backups.go
```diff
@@ -0,0 +1,235 @@
...
+	if mc.ObserveRequest(err) != nil {
+		klog.Errorf("Failed to get backup: %v", err)
+		return nil, err
+	}
+	return backup, nil
+}
+
+func (os *OpenStack) BackupsAreEnabled() (bool, error) {
+	// Check if the backup service is enabled
+	allPages, err := services.List(os.blockstorage, services.ListOpts{Binary: backupBinary}).AllPages()
```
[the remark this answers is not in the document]
Sebastian-RG: There is no way of checking if it's enabled or not by getting the extensions. I've tested cases environments where backups are not enabled and openstack still lists the `Backups` and `CreateBackup` extensions :(

Listing services was the only reliable way i found to tell if backups are enabled.

I also thought about just listing all backups and seeing if it returns an error but it just returns an empty list. I'll set it to `return true, nil` as you say

pkg/csi/cinder/controllerserver.go
```diff
@@ -110,36 +112,66 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 	content := req.GetVolumeContentSource()
 	var snapshotID string
-	var sourcevolID string
+	var sourceVolID string
+	var sourceBackupID string
+	var backupsAreEnabled bool
+	backupsAreEnabled, err = cloud.BackupsAreEnabled()
```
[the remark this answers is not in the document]
Sebastian-RG: I think the added delay for a createVolume or createSnapshot action should not be noticeable. Also if we do this at controller startup it might be hard to debug when adding backups capability to an existing cluster.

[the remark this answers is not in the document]
dulek: This can be done as a second step, I think.

pkg/csi/cinder/openstack/openstack_backups.go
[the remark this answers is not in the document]
dulek: I know, admins tend to forget about configuring the extensions correctly. Well, we can't do much about environments that return incorrect data, but it's probably the thing to look at. Anyway it's not a big deal at this state.

dulek:
I'll let @kayrus and @zetaab have another look before lifting the hold.

Sebastian-RG:
Hi @kayrus, @zetaab. Do you have any additional comments?

dulek:
/retest

It should be good now.

dulek (approved):

zetaab:
@Sebastian-RG thanks for driving this through!

/hold cancel

Issue #2481, opened 2023-11-27
[cinder-csi-plugin] Missing dnsPolicy when running with hostNetwork: true


gman0:
> /kind bug

cinder-csi-plugin node plugin DaemonSet definition (the manifests and the chart) deploys with `hostNetwork: true`, but it is missing `dnsPolicy: ClusterFirstWithHostNet`, so node's DNS config is always used, bypassing cluster's DNS cache. This can negatively impact performance of the external DNS.

I'll attach a PR with optional `dnsPolicy` parameter for the Helm chart, allowing users to set the policy they prefer.

PR #2483, opened 2023-11-28
[cinder-csi-plugin] added dnsPolicy value for nodeplugin

gman0:

This PR adds `dnsPolicy` value to cinder-csi-plugin's Helm chart. The default value is set to `ClusterFirst`, which the current default.

fixes https://github.com/kubernetes/cloud-provider-openstack/issues/2481

**Release note**:

```release-note
[cinder-csi-plugin] Added .csi.plugin.nodePlugin.dnsPolicy value
```

### DIFF
--- a/charts/cinder-csi-plugin/templates/nodeplugin-daemonset.yaml
+++ b/charts/cinder-csi-plugin/templates/nodeplugin-daemonset.yaml
@@ -24,6 +24,7 @@ spec:
     spec:
       serviceAccount: csi-cinder-node-sa
       hostNetwork: true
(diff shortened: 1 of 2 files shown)


dulek:
Any reason to not just set this directly to `ClusterFirstWithHostNet` as in Manila chart?

jichenjc:
>>>Any reason to not just set this directly to ClusterFirstWithHostNet as in Manila chart?

gman0:
@dulek Sure, changed as requested. Initially I've set it to `ClusterFirst` as that's what currently used by the DaemonSet.

dulek:
I wasn't sure this is the right way, but it makes a lot of sense to be consistent here.
/approve

PR #2484, opened 2023-11-28
[occm] Improve route controller reconciling to ensure the cluster's nodes can access each other

### LINKED ISSUE #2482: [occm] The pods cannot access each other when across nodes

/kind bug

**What happened**:

I useed [CAPO](https://github.com/kubernetes-sigs/cluster-api-provider-openstack) deployed a two nodes cluser. And start OCCM with below configurations:

```
[Global]
auth-url=http://192.168.1.29:5000
username=admin
password=123456
region=RegionOne
tenant-id=666853c5323d486cbd81d372359f8942
domain-id=default

[Networking]
public-network-name=ext-net

[LoadBalancer]
enabled=true
floating-network-id=a03550ce-f3ff-4b71-bd37-00c12bcaa3f0
floating-subnet-id=f0db7750-6d43-4259-a9f5-4b171469b2c7
lb-provider=amphora
subnet-id=6cf6198d-8e89-4bf3-9c6e-5dc79509118f
network-id=c2d42f5b-1b1e-4d25-94eb-ea1a7449a8da

[Metadata]
search-order=metadataService,configDrive

[Route]
router-id=89aac4b9-9004-464d-8e01-409149144e28  # Note: router controller be enabled here. And OCCM start with --allocate-node-cidrs=true and --cluster-cidr=10.255.0.0/16
```

And the CNI configurations like below:

```
{
    "cniVersion": "0.3.1",
    "name": "mynet",
    "type": "bridge",
    "bridge": "mynet0",
    "isDefaultGateway": true,
    "forceAddress": false,
    "ipMasq": true,
    "hairpinMode": true,
    "ipam": {
        "type": "host-local",
        "subnet": "10.255.0.0/24"     # Node 1
        "subnet": "10.255.1.0/24"     # Node 2
    }
}
```

But, the pods of node1 cannot access the pods of node 2 and vice versa.

All of the pods can access each other.

**How to reproduce it**:

Deploy a multiple nodes cluser and config OCCM and CNI plugin as above.

IMO, This due to the node's security group has no ingress rule to permit the network packet of other node's pods through directly

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: xxx
- OpenStack version: xxx
- Others: xxx

jeffyjf:

In order to the nodes of one cluster can access each other, route controller need to check and set three things:

1. The openstack router's route rules, so that the packets can be forwarded to correct nodes.
2. The node port's AllowAddressPair, to ensure the node permit the packets that access the node's pods/services pass through.
3. The openstack security group's rules, so that the nodes that bind the security group permit the packets from other nodes enter into.

The current codes just check router's route rule when controller call `ListRoutes` and just set `Route` and `AllowAddressPair` when controller call `CreateRoute`. This PR complements all of the other works.

fixes #2482

This PR lack of unit tests, because of I found occm lack of a mechanism to mock openstack client. Further more the whole occm lack of lots of unit tests  due to the reason. I plan to dive deeper into gophercloud in the next several days try to study whether it is possible  to mock it. If it is possible I'll commit anohter PR to add the unit tests.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/openstack.go
+++ b/pkg/openstack/openstack.go
@@ -149,7 +149,8 @@ type NetworkingOpts struct {
 
 // RouterOpts is used for Neutron routes
 type RouterOpts struct {
-	RouterID string `gcfg:"router-id"`
+	RouterID                string `gcfg:"router-id"`
+	AutoConfigSecurityGroup bool   `gcfg:"auto-config-node-security-group"`
 }
 
 // OpenStack is an implementation of cloud provider Interface for OpenStack.
@@ -230,6 +231,7 @@ func ReadConfig(config io.Reader) (Config, error) {
 	cfg.LoadBalancer.ContainerStore = "barbican"
 	cfg.LoadBalancer.MaxSharedLB = 2
 	cfg.LoadBalancer.ProviderRequiresSerialAPICalls = false
+	cfg.Route.AutoConfigSecurityGroup = false
 
 	err := gcfg.FatalOnly(gcfg.ReadInto(&cfg, config))
 	if err != nil {
--- a/pkg/openstack/routes.go
+++ b/pkg/openstack/routes.go
@@ -18,7 +18,10 @@ package openstack
 
 import (
 	"context"
+	"fmt"
 	"net"
+	"net/netip"
+	"slices"
 	"sync"
 
 	v1 "k8s.io/api/core/v1"
@@ -27,7 +30,10 @@ import (
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/layer3/extraroutes"
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/layer3/routers"
+	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/security/groups"
+	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/security/rules"
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/ports"
+	secgroups "github.com/gophercloud/utils/v2/openstack/networking/v2/extensions/security/groups"
 
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
@@ -41,12 +47,16 @@ import (
 type Routes struct {
 	network *gophercloud.ServiceClient
 	os      *OpenStack
+	// The ID of the node security group
+	nodeSecurityGroupId string
 	// router's private network IDs
 	networkIDs []string
 	// whether Neutron supports "extraroute-atomic" extension
 	atomicRoutes bool
 	// whether Neutron supports "allowed-address-pairs" extension
 	allowedAddressPairs bool
+	// The auto config node security group feature whether enabled
+	autoConfigSecurityGroup bool
 	// Neutron with no "extraroute-atomic" extension can modify only one route at
 	// once
 	sync.Mutex
@@ -61,11 +71,82 @@ func NewRoutes(os *OpenStack, network *gophercloud.ServiceClient, atomicRoutes b
 	}
 
 	return &Routes{
-		network:             network,
-		os:                  os,
-		atomicRoutes:        atomicRoutes,
-		allowedAddressPairs: allowedAddressPairs,
+		network:                 network,
+		os:                      os,
+		atomicRoutes:            atomicRoutes,
+		allowedAddressPairs:     allowedAddressPairs,
+		autoConfigSecurityGroup: os.routeOpts.AutoConfigSecurityGroup,
 	}, nil
+}
+
+func (r *Routes) getOrCreateNodeSecurityGroup(ctx context.Context, clusterName string) error {
+	r.Lock()
+	defer r.Unlock()
+	sgName := fmt.Sprintf("k8s-node-sg-for-cluster-%s", clusterName)
+	sgId, err := secgroups.IDFromName(ctx, r.network, sgName)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
+		mc := metrics.NewMetricContext("security_group", "create")
+		klog.V(2).InfoS("Create node security group", "name", sgName)
+		group, err := groups.Create(ctx, r.network, groups.CreateOpts{Name: sgName}).Extract()
+		if mc.ObserveRequest(err) != nil {
+			return err
+		}
+		sgId = group.ID
+
+	}
+	r.nodeSecurityGroupId = sgId
+	return nil
+}
+
+// checkSecurityGroupRulesExist check whether the ingress security group rules that related to the podCidr and the node (nodeName) are existing,
+// these security group rules use to ensure other nodes permit the traffic from this node (nodeName) or this node's pods pass through.
+func (r *Routes) checkSecurityGroupRulesExist(podCidr string, nodeName types.NodeName, existingRules []rules.SecGroupRule) (bool, bool, error) {
+	nodes, err := r.os.nodeInformer.Lister().List(labels.Everything())
+	if err != nil {
+		return false, false, err
+	}
+	sgRuleForPodCidrFound := false
+	sgRuleForNodeAddressFound := false
+	podPrefix, _ := netip.ParsePrefix(podCidr)
+	nodeAddr := getAddrByNodeName(nodeName, podPrefix.Addr().Is6(), nodes)
+	nodeIp, _ := netip.ParseAddr(nodeAddr)
+
+	for _, rule := range existingRules {
+		sgPrefix, _ := netip.ParsePrefix(rule.RemoteIPPrefix)
+		if sgPrefix.Overlaps(podPrefix) && podPrefix.Bits() >= sgPrefix.Bits() {
+			sgRuleForPodCidrFound = true
+		}
+		if rule.RemoteGroupID == r.nodeSecurityGroupId || sgPrefix.Contains(nodeIp) {
+			sgRuleForNodeAddressFound = true
+		}
+		if sgRuleForPodCidrFound && sgRuleForNodeAddressFound {
+			break
+		}
+	}
+	return sgRuleForPodCidrFound, sgRuleForNodeAddressFound, nil
+}
+
+// checkPortSecurityRules check the port's security rules (SecurityGroups and AllowAddressPairs) whether are valid
+func (r *Routes) checkPortSecurityRules(port *PortWithPortSecurity, route routers.Route) bool {
+
+	// check if the node security group bind to the port
+	if r.autoConfigSecurityGroup && !slices.Contains(port.SecurityGroups, r.nodeSecurityGroupId) {
+		return false
+	}
+
+	if r.allowedAddressPairs {
+		// check whether the related AllowAddressPair is existing
+		for _, addrPair := range port.AllowedAddressPairs {
+			if addrPair.IPAddress == route.DestinationCIDR {
+				return true
+			}
+		}
+		return false
+	}
+	return true
 }
 
 // ListRoutes lists all managed routes that belong to the specified clusterName
@@ -88,8 +169,49 @@ func (r *Routes) ListRoutes(ctx context.Context, clusterName string) ([]*cloudpr
 	}
 
 	routes := make([]*cloudprovider.Route, 0, len(router.Routes))
+
+	// detect router's private network ID for further VM ports filtering
+	r.networkIDs, err = getRouterNetworkIDs(ctx, r.network, r.os.routeOpts.RouterID)
+	if err != nil {
+		return nil, err
+	}
+
+	var existingRules []rules.SecGroupRule
+	if r.autoConfigSecurityGroup {
+		if r.nodeSecurityGroupId == "" {
+			klog.V(5).Info("ListRoutes: Cannot found node security group")
+			return routes, nil
+		}
+		existingRules, err = openstackutil.GetSecurityGroupRules(r.network, rules.ListOpts{SecGroupID: r.nodeSecurityGroupId, Direction: string(rules.DirIngress)})
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	for _, item := range router.Routes {
 		nodeName, foundNode := getNodeNameByAddr(item.NextHop, nodes)
+		// If the node found, check whether the necessary rules (SecurityGroup and AllowAddressPair) was set. If
+		// not, the connectivity of different nodes can't be ensure. CreateRoutes need be to trigger to set these
+		// rules.
+		if foundNode {
+			if r.autoConfigSecurityGroup {
+				podCidrRuleExist, nodeAddrRuleExit, err := r.checkSecurityGroupRulesExist(item.DestinationCIDR, nodeName, existingRules)
+				if err != nil {
+					return nil, err
+				}
+				if !(podCidrRuleExist && nodeAddrRuleExit) {
+					continue
+				}
+			}
+			// get the node port that the route next hop addr belong to
+			port, err := r.getPortByIP(ctx, item.NextHop)
+			if err != nil {
+				return nil, err
+			}
+			if port.PortSecurityEnabled && !r.checkPortSecurityRules(port, item) {
+				continue
+			}
+		}
 		route := cloudprovider.Route{
 			Name:            item.DestinationCIDR,
 			TargetNode:      nodeName, //contains the nexthop address if node name was not found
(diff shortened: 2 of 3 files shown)


dulek:
Why did this ended up in Routes controller? The issue description feels like this is a general issue in cases CAPO is missing SG rules to allow interconnectivity of Node.

pkg/openstack/routes.go
```diff
@@ -264,6 +268,70 @@ func updateAllowedAddressPairs(network *gophercloud.ServiceClient, port *ports.P
...
+			res := groups.Create(network, groups.CreateOpts{Name: segName})
+			if res.Err != nil {
+				return nil, res.Err
+			}
+			sgGroup, err := res.Extract()
+			if err != nil {
+				return nil, err
+			}
+			sgId = sgGroup.ID
+			ruleRes := rules.Create(network, rules.CreateOpts{Direction: rules.DirIngress, SecGroupID: sgId, RemoteGroupID: sgId})
```
dulek, line 288: We've learned that using `RemoteGroupID` causes performance issues in Neutron. I'd rather opt to use CIDRs instead.

jeffyjf:
> Why did this ended up in Routes controller? The issue description feels like this is a general issue in cases CAPO is missing SG rules to allow interconnectivity of Node.

According to the [offical document](https://kubernetes.io/docs/concepts/architecture/cloud-controller/#route-controller):
```
The route controller is responsible for configuring routes in the cloud appropriately so that containers on different nodes in your Kubernetes cluster can communicate with each other.
```
I think that this is route controller's duty. I user don't activate route controller, the interconnectivity of pods should be ensured by other mechanisms.

jeffyjf:
@dulek @jichenjc @kayrus@zetaab asking for review

kayrus:
The PR needs a rebase. However the #2499 is on the way, and I adopted your `getSecurityGroupRules` change there as well.

pkg/openstack/routes.go
```diff
@@ -87,8 +124,77 @@ func (r *Routes) ListRoutes(ctx context.Context, clusterName string) ([]*cloudpr
...
+	r.networkIDs, err = getRouterNetworkIDs(r.network, r.os.routeOpts.RouterID)
+	if err != nil {
+		return nil, err
+	}
+
+	var sgRuleForPodCidrFound, sgRuleForNodeAddressFound, nodePortBindSecurityGroup, allowAddressPairFound bool
+out:
 	for _, item := range router.Routes {
 		nodeName, foundNode := getNodeNameByAddr(item.NextHop, nodes)
+		if foundNode {
```
kayrus, line 138: this looks like a complex code with lots of nesting. can you split it up to several functions?

kayrus (changes requested):

pkg/openstack/routes.go
jeffyjf: Of course

pkg/openstack/routes.go
```diff
@@ -59,12 +67,93 @@ func NewRoutes(os *OpenStack, network *gophercloud.ServiceClient, atomicRoutes b
...
 		allowedAddressPairs: allowedAddressPairs,
-	}, nil
+	}
+
+	err := routes.getOrCreateNodeSecurityGroup(os.routeOpts.RouterID)
+	if err != nil {
+		return nil, err
+	}
+
+	return &routes, nil
```
dulek, line 82: I strongly think this shouldn't be done in the initialization of the controller, but rather on `CreateRoute()`, each time checking if SG exists so that it's existence can be reconciled if deleted. We can cache it's ID if needed.

pkg/openstack/routes.go
```diff
@@ -59,12 +67,93 @@ func NewRoutes(os *OpenStack, network *gophercloud.ServiceClient, atomicRoutes b
...
+	ip, _, _ := net.ParseCIDR(route.DestinationCIDR)
+	isCIDRv6 := ip.To4() == nil
+	nodeAddr := getAddrByNodeName(nodeName, isCIDRv6, nodes)
+	nodeAddrCidr := fmt.Sprintf("%s/32", nodeAddr)
+
+	for _, rule := range existingRules {
+		// route.DestinationCIDR is equivalent with the node (nodeName)'s PodCidr
+		if route.DestinationCIDR == rule.RemoteIPPrefix {
+			sgRuleForPodCidrFound = true
+		}
```
dulek, line 122: What if a rule opening broader traffic is there already? Should we just check if `DestinationCIDR` is in `RemoteIPPrefix`, not is equal?

pkg/openstack/routes.go
```diff
@@ -59,12 +67,93 @@ func NewRoutes(os *OpenStack, network *gophercloud.ServiceClient, atomicRoutes b
...
+			return err
+		}
+	}
+	r.nodeSecurityGroupId = sgId
+	return nil
+}
+
+// checkSecurityGroupRules check whether the ingress security group rules that related to the route is existing.
+// these security group rules use to ensure other nodes permit the traffic from this node (nodeName) pass through.
+func (r *Routes) checkSecurityGroupRules(route routers.Route, existingRules []rules.SecGroupRule, nodeName types.NodeName, nodes []*v1.Node) bool {
```
dulek, line 110: IMO this should be named `checkSecurityGroupRulesExist()`.

pkg/openstack/routes.go
```diff
@@ -87,8 +176,35 @@ func (r *Routes) ListRoutes(ctx context.Context, clusterName string) ([]*cloudpr
...
+
+			// get the node port that the route next hop addr belong to
+			port, err := getPortByIP(r.network, item.NextHop, r.networkIDs)
+			if err != nil {
+				return nil, err
+			}
+
+			if !r.checkPortSecurityRules(port, item) {
+				continue
+			}
```
dulek, line 206: Why do we do that on a `ListRoutes()` call?

pkg/openstack/routes.go
```diff
@@ -268,6 +378,65 @@ func updateAllowedAddressPairs(network *gophercloud.ServiceClient, port *PortWit
...
 }
 
+// updateSecurityGroup update the port's security groups
+func updateSecurityGroup(network *gophercloud.ServiceClient, port *PortWithPortSecurity, sgs []string) (func(), error) {
+	origSgs := port.SecurityGroups
+	mc := metrics.NewMetricContext("port", "update")
+	_, err := ports.Update(network, port.ID, ports.UpdateOpts{
+		SecurityGroups: &sgs,
+	}).Extract()
+	if mc.ObserveRequest(err) != nil {
```
dulek, line 388: This is prone to a lost update race condition, but it's not a big deal IMO.

pkg/openstack/routes.go
```diff
@@ -268,6 +378,65 @@ func updateAllowedAddressPairs(network *gophercloud.ServiceClient, port *PortWit
...
+		return nil, err
+	}
+	unwinder := func() {
+		klog.V(4).Infof("Reverting security-group-rule creation %v", newRule.ID)
+		mc := metrics.NewMetricContext("security_group_rule", "delete")
+		err := rules.Delete(network, newRule.ID).ExtractErr()
+		if mc.ObserveRequest(err) != nil {
+			klog.Warningf("Unable to revert security-group-rule creation during error unwind: %v", err)
+		}
+	}
```
dulek, line 419: Okay, this is actually pretty cool.

pkg/openstack/routes.go
```diff
@@ -288,82 +457,161 @@ func (r *Routes) CreateRoute(ctx context.Context, clusterName string, nameHint s
...
+		if rule.RemoteIPPrefix == route.DestinationCIDR {
+			sgRuleForPodCidrFound = true
+		}
+		if rule.RemoteGroupID == r.nodeSecurityGroupId || rule.RemoteIPPrefix == nodeAddrCidr {
+			sgRuleForNodeAddressFound = true
+		}
+		if sgRuleForPodCidrFound && sgRuleForNodeAddressFound {
+			break
+		}
+	}
```
dulek, line 525: This looks like a duplicated logic of `checkSecurityGroupRules()`?

pkg/openstack/routes.go
```diff
@@ -288,82 +457,161 @@ func (r *Routes) CreateRoute(ctx context.Context, clusterName string, nameHint s
...
-
 	// get the port of addr on target node.
 	port, err := getPortByIP(r.network, addr, r.networkIDs)
 	if err != nil {
 		return err
 	}
+
 	if !port.PortSecurityEnabled {
-		klog.Warningf("Skipping allowed_address_pair for port: %s", port.ID)
+		klog.Warningf("Skipping updation of the port : %s (allowed_address_pairs and node security_group)", port.ID)
```
dulek, line 570: "Skipping update" is the correct form.

pkg/openstack/routes.go
```diff
@@ -288,82 +457,161 @@ func (r *Routes) CreateRoute(ctx context.Context, clusterName string, nameHint s
...
-	for _, item := range port.AllowedAddressPairs {
-		if item.IPAddress == route.DestinationCIDR {
-			klog.V(4).Infof("Found existing allowed-address-pair: %v", item)
-			found = true
-			break
+	if !r.allowedAddressPairs {
+		klog.V(4).Infof("Route created (skipping the allowed_address_pairs update): %v", route)
+	} else {
+		// update node port's AllowedAddressPairs, so that the traffic access the pods/services of this
+		// node (route.TargetNode) can enter into this node. in other words, permitting the packets the
```
dulek, line 579: `AllowedAddressPairs` is about traffic leaving the node, not entering, I think.

dulek (changes requested):
Some remarks inline. In general I'm not totally against it, but I think the code should introduce a new option that will control the new behavior of creating and updating the SG. I don't think this works for every user here.

pkg/openstack/routes.go
```diff
@@ -59,12 +67,93 @@ func NewRoutes(os *OpenStack, network *gophercloud.ServiceClient, atomicRoutes b
...
+
+	return &routes, nil
+}
+
+func (r *Routes) getOrCreateNodeSecurityGroup(router_id string) error {
+	// TODO(jeffyjf) Maybe one router can be used to support multiple
+	// clusters. Whether it is appropriate that use router id as the security
+	// group name's identity. Maybe we should introduce other mechanism to identify
+	// the security group.
+	sgName := fmt.Sprintf("k8s-node-sg-%s", router_id)
```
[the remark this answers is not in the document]
jeffyjf: I'm afraid it can't, Because the clusterName isn't unique.

pkg/openstack/routes.go
jeffyjf: Yep, you are right.  I'll update it.

pkg/openstack/routes.go
jeffyjf: For reconciling. If the security group and related rules aren't set, the connectivity between different node can't be ensure, in this situation, CreateRoutes() should be triggered to set security group and related rules. In my opinion, it is useful when users need the feature and upgrade their cluster from previous edition to the lastest(Include this PR).

dulek: Yeah, makes sense now, thanks!

pkg/openstack/routes.go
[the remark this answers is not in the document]
jeffyjf: Got it. I will submit a new patch to update it in next several days.

pkg/openstack/openstack.go
```diff
@@ -483,7 +485,7 @@ func (os *OpenStack) Routes() (cloudprovider.Routes, bool) {
 		return nil, false
 	}
 
-	r, err := NewRoutes(os, network, netExts["extraroute-atomic"], netExts["allowed-address-pairs"])
+	r, err := NewRoutes(os, network, netExts["extraroute-atomic"], netExts["allowed-address-pairs"], os.routeOpts.AutoConfigSecurityGroup)
```
kayrus, line 488: the `os` is already passed to the `NewRoutes` func. There is no need to pass the `os.routeOpts.AutoConfigSecurityGroup`.

pkg/openstack/routes.go
```diff
@@ -54,19 +64,98 @@ type Routes struct {
...
+			klog.V(2).InfoS("Create node security group", "name", sgName)
+			group, err := groups.Create(r.network, groups.CreateOpts{Name: sgName}).Extract()
+			if mc.ObserveRequest(err) != nil {
+				return err
+			}
+			sgId = group.ID
+		} else {
+			return err
+		}
+	}
```
kayrus, line 98:
```suggestion
	if err != nil {
		if !errors.IsNotFound(err) {
			return err
		}
		mc := metrics.NewMetricContext("security_group", "create")
		klog.V(2).InfoS("Create node security group", "name", sgName)
		group, err := groups.Create(r.network, groups.CreateOpts{Name: sgName}).Extract()
		if mc.ObserveRequest(err) != nil {
			return err
		}
		r.nodeSecurityGroupId = group.ID
		return nil
	}
```

pkg/openstack/routes.go
```diff
@@ -54,19 +64,98 @@ type Routes struct {
...
+	if r.allowedAddressPairs {
+		// check whether the related AllowAddressPair is existing
+		for _, addrPair := range port.AllowedAddressPairs {
+			if addrPair.IPAddress == route.DestinationCIDR {
+				return true
+			}
+		}
+		return false
+	}
+	return true
```
kayrus, line 156:
```suggestion
	if !r.allowedAddressPairs {
		return true
	}

	// check whether the related AllowAddressPair is existing
	for _, addrPair := range port.AllowedAddressPairs {
		if addrPair.IPAddress == route.DestinationCIDR {
			return true
		}
	}
	return false
```

kayrus (changes requested):
thanks for the PR. Could you take a look at comments I just left?

pkg/openstack/routes.go
```diff
@@ -54,19 +64,98 @@ type Routes struct {
...
+		for _, sg := range port.SecurityGroups {
+			if sg == r.nodeSecurityGroupId {
+				nodePortBindSecurityGroup = true
+				break
+			}
+		}
+		if !nodePortBindSecurityGroup {
+			return false
+		}
+	}
```
kayrus, line 145:
```suggestion
	// check if the node security group bind to the port
	if if r.autoConfigSecurityGroup && !slices.Contains(port.SecurityGroups, r.nodeSecurityGroupId)
		return false
	}
```

pkg/openstack/routes.go
kayrus, line 156:
```suggestion
	// check whether the related AllowAddressPair is existing
	if r.allowedAddressPairs && !slices.Contains(port.AllowedAddressPairs, route.DestinationCIDR) {
		return false
	}

	return true
```

kayrus (commented):
@jeffyjf could you please rebase + see code review comments?

pkg/openstack/routes.go
```diff
@@ -87,8 +176,51 @@ func (r *Routes) ListRoutes(ctx context.Context, clusterName string) ([]*cloudpr
...
+			// get the node port that the route next hop addr belong to
+			port, err := getPortByIP(r.network, item.NextHop, r.networkIDs)
+			if err != nil {
+				return nil, err
+			}
+			if port.PortSecurityEnabled {
+				if !r.checkPortSecurityRules(port, item) {
+					continue
+				}
+			}
```
kayrus, line 222:
```suggestion
			if port.PortSecurityEnabled && !r.checkPortSecurityRules(port, item) {
				continue
			}
```

jeffyjf:
> @jeffyjf could you please rebase + see code review comments?

@kayrus Thanks for your review. I will do these as soon as possible.

pkg/openstack/routes.go
jeffyjf: `port.AllowedAddressPairs` is a struct object while `route.DestinationCIDR` is a string object. So, `slices.Contains` cannot directly be applied.

pkg/openstack/routes.go
```diff
@@ -288,55 +473,107 @@ func (r *Routes) CreateRoute(ctx context.Context, clusterName string, nameHint s
 
 	klog.V(4).Infof("Using nexthop %v for node %v", addr, route.TargetNode)
 
-	if !r.atomicRoutes {
-		// classical logic
-		r.Lock()
-		defer r.Unlock()
+	mc := metrics.NewMetricContext("router", "get")
+	router, err := routers.Get(r.network, r.os.routeOpts.RouterID).Extract()
```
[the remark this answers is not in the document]
jeffyjf: However, I think this is necessary. The `check` step is essential for reconciling. This can ensure CPO can repair a wider range of `route` issues. For example, if a user accidentally and manually deletes a route entity, it can be repaired simply by restarting CPO.

PR #2489, opened 2023-12-04
[occm] update docs in order to reflect the current state of the ovn lb-provider support

### LINKED ISSUE #2225: [occm] octavia: support fallback to TCP health monitor

ProbstDJakob:
This PR updates the occm docs regarding the current state of the ovn load balancer provider support.

fixes #2225

**Special notes for reviewers:**

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
+++ b/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
@@ -180,6 +180,8 @@ Request Body:
 
   Defines whether to create health monitor for the load balancer pool, if not specified, use `create-monitor` config. The health monitor can be created or deleted dynamically. A health monitor is required for services with `externalTrafficPolicy: Local`.
 
+  NOTE: Health monitors for the `ovn` provider are only supported on OpenStack Wallaby and later.
+
 - `loadbalancer.openstack.org/health-monitor-delay`
 
   Defines the health monitor delay in seconds for the loadbalancer pools.
--- a/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
+++ b/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
@@ -180,8 +180,6 @@ Request Body:
 
   Defines whether to create health monitor for the load balancer pool, if not specified, use `create-monitor` config. The health monitor can be created or deleted dynamically. A health monitor is required for services with `externalTrafficPolicy: Local`.
 
-  Not supported when `lb-provider=ovn` is configured in openstack-cloud-controller-manager.
-
 - `loadbalancer.openstack.org/health-monitor-delay`
 
   Defines the health monitor delay in seconds for the loadbalancer pools.
--- a/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
+++ b/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
@@ -211,7 +211,7 @@ Although the openstack-cloud-controller-manager was initially implemented with N
   If `lb-provider` is set to "ovn" the value must be set to `SOURCE_IP_PORT`.
 
 * `lb-provider`
-  Optional. Used to specify the provider of the load balancer, e.g. "amphora" (default), "octavia", or "ovn". Only the "amphora", "octavia", and "ovn" providers are officially tested, other providers will cause a warning log.
+  Optional. Used to specify the provider of the load balancer, e.g. "amphora" (default), "octavia" (deprecated alias for "amphora"), or "ovn". Only the "amphora", "octavia", and "ovn" providers are officially tested, other providers will cause a warning log.
 
 * `lb-version`
   Optional. If specified, only "v2" is supported.
--- a/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
+++ b/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
@@ -231,6 +231,8 @@ Although the openstack-cloud-controller-manager was initially implemented with N
 * `create-monitor`
   Indicates whether or not to create a health monitor for the service load balancer. A health monitor required for services that declare `externalTrafficPolicy: Local`. Default: false
 
+  NOTE: Health monitors for the `ovn` provider are only supported on OpenStack Wallaby and later.
+
 * `monitor-delay`
   The time, in seconds, between sending probes to members of the load balancer. Default: 5
 
--- a/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
+++ b/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
@@ -201,10 +201,17 @@ Although the openstack-cloud-controller-manager was initially implemented with N
   Optional. Tags for the external network subnet used to create floating IP for the load balancer VIP. Can be overridden by the Service annotation `loadbalancer.openstack.org/floating-subnet-tags`. If multiple subnets match the first one with still available IPs is used.
 
 * `lb-method`
-  The load balancing algorithm used to create the load balancer pool. The value can be `ROUND_ROBIN`, `LEAST_CONNECTIONS`, or `SOURCE_IP`. Default: `ROUND_ROBIN`
+  The load balancing algorithm used to create the load balancer pool.
+
+  If `lb-provider` is set to "amphora" or "octavia" the value can be one of:
+    * `ROUND_ROBIN` (default)
+    * `LEAST_CONNECTIONS`
+    * `SOURCE_IP`
+    
+  If `lb-provider` is set to "ovn" the value must be set to `SOURCE_IP_PORT`.
 
 * `lb-provider`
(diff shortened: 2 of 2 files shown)


ProbstDJakob:
As discussed within the thread https://github.com/kubernetes/cloud-provider-openstack/pull/2234#discussion_r1408124865 (last sentence), this PR is a replacement for the PR https://github.com/kubernetes/cloud-provider-openstack/pull/2234. None the less until the PR https://github.com/kubernetes/cloud-provider-openstack/pull/2234 has been closed, this should probably not be merged to prevent misunderstandings/open questions.

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
```diff
@@ -180,8 +180,6 @@ Request Body:
 
   Defines whether to create health monitor for the load balancer pool, if not specified, use `create-monitor` config. The health monitor can be created or deleted dynamically. A health monitor is required for services with `externalTrafficPolicy: Local`.
 
-  Not supported when `lb-provider=ovn` is configured in openstack-cloud-controller-manager.
```
kayrus, line 183: if the `externalTrafficPolicy: Local` is set, HTTP monitors are being used. If OVN doesn't support HTTP monitors, why do you remove this string?

ProbstDJakob: Because the current behaviour is that the health monitor type will initially be set to `port.Protocol`. Later a check will be made if `HTTP` health monitors are supported and if that is the case the type will be changed to `HTTP`, but since OVN does not support it the type will stay at `TCP`.
https://github.com/kubernetes/cloud-provider-openstack/blob/2a74c6c55efba22eec2ad4657025476e5a731347/pkg/openstack/loadbalancer.go#L1153
https://github.com/kubernetes/cloud-provider-openstack/blob/2a74c6c55efba22eec2ad4657025476e5a731347/pkg/openstack/loadbalancer.go#L1162-L1167
Thus the health monitors will be created with TCP which OVN does support (as described [here](https://github.com/kubernetes/cloud-provider-openstack/pull/2234#discussion_r1406114979)). Though there has been an issue with floating IPs but that [issue](https://bugs.launchpad.net/neutron/+bug/1997418) has been fixed earlier this year.

kayrus: Right, so the monitor type will be TCP, which means that `externalTrafficPolicy: Local` is not supported in OVN. The idea of the `externalTrafficPolicy: Local` is to monitor each node that contains a pod's endpoint. Simple TCP check will mark all endpoints as active even though the node doesn't contain a pod.

kayrus: Were you able to capture a healthmonitor check tcpdump on the node host net level?
This is how curl check looks like:

```
curl -s 10.0.13.67:31823 | jq -r
{
  "service": {
    "namespace": "default",
    "name": "nginx"
  },
  "localEndpoints": 1,
  "serviceProxyHealthy": true
}
```

with a 200 response code.

this is how failed check looks like:

```
curl 10.0.13.69:31823
{
        "service": {
                "namespace": "default",
                "name": "nginx"
        },
        "localEndpoints": 0,
        "serviceProxyHealthy": true
}
```

with a 503 response code.

Regular netcat behaves identically for both failed or not failed node:

```
root@busybox:/# echo "" | nc 10.0.13.69 31823
HTTP/1.1 400 Bad Request
Content-Type: text/plain; charset=utf-8
Connection: close

root@busybox:/# echo $?
0
```

ProbstDJakob: But what you are connecting to is not the node port but the healthcheck port. This port should always be open. The node port on the other side does behave different. In our case I tested it with an OKD 4.14 cluster with ovn-kubernetes. With `netstat` you can see that the node port is open on every node, but it will reject traffic if no pod is scheduled on that node. I verified that with `nc` days ago and that is also what you can see in the pictures. Only the three nodes running a Nginx ingress controller pod are online. All other nodes are in error state, thus the healthcheck works. And yes `master-2` is in state online for port 80 (node port 30507), but I used `nc` in a loop to connect to the floating ip of the load balancer on port 80 multiple hundred times and never got rejected or timed out (and yes the traffic got routed to all pods). Also the `nc` call directly to the `master-2` on port 80 does get rejected, so it is not a problem of kubernetes, openshift, or ovn-kubernetes, it is a problem of OpenStack and I do not know why or how it is in online state but also does not get selected for routing, but that is also not a problem for OCCM.

kayrus: you're right, I got your point. Basically `kube-proxy` doesn't route the traffic to another nodes, `kubelet` listens the  healtchcheck endpoint, but no healthmonitors check it.

kayrus (approved):
/lgtm Thanks!

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
ProbstDJakob: Yes, `kube-proxy` does not route the traffic since `externalTrafficPolicy: Local` has been set and the `healthCheckNodePort` is not used. The health monitors directly connect to the node port instead of the `healthCheckNodePort`.

dulek: @kayrus: Yup, that's how it was designed in https://github.com/kubernetes/cloud-provider-openstack/pull/2000 and I can assure it works as we test this in OpenShift CI.

docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
```diff
@@ -201,10 +201,17 @@ Although the openstack-cloud-controller-manager was initially implemented with N
...
+  If `lb-provider` is set to "amphora" or "octavia" the value can be one of:
+    * `ROUND_ROBIN` (default)
+    * `LEAST_CONNECTIONS`
+    * `SOURCE_IP`
+    
+  If `lb-provider` is set to "ovn" the value must be set to `SOURCE_IP_PORT`.
 
 * `lb-provider`
-  Optional. Used to specify the provider of the load balancer, e.g. "amphora" or "octavia". Only "amphora" or "octavia" provider are officially tested, other provider will cause a warning log.
+  Optional. Used to specify the provider of the load balancer, e.g. "amphora" (default), "octavia", or "ovn". Only the "amphora", "octavia", and "ovn" providers are officially tested, other providers will cause a warning log.
```
dulek, line 214: Could we add a note that "octavia" is a deprecated alias of "amphora"?

docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
```diff
@@ -295,8 +302,6 @@ Although the openstack-cloud-controller-manager was initially implemented with N
 
 NOTE:
 
-* When using `ovn` provider service has limited scope - `create_monitor` is not supported and only supported `lb-method` is `SOURCE_IP`.
```
dulek, line 298: Can you keep the note extended with the information that OVN Octavia Provider supports health monitors only starting from Wallaby [1]?

[1] https://opendev.org/openstack/ovn-octavia-provider/commit/8bbd8f1b62d7811433079bff8ce940fbdd041a0c

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
dulek, line 183: Can we change this to say that ETP=Local is supported with OVN starting from Wallaby? https://opendev.org/openstack/ovn-octavia-provider/commit/8bbd8f1b62d7811433079bff8ce940fbdd041a0c

dulek (changes requested):

docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
ProbstDJakob: Keep this node or add a note to the `loadbalancer.openstack.org/enable-health-monitor` annotation doc (or both)? I would add it (only) to the annotation to keep them close together.

dulek: Sure, we can put it at the annotation. Thanks!

dulek (approved):

PR #2492, opened 2023-12-05
[occm] Not all OpenStack providers support AllowedAddressPairs

### LINKED ISSUE #2491: [occm] Not all OpenStack providers support AllowedAddressPairs
FEATURE
/kind feature
**What happened**:
I've experienced an issue with routing controller on one of OpenStack cloud providers. This provider has disabled `allowed_address_pairs` for network ports, whereas OCCM tries to install them and reverts all changes (including created route in router).

I can implement this if maintainers agree with this feature.

Config option to disable adding `allowed_address_pairs` for network port, so that users could disable this feature as it is not necessary on VMs of such providers.

**How to reproduce it**:
Have no idea, as I'm just a customer.

Native routing works fine without altering `allowed_address_pairs` on VMs' ports.

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: 1.28.1
- OpenStack version: Neutron v2.0
- Others: Kubernetes 1.28.0

sircthulhu:
Without this PR controller will not be able to create routes and end up errors with message like
```json
{"NeutronError": {"type": "HTTPBadRequest", "message": "Unrecognized attribute(s) 'allowed_address_pairs'", "detail": ""}}
```
This PR makes route controller skip updates to `allowed_address_pairs` if such extensions is not available or `PortSecurityEnabled` is false

fixes #2491

Bootstrap k8s cluster on OpenStack with disabled `allowed_address_pairs` feature. 

**Release note**:
```release-note
[openstack-cloud-controller-manager] skip editing allowed_address_pairs on routes updates if not supported by Neutron 
```

### DIFF
--- a/pkg/openstack/openstack.go
+++ b/pkg/openstack/openstack.go
@@ -483,7 +483,7 @@ func (os *OpenStack) Routes() (cloudprovider.Routes, bool) {
 		return nil, false
 	}
 
-	r, err := NewRoutes(os, network, netExts["extraroute-atomic"])
+	r, err := NewRoutes(os, network, netExts["extraroute-atomic"], netExts["allowed-address-pairs"])
 	if err != nil {
 		klog.Warningf("Error initialising Routes support: %v", err)
 		return nil, false
--- a/pkg/openstack/routes.go
+++ b/pkg/openstack/routes.go
@@ -43,7 +43,7 @@ type Routes struct {
 	networkIDs []string
 	// whether Neutron supports "extraroute-atomic" extension
 	atomicRoutes bool
-	// whether cloud provider supports allowed_address_pairs
+	// whether Neutron supports "allowed-address-pairs" extension
 	allowedAddressPairs bool
 	// Neutron with no "extraroute-atomic" extension can modify only one route at
 	// once
--- a/pkg/openstack/routes.go
+++ b/pkg/openstack/routes.go
@@ -18,6 +18,7 @@ package openstack
 
 import (
 	"context"
+	openstackutil "k8s.io/cloud-provider-openstack/pkg/util/openstack"
 	"net"
 	"sync"
 
@@ -242,7 +243,7 @@ func removeRoute(network *gophercloud.ServiceClient, routerID string, oldRoute [
 	return unwinder, nil
 }
 
-func updateAllowedAddressPairs(network *gophercloud.ServiceClient, port *ports.Port, newPairs []ports.AddressPair) (func(), error) {
+func updateAllowedAddressPairs(network *gophercloud.ServiceClient, port *PortWithPortSecurity, newPairs []ports.AddressPair) (func(), error) {
 	origPairs := port.AllowedAddressPairs // shallow copy
 
 	mc := metrics.NewMetricContext("port", "update")
@@ -342,6 +343,11 @@ func (r *Routes) CreateRoute(ctx context.Context, clusterName string, nameHint s
 	port, err := getPortByIP(r.network, addr, r.networkIDs)
 	if err != nil {
 		return err
+	}
+	if !port.PortSecurityEnabled {
+		klog.Warningf("Skipping allowed_address_pair for port: %s", port.ID)
+		onFailure.disarm()
+		return nil
 	}
 
 	found := false
@@ -457,6 +463,11 @@ func (r *Routes) DeleteRoute(ctx context.Context, clusterName string, route *clo
 	if err != nil {
 		return err
 	}
+	if !port.PortSecurityEnabled {
+		klog.Warningf("Skipping allowed_address_pair for port: %s", port.ID)
+		onFailure.disarm()
+		return nil
+	}
 
 	addrPairs := port.AllowedAddressPairs
 	index := -1
@@ -484,7 +495,7 @@ func (r *Routes) DeleteRoute(ctx context.Context, clusterName string, route *clo
 	return nil
 }
 
-func getPortByIP(network *gophercloud.ServiceClient, addr string, networkIDs []string) (*ports.Port, error) {
+func getPortByIP(network *gophercloud.ServiceClient, addr string, networkIDs []string) (*PortWithPortSecurity, error) {
 	for _, networkID := range networkIDs {
 		opts := ports.ListOpts{
 			FixedIPs: []ports.FixedIPOpts{
@@ -494,12 +505,7 @@ func getPortByIP(network *gophercloud.ServiceClient, addr string, networkIDs []s
 			},
 			NetworkID: networkID,
 		}
-		mc := metrics.NewMetricContext("port", "list")
-		pages, err := ports.List(network, opts).AllPages()
-		if mc.ObserveRequest(err) != nil {
-			return nil, err
-		}
-		ports, err := ports.ExtractPorts(pages)
+		ports, err := openstackutil.GetPorts[PortWithPortSecurity](network, opts)
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/openstack/routes.go
+++ b/pkg/openstack/routes.go
@@ -43,6 +43,8 @@ type Routes struct {
 	networkIDs []string
 	// whether Neutron supports "extraroute-atomic" extension
 	atomicRoutes bool
+	// whether cloud provider supports allowed_address_pairs
+	allowedAddressPairs bool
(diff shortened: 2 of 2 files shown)


pkg/openstack/openstack.go
```diff
@@ -483,7 +484,7 @@ func (os *OpenStack) Routes() (cloudprovider.Routes, bool) {
 		return nil, false
 	}
 
-	r, err := NewRoutes(os, network, netExts["extraroute-atomic"])
+	r, err := NewRoutes(os, network, netExts["extraroute-atomic"], !os.routeOpts.SkipAddressAllowedPairs)
```
kayrus, line 487: you can check for `netExts["allowed-address-pairs"]`

kayrus:
@sircthulhu basically you'd like to disable routing support for your k8s installation? Is it an option for you just to skip the `[Routes]` config section?

sircthulhu:
@kayrus 
I'd like to use networking but with `allowed_address_pairs` I cannot create via controller. If I create routes manually the same way as controller does, networking works fine

kayrus:
@sircthulhu networking will work, but with the overlay network. Adding routes allowes to avoid overlay network and set custom routes per each node's CIDR using a router. If you skip the `[Routes]` config section, k8s will use overlay network with kube-proxy based routing. Can you give it a try?

pkg/openstack/routes.go
```diff
@@ -43,6 +43,8 @@ type Routes struct {
 	networkIDs []string
 	// whether Neutron supports "extraroute-atomic" extension
 	atomicRoutes bool
+	// whether cloud provider supports allowed_address_pairs
+	addressAllowedPairsSupported bool
```
kayrus, line 47: I'd rename it to
```suggestion
	allowedAddressPairs bool
```

pkg/openstack/routes.go
```diff
@@ -329,30 +332,32 @@ func (r *Routes) CreateRoute(ctx context.Context, clusterName string, nameHint s
 		defer onFailure.call(unwind)
 	}
 
-	// get the port of addr on target node.
-	port, err := getPortByIP(r.network, addr, r.networkIDs)
-	if err != nil {
-		return err
-	}
+	if r.addressAllowedPairsSupported {
```
kayrus, line 335: you can just return in case the address pairs are not supported:

```golang
if !r.allowedAddressPairs {
        klog.V(4).Infof("Route created (skipping the allowed_address_pairs update): %v", route)
        onFailure.disarm()
        return nil
}
```

and please add the same early return into the `DeleteRoute` method:

```golang
if !r.allowedAddressPairs {
        klog.V(4).Infof("Route deleted (skipping the allowed_address_pairs update): %v", route)
        onFailure.disarm()
        return nil
}
```

kayrus (changes requested):

sircthulhu:
@kayrus 
Returned with test results :)

- Bootstrapped talos linux cluster on my cloud provider with cilium in kube proxy replacement mode with native routing cidr equal to pod cidr.
- Created route manually in openstack via cli
- I can communicate from pod-to-pod and pod-to-world finely
On tcpdump I see bare pod ip addresses.

So, I can confirm, this change won't break anything but provide an opportunity to use native routing for providers with disabled support of allowed address pairs

pkg/openstack/routes.go
```diff
@@ -437,6 +446,12 @@ func (r *Routes) DeleteRoute(ctx context.Context, clusterName string, route *clo
 		defer onFailure.call(unwind)
 	}
 
+	if !r.allowedAddressPairs {
+		klog.V(4).Infof("Route created (skipping the allowed_address_pairs update): %v", route)
```
kayrus, line 450:
```suggestion
		klog.V(4).Infof("Route deleted (skipping the allowed_address_pairs update): %v", route)
```

pkg/openstack/routes.go
```diff
@@ -43,6 +43,8 @@ type Routes struct {
 	networkIDs []string
 	// whether Neutron supports "extraroute-atomic" extension
 	atomicRoutes bool
+	// whether cloud provider supports allowed_address_pairs
```
kayrus, line 46:
```suggestion
	// whether Neutron supports "allowed-address-pairs" extension
```

kayrus (approved):
LGTM now, thanks. Please fix the typos and nits, and it should be ok to test.

kayrus:
@sircthulhu can you adjust the release notes in the initial message?

sircthulhu:
@kayrus 
Sure! Is it correct now?

pkg/openstack/routes.go
```diff
@@ -329,6 +332,12 @@ func (r *Routes) CreateRoute(ctx context.Context, clusterName string, nameHint s
 		defer onFailure.call(unwind)
 	}
 
+	if !r.allowedAddressPairs {
```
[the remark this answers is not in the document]
sircthulhu: ref [OpenStack Neutron](https://github.com/openstack/neutron/blob/58dcd30dbba67464f6fd7880ce7aee543156af65/neutron/plugins/ml2/plugin.py#L1510)

I think in this case we must skip editing `address_allowed_pairs` too, as subsequent calls will return with errors

sircthulhu:
Should we log a warning message, if the port has `port_security_enabled=true` while `allowed_address_pairs` extension is disabled?
If I'm not mistaken, Neutron will block traffic from pods, as it is not aware of such ips.

pkg/openstack/routes.go
```diff
@@ -437,11 +450,20 @@ func (r *Routes) DeleteRoute(ctx context.Context, clusterName string, route *clo
...
+		return nil
+	}
+
 	// get the port of addr on target node.
 	port, err := getPortByIP(r.network, addr, r.networkIDs)
 	if err != nil {
 		return err
 	}
+	if !port.PortSecurityEnabled {
+		klog.V(4).Infof("Skipping allowed_address_pair for port: %v", port.ID)
```
jichenjc, line 465: not sure why the `port.PortSecurityEnabled ` is involved here and seems only a log for it 
nothing impact logic, is this expected code logic?

sircthulhu: As I can see from [Neutron source code](https://github.com/openstack/neutron/blob/58dcd30dbba67464f6fd7880ce7aee543156af65/neutron/plugins/ml2/plugin.py#L1510), it will reject changes to `allowed_address_pairs` if port security is disabled for port

jichenjc: ok, thanks for the link

kayrus: if the port security is disabled, why don't you just return?

sircthulhu: @kayrus 
Should we log a warning message before returning, if the port has `port_security_enabled=true` while `allowed_address_pairs` extension is disabled? If I'm not mistaken, Neutron will block traffic from pods, as it is not aware of such ips.

kayrus: I propose to change the `Infof` with a warning, then just return.

kayrus (approved):
/lgtm
@dulek  @zetaab @jichenjc is the `PortSecurityEnabled` something that OCCM should control? if not, then I'm ok with the current PR state.

jeffyjf:
> This PR introduces new config option to disable altering allowed_address_pairs for VM ports, so that routes could be created successfully.

I didn't find the new config option after I walked through this PR's code. Could you update the description?

jichenjc:
I think the config is here which is a param to be more specific
https://github.com/kubernetes/cloud-provider-openstack/pull/2492/files#diff-cff10f4fa8015498d12b1ee3f0d27d37b4c21f8f405405da80e1b4c10b5629f1R47

sircthulhu:
@jeffyjf 
Sorry, PR description was a bit outdated :)
Updated PR description to better align with current state of code.

sircthulhu:
Looks like strange connectivity issue to ubuntu mirror. Should I just rerun tests?

jichenjc:
/test openstack-cloud-controller-manager-ovn-e2e-test

um.. ovn is new CI to me :)

kayrus:
Ubuntu serves experience some turbulence this week...
```
curl -sSL https://cloud-images.ubuntu.com/releases/jammy/release/ubuntu-22.04-server-cloudimg-amd64.img -o ubuntu-jammy.img
curl: (56) OpenSSL SSL_read: Connection reset by peer, errno 104
```

/test openstack-cloud-controller-manager-e2e-test

PR #2510, opened 2023-12-14
[tests] don't use static version in sed replacement rules

kayrus:

don't use static version in sed replacement rules

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/playbooks/roles/install-cpo-occm/tasks/main.yaml
+++ b/tests/playbooks/roles/install-cpo-occm/tasks/main.yaml
@@ -82,7 +82,7 @@
       fi
       cd $GOPATH/src/k8s.io/cloud-provider-openstack
       # replace image with built image
-      sed -i "s#registry.k8s.io/provider-os/openstack-cloud-controller-manager:v1.28.0#{{ remote_registry_host }}/openstack-cloud-controller-manager:${VERSION}#" manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml
+      sed -i "s#registry.k8s.io/provider-os/openstack-cloud-controller-manager:[^'\"]\+#{{ remote_registry_host }}/openstack-cloud-controller-manager:${VERSION}#" manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml
       sed -i "s#node-role.kubernetes.io/control-plane: \"\"#node-role.kubernetes.io/control-plane: \"true\"#" manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml
       sed -i "s#--v=1#--v=5#" manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml
       cat manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml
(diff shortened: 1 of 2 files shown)


tests/playbooks/roles/install-cpo-occm/tasks/main.yaml
```diff
@@ -82,7 +82,7 @@
       fi
       cd $GOPATH/src/k8s.io/cloud-provider-openstack
       # replace image with built image
-      sed -i "s#registry.k8s.io/provider-os/openstack-cloud-controller-manager:v1.28.0#{{ remote_registry_host }}/openstack-cloud-controller-manager:${VERSION}#" manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml
+      sed -i "s#registry.k8s.io/provider-os/openstack-cloud-controller-manager:.*#{{ remote_registry_host }}/openstack-cloud-controller-manager:${VERSION}#" manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml
```
dulek, line 85: Hm, this would most likely work, but how about a proper regex instead?

`v[0-9]+\.[0-9]+\.[0-9]+` should be fine?

kayrus: I don't want to do this by purpose, since there could be potential `-alpha`, etc. The wildcard covers everything and you don't need to think about rare cases.

dulek: Hm, then let's at least use `.+` or even `.+$` here?

kayrus: I'm ok with `.+$`, however in this particular situation it doesn't make sense :see_no_evil:

dulek: Wait, but `.*`, this will stop working once someone puts image pullspec into quotation marks here: https://github.com/kubernetes/cloud-provider-openstack/blob/f806574b31aa32a269370b3c1a2ed73fc161a498/manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml#L41

I understand the intent, but this should be a bit smarter to not bite us in the future.

`[^\'\"]+`?

kayrus: good point. and nice critical thinking ;)

tests/playbooks/roles/install-cpo-occm/tasks/main.yaml
```diff
@@ -82,7 +82,7 @@
       fi
       cd $GOPATH/src/k8s.io/cloud-provider-openstack
       # replace image with built image
-      sed -i "s#registry.k8s.io/provider-os/openstack-cloud-controller-manager:v1.28.0#{{ remote_registry_host }}/openstack-cloud-controller-manager:${VERSION}#" manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml
+      sed -i "s#registry.k8s.io/provider-os/openstack-cloud-controller-manager:[^'\"]\+#{{ remote_registry_host }}/openstack-cloud-controller-manager:${VERSION}#" manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml
```
dulek, line 85: Wait, I'm confused now, you're escaping the `+` here?

dulek: According to logs it works, okay.

PR #2511, opened 2023-12-20
[fix] Added call to WaitGroup's Done method in csi package

meetmorrowsolonmars:

If someone wants to use one of the CSI packages and implement gracefull shutdown correctry, they will not succeed because Wait does not return control.

fixes #

**Release note**:

```release-note
NONE
```


meetmorrowsolonmars:
I propose using atomic to store a pointer to grpc.Server and adding a check before calling GracefulStop. What do you think about this, @zetaab?

pkg/csi/cinder/server.go
```diff
@@ -69,6 +69,7 @@ func (s *nonBlockingGRPCServer) ForceStop() {
 }
 
 func (s *nonBlockingGRPCServer) serve(endpoint string, ids csi.IdentityServer, cs csi.ControllerServer, ns csi.NodeServer) {
+	defer s.wg.Done()
```
jichenjc, line 72: just for education, can you help point the reason behind this ? 
curious about the reason of doing this and whether other CSI driver doing this as well ? Thanks for the help and PR~

dulek: I'm guessing this will make `serve()` wait for all the threads before allowing the server to shut down, so that all requests are processed?

dulek: I don't see it in AWS EBS CSI driver though.

meetmorrowsolonmars: I've fixed an issue in the code that launches drivers in the CSI package. The current code version uses sync.WaitGroup to wait for the driver's completion. It is assumed that upon receiving a signal to terminate the application, the server should stop. However, the current code fails to release sync.WaitGroup upon exiting the function, which may lead to a deadlock in the calling code, even though the server has already stopped. I believe this is an implicit behavior that should be addressed.

jichenjc: ok, thanks for the detailed info, that helps a lot
I personal don't have strong expertise in this area so it make sense to me while I hope @dulek @zetaab or anyone else can help provide more comments, thanks~

zetaab (approved):
/approve

PR #2512, opened 2023-12-21
[occm] Fix panic on failure to get loadbalancer status

pierreprinetti:
After loadbalancer creation, the controller calls a wait function that polls the Octavia API until its state turns ACTIVE, or timeout is reached.

Before this patch, a failure to GET the loadbalancer in the wait function would result in a nil loadbalancer to be returned to the caller, resulting in an immediate panic when accessing its members.

With this patch:
* GET failures are logged and don't break the polling;
* if the wait timeout is reached while GET is failing, the behaviour exactly matches what happens when reaching the timeout.

**Release note**:
```release-note
NONE
```


dulek:
/lgtm

Looks good, thank you!

zetaab (approved):
/approve

PR #2515, opened 2023-12-29
filter node for etp service

smartwang:

For services with the externalTrafficPolicy type, a better implementation of pool members would be to include only those nodes that have service Pods, ensuring that service requests are properly routed even without configured health checks.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -34,6 +34,8 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
 	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


jichenjc:
Thanks for the PR, I assume you want to exclude the nodes that doesn't have the service pod running on when ensure the LB with externalTrafficPolicy  = ServiceExternalTrafficPolicyTypeLocal ? 

could you please help create an issue and provide detailed info so we can track why we need such change and use case etc? Thanks

smartwang:
> Thanks for the PR, I assume you want to exclude the nodes that doesn't have the service pod running on when ensure the LB with externalTrafficPolicy = ServiceExternalTrafficPolicyTypeLocal ?
> 
> could you please help create an issue and provide detailed info so we can track why we need such change and use case etc? Thanks

Oh, I found a fatal bug in my PR:  the controller didn't handle  svc pod's update event。 I'll close this PR and fix it in my fork.  

The initial intent of this PR was to:

1. Ensure that the system can still function as normally as possible without a configured health-monitor.
2. Mitigate the issue of traffic loss during rolling updates (due to the intervals in health checks).

However, my PR overlooked the logic for handling post-update pods, which requires adding a servicePodInformer and related processing logic in the ingress controller. I will close this PR and attempt to resolve this issue in my fork with additional testing.

Nevertheless, I've noticed that Kubernetes seems to be addressing this issue as well:
https://kubernetes.io/blog/2022/12/30/advancements-in-kubernetes-traffic-engineering

Therefore, I am now unsure whether this should be resolved within OCCM.

dulek:
@smartwang: It won't be that simple. The Ingress controller is a separate entity that handles [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/). You'd need to create a whole new controller that will run in the openstack-cloud-controller-manager and would watch for Pods. The code you're modifying is [called from the cloud-provider and only gets informed about Service or Node changes](https://github.com/kubernetes/kubernetes/blob/80f2101da12e8ba27a81d15955d84c78534147cf/staging/src/k8s.io/cloud-provider/cloud.go#L117-L162).

smartwang:
> @smartwang: It won't be that simple. The Ingress controller is a separate entity that handles [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/). You'd need to create a whole new controller that will run in the openstack-cloud-controller-manager and would watch for Pods. The code you're modifying is [called from the cloud-provider and only gets informed about Service or Node changes](https://github.com/kubernetes/kubernetes/blob/80f2101da12e8ba27a81d15955d84c78534147cf/staging/src/k8s.io/cloud-provider/cloud.go#L117-L162).

Indeed, "ingress controller" was a typographical error. What I meant to convey is the same as you, that it is somewhat troublesome. I can roughly understand why the health check implementation was continuously used before; it's sufficiently simple and can handle most scenarios. Striving for perfection comes at a cost; it's a matter of whether it's worth it or not.

smartwang:
> @smartwang: It won't be that simple. The Ingress controller is a separate entity that handles [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/). You'd need to create a whole new controller that will run in the openstack-cloud-controller-manager and would watch for Pods. The code you're modifying is [called from the cloud-provider and only gets informed about Service or Node changes](https://github.com/kubernetes/kubernetes/blob/80f2101da12e8ba27a81d15955d84c78534147cf/staging/src/k8s.io/cloud-provider/cloud.go#L117-L162).

Regardless, thanks for your response and guidance.

dulek:
> > @smartwang: It won't be that simple. The Ingress controller is a separate entity that handles [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/). You'd need to create a whole new controller that will run in the openstack-cloud-controller-manager and would watch for Pods. The code you're modifying is [called from the cloud-provider and only gets informed about Service or Node changes](https://github.com/kubernetes/kubernetes/blob/80f2101da12e8ba27a81d15955d84c78534147cf/staging/src/k8s.io/cloud-provider/cloud.go#L117-L162).
> 
> Indeed, "ingress controller" was a typographical error. What I meant to convey is the same as you, that it is somewhat troublesome. I can roughly understand why the health check implementation was continuously used before; it's sufficiently simple and can handle most scenarios. Striving for perfection comes at a cost; it's a matter of whether it's worth it or not.

The real reason we use healthchecks is pretty simple - other clouds do it the same way. It's far from ideal, but it's an easy way to do it regardless of the cloud K8s is running on. While I'm not opposed to try to hack our way out of this, I'd totally prefer discussing and finding a way to do it in a way that can be supported by other clouds in an unified way. This would probably involve a KEP and broader discussion with SIG Cloud Provider. I wouldn't be surprised if there were similar ideas proposed already.

All that being said, please also check if Gateway API doesn't actually solve what you're looking for.

PR #2518, opened 2024-01-09
[octavia-ingress-controller] Ingress: Fix listener timeout updates

dulek:
PR #2468 used an incorrect condition when updating the listener timeout values and only updates them when listener's allowedCIDRs are updated. This commit fixes this by making sure timeout values are checked too.

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/ingress/controller/openstack/octavia.go
+++ b/pkg/ingress/controller/openstack/octavia.go
@@ -365,19 +365,28 @@ func (os *OpenStack) EnsureListener(name string, lbID string, secretRefs []strin
 
 		log.WithFields(log.Fields{"lbID": lbID, "listenerName": name}).Info("listener created")
 	} else {
+		updateOpts := listeners.UpdateOpts{}
 		if len(listenerAllowedCIDRs) > 0 && !reflect.DeepEqual(listener.AllowedCIDRs, listenerAllowedCIDRs) {
-			_, err := listeners.Update(os.Octavia, listener.ID, listeners.UpdateOpts{
-				AllowedCIDRs:         &listenerAllowedCIDRs,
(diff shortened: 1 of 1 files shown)


pkg/ingress/controller/openstack/octavia.go
```diff
@@ -365,19 +365,28 @@ func (os *OpenStack) EnsureListener(name string, lbID string, secretRefs []strin
...
-				TimeoutMemberConnect: timeoutMemberConnect,
-				TimeoutTCPInspect:    timeoutTCPInspect,
-			}).Extract()
+			updateOpts.AllowedCIDRs = &listenerAllowedCIDRs
+		}
+
+		if timeoutClientData == nil && listener.TimeoutClientData != 0 || timeoutClientData != nil && *timeoutClientData != listener.TimeoutClientData ||
+			timeoutMemberData == nil && listener.TimeoutMemberData != 0 || timeoutMemberData != nil && *timeoutMemberData != listener.TimeoutMemberData ||
+			timeoutMemberConnect == nil && listener.TimeoutMemberConnect != 0 || timeoutMemberData != nil && *timeoutMemberConnect != listener.TimeoutMemberConnect ||
+			timeoutTCPInspect == nil && listener.TimeoutTCPInspect != 0 || timeoutTCPInspect != nil && *timeoutTCPInspect != listener.TimeoutTCPInspect {
```
dulek, line 376: If someone knows how to do this in a cleaner way, I'm happy to hear ideas. I spent quite some time on this and I still don't like it.

jichenjc:
/lgtm

I don't have better way for the proposal but I think it's clear to read the logic it should be good

PR #2519, opened 2024-01-10
fix: add missing labels to resources

cwrau:
Currently the chart doesn't contain the required labels on the applied resources, this makes it impossible to update the chart without manual intervention

I introduced the common chart from bitnami to standardly handle labels. The chart could also be used to handle resource names, secrets, matchlabels, ... . But I wanted to have this change be backwards-compatible and non-intrusive

**Release note**:
```release-note
- adds missing labels to enable helm upgrades
```

### DIFF
--- a/charts/openstack-cloud-controller-manager/Chart.yaml
+++ b/charts/openstack-cloud-controller-manager/Chart.yaml
@@ -1,11 +1,15 @@
-apiVersion: v1
+apiVersion: v2
 appVersion: v1.29.0
 description: Openstack Cloud Controller Manager Helm Chart
 icon: https://object-storage-ca-ymq-1.vexxhost.net/swift/v1/6e4619c416ff4bd19e1c087f27a43eea/www-images-prod/openstack-logo/OpenStack-Logo-Vertical.png
 home: https://github.com/kubernetes/cloud-provider-openstack
 name: openstack-cloud-controller-manager
-version: 2.29.0
+version: 2.29.0-1
 maintainers:
   - name: eumel8
     email: f.kloeker@telekom.de
     url: https://www.telekom.com
+dependencies:
+  - name: common
+    version: 2.14.1
+    repository: https://charts.bitnami.com/bitnami
--- a/charts/openstack-cloud-controller-manager/Chart.yaml
+++ b/charts/openstack-cloud-controller-manager/Chart.yaml
@@ -4,7 +4,7 @@ description: Openstack Cloud Controller Manager Helm Chart
 icon: https://object-storage-ca-ymq-1.vexxhost.net/swift/v1/6e4619c416ff4bd19e1c087f27a43eea/www-images-prod/openstack-logo/OpenStack-Logo-Vertical.png
 home: https://github.com/kubernetes/cloud-provider-openstack
 name: openstack-cloud-controller-manager
-version: 2.29.0-1
+version: 2.29.1
 maintainers:
   - name: eumel8
     email: f.kloeker@telekom.de
--- a/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
@@ -2,9 +2,8 @@ apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   name: {{ include "occm.name" . }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
   namespace: {{ .Release.Namespace }}
-  labels:
-    {{- include "occm.labels" . | nindent 4 }}
   annotations:
     {{- with .Values.commonAnnotations }}
     {{- toYaml . | nindent 4 }}
--- a/charts/openstack-cloud-controller-manager/templates/_helpers.tpl
+++ b/charts/openstack-cloud-controller-manager/templates/_helpers.tpl
@@ -12,19 +12,6 @@ Create chart name and version as used by the chart label.
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
-{{/*
-Common labels and app labels
-*/}}
-{{- define "occm.labels" -}}
-app.kubernetes.io/name: {{ include "occm.name" . }}
-helm.sh/chart: {{ include "occm.chart" . }}
-app.kubernetes.io/instance: {{ .Release.Name }}
-{{- if .Chart.AppVersion }}
-app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
-{{- end }}
-app.kubernetes.io/managed-by: {{ .Release.Service }}
-{{- end -}}
-
 {{- define "occm.common.matchLabels" -}}
 app: {{ template "occm.name" . }}
 release: {{ .Release.Name }}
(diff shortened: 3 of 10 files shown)


cwrau:
Also, I'd like to backport this to at least 1.27 and up, do I still just open a PR with a cherry pick?

charts/openstack-cloud-controller-manager/templates/daemonset.yaml
```diff
@@ -2,9 +2,8 @@ apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   name: {{ include "occm.name" . }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
   namespace: {{ .Release.Namespace }}
-  labels:
-    {{- include "occm.labels" . | nindent 4 }}
```
dulek, line 7: Will this give use the same labels here?

charts/openstack-cloud-controller-manager/Chart.yaml
```diff
@@ -9,3 +9,7 @@ maintainers:
   - name: eumel8
     email: f.kloeker@telekom.de
     url: https://www.telekom.com
+dependencies:
+  - name: common
+    version: 2.14.1
+    repository: https://charts.bitnami.com/bitnami
```
dulek, line 15: I think I'd rather prefer to just copy this single function here.

charts/openstack-cloud-controller-manager/Chart.yaml
```diff
@@ -9,3 +9,7 @@ maintainers:
   - name: eumel8
```
dulek, line 9: You need to bump version in line 7.

jichenjc:
>>>I introduced the common chart from bitnami to standardly handle labels. The chart could also be used to handle resource names, secrets, matchlabels, ... . But I wanted to have this change be backwards-compatible and non-intrusive

not sure ,is it a bug or label ? usually we only backport bugs

cwrau:
> >>>I introduced the common chart from bitnami to standardly handle labels. The chart could also be used to handle resource names, secrets, matchlabels, ... . But I wanted to have this change be backwards-compatible and non-intrusive
> 
> not sure ,is it a bug or label ? usually we only backport bugs 

It's both, the missing labels are a major bug. To be honest I'm really surprised that noone noticed this before

charts/openstack-cloud-controller-manager/templates/daemonset.yaml
cwrau: Yes, this will result in the same labels

charts/openstack-cloud-controller-manager/Chart.yaml
cwrau: Uff, I'm highly against copying code, that's the whole point of using a central, standard library for this; to remove code duplication.

Then you'd have to keep track of the upstream changes and copy them as well.

dulek: Sure, but how "standard" is this standard library? I see it's used extensively in charts around GitHub, but it still seems to be from a third party?

cwrau: Maybe it's not a "standard" by definition, but as you said it's extremely widely used across helm charts and bitnami, even though they're a third party, has made themselves a recognizable helm chart writer and supporter known for their quality, at least in my knowledge circle.

dulek:
If this is to be backported it'll need an issue reported.

charts/openstack-cloud-controller-manager/Chart.yaml
```diff
@@ -1,11 +1,15 @@
-apiVersion: v1
+apiVersion: v2
```
jichenjc, line 1: I think it's not something we want to change

charts/openstack-cloud-controller-manager/Chart.yaml
```diff
@@ -1,11 +1,15 @@
...
 appVersion: v1.28.0
 description: Openstack Cloud Controller Manager Helm Chart
 icon: https://object-storage-ca-ymq-1.vexxhost.net/swift/v1/6e4619c416ff4bd19e1c087f27a43eea/www-images-prod/openstack-logo/OpenStack-Logo-Vertical.png
 home: https://github.com/kubernetes/cloud-provider-openstack
 name: openstack-cloud-controller-manager
-version: 2.29.0-alpha.5
+version: 2.29.1-alpha.5
```
jichenjc, line 7: should this be .6 instead ?

charts/openstack-cloud-controller-manager/Chart.yaml
```diff
@@ -1,11 +1,15 @@
-apiVersion: v1
+apiVersion: v2
 appVersion: v1.29.0
 description: Openstack Cloud Controller Manager Helm Chart
 icon: https://object-storage-ca-ymq-1.vexxhost.net/swift/v1/6e4619c416ff4bd19e1c087f27a43eea/www-images-prod/openstack-logo/OpenStack-Logo-Vertical.png
 home: https://github.com/kubernetes/cloud-provider-openstack
 name: openstack-cloud-controller-manager
-version: 2.29.0
+version: 2.29.0-1
```
cwrau, line 7: Is this correct?

dulek:
These CI errors look unrelated, I see them on #2316 too.

/retest

charts/openstack-cloud-controller-manager/templates/daemonset.yaml
dulek: I'm looking at CI results and I only see this label added to the DaemonSet:
```
Labels:         k8s-app=openstack-cloud-controller-manager
```
See for yourself in the "Describe failed openstack-cloud-controller-manager" step: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/cloud-provider-openstack/2519/openstack-cloud-controller-manager-e2e-test/1749846521752326144/build-log.txt

cwrau: I don't know how CI comes up with that, but I would say it's technically impossible for this label to exist, neither in the old version nor my new one 😅

jichenjc:
e2e report this error 

```
Error from server (BadRequest): container "openstack-cloud-controller-manager" in pod "openstack-cloud-controller-manager-ghc27" is waiting to start: ImageInspectError
```

I googled this but seems the related report is pointing to image has some problem ,may need check more

dulek:
> e2e report this error
> 
> ```
> Error from server (BadRequest): container "openstack-cloud-controller-manager" in pod "openstack-cloud-controller-manager-ghc27" is waiting to start: ImageInspectError
> ```
> 
> I googled this but seems the related report is pointing to image has some problem ,may need check more

Yes, I'm on this.

charts/openstack-cloud-controller-manager/templates/daemonset.yaml
dulek: Okay, CI gets that from https://github.com/kubernetes/cloud-provider-openstack/blob/f7abe101ed041c8867d5fd1faa8351044f40a9cb/manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml, so I guess we can disregard this.

dulek:
/lgtm

@jichenjc, @kayrus, @zetaab, could you take a look here? I'm not that experienced with Helm to be confident to approve this myself.

jichenjc:
/approve 
/hold 

also ok to me and leave unhold to someone might have further experience on helm than me :)

zetaab:
imo the change looks ok, the labels should stay as is

PR #2533, opened 2024-02-07
[charts] add support for podLabels and podAnnotations to occm and cinder-csi-plugin charts

roehlc:
* feat: add support for defining `podLabels` and `podAnnotations` in the openstack-cloud-controller-manager and cinder-csi-plugin charts
* fix: explain where `commonAnnotations` are applied in cinder-csi-plugin chart

This may be used to add custom labels and annotations to theses resources. In environments using vault-injector to  add the OpenStack credentials to occm and the csi-plugin this is typically used to specify in which pods the secret should be injected. I'd rather not use already implemented `commonAnnotations` for this, as these will add the annotations not exclusively to the pods but also to the deployment, daemonset and podmonitor where they do not make any sense.

* fixes -

* please let me know if I chose the correct chart versions - I did not find any documentation about the conventions used by the charts in this repo. My interpretation was the following: the minor chart version is in sync with the appVersion which in turn is in sync with the minor k8s release

**Release note**:

```release-note
[charts/occm] add support for podLabels and podAnnotations in helm chart
[charts/cinder-csi-plugin] add support for podLabels and podAnnotations in helm chart
```

### DIFF
--- a/charts/cinder-csi-plugin/values.yaml
+++ b/charts/cinder-csi-plugin/values.yaml
@@ -76,6 +76,10 @@ csi:
         readOnly: true
     nodePlugin:
       dnsPolicy: ClusterFirstWithHostNet
+      # Optional additional annotations to add to the nodePlugin Pods.
+      podAnnotations: {}
+      # Optional additional labels to add to the nodePlugin Pods.
+      podLabels: {}
       podSecurityContext: {}
       securityContext: {}
         # capabilities:
@@ -106,6 +110,10 @@ csi:
           # maxSurge is the maximum number of pods that can be
           # created over the desired number of pods.
           maxSurge: 1
+      # Optional additional annotations to add to the controllerPlugin Pods.
+      podAnnotations: {}
+      # Optional additional labels to add to the controllerPlugin Pods.
+      podLabels: {}
       podSecurityContext: {}
         # runAsNonRoot: true
         # runAsUser: 65532
--- a/charts/cinder-csi-plugin/values.yaml
+++ b/charts/cinder-csi-plugin/values.yaml
@@ -204,5 +204,5 @@ priorityClassName: ""
 imagePullSecrets: []
 # - name: my-imagepull-secret
 
-# add annotations to all pods
+## Annotations to apply to all resources
 commonAnnotations: {}
--- a/charts/cinder-csi-plugin/values.yaml
+++ b/charts/cinder-csi-plugin/values.yaml
@@ -212,5 +212,5 @@ priorityClassName: ""
 imagePullSecrets: []
 # - name: my-imagepull-secret
 
-# add annotations to all pods
+# add annotations to all pods, deployment, daemonset and podmonitor
 commonAnnotations: {}
--- a/charts/cinder-csi-plugin/values.yaml
+++ b/charts/cinder-csi-plugin/values.yaml
@@ -212,5 +212,5 @@ priorityClassName: ""
 imagePullSecrets: []
 # - name: my-imagepull-secret
 
-## Annotations to apply to all resources
+# add annotations to all pods
 commonAnnotations: {}
--- a/charts/openstack-cloud-controller-manager/templates/_helpers.tpl
+++ b/charts/openstack-cloud-controller-manager/templates/_helpers.tpl
@@ -30,6 +30,9 @@ component: controllermanager
 {{- define "occm.controllermanager.labels" -}}
 {{ include "occm.controllermanager.matchLabels" . }}
 {{ include "occm.common.metaLabels" . }}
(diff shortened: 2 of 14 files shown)


charts/cinder-csi-plugin/values.yaml
```diff
@@ -196,5 +204,5 @@ priorityClassName: ""
 imagePullSecrets: []
 # - name: my-imagepull-secret
 
-# add annotations to all pods
+## Annotations to apply to all resources
```
dulek, line 207: Looks like we're changing the behavior of this field here in a non backward compatible way. It would at least require a proper release note, but I'd prefer to do it without a change to how the field is interpreted.

dulek (changes requested):
Could we wrap all these into some template functions, so that we don't need to write whole with-toYaml-end? Also a bunch of remarks inline.

charts/cinder-csi-plugin/values.yaml
roehlc: I added these changes because I was confused why a field with the same name does different things in the occm and cinder-csi-plugin chart. Thats why I was trying to align these annotations with the way `commonAnnotations` work in the occm chart:
See also: https://github.com/kubernetes/cloud-provider-openstack/blob/e87f50614b96a6af6095660aa6a77457223b344b/charts/openstack-cloud-controller-manager/values.yaml#L5-L6

If you would like to keep the way this field is interpreted in the cinder-csi-plugin chart the way it is right now, I would rather drop these changes from this PR instead of investing into doing this in a backward compatible way. I don't need the changes and this allows for somebody else who needs annotations on all resources to open a separate PR to solve this in a backward compatible manner.

Is this ok with you?

roehlc: After further review, I opted for changing the comment explaining what this field does instead of the behavior.

charts/openstack-cloud-controller-manager/templates/_helpers.tpl
```diff
@@ -43,6 +43,9 @@ component: controllermanager
 {{- define "occm.controllermanager.labels" -}}
 {{ include "occm.controllermanager.matchLabels" . }}
 {{ include "occm.common.metaLabels" . }}
+{{- range $key, $val := .Values.podLabels }}
+{{ $key }}: {{ $val }}
+{{- end }}
```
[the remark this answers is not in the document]
roehlc: As far as I understand, the `include` keyword for helm templating only works for including templates (i.e. `occm.common.metaLabels` defined a few lines above) and not for using values. I found this explanation in the helm docs: https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-include-function
I tried a few ways of using `include` but failed to use it with build in functions (i.e. "toYaml") or to use it to include the values directly.

I opted for changing the syntax to a conditional block and using the builtin `toYaml`, which looks cleaner in my opinion.

Another idea would be to define our own generic `toYaml` template:
```
{{- define "toYaml" -}}
{{- if . }}
{{- toYaml . }}
{{- end }}
{{- end -}}
```

which would then work with include, i.e.:
`{{ include "toYaml" .Values.podLabels }}`

But in my opinion the overhead of recreating builtins like this might not be worth it.
Please let me know if the implementation I chose works for you, or if you'd like me to add the generic "toYaml" describe above.

roehlc:
> Could we wrap all these into some template functions, so that we don't need to write whole with-toYaml-end? Also a bunch of remarks inline.

I added helper templates to remove the need to write "with-toYaml-end" and addressed the remarks inline.

@dulek Would you please take another look?

dulek:
/lgtm

Leaving approval for someone more experienced with Helm.

roehlc:
@dulek , @kayrus , @zetaab 
do any of you have capacity to take a look at this PR?

As far as I can tell I implemented the requested changes. As described in my first comment I'm unsure about the chart version and would suggest squashing the commits after your review is finished.

roehlc:
@dulek , @kayrus , @zetaab, @jichenjc
Sorry for pinging you again. Unfortunately the PR was closed when I failed to correctly rebase my branch. This also removed the `lgtm` labels.

Do any of you have bandwidth to take another look at this PR?

As described above I think I implemented all requested changes from the review.
I also added an explanation to the first comment, describing why I think this change might be useful.

kayrus:
have you tested this PR in your env? I don't use this chart and it's hard to say whether everything works as expected.

roehlc:
> have you tested this PR in your env? I don't use this chart and it's hard to say whether everything works as expected.

Yes I tested these changes. We are using this in production since I opened this PR.
I wanted to contribute these changes, so we don't have to use our own fork anymore.

PR #2551, opened 2024-02-14
[cinder-csi-plugin] Multi region/clouds support for controllerServer

### LINKED ISSUE #2035: [cinder-csi-plugin] Multi-region support for provisioning Cinder storage.
/kind feature

I have a scenario where we would need to implement high-availability across two datacenters, each operating as their own OpenStack region. As both regions are in the same City with a high speed low latency (<1ms) network between then it would be possible to create a stretched cluster between the regions. (I'm aware it's not possible to create proper high-availability of etcd with just two regions, but I belive it can be managed and it's a separate issue).

We do have some components that need persistent storage in our cluster, and we would like to be able to deploy these components in Kubernetes and use the built-in region awareness to deploy apps and their volumes.

Seems it should be possible to label each node with an appropriate `topology.kubernetes.io/region` label and then configure the Cinder CSI to map regions to a set of OpenStack credentials. Then get kind of similar to how for example the AWS EBS CSI manages different availability zones, with the awareness that only nodes in the correct region will be able to mount the volumes.

This might be a duplicate of #1924 , but I only really care about the storage part, we wouldn't need for instance the LoadBalancer support.

MatthieuFin:
**Affected binaries:**
- cinder-csi-plugin

Enable multi region / multi clouds support on cinder-csi-plugin controllerServer.

fixes #2035

I have a kubernetes cluster stretch on multiples openstack clusters (3 OVH and 1 onPremise linked by a private dark fiber).
That why my approche is more "multi clouds" than only "multi region"

I don't touch nodeServer behavior (I simply deploy a Daemonset with a nodeSelector on a label whitch select nodes based on their hypervisor, and mount a dedicated secret with openstack creds associated)

The purpose is on controllerServer which handle all kube requests to manage pvc, he had be able to CRUD volumes on all managed openstack cluster.

I propose to use [gcfg subsection](https://pkg.go.dev/gopkg.in/gcfg.v1#example-ReadStringInto-Subsections) feature to be abble to handle multiple "sections" `Global` in config files. This permit to be backaward compatible with configfiles syntax.

I choose to use StorageClass `parameters` to deal with cloud selection (i add an optionnal field `cloud` which containt config Global subsection name). In this way when a [createVolumeRequest](https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume) income controllerServer could identify OSInstance to used based on match between SC field `parametes.cloud` and his config Global subsections.

**Issue**
This PR is currently a MVP, which permit to create volumes and attach them to correct VM / OpenStack cluster, but i have a bit issue to troubleshoot, indeed all [csi spec message](https://github.com/container-storage-interface/spec/blob/master/spec.md) don't contains `volume_context` or `parameters`, especially for [DeleteVolumeRequest](https://github.com/container-storage-interface/spec/blob/master/spec.md#deletevolume) which only contain volumeID.

I have 2 ideas to troubleshoot this issue:
- query each managed OS clouds to find one who had a volume with searched ID (That's linear complexity)
- implement a kubernetes client and as kubernetes API to find (bad idea, I just realized that volume_id is openstack id not pv name)
- maybe use secrets field as it seems suggested in [external-provisioner discussion](https://github.com/kubernetes-csi/external-provisioner/issues/87)

**Release note**:

```release-note
feat add multi regions/clouds support.
```

### DIFF
--- a/cmd/cinder-csi-plugin/main.go
+++ b/cmd/cinder-csi-plugin/main.go
@@ -34,6 +34,7 @@ var (
 	endpoint                 string
 	nodeID                   string
 	cloudConfig              []string
+	cloudNames               []string
 	cluster                  string
 	httpEndpoint             string
 	provideControllerService bool
@@ -65,6 +66,8 @@ func main() {
 		klog.Fatalf("Unable to mark flag cloud-config to be required: %v", err)
 	}
 
+	cmd.PersistentFlags().StringSliceVar(&cloudNames, "cloud-name", []string{""}, "CSI driver cloud name in config files. This option can be given multiple times to manage multiple openstack clouds")
+
 	cmd.PersistentFlags().StringVar(&cluster, "cluster", "", "The identifier of the cluster that the plugin is running in.")
 	cmd.PersistentFlags().StringVar(&httpEndpoint, "http-endpoint", "", "The TCP network address where the HTTP server for providing metrics for diagnostics, will listen (example: `:8080`). The default is empty string, which means the server is disabled.")
 
@@ -82,14 +85,18 @@ func handle() {
 	d := cinder.NewDriver(&cinder.DriverOpts{Endpoint: endpoint, ClusterID: cluster})
 
 	openstack.InitOpenStackProvider(cloudConfig, httpEndpoint)
-	cloud, err := openstack.GetOpenStackProvider()
-	if err != nil {
-		klog.Warningf("Failed to GetOpenStackProvider: %v", err)
-		return
+	var err error
+	clouds := make(map[string]openstack.IOpenStack)
+	for _, cloudName := range cloudNames {
+		clouds[cloudName], err = openstack.GetOpenStackProvider(cloudName)
+		if err != nil {
+			klog.Warningf("Failed to GetOpenStackProvider %s: %v", cloudName, err)
+			return
+		}
 	}
 
 	if provideControllerService {
-		d.SetupControllerService(cloud)
+		d.SetupControllerService(clouds)
 	}
 
 	if provideNodeService {
@@ -97,9 +104,9 @@ func handle() {
 		mount := mount.GetMountProvider()
 
 		//Initialize Metadata
-		metadata := metadata.GetMetadataProvider(cloud.GetMetadataOpts().SearchOrder)
+		metadata := metadata.GetMetadataProvider(clouds[cloudNames[0]].GetMetadataOpts().SearchOrder)
 
-		d.SetupNodeService(cloud, mount, metadata)
+		d.SetupNodeService(clouds[cloudNames[0]], mount, metadata)
 	}
 
 	d.Run()
--- a/cmd/cinder-csi-plugin/main.go
+++ b/cmd/cinder-csi-plugin/main.go
@@ -35,6 +35,7 @@ var (
 	nodeID                   string
 	cloudConfig              []string
 	cloudNames               []string
+	additionalTopologies     map[string]string
 	cluster                  string
 	httpEndpoint             string
 	provideControllerService bool
@@ -67,6 +68,7 @@ func main() {
 	}
 
 	cmd.PersistentFlags().StringSliceVar(&cloudNames, "cloud-name", []string{""}, "CSI driver cloud name in config files. This option can be given multiple times to manage multiple openstack clouds")
+	cmd.PersistentFlags().StringToStringVar(&additionalTopologies, "additionnal-topology", map[string]string{}, "CSI driver topology example topology.kubernetes.io/region=REGION1. This option can be given multiple times to manage multiple topology keys")
 
 	cmd.PersistentFlags().StringVar(&cluster, "cluster", "", "The identifier of the cluster that the plugin is running in.")
 	cmd.PersistentFlags().StringVar(&httpEndpoint, "http-endpoint", "", "The TCP network address where the HTTP server for providing metrics for diagnostics, will listen (example: `:8080`). The default is empty string, which means the server is disabled.")
@@ -106,7 +108,7 @@ func handle() {
 		//Initialize Metadata
 		metadata := metadata.GetMetadataProvider(clouds[cloudNames[0]].GetMetadataOpts().SearchOrder)
 
-		d.SetupNodeService(clouds[cloudNames[0]], mount, metadata)
+		d.SetupNodeService(clouds[cloudNames[0]], mount, metadata, additionalTopologies)
 	}
 
 	d.Run()
--- a/cmd/cinder-csi-plugin/main.go
+++ b/cmd/cinder-csi-plugin/main.go
@@ -67,8 +67,8 @@ func main() {
 		klog.Fatalf("Unable to mark flag cloud-config to be required: %v", err)
 	}
 
-	cmd.PersistentFlags().StringSliceVar(&cloudNames, "cloud-name", []string{""}, "CSI driver cloud name in config files. This option can be given multiple times to manage multiple openstack clouds")
-	cmd.PersistentFlags().StringToStringVar(&additionalTopologies, "additionnal-topology", map[string]string{}, "CSI driver topology example topology.kubernetes.io/region=REGION1. This option can be given multiple times to manage multiple topology keys")
+	cmd.PersistentFlags().StringSliceVar(&cloudNames, "cloud-name", []string{""}, "Cloud name to instruct CSI driver to read additional OpenStack cloud credentials from the configuration subsections. This option can be specified multiple times to manage multiple OpenStack clouds.")
+	cmd.PersistentFlags().StringToStringVar(&additionalTopologies, "additional-topology", map[string]string{}, "Additional CSI driver topology keys, for example topology.kubernetes.io/region=REGION1. This option can be specified multiple times to add multiple additional topology keys.")
 
 	cmd.PersistentFlags().StringVar(&cluster, "cluster", "", "The identifier of the cluster that the plugin is running in.")
 	cmd.PersistentFlags().StringVar(&httpEndpoint, "http-endpoint", "", "The TCP network address where the HTTP server for providing metrics for diagnostics, will listen (example: `:8080`). The default is empty string, which means the server is disabled.")
--- a/docs/cinder-csi-plugin/multi-region-clouds.md
+++ b/docs/cinder-csi-plugin/multi-region-clouds.md
@@ -1,0 +1,316 @@
+# Multi Az/Region/Openstack Configuration
+
+### Multi cluster Configuration file
+
+Create a configuration file with a subsection per openstack cluster to manage (pay attention to enable ignore-volume-az in BlockStorage section).
+
+Example of configuration with 3 regions (The default is backward compatible with mono cluster configuration but not mandatory).
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cloud-config
+  namespace: kube-system
+type: Opaque
+stringData:
+  cloud.conf: |-
+    [BlockStorage]
+    bs-version=v3
+    ignore-volume-az=True
+    
+    [Global]
+    auth-url="https://auth.cloud.openstackcluster.region-default.local/v3"
+    username="region-default-username"
+    password="region-default-password"
+    region="default"
+    tenant-id="region-default-tenant-id"
+    tenant-name="region-default-tenant-name"
+    domain-name="Default"
+    
+    [Global "region-one"]
+    auth-url="https://auth.cloud.openstackcluster.region-one.local/v3"
+    username="region-one-username"
+    password="region-one-password"
+    region="one"
+    tenant-id="region-one-tenant-id"
+    tenant-name="region-one-tenant-name"
+    domain-name="Default"
+    
+    [Global "region-two"]
+    auth-url="https://auth.cloud.openstackcluster.region-two.local/v3"
+    username="region-two-username"
+    password="region-two-password"
+    region="two"
+    tenant-id="region-two-tenant-id"
+    tenant-name="region-two-tenant-name"
+    domain-name="Default"
+```
+
+
+
+### Create region/cloud secrets
+
+Create a secret per openstack cluster which contains a key `cloud` and as value the subsection's name of corresponding openstack cluster in configuration file.
+
+These secrets are referenced in storageClass definitions to identify openstack cluster associated to the storageClass.
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: openstack-config-region-one
+  namespace: kube-system
+type: Opaque
+stringData:
+  cloud: region-one
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: openstack-config-region-two
+  namespace: kube-system
+type: Opaque
+stringData:
+  cloud: region-two
+```
+
+### Create storage Class for dedicated cluster
+
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  annotations:
+    storageclass.kubernetes.io/is-default-class: "true"
+  name: sc-region-one
+allowVolumeExpansion: true
+allowedTopologies:
+- matchLabelExpressions:
+  - key: topology.cinder.csi.openstack.org/zone
+    values:
+    - nova
+  - key: topology.kubernetes.io/region
+    values:
+    - region-one
+parameters:
+  csi.storage.k8s.io/controller-publish-secret-name: openstack-config-region-one
+  csi.storage.k8s.io/controller-publish-secret-namespace: kube-system
+  csi.storage.k8s.io/node-publish-secret-name: openstack-config-region-one
+  csi.storage.k8s.io/node-publish-secret-namespace: kube-system
+  csi.storage.k8s.io/node-stage-secret-name: openstack-config-region-one
+  csi.storage.k8s.io/node-stage-secret-namespace: kube-system
+  csi.storage.k8s.io/provisioner-secret-name: openstack-config-region-one
+  csi.storage.k8s.io/provisioner-secret-namespace: kube-system
+provisioner: cinder.csi.openstack.org
+reclaimPolicy: Delete
+volumeBindingMode: Immediate
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: sc-region-two
+allowVolumeExpansion: true
+allowedTopologies:
+- matchLabelExpressions:
+  - key: topology.cinder.csi.openstack.org/zone
+    values:
+    - nova
+  - key: topology.kubernetes.io/region
+    values:
+    - region-two
+parameters:
+  csi.storage.k8s.io/controller-publish-secret-name: openstack-config-region-two
+  csi.storage.k8s.io/controller-publish-secret-namespace: kube-system
+  csi.storage.k8s.io/node-publish-secret-name: openstack-config-region-two
+  csi.storage.k8s.io/node-publish-secret-namespace: kube-system
+  csi.storage.k8s.io/node-stage-secret-name: openstack-config-region-two
+  csi.storage.k8s.io/node-stage-secret-namespace: kube-system
+  csi.storage.k8s.io/provisioner-secret-name: openstack-config-region-two
+  csi.storage.k8s.io/provisioner-secret-namespace: kube-system
+provisioner: cinder.csi.openstack.org
+reclaimPolicy: Delete
+volumeBindingMode: Immediate
+```
+
+### Create a csi-cinder-nodeplugin daemonset per cluster openstack
+
+Daemonsets should deploy pods on nodes from proper openstack context. We suppose that the node have a label `topology.kubernetes.io/region` with the openstack cluster name as value (you could manage this with kubespray, manually, whatever, it should be great to implement this in openstack cloud controller manager).
+
+Do as follows:
+- Use nodeSelector to match proper nodes labels
+- Add cli argument `--additionnal-topology topology.kubernetes.io/region=region-one`, which should match node labels, to container cinder-csi-plugin
+- Add cli argument `--cloud-name="region-one"`, which should match configuration file subsection name, to container cinder-csi-plugin.
+
+```yaml
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: csi-cinder-nodeplugin-region-one
+  namespace: kube-system
+spec:
+  selector:
+    matchLabels:
+      app: csi-cinder-nodeplugin-region-one
+  template:
+    metadata:
+      labels:
+        app: csi-cinder-nodeplugin-region-one
+    spec:
+      containers:
+      - name: node-driver-registrar
+        ...
+      - name: liveness-probe
+        ...
+      - name: cinder-csi-plugin
+        image: docker.io/k8scloudprovider/cinder-csi-plugin:v1.31.0
+        args:
+        - /bin/cinder-csi-plugin
+        - --endpoint=$(CSI_ENDPOINT)
+        - --cloud-config=$(CLOUD_CONFIG)
+        - --cloud-name="region-one"
+        - --additionnal-topology
+        - topology.kubernetes.io/region=region-one
+        env:
+        - name: CSI_ENDPOINT
+          value: unix://csi/csi.sock
+        - name: CLOUD_CONFIG
+          value: /etc/config/cloud.conf
+        ...
+        volumeMounts:
+        ...
+        - mountPath: /etc/config
+          name: secret-cinderplugin
+          readOnly: true
+        ...
+      nodeSelector:
+        topology.kubernetes.io/region: region-one
+      volumes:
+      ...
+      - name: secret-cinderplugin
+        secret:
+          defaultMode: 420
+          secretName: cloud-config
+      ...
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: csi-cinder-nodeplugin-region-two
+  namespace: kube-system
+spec:
+  selector:
+    matchLabels:
+      app: csi-cinder-nodeplugin-region-two
+  template:
+    metadata:
+      labels:
+        app: csi-cinder-nodeplugin-region-two
+    spec:
+      containers:
+      - name: node-driver-registrar
+        ...
+      - name: liveness-probe
+        ...
+      - name: cinder-csi-plugin
+        image: docker.io/k8scloudprovider/cinder-csi-plugin:v1.31.0
+        args:
+        - /bin/cinder-csi-plugin
+        - --endpoint=$(CSI_ENDPOINT)
+        - --cloud-config=$(CLOUD_CONFIG)
+        - --cloud-name="region-two"
+        - --additionnal-topology
+        - topology.kubernetes.io/region=region-two
+        env:
+        - name: CSI_ENDPOINT
+          value: unix://csi/csi.sock
+        - name: CLOUD_CONFIG
+          value: /etc/config/cloud.conf
+        ...
+        volumeMounts:
+        ...
+        - mountPath: /etc/config
+          name: secret-cinderplugin
+          readOnly: true
+        ...
+      nodeSelector:
+        topology.kubernetes.io/region: region-two
+      volumes:
+      ...
+      - name: secret-cinderplugin
+        secret:
+          defaultMode: 420
+          secretName: cloud-config
+      ...
+```
+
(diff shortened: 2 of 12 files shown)


MatthieuFin:
As mentioned [here](https://github.com/kubernetes-csi/external-provisioner/issues/87#issuecomment-411880948) and after some local tests it seems to be the proper implementation way.

I propose to avoid usage of storageClass `parameter.cloud` and volume context to propagate config cloud name between gRPC calls, and instead use a secret with a key `cloud` which contain the Global config subsection name, which permit to controller to retrieve proper credentials from his configuration.

dulek:
Looks like the tests have to be fixed here.

MatthieuFin:
Hi,
After a few days of use on my cluster, I encountered an issue, regard `allowedTopologies` set on StorageClass (first of all don't forget to enable feature gate Topology add argument `--feature-gates Topology=true` on container `csi-provisioner`) which permit to push allowedTopologies constraints on pv .spec.nodeAffinity and keep affinity on pod reschedule.

My issue concern topology keys, currently only available is topology.cinder.csi.openstack.org/zone and obviously my provider OVH use same zone name (nova) on each of their openstack cluster, so I need to add possibility to manage another key to differentiate them.

sebastienmusso:
> Hi, After a few days of use on my cluster, I encountered an issue, regard `allowedTopologies` set on StorageClass (first of all don't forget to enable feature gate Topology add argument `--feature-gates Topology=true` on container `csi-provisioner`) which permit to push allowedTopologies constraints on pv .spec.nodeAffinity and keep affinity on pod reschedule.
> 
> My issue concern topology keys, currently only available is topology.cinder.csi.openstack.org/zone and obviously my provider OVH use same zone name (nova) on each of their openstack cluster, so I need to add possibility to manage another key to differentiate them.

we have exactly the same problem (several openstack with same availability zone name)

I'm interesseted by this feature let me know if you need some test feedback

regards

MatthieuFin:
> do we aggre that I have to deploy the whole cinder csi (daemonset+deployment) for both of my openstack ? with field "region" different for each openstack right ?

you have to provide a secret with your differents openstack clusters credentials 
```
[BlockStorage]
bs-version=v3
ignore-volume-az=True

[Global "OVH-GRA11"]
auth-url="https://auth.cloud.ovh.net/v3"
username="****"
password="****"
region="GRA11"
tenant-id="****"
tenant-name="****"
domain-name="Default"

[Global "OVH-GRA9"]
auth-url="https://auth.cloud.ovh.net/v3"
username="***"
password=""
region="GRA9"
tenant-id="****"
tenant-name="****"
domain-name="Default"

...
```

One deployement for controller with argument `cloud-name` referencing your configuration in that case : `--cloud-name=GRA9 --cloud-name=GRA11` which permit to your controller to manager all your OS clusters

And one daemonset nodeServer with nodeselector per openstack cluster with `--cloud-name=GRA9` option for corresponding OS cluster and `--additionnal-topology topology.kubernetes.io/region=*GRA9`

I share you my local helm chart to deploy this stack it will be probably simpler than words ;-)
[csi-cinder.tar.gz](https://github.com/kubernetes/cloud-provider-openstack/files/14407628/csi-cinder.tar.gz)

sebastienmusso:
did you add something specific to the build step I have this error:

Args (comma-delimited): /bin/cinder-csi-plugin,-v=2,--endpoint=unix://csi/csi.sock,--cloud-config=/etc/kubernetes/cloud.conf,--additionnal-topology=topology.kubernetes.io/region=dc1-int
/bin/cinder-csi-plugin: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /bin/cinder-csi-plugin)
2024/02/27 14:20:19 Now listening for interrupts
/bin/cinder-csi-plugin: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /bin/cinder-csi-plugin)

MatthieuFin:
I build binary locally with:
```bash
VERSION=v1.29.1-rc.16 make build cinder-csi-plugin
```
and docker image with:
```bash
REGISTRY='0g4p75r8.c1.gra9.container-registry.ovh.net/infra' VERSION=v1.29.1-rc.16 make build-local-image-cinder-csi-plugin
```

You need docker buildx which should embed right verison of glibc on my local system i have glibc 2.39:
```bash
$ ldd --version
ldd (GNU libc) 2.39
```

sebastienmusso:
I took your latest commit and rebuild image.
controller successfully list volumes on both openstack and daemon are up and running with right region and tolopology labels.

But when provisionner cannot create volume with this error:


I0229 09:08:39.891611       1 event.go:298] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"kube-system", Name:"pod-with-volume", UID:"c8694fba-26a7-4288-a9b9-9b065023585a", APIVersion:"v1", ResourceVersion:"1746245", FieldPath:""}): type: 'Warning' reason: 'ProvisioningFailed' failed to provision volume with StorageClass "region-1": rpc error: code = InvalidArgument desc = [CreateVolume] specified cloud undefined

event with 
--additionnal-topology=topology.kubernetes.io/region=region-2
--cloud-name=region-2

extra arg in cinder-csi-plugin container

with this cloud config:
`[Global "region-1"]
region=region-1
tls-insecure=true
auth-url=https://keystone.region-1.example:5000/v3
username=region-1-member
password=region-1-password
tenant-name=region-1-tenant
[Global "region-2"]
tls-insecure=true
region=region-2
auth-url=https://keystone.region-2.example:5000/v3
username=region-2-member
password=region-2-password
tenant-name=region-2-tenant
[BlockStorage]
bs-version=v3
ignore-volume-az=true
rescan-on-resize=true
[Metadata]
search-order=metadataService`

MatthieuFin:
Yes, I think you didn't create appropriate secret to reference you Global section "region-1"
Create following secret in kube-system namespace:
```yaml
apiVersion: v1
stringData:
  cloud: region-1
kind: Secret
metadata:
  name: cloud-config-sc-region-1
  namespace: kube-system
type: Opaque
```

And link this secret to your storageClass in paramaters as follow:

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    storageclass.kubernetes.io/is-default-class: "false"
  name: cinder-region-1
allowVolumeExpansion: true
allowedTopologies:
- matchLabelExpressions:
  - key: topology.cinder.csi.openstack.org/zone
    values:
    - nova
  - key: topology.kubernetes.io/region
    values:
    - region-1
parameters:
  # Create/Delete Volume Secret
  csi.storage.k8s.io/controller-publish-secret-name: cloud-config-sc-region-1
  csi.storage.k8s.io/controller-publish-secret-namespace: kube-system
  # Controller Publish/Unpublish Secret
  csi.storage.k8s.io/node-publish-secret-name: cloud-config-sc-region-1
  csi.storage.k8s.io/node-publish-secret-namespace: kube-system
  # Node Stage Secret
  csi.storage.k8s.io/node-stage-secret-name: cloud-config-sc-region-1
  csi.storage.k8s.io/node-stage-secret-namespace: kube-system
  # Node Publish Secret
  csi.storage.k8s.io/provisioner-secret-name: cloud-config-sc-region-1
  csi.storage.k8s.io/provisioner-secret-namespace: kube-system
reclaimPolicy: Delete
volumeBindingMode: WaitForFirstConsumer

```


Process same for region-2 (and adapt to your case)

MatthieuFin:
Hi this version is deployed on our cluster from a couple of weeks it seems to be stable.

@dulek @mdbooth @jichenjc @kayrus  Please take a look.

cmd/cinder-csi-plugin/main.go
```diff
@@ -67,6 +68,7 @@ func main() {
 	}
 
 	cmd.PersistentFlags().StringSliceVar(&cloudNames, "cloud-name", []string{""}, "CSI driver cloud name in config files. This option can be given multiple times to manage multiple openstack clouds")
+	cmd.PersistentFlags().StringToStringVar(&additionalTopologies, "additionnal-topology", map[string]string{}, "CSI driver topology example topology.kubernetes.io/region=REGION1. This option can be given multiple times to manage multiple topology keys")
```
dulek, line 71: Let's rephrase this a bit:

> Additional CSI driver topology keys, for example topology.kubernetes.io/region=REGION1. This option can be specified multiple times to add multiple additional topology keys.

Also there's a typo in the option name, please fix.

cmd/cinder-csi-plugin/main.go
```diff
@@ -65,6 +66,8 @@ func main() {
 		klog.Fatalf("Unable to mark flag cloud-config to be required: %v", err)
 	}
 
+	cmd.PersistentFlags().StringSliceVar(&cloudNames, "cloud-name", []string{""}, "CSI driver cloud name in config files. This option can be given multiple times to manage multiple openstack clouds")
```
dulek, line 70: How about:

> Cloud name to instruct CSI driver to read additional OpenStack cloud credentials from the configuration subsections. This option can be specified multiple times to manage multiple OpenStack clouds.

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -197,24 +207,25 @@ func CreateOpenStackProvider() (IOpenStack, error) {
...
 
 // GetOpenStackProvider returns Openstack Instance
-func GetOpenStackProvider() (IOpenStack, error) {
-	if OsInstance != nil {
+func GetOpenStackProvider(cloudName string) (IOpenStack, error) {
+	OsInstance, OsInstanceDefined := OsInstances[cloudName]
+	if OsInstanceDefined {
 		return OsInstance, nil
 	}
 	var err error
```
dulek, line 227: You can use := in the next line to skip this declaration.

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -90,13 +90,16 @@ type BlockStorageOpts struct {
 }
 
 type Config struct {
-	Global       client.AuthOpts
+	Global       map[string]*client.AuthOpts
```
dulek, line 93: Why a pointer?

pkg/csi/cinder/controllerserver.go
```diff
@@ -315,52 +346,130 @@ func (cs *controllerServer) ControllerUnpublishVolume(ctx context.Context, req *
...
+	var cloudsVentries []*csi.ListVolumesResponse_Entry
+	var vlist []volumes.Volume
+	var nextPageToken string
 
-		status := &csi.ListVolumesResponse_VolumeStatus{}
-		status.PublishedNodeIds = make([]string, 0, len(v.Attachments))
-		for _, attachment := range v.Attachments {
-			status.PublishedNodeIds = append(status.PublishedNodeIds, attachment.ServerID)
+	startIdx := 0
+	for idx := 0; idx < len(cloudsNames); idx++ {
```
dulek, line 387: Why not iterate using `range()`?

pkg/csi/cinder/controllerserver.go
```diff
@@ -48,6 +51,13 @@ const (
 func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
 	klog.V(4).Infof("CreateVolume: called with args %+v", protosanitizer.StripSecrets(*req))
 
+	// Volume cloud
+	volCloud := req.GetSecrets()["cloud"]
```
dulek, line 55: How does `GetSecrets()` work?

pkg/csi/cinder/controllerserver.go
```diff
@@ -777,12 +907,19 @@ func (cs *controllerServer) ControllerGetVolume(ctx context.Context, req *csi.Co
...
+	var err error
+	for _, cloud := range cs.Clouds {
+		volume, err = cloud.GetVolume(volumeID)
+		if err != nil {
+			if cpoerrors.IsNotFound(err) {
+				continue
+			}
+			return nil, status.Errorf(codes.Internal, "ControllerGetVolume failed with error %v", err)
 		}
-		return nil, status.Errorf(codes.Internal, "ControllerGetVolume failed with error %v", err)
```
dulek, line 785: So we don't know which cloud each volume is on and we need to iterate over all of them? This feels a bit inefficient here.

dulek (changes requested):
I've listed a bunch of remarks inline. Besides that this surely needs documentation before it's merged.

I'm a bit inclined to at least request investigating a possibility to test this in the gate, even if just by having credentials for 2 separate projects on a single DevStack installation.

sebastienmusso:
> Hi this version is deployed on our cluster from a couple of weeks it seems to be stable.
> 
> 
> 
> @dulek @mdbooth @jichenjc @kayrus  Please take a look.

With the help of Mathieu I successfuly deployed this version on one of our cluster spread other several openstack. It work like. Charm . We plan to test it on a single openstack in order to Check regression.

cmd/cinder-csi-plugin/main.go
MatthieuFin: c98dfe4cbbd70dce0ab0260180ed91c35edb3aec

cmd/cinder-csi-plugin/main.go
MatthieuFin: c98dfe4cbbd70dce0ab0260180ed91c35edb3aec

pkg/csi/cinder/openstack/openstack.go
MatthieuFin: 00bb738a26f48740410451f4e7e635643047995f

pkg/csi/cinder/openstack/openstack.go
MatthieuFin: Because it's a requirement of gcfg library for subsections handle https://pkg.go.dev/gopkg.in/gcfg.v1#example-ReadStringInto-Subsections
If I use type `map[string]client.AuthOpts` type, I got following error on execution of function `gcfg.FatalOnly(gcfg.ReadInto(&cfg, config))`:
```
panic: map field for section must have string keys and  pointer-to-struct values: section "Global"
```
Do you have a better solution in mind ?

pkg/csi/cinder/controllerserver.go
MatthieuFin: 0128e96cebabcd2c44efce3307802c834026392b

pkg/csi/cinder/controllerserver.go
MatthieuFin: CSI [spec](https://github.com/container-storage-interface/spec/blob/master/spec.md) handle passing secrets through RPC calls, so some calls have a `secrets` field implemented in request.

In this case [CreateVolume](https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume) request support passing some secret data:
image

How to use it:
- You have to create your k8s secret in namespace you want with your data. In our case secret contains only a key `cloud` with the "could name" as value nothing really sensitive but it is the only way I found to be able to propagate cloud hosting a volume.
- Then you have to refer this secret in your StorageClass object as documented [here](https://kubernetes-csi.github.io/docs/secrets-and-credentials-storage-class.html) which permit to specify to external-provisioner to feed RPC call secrets with your kubernetes secret data.
- In this way CSI spec offert a method [GetSecrets](https://github.com/container-storage-interface/spec/blob/v1.8.0/lib/go/csi/csi.pb.go#L979) to retrieve this data from csi-plugin implmentation (in our case cinder-csi-plugin).

This is my third idea to resolve this issue (mention in issue description)

> I have 2 ideas to troubleshoot this issue:
> - query each managed OS clouds to find one who had a volume with searched ID (That's linear complexity)
> - implement a kubernetes client and as kubernetes API to find (bad idea, I just realized that volume_id is openstack id not pv name)
> - maybe use secrets field as it seems suggested in [external-provisioner discussion](https://github.com/kubernetes-csi/external-provisioner/issues/87)

pkg/csi/cinder/controllerserver.go
MatthieuFin: Indeed !
In this case we wanna implement [ControllerGetVolumeRequest](https://github.com/container-storage-interface/spec/blob/master/spec.md#controllergetvolume) RPC message unfortunately this message don't provide `secrets` field, so I can't use my trick describe above with [StorageClasse Secrets](https://kubernetes-csi.github.io/docs/secrets-and-credentials-storage-class.html).

And as you can figure it doesn't handle secrets for operation `ControllerGetVolume`
image

Finally I had to choose one of my 2 first ideas and I choose to iterate over each cloud.
I don't know OS call complexity but if it is a hash table or equivalent this call should be o(1) so iterate over cloud provide a linear complexity o(n) with n the number of cloud configured.
I find this acceptable, the best way should probably to add support of operation `ControllerGetVolume` in [external-provisioner](https://kubernetes-csi.github.io/docs/external-provisioner.html) to being able to propagate cloud name in RPC call and retrieve a complexity o(1) (or flat OS complexity). But I think that is a hard work for a low gain.

MatthieuFin:
Sorry for latency, I had some emergencies to resolved first. @dulek thanks for your review I tried to answer to your questions and made changes requested.
I will look to write some documentation, If anyone wanna help on documentation writing I could appreciate <3.

MatthieuFin:
Thanks to @sebastienmusso, @fllaap and @Krast76 who offered to help write documentation.

@dulek, @jichenjc, @kayrus, @mdbooth how can we advance this PR ?

sergelogvinov:
Thank you for this PR, very interesting idea. 👍 

So the complexity of this PR is that node-pluging needs openstack api credentials for ephemeral storage which is deprecated. Perhaps we should ask about the removal process for it?

The killer feature is to use one name of storageClass in stetefulset deployment. Base on nodAffinity/podAntiAffinity kubernetes scheduler spread the pods across the regions/zones.

I quit to using allowedTopologies in storageClass they are immutable. It is very hard to maintain. It's better to use nodeSelector/nodeAffinity instead; you can change these parameters during the lifetime.

To help kubernetes scheduler to choose the best region/zone (if not set) we need to expose free disk capacity.

sergelogvinov:
I've added a capacity  capabilities - https://github.com/kubernetes/cloud-provider-openstack/pull/2597

sebastienmusso:
@dulek, @jichenjc, @kayrus, @mdbooth how can we advance this PR ?
And if this capabilities is not available within openstack cinder-csi-plugin, how can we handle
the stateful storage in a kubernetes cluster spread over multiple openstack cluster ?

jichenjc:
>>>And if this capabilities is not available within openstack cinder-csi-plugin, how can we handle
the stateful storage with cinder in a kubernetes cluster spread over multiple openstack cluster ?

I don't know how other cloud are handling this kind of cross region (e.g AWS cloud provider handle this?) maybe someone can help comment 

and I knew openstack itself handling multiple cloud (region) is that not good
especialyly I don't think we knew the free size of the storage cinder backend IIRC
and neutron might give us additional issue as cross region usually means no connection (not sure the impact to our network related things implementation) 

so overall I think we may easily focus on compute (VM) provisioning and claim others as gap/follow up?

MatthieuFin:
Hi @jichenjc ,
Other cloud provider (I'll take example of GCP) offer a unique storage class which is able to create pods across multiple "regions".
To do that they implements [topology spec](https://kubernetes-csi.github.io/docs/topology.html?highlight=master#implementing-topology-in-your-csi-driver) in their [csi implementation](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/586f6c84fc0b27f5b9da11eda29faa2bf9abf8d4/pkg/gce-pd-csi-driver/controller.go#L345) with help of `PluginCapability` `VOLUME_ACCESSIBILITY_CONSTRAINTS`

Technically our [openstack implementation](https://github.com/kubernetes/cloud-provider-openstack/blob/7f1daa8b0edacabda1a3bd16f87377a5e767e241/pkg/csi/cinder/driver.go#L89) doesn't support `PluginCapability` `VOLUME_ACCESSIBILITY_CONSTRAINTS` so GRPc calls [createVolume](https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume) doesn't have field `accessibility_requirements` fill.

We can implement support of `PluginCapability` `VOLUME_ACCESSIBILITY_CONSTRAINTS` to get informations in field `accessibility_requirements` and retreive information from scheduler or git topology choose by csi plugin to scheduler (depending of volumeBindingMode). anyway this approach doesn't help us here.

Indeed, GCP or AWS could implement that because all regions are uniformly managed by cloud API, so when you ask for a volume_id in [DeleteVolumeRequest](https://github.com/container-storage-interface/spec/blob/master/spec.md#deletevolume) for example (same for others calls) to you cloud API you got response never mind in which region is the volume.

In our case, manage multi openstack region is the same approach than manage multiple openstack clusters. So I have many n clouds api to n regions, where in GCP you have n region for only 1 cloud API.

The "GCP/AWS like" behavior is probably to manage "regions" with AZ zone notion in openstack. But that is not my use case, I need to be able to support multiple openstack clusters not multiple AZ in one openstack cluster. And I don't know public cloud provider based on openstack which properly implement openstack availability zones...

Moreover if we wanna handle multi clouds (openstack regions) with the help of `PluginCapability` `VOLUME_ACCESSIBILITY_CONSTRAINTS` we have 2 possibility:
- modify [csi spec](https://github.com/container-storage-interface/spec/blob/master/spec.md) to add some additional data which permit to trace cloud volume owner in GRPc calls (I don't think that will be possible/accepted)
- Implement an "openstack api like" software which expose a unique API to csi plugin and manage multiple clouds in backend and maintain an internal database to know which cloud has each volume_id, and ovoid possible collision in volume_id between 2 differents clouds.

Personally these 2 solution really to complex to implement and maintain.

This PR permit to do it easier way to understand and maintain imo.

I have a storageClass per regions/clouds. and if i wanna deploy 1 sts cross 3 regions I create 3 pvc one per regions (each with a different SC: `data-0`, `data-1`, `data-2`) and I use sts:
```yaml
spec.volumeClaimTemplates:
  - apiVersion: v1
    kind: 
      name: data
```

In that way I'm able to dispatch pods from same sts cross different sc and different openstack cluster.

If you wanna a "multi-region" implementation like GCP or AWS
- Currently I couldn't implement it because I haven't openstack cluster environment with multi AZ.
- You miss a feature that GCP/AWS doesn't offer which is storage support cross multi openstack cluster.


From my limited experience as a cloud consumer, openstack providers offer openstacks with multiple regions and not with multiple AZs. It's sad I know, but it seems to be the public offerings.

Concerning network cross region indeed it shouldn't have to be managed by neutron, personally i managed it by myself out of openstack with some BGP routing and physical interconnections L2 (dark fiber) cross my DCs and ipsec tunnels for dev/poc links cross different DCs not physically linked (I shared my firsts reflections about it [here](https://github.com/MatthieuFin/ovh-public-cloud-network) but that's out of this PR scope). I know that @sebastienmusso run k8s clusters spread on multiples openstack clusters too, I don't know how you handle this point ?

Implement capability

sergelogvinov:
+1

I think we do not break anything with `VOLUME_ACCESSIBILITY_CONSTRAINTS` capability, and in multi-region setup, we need to add the extra label to the volume topology.

So we need to decide what the label we need to use `topology.kubernetes.io/region` or csi-specific label like `cinder.csi.openstack.org/cluster`. In second case csi-controller should label the nodes by itself.

I prefer to use well-known label  `topology.kubernetes.io/region` but it adds limitations -> region == openstack-cluster-name 
As 'openstack-cluster-name' is a string, it should not be a problem.

MatthieuFin:
I agree, support for VOLUME_ACCESSIBILITY_CONSTRAINTS capability should be compatible with this multi-cloud PR implementation. If anyone has access to a multi-AZ openstack cluster :crossed_fingers: .

I intentionally let the user choice of which label to use, `topology.kubernetes.io/region` is just mentioned as an example in the cli argument `--additional-topology` that way we are not stuck if we wanna manage this label with occm or manually by the user.

jichenjc:
ok, thanks for the detailed info, now I have better understanding of the desire with the info got from `VOLUME_ACCESSIBILITY_CONSTRAINTS` , it's really helpful

Hybrid512:
Hi,
Any chance to see this PR merged ?
We tested it and it definitely fits our needs.
As @MatthieuFin explained, it is quite common to have "multiple regions" (aka. multiple independant Openstack clusters, and thus, multiple API endpoints) rather than having big, distributed, Openstack clusters with AZs managed by a single API.
We use kubernetes to "stretch" our workloads on top our multiple standalone Openstack clusters so being able to manage their "region locked" storage is mandatory.

We would love to help in any way if that can make this PR merged soon.

docs/cinder-csi-plugin/multi-region-clouds.md
```diff
@@ -0,0 +1,316 @@
...
+    tenant-id="region-two-tenant-id"
+    tenant-name="region-two-tenant-name"
+    domain-name="Default"
+```
+
+
+
+### Create region/cloud secrets
+
+Create a secret per openstack cluster which contains a key `cloud` and as value the subsection's name of corresponding openstack cluster in configuration file.
```
zetaab, line 53: I would still use the correct terminology in everywhere in this PR. You call openstack region as "openstack cluster". It will confuse users in long run. Could you modify terminology from "openstack cluster" -> "openstack region".

pkg/csi/cinder/controllerserver.go
```diff
@@ -274,14 +297,21 @@ func (cs *controllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 func (cs *controllerServer) ControllerUnpublishVolume(ctx context.Context, req *csi.ControllerUnpublishVolumeRequest) (*csi.ControllerUnpublishVolumeResponse, error) {
 	klog.V(4).Infof("ControllerUnpublishVolume: called with args %+v", protosanitizer.StripSecrets(*req))
 
+	// Volume cloud
+	volCloud := req.GetSecrets()["cloud"]
```
zetaab, line 301: this change is breaking? Means that existing storageClasses does not work? Everyone should now add new parameters to their classes?

zetaab (changes requested):

pkg/csi/cinder/controllerserver.go
MatthieuFin: There is no breaking change.
You could upgrade from last version to this PR version without configuration changes, it is backward compatible.
I use it to migrate my already created PVC on old SC without secrets reference on my new SC. Old PVC still works and new created PVC on old SC without secrets reference still works too.
GetSecrets() method return type `map[string]string` accessing to an undefined key of this kind of map return an empty string `""`.
Simple example to run locally:
```golang
package main

import "fmt"

func main() {
    secrets := make(map[string]string)
    fmt.Println(secrets["cloud"])
    fmt.Println(secrets["cloud"] == "")
}
```
```bash
$ go run ./main.go

true
```
and empty string in this `volCloud` variable refer to the actual configuration cloud.conf section `[Global]`  (section without subsection or section with subsection `""` is the same behavior)

MatthieuFin:
Thanks to @zetaab on slack #provider-openstack we identify an issue on listVolumes requests when `--max-entries` is set on external-attacher and lower than openstack volumes number. I will try to troubleshoot that.

MatthieuFin:
> Thanks to @zetaab on slack #provider-openstack we identify an issue on listVolumes requests when `--max-entries` is set on external-attacher and lower than openstack volumes number. I will try to troubleshoot that.

I just added some unit tests on the `ListVolumes` function with and without using `--max-entries` usage in a multi-cloud configuration. To be able to say that this function should work well.

Moreover, I tested it on my cluster and all volumes are correctly listed with pagination feature.

Implementation sounds good now for me !

@dulek , @kayrus , @zetaab , @jichenjc , @mdbooth Please could we have feedbacks ?

Hybrid512:
Looks like @MatthieuFin fixed most of the issues, what are the reasons why this code can't be merged ?

jichenjc:
looks like there is a merge conflict, the others ok to me now

/lgtm

MatthieuFin:
> looks like there is a merge conflict, the others ok to me now
> 
> /lgtm

Rebase done

MatthieuFin:
Hello there !
Can we hope that this PR will be merged ?
Personally, I need this feature in my production environment since couple month, in my POV it works pretty well, and maintaining a fork with rebase to do and conflicts to resolve is quite painful.

jichenjc:
I can approve but I still need folks to take a look
so if no objection until 3-5 days later then I will go ahead and merge this

MatthieuFin:
No problem !
I just wanted an update on this PR state. Thanks for it.
Take your time, I prefer a proper review

devfaz:
Hi, is there any todo left to get this merged?

jichenjc:
/approve

good enough time for people to comment if they have any objection

PR #2552, opened 2024-02-16
[occm] Allow changing cluster-name on existing deployments

### LINKED ISSUE #2241: [openstack-cloud-controller-manager] Shared tenancy OpenStack project service type LoadBalancer collision
/kind bug

**What happened**:
Two Kubernetes clusters in the same project were created. The following steps  were taken along with their timeline.
- Two different Kubernetes clusters were created using RKE2 in the same OpenStack project.
- Both Kubernetes clusters came online and were fully functional.
- Kubernetes cluster A had the `default:kubernetes` service changed from `ClusterIP` to `LoadBalancer`.
- A load balancer was created and the Kubernetes service was made public successfully.
- Kubernetes cluster B had the `default:kubernetes` service changed from `ClusterIP` to `LoadBalancer`.
- OCCM identified the first load balancer in Octavia as it's own because the name was `kube_service_kubernetes_default_kubernetes` and rebuilt it to point to Kubernetes cluster B.
- Kubernetes cluster A is now desynchronized and has a LoadBalancer entry which no longer exists.

Ideally, I would like to see the load balancer be created with a random suffix in OpenStack (IE: kube_service_kubernetes_default_kubernetes_vxwmx). We use suffixes in other places in Kubernetes, makes sense to use them here too. Further, if a load balancer collision were to exist, instead of deleting/recreating a load balancer of the same name, perhaps an error event should be fired instead if the suffix option is unavailable?

Also, if the occm is going to delete an existing load balancer, there should be a more obvious log entry for this. Currently it just transitions from `ACTIVE` to `PENDING_CREATE` in the logs. It's very ambiguous and should have more detailed output.

**How to reproduce it**:
- Create two Kubernetes clusters in OpenStack in the same project, change the `default:kubernetes` service on each one from ClusterIP to LoadBalancer with a 5 minute gap between the first cluster being fully provisioned.
- Check the Octavia API, only one load balancer will exist pointed to the second cluster.

- Kubernetes cluster A and Kubernetes cluster B have two different Octavia UUIDs, so the LoadBalancer was deleted/created, not simply modified.
- Container image: docker.io/k8scloudprovider/openstack-cloud-controller-manager:latest

**Environment**:
- openstack-cloud-controller-manager version: latest
- OpenStack version: Zed
- Others:

dulek:
It's a common issue that clusters are deployed with the default `--cluster-name=kubernetes` and later on it's discovered that next deployments on the same cloud will have conflicts when trying to manage LBs of the same namespace and name.

This commit aims at allowing to change the cluster-name on a running environment and handling all the renames of the LB resources and their tags.

fixes #2241

**Release note**:
```release-note
It's now allowed to change the `--cluster-name` on existing deployments and OCCM will handle all the LB renames.
```

### DIFF
--- a/pkg/openstack/loadbalancer_rename.go
+++ b/pkg/openstack/loadbalancer_rename.go
@@ -1,0 +1,149 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package openstack
+
+import (
+	"fmt"
+	"k8s.io/cloud-provider-openstack/pkg/util"
+	"regexp"
+	"strings"
+
+	"github.com/gophercloud/gophercloud"
+
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/listeners"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/loadbalancers"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/monitors"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/pools"
+	openstackutil "k8s.io/cloud-provider-openstack/pkg/util/openstack"
+)
+
+// lbHasOldClusterName checks if the OCCM LB prefix is present and if so, validates the cluster-name
+// component value. Returns true if the cluster-name component of the loadbalancer's name doesn't match
+// clusterName.
+func lbHasOldClusterName(loadbalancer *loadbalancers.LoadBalancer, clusterName string) bool {
+	if !strings.HasPrefix(loadbalancer.Name, servicePrefix) {
+		// This one was probably not created by OCCM, let's leave it as is.
+		return false
+	}
+	existingClusterName := getClusterName("", loadbalancer.Name)
+
+	return existingClusterName != clusterName
+}
+
+// decomposeLBName returns clusterName based on object name
+func getClusterName(resourcePrefix, objectName string) string {
+	// This is not 100% bulletproof when string was cut because full name would exceed 255 characters, but honestly
+	// it's highly unlikely, because it would mean cluster name, namespace and name would together need to exceed 200
+	// characters. As a precaution the _<name> part is treated as optional in the regexp, assuming the longest trim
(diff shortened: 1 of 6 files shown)


dulek:
This is a WiP, definitely needs unit tests and some more manual testing from me.

jichenjc:
>>>It's a common issue that clusters are deployed with the default --cluster-name=kubernetes and later on it's discovered that next deployments on the same cloud will have conflicts when trying to manage LBs of the same namespace and name.

I remember it's been discussed multiple times before (with --cluster-name=kubernetes) 
at that time seems the conclusion is related to cloud provider itself or I might remember wrong thing
but I think it's worthy to open an issue to have some discussion there to get agreement on the goal and fix proposal..

dulek:
@jichenjc: I dug out #2241 and added it to this PR. Citing @zetaab from there:

> the problem is that quite many cluster still uses default clusterName which is kubernetes. It is also difficult to migrate away from it.

This is what the PR proposes - to make it easier to migrate away from a wrong `clusterName`.

dulek:
Hey, @stephenfin, could you take a preliminary look?

dulek:
And you @gryf.

gryf:
Looks good from my side as well. Good job, @dulek!

pkg/openstack/loadbalancer_rename.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/loadbalancers"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/monitors"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/pools"
+	openstackutil "k8s.io/cloud-provider-openstack/pkg/util/openstack"
+)
+
+// lbHasOldClusterName checks if the OCCM LB prefix is present and if so, validates the cluster-name
+// component value. Returns true if the cluster-name component of the loadbalancer's name doesn't match
+// clusterName.
+func lbHasOldClusterName(loadbalancer *loadbalancers.LoadBalancer, clusterName string) bool {
```
zetaab, line 37: so basically this will think that all "kubernetes" loadbalancers are from old clustername? This is quite risky if someone have now two kubernetes clusters named "kubernetes" in same openstack project. However, that kind of setup is already problematic but I think someone is doing that

dulek: I think this approach will still be safe when there are multiple clusters using default cluster name, assuming that there were no conflicting LBs, which would be not functional anyway. Note that `renameLoadBalancer()` is only called from `ensureLoadBalancer()` [1] meaning that it's not looking up all LBs the tenant sees in OpenStack, but rather reacts on Services that were supposed to have LBs, so renames should be constrained to a single K8s cluster.

[1] https://github.com/kubernetes/cloud-provider-openstack/pull/2552/files#diff-89af30f72ae5c8aefb464330d654b95dc41544d558d9c04c9e728a8ea9a94793R1617

zetaab (approved):

stephenfin:
/lgtm

...is what I meant :sweat_smile:

jichenjc:
/approve

I think @zetaab you approved this :) 
will read this later and input if I have more time

PR #2557, opened 2024-02-21
[openstack-cloud-controller-manager] Remove enforcement of IPv6 LB as internal

MaysaMacedo:

In OpenStack IPv6 that uses GUAs don't require NAT
to access the outside world, so IPv6 can be
rechable without Floating IPs, which makes the
enforcement of IPv6 LB as internal in CPO not necessary.
This commit removes this enforcement, which results in
IPv6 load-balancers being allowed to be shared between
Services. Also, now it's possible to make the load-balancer
use the IPv6 stateful address defined in the loadBalancerIP
of the Service.

fixes #

**Release note**:

```release-note
It's now possible to share the load balancer for IPv6 Services and to use IPv6 addresses pre-defined in the `loadBalancerIP`. When using `loadBalancerIP` with IPv6, only stateful addresses are allowed.
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -259,8 +259,10 @@ func (lbaas *LbaasV2) createOctaviaLoadBalancer(name, clusterName string, servic
 
 	// For external load balancer, the LoadBalancerIP is a public IP address.
 	loadBalancerIP := service.Spec.LoadBalancerIP
-	if loadBalancerIP != "" && svcConf.internal {
-		createOpts.VipAddress = loadBalancerIP
+	if loadBalancerIP != "" {
+		if svcConf.internal || (svcConf.preferredIPFamily == corev1.IPv6Protocol) {
+			createOpts.VipAddress = loadBalancerIP
+		}
 	}
 
 	if !lbaas.opts.ProviderRequiresSerialAPICalls {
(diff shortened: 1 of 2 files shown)


pkg/openstack/loadbalancer.go
```diff
@@ -1711,9 +1708,13 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 		}
 	}
 
-	addr, err := lbaas.ensureFloatingIP(clusterName, service, loadbalancer, svcConf, isLBOwner)
-	if err != nil {
-		return nil, err
+	addr := loadbalancer.VipAddress
```
dulek, line 1711: I'm a bit uneasy of using `VipAddress` here because of [1]. I think what I'd do is to make sure we're explicitly logging and maybe adding an event about skipping FIP creation.

I also think that code fragment [1] needs a change to make sure we're supporting LoadBalancerIP enforcement for IPv6.

[1] https://github.com/kubernetes/cloud-provider-openstack/blob/8245f5385c4845f54df8bf95f8caec571aa7fc5d/pkg/openstack/loadbalancer.go#L260-L264

dulek (changes requested):
One comment inline. I also thing [1] should be updated to not lookup FIP network for IPv6 services.

[1] https://github.com/kubernetes/cloud-provider-openstack/blob/8245f5385c4845f54df8bf95f8caec571aa7fc5d/pkg/openstack/loadbalancer.go#L1397

pkg/openstack/loadbalancer.go
MaysaMacedo: I just hit an internal server error when using a IPv6 from the nodes subnet to as a loadbalancerIP. The subnet was configured with slaac. I will try with status to see if there is any difference

MaysaMacedo: done. Note that this will only work with addresses that were generated from a stateful subnet and not slaac.
Otherwise we would hit the following issue: `cannot be directly assigned to a port on subnet a215f064-5fe6-4a15-856a-a7e70912e510 as the subnet is configured for automatic addresses`

MaysaMacedo:
@dulek this is ready for another look, whenever possible.

dulek (approved):
/approve

PR #2559, opened 2024-02-29
Cut 1.28.2

### LINKED ISSUE #2531: [barbican-kms-plugin] Container broken in v1.28.1
/kind bug

The `barbican-kms-plugin:v1.28.1` does not start as the inner command does not recognize the flags correctly. This is due to the fact that flags are parsed before they are declared in the `main.go` code. This behaviour has been solved by PR #2325  , but the fix has not been back-ported to v1.28.1.

dulek:
This is to include fixes to #2531 in a release.

fixes #2531

```release-note
NONE
```

### DIFF
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: v1.28.1
+appVersion: v1.28.2
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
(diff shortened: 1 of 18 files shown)


dulek:
/retest

Looks like we just suddenly lost a VM, shouldn't be caused by this patch.

dulek:
@kayrus, @jichenjc, @zetaab: This is ready for a review.

kayrus (approved):
LGTM, thanks!
Do you plan to make 1.27 release as well?

PR #2561, opened 2024-03-05
[barbican-kms-plugin] Update KMS API to v2

dulek:
KMS v1 API is deprecated and in v1.29 core K8s won't allow it with default feature gates set. This commit makes sure we're proposing configuration of v2 API in example EncryptionConfigs.

fixes #

**Release note**:
```release-note
KMS EncryptionConfig examples are updated to configure barbican-kms-plugin with v2 API.
```


GlassOfWhiskey:
I cannot approve PRs, but I think this is a very good idea :)
KMSv2 is way better than v1

PR #2565, opened 2024-03-28
[docs] Update the release procedure

dulek:
Seems like the point explaining how to promote images from staging is gone, this commit adds it back, fixes formatting and moves the file into the `docs` directory.

fixes #

**Release note**:

```release-note
NONE
```


docs/release-procedure.md
```diff
@@ -0,0 +1,44 @@
+# Release Procedure
```
jichenjc, line 1: I didn't have chance to try this procedure , @dulek I assume you will try this or already tried?

dulek: Yes, I made releases a few times. That's why I'm adding back the point about updating `images.yaml` in `kubernetes/k8s.io`, because that one was have been removed by mistake.

PR #2566, opened 2024-03-28
[release-1.29] Remove enforcement of IPv6 LB as internal (#2557)

MaysaMacedo:
In OpenStack IPv6 that uses GUAs don't require NAT to access the outside world, so IPv6 can be
rechable without Floating IPs, which makes the
enforcement of IPv6 LB as internal in CPO not necessary. This commit removes this enforcement, which results in IPv6 load-balancers being allowed to be shared between Services. Also, now it's possible to make the load-balancer use the IPv6 stateful address defined in the loadBalancerIP of the Service.

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/events.go
+++ b/pkg/openstack/events.go
@@ -21,4 +21,5 @@ const (
 	eventLBExternalNetworkSearchFailed = "LoadBalancerExternalNetworkSearchFailed"
 	eventLBSourceRangesIgnored         = "LoadBalancerSourceRangesIgnored"
 	eventLBAZIgnored                   = "LoadBalancerAvailabilityZonesIgnored"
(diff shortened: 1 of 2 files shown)


MaysaMacedo:
This is a cherry-pick of https://github.com/kubernetes/cloud-provider-openstack/pull/2557 to release-1.29

dulek:
/retest

This looks irrelevant.

dulek:
/approve

Looks like a clear cherry-pick, you probably could have used the bot and `/cherry-pick release-1.29`.

dulek:
Okay, I think #2567 is needed first.

PR #2572, opened 2024-04-10
[occm] Set `--use-service-account-credentials=false`

### LINKED ISSUE #2560: service account cloud-controller-manager cannot patch service object

>
> /kind bug
> /kind feature

**What happened**:
create a LB follow : https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
```yaml
kind: Service
apiVersion: v1
metadata:
  name: external-http-nginx-service
  annotations:
    service.beta.kubernetes.io/openstack-internal-load-balancer: "false"
    loadbalancer.openstack.org/floating-network-id: "7d304522-1456-4aef-b953-c9775e3064f6"
    loadbalancer.openstack.org/floating-subnet-id: "346c6134-c0fd-46ee-9e15-7ddcc9ce18d6"
spec:
  selector:
    app: nginx
  type: LoadBalancer
  ports:
  - name: http
    port: 80
    targetPort: 80
```

log 

```bash
  Warning  SyncLoadBalancerFailed  15s (x4 over 52s)    service-controller  Error syncing load balancer: failed to ensure load balancer: failed to patch service object default/external-http-nginx-service: services "external-http-nginx-service" is forbidden: User "system:serviceaccount:kube-system:cloud-controller-manager" cannot patch resource "services" in API group "" in the namespace "default"
```

Install OCCM using Helm.

```bash
$ helm install openstack-ccm cpo/openstack-cloud-controller-manager --version  2.29.1 -n cpo --create-namespace --values openstack-ccm.yaml
```

**How to reproduce it**:

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: chart version is 2.29.1
- OpenStack version:
- Others:

dulek:
The above option seems to be causing CCM to create clients using ServiceAccount from the `kube-system` namespace, so requires users to either run in `kube-system` namespace, or manage 2 ServiceAccounts, one in `kube-system` and other in regular CCM namespace. See [1].

This commit changes this setting.

/hold

[1] https://github.com/kubernetes/cloud-provider/blob/c3862938334ba18226098015193374fda40ab7a9/options/options.go#L230-L237

fixes #2560

**Release note**:
```release-note
OCCM is by default no longer run with `--use-service-account-credentials=true`, meaning that it will use the ServiceAccount specified in the DaemonSet. This means that you can stop managing the additional ServiceAccount in the `kube-system` namespace.
```

### DIFF
--- a/charts/openstack-cloud-controller-manager/Chart.yaml
+++ b/charts/openstack-cloud-controller-manager/Chart.yaml
@@ -4,7 +4,7 @@ description: Openstack Cloud Controller Manager Helm Chart
 icon: https://object-storage-ca-ymq-1.vexxhost.net/swift/v1/6e4619c416ff4bd19e1c087f27a43eea/www-images-prod/openstack-logo/OpenStack-Logo-Vertical.png
 home: https://github.com/kubernetes/cloud-provider-openstack
 name: openstack-cloud-controller-manager
-version: 2.29.1
+version: 2.29.2
 maintainers:
   - name: eumel8
     email: f.kloeker@telekom.de
--- a/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
@@ -54,7 +54,7 @@ spec:
             - --cloud-config=$(CLOUD_CONFIG)
             - --cluster-name=$(CLUSTER_NAME)
             - --cloud-provider=openstack
-            - --use-service-account-credentials=true
+            - --use-service-account-credentials=false
             - --controllers={{- trimAll "," (include "occm.enabledControllers" . ) -}}
             {{- if .Values.serviceMonitor.enabled }}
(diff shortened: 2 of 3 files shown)


dulek:
I still have to validate that this is the culprit, but there's a high chance it is. We've copied AWS manifest, blindly copying this option, but AWS manifest seems to override it in a patch: https://github.com/kubernetes/cloud-provider-aws/blob/234a39835a0ef28eea15424f83e9cb96be37d386/examples/existing-cluster/overlays/superset-role/aws-cloud-controller-manager-daemonset-patch.yaml#L16

dulek:
/hold cancel

I just validated that this makes sense for OpenShift.

dulek:
/hold cancel

Okay, this might be the way to go upstream here.

zetaab (approved):
/approve

PR #2577, opened 2024-04-18
[occm] add tag on floating ip create

lidongshengxdayu:

In some scenarios, tags are used as resource tags, but occm creates a floating ip without a tag. This PR adds the same tag as the LB when the floating ip created.

```release-note
Add tag same as LB for floating ip create
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -31,6 +31,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/loadbalancers"
 	v2monitors "github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/monitors"
 	v2pools "github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/pools"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/attributestags"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
 	corev1 "k8s.io/api/core/v1"
@@ -563,7 +564,7 @@ func (lbaas *LbaasV2) deleteOctaviaListeners(lbID string, listenerList []listene
 	return nil
 }
 
-func (lbaas *LbaasV2) createFloatingIP(msg string, floatIPOpts floatingips.CreateOpts) (*floatingips.FloatingIP, error) {
+func (lbaas *LbaasV2) createFloatingIP(msg string, floatIPOpts floatingips.CreateOpts, tag *string) (*floatingips.FloatingIP, error) {
 	klog.V(4).Infof("%s floating ip with opts %+v", msg, floatIPOpts)
 	mc := metrics.NewMetricContext("floating_ip", "create")
 	floatIP, err := floatingips.Create(lbaas.network, floatIPOpts).Extract()
@@ -571,6 +572,16 @@ func (lbaas *LbaasV2) createFloatingIP(msg string, floatIPOpts floatingips.Creat
 	if mc.ObserveRequest(err) != nil {
 		return floatIP, fmt.Errorf("error creating LB floatingip: %v", err)
 	}
+
+	if tag != nil {
+		err = attributestags.Add(lbaas.network, "floatingips", floatIP.ID, *tag).ExtractErr()
+		err = PreserveGopherError(err)
+		if mc.ObserveRequest(err) != nil {
+			floatingips.Delete(lbaas.network, floatIP.ID)
+			return floatIP, fmt.Errorf("error update LB floatingip tag: %v", err)
+		}
+	}
+
 	return floatIP, err
 }
 
@@ -704,7 +715,7 @@ func (lbaas *LbaasV2) ensureFloatingIP(clusterName string, service *corev1.Servi
 					svcConf.lbPublicSubnetSpec, svcConf.lbPublicNetworkID)
 				for _, subnet := range foundSubnets {
 					floatIPOpts.SubnetID = subnet.ID
-					floatIP, err = lbaas.createFloatingIP(fmt.Sprintf("Trying subnet %s for creating", subnet.Name), floatIPOpts)
+					floatIP, err = lbaas.createFloatingIP(fmt.Sprintf("Trying subnet %s for creating", subnet.Name), floatIPOpts, &svcConf.lbName)
 					if err == nil {
 						foundSubnet = subnet
 						break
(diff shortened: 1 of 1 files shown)


pkg/openstack/loadbalancer.go
```diff
@@ -563,14 +564,24 @@ func (lbaas *LbaasV2) deleteOctaviaListeners(lbID string, listenerList []listene
...
 	err = PreserveGopherError(err)
 	if mc.ObserveRequest(err) != nil {
 		return floatIP, fmt.Errorf("error creating LB floatingip: %v", err)
 	}
+
+	if tag != nil {
+		err = attributestags.Add(lbaas.network, "floatingips", floatIP.ID, *tag).ExtractErr()
+		err = PreserveGopherError(err)
+		if mc.ObserveRequest(err) != nil {
+			floatingips.Delete(lbaas.network, floatIP.ID)
```
dulek, line 580: I'm not really sure a failure to add a tag should be fatal and cause a full rollback. This most likely won't work either. I'd rather have the code follow K8s reconciling principles and make it fetch the FIP, check if tag exists and if not, add it. This would also handle the upgrade situations, so that after upgrade all FIPs without tags would get one.

dulek (changes requested):

dulek:
This looks like a sane request to be able to tag the FIPs created by CCM. This needs to handle one more case though:

`loadbalancer.openstack.org/keep-floatingip` annotation will force CCM to not delete the FIP. In [such](https://github.com/kubernetes/cloud-provider-openstack/blob/290e7c72cb689141ab7e3eeae7edacb1b993c818/pkg/openstack/loadbalancer.go#L2051-L2067) [two](https://github.com/kubernetes/cloud-provider-openstack/blob/290e7c72cb689141ab7e3eeae7edacb1b993c818/pkg/openstack/loadbalancer.go#L626-L632) cases we should untag the FIP.

While you're on this, be informed that a bunch of other folks are working on resource tagging, @stephenfin and @dd-georgiev.

pkg/openstack/loadbalancer.go
qinghon: sure ，this is indeed not very clean. Now it is modified to check the tag when confirming the floating ip. If it does not exist, it will be added to the floating ip.

dulek (changes requested):
This is still not untagging FIPs we don't delete properly.

lidongshengxdayu:
> This is still not untagging FIPs we don't delete properly.

it mean that you need to untag it when  cancel the association?

dulek:
See this comment for details: https://github.com/kubernetes/cloud-provider-openstack/pull/2577#issuecomment-2075394150

pkg/openstack/loadbalancer.go
```diff
@@ -603,6 +604,42 @@ func (lbaas *LbaasV2) updateFloatingIP(floatingip *floatingips.FloatingIP, portI
 	return floatingip, nil
 }
 
+func (lbaas *LbaasV2) updateFloatingIPTag(floatingip *floatingips.FloatingIP, Tag string, delete bool) error {
```
kayrus, line 607: Can you rebase on master and pass ctx as the first argument in this method?

pkg/openstack/loadbalancer.go
```diff
@@ -663,6 +697,7 @@ func (lbaas *LbaasV2) ensureFloatingIP(ctx context.Context, clusterName string,
 				if err != nil {
 					return "", err
 				}
+				_ = lbaas.updateFloatingIPTag(ctx, floatIP, svcConf.lbName, true)
```
kayrus, line 700: I wonder why do you skip `err`?

pkg/openstack/loadbalancer.go
```diff
@@ -618,6 +619,39 @@ func (lbaas *LbaasV2) updateFloatingIP(ctx context.Context, floatingip *floating
 	return floatingip, nil
 }
 
+func (lbaas *LbaasV2) updateFloatingIPTag(ctx context.Context, floatingip *floatingips.FloatingIP, Tag string, delete bool) error {
```
kayrus, line 622: `delete` is a reserved golang keyword, I'd rather use `del`, or even better, use `tag` as a pointer. if the pointer is `nil`, then delete

```suggestion
func (lbaas *LbaasV2) updateFloatingIPTag(ctx context.Context, fip *floatingips.FloatingIP, tag *string) error {
```

pkg/openstack/loadbalancer.go
```diff
@@ -618,6 +619,39 @@ func (lbaas *LbaasV2) updateFloatingIP(ctx context.Context, floatingip *floating
...
+	if Tag == "" {
+		return fmt.Errorf("Error input tag argument ")
+	}
+	tags, err := attributestags.List(ctx, lbaas.network, "floatingips", floatingip.ID).Extract()
+	if err != nil {
+		klog.V(3).Infof("Cannot get floatIP tags for floating %s", floatingip.ID)
```
kayrus, line 628:
```suggestion
		klog.V(2).Errorf("Cannot get floating IP tags for floating %s", fip.ID)
```

kayrus (changes requested):
regarding the conditions logic, I think it makes sense to update it to the pattern below:

```
        if found && !delete {
                return nil
        }
        
        if delete { 
                DELETE
        }
                
        ADD
```

UPD: not all openstack deployments have the `standard-attr-tag` extension enabled. Can you add a check for tags extensions? See https://github.com/kubernetes/cloud-provider-openstack/blob/733195a4ef8ee6e959f0414848d31ac508150af0/pkg/openstack/openstack.go#L466-L470 code for example. We can run this check once during the start and keep the bool value in the `LoadBalancer` struct

pkg/openstack/loadbalancer.go
lidongshengxdayu: Need to search and delete based on tags, and  cannot delete all the tags , pointer cannot be used,

pkg/openstack/loadbalancer.go
lidongshengxdayu: If you think this and the `add` below need to return err, tell me and I will add it.

kayrus:
@lidongshengxdayu have you considered to change the logic to the suggested pattern? And what about my `standard-attr-tag` comment? I don't see that the extension check was implemented.

lidongshengxdayu:
> @lidongshengxdayu have you considered to change the logic to the suggested pattern? And what about my `standard-attr-tag` comment? I don't see that the extension check was implemented.

sorry, eyes are drowning in code.
I've tried to add the `standard-attr-tag` feature check to the code, unfortunately..., it's been too long since the PR was proposed and I don't have the environment to do this test anymore, so I'm still adding extra code to avoid errors affecting the original logic

PR #2578, opened 2024-04-19
[cinder-csi-plugin] Enable optional storageclasses.

powellchristoph:
[cinder-csi-plugin] Enable Optional StorageClasses

If you enable the included storageclasses, you get both the `retain` and `delete`. This PR allows each class to be enabled individually.

na

na

**Release note**:
```release-note
[cinder-csi-plugin] Included storageclasses can be individually enabled.
```

### DIFF
--- a/charts/cinder-csi-plugin/values.yaml
+++ b/charts/cinder-csi-plugin/values.yaml
@@ -160,12 +160,13 @@ secret:
 #      ca-file=/etc/cacert/ca-bundle.crt
 
 storageClass:
+  enabled: true
   delete:
-    enabled: false
+    enabled: true
     isDefault: false
     allowVolumeExpansion: true
   retain:
-    enabled: false
+    enabled: true
     isDefault: false
     allowVolumeExpansion: true
 # any kind of custom StorageClasses
--- a/charts/cinder-csi-plugin/values.yaml
+++ b/charts/cinder-csi-plugin/values.yaml
@@ -160,11 +160,12 @@ secret:
 #      ca-file=/etc/cacert/ca-bundle.crt
 
 storageClass:
-  enabled: true
   delete:
+    enabled: false
     isDefault: false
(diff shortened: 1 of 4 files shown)


dulek:
@powellchristoph: Before jumping into the code - you need to retain the release note box from the PR template: https://github.com/kubernetes/cloud-provider-openstack/blob/66624e34b1884781dca75f2ae6dc9736c0de9f28/.github/PULL_REQUEST_TEMPLATE.md

charts/cinder-csi-plugin/values.yaml
```diff
@@ -160,11 +160,12 @@ secret:
 #      ca-file=/etc/cacert/ca-bundle.crt
 
 storageClass:
-  enabled: true
```
dulek, line 163: Okay, this needs a release note for sure to show that the options are moved. I think I'd also keep the global option here and treat individual options as ones overriding the global one.

dulek (changes requested):

dulek:
I'm looking into chart lint failure.

dulek:
See #2582 for chart linting failure explanation. It's not fault of this patch, we'll fix it soon.

/ok-to-test

dulek:
Ah, looks like you need to make EasyCLA happy. Just click "Details" in its test table row and follow the steps.

charts/cinder-csi-plugin/values.yaml
powellchristoph: I have updated the PR with the release notes and kept the default with an override.

powellchristoph:
> Ah, looks like you need to make EasyCLA happy. Just click "Details" in its test table row and follow the steps.

I've followed the link and filled it out twice now. Its sent me the email. I'm not sure what else I'm supposed to do.

dulek:
/approve

Can you squash the commits before the final `/lgtm`?

PR #2580, opened 2024-04-22
[occm] - add annotation for custom octavia listener tags

dd-georgiev:
Added support for new annotation which provides the ability to add tags when the load balancer is created. Those tags affect the load balancer and the listener. The rest of the resources are not tagged

https://github.com/kubernetes/cloud-provider-openstack/issues/2327

1. As far as  I understood(based on Slack conversation) this functionality will be expanded. That is more features will be added.
2. Testing notes - There is e2e test, but basically adding the annotation specified in the docs, must result in having one or multiple tags on the load balancer and listener resource in openstack.
3. No extra validation is done, as the openstack document doesn't mention any excluded chars(as far as I managed to understand). There is edge-case mentioned in the openstack docs, but I don't think they apply for the LB resources. Namely:
> Filtering resources with a tag whose name contains a comma is not supported. Thus, do not put such a tag name to resources.
> Source: https://docs.openstack.org/mitaka/networking-guide/ops-resource-tags.html#limitations

But I don't think this will impact us, since "," is used to split the tags. Hence tag containing "," cannot exist. I didn't document this behavior as it fells like common sense. Also the docs are coming from the Mitaka project. 
4. I'm not sure if this change is big enough for release notes, but I can add them if necessary 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
+++ b/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
@@ -236,6 +236,11 @@ Request Body:
   This annotation is automatically added and it contains the floating ip address of the load balancer service.
   When using `loadbalancer.openstack.org/hostname` annotation it is the only place to see the real address of the load balancer.
 
+- `loadbalancer.openstack.org/custom-tags`
+
+  Allows to specify custom tags for the *load balancer* and the *listeners* resources for that Service. *Pools* do not have any tags at the moment. Other than the custom tags, each resource has a tag in the format of "$NAMESPACE-$SERVICE_NAME'.
+  Tags are arbitrary strings, to specify multiple tags separate them using a comma , in the annotation.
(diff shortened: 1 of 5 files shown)


dd-georgiev:
I've converted this into draft.
Reason being I'm working on the extended requirement to support the ability to change the tags when the annotation is changed. This functionality will require changes to this code as well.

dd-georgiev:
Hello @m-kratochvil, 
The idea was as mentioned in the PR to expand the code to support more requirements. 
Again as mentioned in the PR:
> Those tags affect the load balancer and the listener. The rest of the resources are not tagged.

Unfortunately life happened and I'm unable to continue working on the extra implementation for the time being. 

You can find more information about the further changes in the following slack discussion: 
https://kubernetes.slack.com/archives/C0LSA3T7C/p1713791973852479 

I can reopen the PR if you find this useful. Some work related to tagging was done in other PR as well:
https://github.com/kubernetes/cloud-provider-openstack/pull/2577

For me, this PR brings value only if somebody is going to extend the work to meet the requirements imposed in the slack discussion, but I will leave this to your judgment.

PR #2586, opened 2024-04-26
fix(occm): Empty nodeSelector by default

### LINKED ISSUE #2550: [occm] The nodeSelector key is not empty in the helm chart in release v1.29.0
**This is a BUG REPORT **:

/kind bug

**What happened**:
Since the release v1.29.0, the _nodeSelector_ key in the file values.yaml of the helm chart is no longer empty and therefore requires positioning the following label `node-role.kubernetes.io/controlplane: ""` on the workers where the daemonSet must be deployed.

The _nodeSelector_ key in the file values.yaml of the helm chart should be empty.

`nodeSelector: {}`

**How to reproduce it**:
Install occm v1.29.0 with the helm chart.

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: v1.29.0

Let me know if you need a PR.
Regards.

babykart:
* keep the nodeSelector empty by default so users can easily use their own.
* add documentation in the README

fixes #2550

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/openstack-cloud-controller-manager/values.yaml
+++ b/charts/openstack-cloud-controller-manager/values.yaml
@@ -46,9 +46,10 @@ livenessProbe: {}
 # Set readinessProbe in the same way like livenessProbe
 readinessProbe: {}
 
-# Set nodeSelector where the controller should run, i.e. controlplane nodes
-nodeSelector:
-  node-role.kubernetes.io/control-plane: ""
+# Set nodeSelector where the controller should run.
+nodeSelector: {}
+  # i.e. controlplane nodes
+  # node-role.kubernetes.io/control-plane: "true"
 
 # Set tolerations for nodes where the controller should run, i.e. node
 # should be uninitialized, controlplane...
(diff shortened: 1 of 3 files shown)


charts/openstack-cloud-controller-manager/values.yaml
```diff
@@ -46,9 +46,10 @@ livenessProbe: {}
 # Set readinessProbe in the same way like livenessProbe
 readinessProbe: {}
 
-# Set nodeSelector where the controller should run, i.e. controlplane nodes
```
zetaab, line 49: you could still override the values using helm, no matter what is defined in this file

dulek:
@wwentland introduced this in https://github.com/kubernetes/cloud-provider-openstack/pull/2346. Why do we want to change the default again? It's possible to override this, as I indicated on the issue itself and I believe CPO is a control plane service anyway. We're also following cloud-provider-aws pattern with this: https://github.com/kubernetes/cloud-provider-aws/blob/master/charts/aws-cloud-controller-manager/values.yaml#L14-L16

This is changing a default behavior, so we can't merge this without a release note. But that being said I'm still opposed to this.

/hold for release note.

zetaab:
yeah actually @dulek had quite good point.

/approve cancel

PR #2584, opened 2024-04-25
Bump versions for 1.30

dulek:
**What this PR does / why we need it**:
Bumping versions for K8s 1.30.

**Release note**:
```release-note
NONE
```

PR #2587, opened 2024-04-28
[occm] KEP-1860: Add support for LoadBalancer ipMode

kbudde:
This PR adds support for ipMode in LoadBalancer status field introduced in KEP-1860.
This enables the usage of Proxy Protocol without workaround (EnableIngressHostname + IngressHostnameSuffix).

The workarounds were introducing new issues, e.g. externalDNS was not functional with a loadbalancer with ingress Hostname set.

There is no open issue as a workaround was introduced. 

Tthe ipMode field is available even if  IpMode is still behind a featureGate. 

**Release note**:

```release-note
[occm] KEP-1860: Add support for LoadBalancer ipMode
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1568,18 +1568,26 @@ func (lbaas *LbaasV2) createLoadBalancerStatus(service *corev1.Service, svcConf 
 		status.Ingress = []corev1.LoadBalancerIngress{{Hostname: hostname}}
 		return status
 	}
-	// If the load balancer is using the PROXY protocol, expose its IP address via
-	// the Hostname field to prevent kube-proxy from injecting an iptables bypass.
-	// This is a workaround until
-	// https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1860-kube-proxy-IP-node-binding
-	// is implemented (maybe in v1.22).
-	if svcConf.enableProxyProtocol && lbaas.opts.EnableIngressHostname {
-		fakeHostname := fmt.Sprintf("%s.%s", addr, lbaas.opts.IngressHostnameSuffix)
-		status.Ingress = []corev1.LoadBalancerIngress{{Hostname: fakeHostname}}
-		return status
+
+	ipMode := corev1.LoadBalancerIPModeVIP
+	if svcConf.enableProxyProtocol {
+		// If the load balancer is using the PROXY protocol, expose its IP address via
+		// the Hostname field to prevent kube-proxy from injecting an iptables bypass.
+		// Setting must be removed by the user to allow the use of the LoadBalancerIPModeProxy.
+		if lbaas.opts.EnableIngressHostname {
+			fakeHostname := fmt.Sprintf("%s.%s", addr, lbaas.opts.IngressHostnameSuffix)
+			status.Ingress = []corev1.LoadBalancerIngress{{Hostname: fakeHostname}}
+			return status
+		}
+		// Set the LoadBalancerIPMode to Proxy to prevent kube-proxy from injecting an iptables bypass.
+		// https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1860-kube-proxy-IP-node-binding
+		ipMode = corev1.LoadBalancerIPModeProxy
 	}
 	// Default to IP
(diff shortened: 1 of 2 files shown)


dulek:
@gryf, you might want to check this out.

zetaab (approved):
/approve

zetaab:
we are soonish releasing 1.30 (https://github.com/kubernetes/cloud-provider-openstack/pull/2584) but IMO this could be interesting feature to new release. 

cc @dulek @jichenjc

jichenjc:
/lgtm
/hold 

@dulek maybe you can check this again before merge

pkg/openstack/loadbalancer.go
```diff
@@ -1568,18 +1568,26 @@ func (lbaas *LbaasV2) createLoadBalancerStatus(service *corev1.Service, svcConf
...
+		// the Hostname field to prevent kube-proxy from injecting an iptables bypass.
+		// Setting must be removed by the user to allow the use of the LoadBalancerIPModeProxy.
+		if lbaas.opts.EnableIngressHostname {
+			fakeHostname := fmt.Sprintf("%s.%s", addr, lbaas.opts.IngressHostnameSuffix)
+			status.Ingress = []corev1.LoadBalancerIngress{{Hostname: fakeHostname}}
+			return status
+		}
+		// Set the LoadBalancerIPMode to Proxy to prevent kube-proxy from injecting an iptables bypass.
+		// https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1860-kube-proxy-IP-node-binding
+		ipMode = corev1.LoadBalancerIPModeProxy
```
dulek, line 1584: Okay, so reading [1] this seems pretty much correct, but reading [2] I would argue we should always set IPMode to Proxy, because following is true for both Amphora and OVN:
> Setting this to "Proxy" indicates that traffic is delivered to the node or pod with the destination set to the node's IP and node port or the pod's IP and port. Service implementations may use this information to adjust traffic routing.

What am I missing?

Also - doesn't this apply to other Amphora functions, like TLS termination or `XForwardedFor` header injection?

[1] https://kubernetes.io/blog/2023/12/18/kubernetes-1-29-feature-loadbalancer-ip-mode-alpha/
[2] https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#loadbalanceringress-v1-core

pkg/openstack/loadbalancer.go
```diff
@@ -1568,18 +1568,26 @@ func (lbaas *LbaasV2) createLoadBalancerStatus(service *corev1.Service, svcConf
...
 	}
 	// Default to IP
-	status.Ingress = []corev1.LoadBalancerIngress{{IP: addr}}
+	status.Ingress = []corev1.LoadBalancerIngress{{
+		IP:     addr,
+		IPMode: &ipMode,
```
dulek, line 1589: What happens when the feature gate is off and this is set? Is it a no-op? Should we have this behind a feature gate?

dulek:
I added a bunch of questions.

pkg/openstack/loadbalancer.go
zetaab: afaik 1 is how I understood this should work

kbudde: Same here. 1 is how it is supposed to work.

pkg/openstack/loadbalancer.go
kbudde: It's save, the field is just ignored if the feature is not enabed.

https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1860-kube-proxy-IP-node-binding#rollout-upgrade-and-rollback-planning
> Disable the feature flag ("downgrade")
    Edit kube-apiserver manifest and disable the feature flag
    Edit kube-proxy configmap, disable the feature and restart kube-proxy Pods
    Confirmed that both iptables rules are present, even if the ipMode field was still set as Proxy, confirming the feature is disabled. Both accesses are working

pkg/openstack/loadbalancer.go
kbudde: The ipMode field should only be set, if the loadbalancer IP is used and not an Hostname.
LoadBalancer with IngressHostname are not optimized in terms of traffic flow as kube-proxy would need to do a dns query before writing the firewall rules (and update it every time dns changes).

Issue #1924, opened 2022-06-16
[occm] Multi region support


sergelogvinov:
/kind feature

I am noticed that openstack-client supports multi-region config.  https://github.com/gophercloud/utils/blob/master/openstack/clientconfig/testing/clouds.yaml#L160-L171

What do you think to add the multi-region support in one OCCM, based on the config-file, after #1900 ?
If the config file has a regions tree, OCCM checks it at boot time and will watch the nodes in those regions.

sergelogvinov:
Yep. You are right. I forgot about LB. 
I spoke about node/node-lifecycle only. 

For LB, it can be done by service annotation property... 

So, we can introduce it in steps.
Unfortunately, I do not have regional openstack with LB. It will be hard to test.

sergelogvinov:
If the node uninitialized, we need to check all regions to fine the node.
If node has providerId, we can get the region name from it.

and if i right, we need to have ```go-client``` connection for each region.

mikejoh:
> If the node uninitialized, we need to check all regions to fine the node. If node has providerId, we can get the region name from it.
> 
> and if i right, we need to have `go-client` connection for each region.

Correct! One compute `ServiceClient` per region basically. We're not completely done with our implementation, it's still in PoC mode, but so far everything looks good. We made sure to implement the `k8s.io/cloud-provider` [`InstancesV2`](https://github.com/kubernetes/cloud-provider/blob/04905862e78d299130f3a7a3dac05d7b40357964/cloud.go#L199-L212) interface which is slimmer and don't require as many queries against the underlying cloud APIs compared to the `Instances` interface.

As a side note: I know that there's a todo on implementing the V2 interface in the OCCM, which I think we could help out with!

dulek:
> What about this feature ? This would clearly be needed in our use case where we have 3 Openstack standalone clusters and want to have kubernetes clusters stretched on top of them. This can be done but things get complicated since CCM can't handle more than 1 Openstack API endpoint and thus, this makes everything related to storage or LBs a hell.
> 
> Is this somewhere in the pipe or maybe there are some alternatives ? Any help would be gladly appreciated.

I don't think CPO was designed with that use case in mind and it might need a lot of work to get right. Happy to help with advice and reviews if you have some development resources to throw at this.

Interestingly - what's the use case to stretch K8s like that?

Hybrid512:
[...]

PR #2590, opened 2024-05-08
[occm] multizone race condition

sergelogvinov:
We cannot definitively determine whether a node exists within our zone without the provider ID string.

It affects if `OS_CCM_REGIONAL` is `true`
The node trying to join to the cluster. It has non ready status and uninitialized taint.
The `cloud-node` controller is attempting to initialize the node simultaneously with the `node-lifecycle` controller's attempt to delete the node, as the node is located in another region

So we need to skip nodes without ProviderID in `OS_CCM_REGIONAL` mode.
Azure does the same [az.IsNodeUnmanaged](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure_instances_v2.go#L37) -> [IsNodeUnmanagedByProviderID] (https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure_wrap.go#L86) - if ProviderID is not azure string (or empty) - set it as unmanaged node.

part of #1924

**Release note**:

`OS_CCM_REGIONAL` is alpha feature, we do not need to update release note.
Feature flag was introduced here #1909 

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/instancesv2.go
+++ b/pkg/openstack/instancesv2.go
@@ -79,9 +79,20 @@ func (os *OpenStack) instancesv2() (*InstancesV2, bool) {
 
 // InstanceExists indicates whether a given node exists according to the cloud provider
 func (i *InstancesV2) InstanceExists(ctx context.Context, node *v1.Node) (bool, error) {
+	if i.regionProviderID {
+		if node.Spec.ProviderID == "" {
+			return false, fmt.Errorf("node %s has empty ProviderID, it should be initialized first", node.Name)
+		}
+
+		if isNodeUnmanaged(node.Spec.ProviderID) {
+			klog.V(4).InfoS("Omitting unmanaged node", "node", klog.KObj(node))
+			return true, nil
+		}
+	}
+
 	_, err := i.getInstance(ctx, node)
 	if err == cloudprovider.InstanceNotFound {
-		klog.V(6).Infof("instance not found for node: %s", node.Name)
+		klog.V(6).InfoS("Node is not found in cloud provider", "node", klog.KObj(node))
 		return false, nil
 	}
 
@@ -94,6 +105,17 @@ func (i *InstancesV2) InstanceExists(ctx context.Context, node *v1.Node) (bool, 
 
 // InstanceShutdown returns true if the instance is shutdown according to the cloud provider.
 func (i *InstancesV2) InstanceShutdown(ctx context.Context, node *v1.Node) (bool, error) {
+	if i.regionProviderID {
+		if node.Spec.ProviderID == "" {
+			return false, fmt.Errorf("node %s has empty ProviderID, it should be initialized first", node.Name)
+		}
+
+		if isNodeUnmanaged(node.Spec.ProviderID) {
+			klog.V(4).InfoS("Omitting unmanaged node", "node", klog.KObj(node))
+			return false, nil
+		}
+	}
+
 	server, err := i.getInstance(ctx, node)
 	if err != nil {
 		return false, err
@@ -101,6 +123,7 @@ func (i *InstancesV2) InstanceShutdown(ctx context.Context, node *v1.Node) (bool
 
 	// SHUTOFF is the only state where we can detach volumes immediately
 	if server.Status == instanceShutoff {
+		klog.V(6).InfoS("Node is shutoff", "node", klog.KObj(node))
 		return true, nil
 	}
 
--- a/pkg/openstack/utils.go
+++ b/pkg/openstack/utils.go
@@ -1,0 +1,53 @@
+/*
+Copyright 2014 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package openstack
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+// If Instances.InstanceID or cloudprovider.GetInstanceProviderID is changed, the regexp should be changed too.
+var providerIDRegexp = regexp.MustCompile(`^` + ProviderName + `://([^/]*)/([^/]+)$`)
+
+// instanceIDFromProviderID splits a provider's id and return instanceID.
+// A providerID is build out of '${ProviderName}:///${instance-id}' which contains ':///'.
+// or '${ProviderName}://${region}/${instance-id}' which contains '://'.
+// See cloudprovider.GetInstanceProviderID and Instances.InstanceID.
+func instanceIDFromProviderID(providerID string) (instanceID string, region string, err error) {
+
+	// https://github.com/kubernetes/kubernetes/issues/85731
+	if providerID != "" && !strings.Contains(providerID, "://") {
+		providerID = ProviderName + "://" + providerID
+	}
+
+	matches := providerIDRegexp.FindStringSubmatch(providerID)
+	if len(matches) != 3 {
+		return "", "", fmt.Errorf("ProviderID \"%s\" didn't match expected format \"openstack://region/InstanceID\"", providerID)
+	}
+	return matches[2], matches[1], nil
+}
+
+// Return true if instance is not openstack.
+func isNodeUnmanaged(providerID string) bool {
+	if providerID != "" {
+		return strings.Contains(providerID, "://") && !strings.HasPrefix(providerID, ProviderName+"://")
+	}
+
+	return false
+}
--- a/pkg/openstack/instances.go
+++ b/pkg/openstack/instances.go
@@ -448,27 +448,6 @@ func isValidLabelValue(v string) bool {
 	return true
 }
 
-// If Instances.InstanceID or cloudprovider.GetInstanceProviderID is changed, the regexp should be changed too.
-var providerIDRegexp = regexp.MustCompile(`^` + ProviderName + `://([^/]*)/([^/]+)$`)
-
-// instanceIDFromProviderID splits a provider's id and return instanceID.
-// A providerID is build out of '${ProviderName}:///${instance-id}' which contains ':///'.
-// or '${ProviderName}://${region}/${instance-id}' which contains '://'.
-// See cloudprovider.GetInstanceProviderID and Instances.InstanceID.
-func instanceIDFromProviderID(providerID string) (instanceID string, region string, err error) {
-
-	// https://github.com/kubernetes/kubernetes/issues/85731
-	if providerID != "" && !strings.Contains(providerID, "://") {
-		providerID = ProviderName + "://" + providerID
-	}
-
-	matches := providerIDRegexp.FindStringSubmatch(providerID)
-	if len(matches) != 3 {
-		return "", "", fmt.Errorf("ProviderID \"%s\" didn't match expected format \"openstack://region/InstanceID\"", providerID)
-	}
-	return matches[2], matches[1], nil
-}
-
 // AddToNodeAddresses appends the NodeAddresses to the passed-by-pointer slice,
 // only if they do not already exist
 func AddToNodeAddresses(addresses *[]v1.NodeAddress, addAddresses ...v1.NodeAddress) {
--- a/pkg/openstack/instances_test.go
+++ b/pkg/openstack/instances_test.go
@@ -22,7 +22,6 @@ import (
 	"reflect"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
 	v1 "k8s.io/api/core/v1"
 )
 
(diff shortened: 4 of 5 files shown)


pkg/openstack/instancesv2.go
```diff
@@ -79,6 +79,18 @@ func (os *OpenStack) instancesv2() (*InstancesV2, bool) {
 
 // InstanceExists indicates whether a given node exists according to the cloud provider
 func (i *InstancesV2) InstanceExists(ctx context.Context, node *v1.Node) (bool, error) {
+	if i.regionProviderID {
+		if node.Spec.ProviderID == "" {
+			klog.V(4).Infof("Instance %s should initialized first", node.Name)
+			return true, nil
```
mdbooth, line 85: I haven't checked the call-sites, but my initial thought is that this should return an error rather than true. It feels like we are potentially asserting that the instance exists when in fact we don't know.

pkg/openstack/instancesv2.go
```diff
@@ -79,6 +79,18 @@ func (os *OpenStack) instancesv2() (*InstancesV2, bool) {
 
 // InstanceExists indicates whether a given node exists according to the cloud provider
 func (i *InstancesV2) InstanceExists(ctx context.Context, node *v1.Node) (bool, error) {
+	if i.regionProviderID {
+		if node.Spec.ProviderID == "" {
+			klog.V(4).Infof("Instance %s should initialized first", node.Name)
```
mdbooth, line 84: nit:
```suggestion
			klog.V(4).Infof("Node %s should be initialized first", node.Name)
```

Incidentally, it looks like you've started moving to [structured logging](https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/1602-structured-logging/README.md) in at least the loadbalancer code.

This might be written as a structured log instead:
```golang
klog.V(4).InfoS("Node is not yet initialised", "node", klog.KObj(node))
```

pkg/openstack/utils.go
```diff
@@ -0,0 +1,53 @@
...
+	if len(matches) != 3 {
+		return "", "", fmt.Errorf("ProviderID \"%s\" didn't match expected format \"openstack://region/InstanceID\"", providerID)
+	}
+	return matches[2], matches[1], nil
+}
+
+// Return true if instance is not openstack.
+func instanceNodeUnmanaged(providerID string) bool {
+	if providerID != "" {
+		return strings.Contains(providerID, "://") && !strings.HasPrefix(providerID, ProviderName+"://")
```
mdbooth, line 49: What does the first assertion (`strings.Contains(providerID, "://")`)buy us?

pkg/openstack/instancesv2.go
```diff
@@ -79,6 +79,18 @@ func (os *OpenStack) instancesv2() (*InstancesV2, bool) {
...
+			return true, nil
+		}
+
+		if instanceNodeUnmanaged(node.Spec.ProviderID) {
+			klog.V(4).Infof("Instance %s is not an OpenStack instance", node.Name)
+			return true, nil
```
mdbooth, line 90: Again, is `true` the correct return value here? I'd have to check the call-sites, but my feeling is that this could lead to incorrect behaviour possibly today, but if not in the future. My feeling is that we should return an error here.

pkg/openstack/instancesv2.go
```diff
@@ -94,6 +106,18 @@ func (i *InstancesV2) InstanceExists(ctx context.Context, node *v1.Node) (bool,
...
+	if i.regionProviderID {
+		if node.Spec.ProviderID == "" {
+			klog.V(4).Infof("Instance %s should initialized first", node.Name)
+			return false, nil
+		}
+
+		if instanceNodeUnmanaged(node.Spec.ProviderID) {
+			klog.V(4).Infof("Instance %s is not an OpenStack instance", node.Name)
+			return false, nil
+		}
```
mdbooth, line 118: Same comments as above re structured logging and returning a non-error result.

pkg/openstack/instancesv2.go
sergelogvinov: cloud-provider does nothing if we return positive (it's alive and exist) response. Otherwise it delete/taint the node. The CCM in another region can make right decision. 

Than ProviderID is empty - yep, better return an error. 
If ProviderID is not openstack magic string - better do nothing. Errors will create noisy logs. 
AzureCCM does the same.

pkg/openstack/utils.go
sergelogvinov: `instanceIDFromProviderID` adds `openstack://` if providerID has string like '/instance-id'.
We are checking whether the providerID is in the old style or not. If yes - node is ours (backward compatibility)

mdbooth:
I feel like the underlying problem here is that `getInstance` doesn't support multiple regions when providerID is not set: https://github.com/sergelogvinov/cloud-provider-openstack/blob/3c97ceb61c74d4a58716b1d3109169d05451ea67/pkg/openstack/instancesv2.go#L153-L175

Would it be safer and less prone to hard-to-fix future edge cases to address that instead?

sergelogvinov:
My changes based on these thoughts:

If we decide to add a new region into the existing cluster, we'll need to redeploy the CCM with the `OS_CCM_REGIONAL` flag. Subsequently, all new nodes will be assigned a region name in their ProviderID string.

CCM will use default region (regionA) for nodes lacking a region in their ProviderID. 
Therefore, two instances of CCM, each configured with different region names, will operate independently.
CCM-regionA will locate instances by ProviderID regardless of region presence, while CCM-regionB will only recognize nodes with a region name in their ProviderID (as all nodes in regionB will have this).

mdbooth:
Got it. However, this would still be baking logic into the CCM based on a particular deployment strategy. I think it would still address your use case if we, e.g. updated getInstance to be multi-region aware in the event that providerID is not set. This would also mean that we would not be asserting that an instance exists when we don't know that it does. I can't help but feel that the fact this this happens not to break anything currently is not something we should rely on indefinitely.

mdbooth:
IIUC, the situation you're describing is one where we have multiple CCMs: one per region. The motivation for doing this is...

OpenStack CCM is not multi-region aware? Is there any other reason?

My initial thought was that it would be required for HA, but as long as your control plane is multi-region, a failure of a complete region where the CCM is running would result in the CCM running in another region.

It's not to reduce east-west traffic, either, because the CCM is only talking to KAS, and etcd data is presumably replicated across your control plane anyway.

That isn't quite where I thought I was going with this comment, btw. I just thought I'd state the motivation for wanting to do this first, only to realise I didn't understand it.

Regardless, I think this is fundamentally a design question of supporting multiple multiple CCMs in a single cluster. I wonder if the node folks have any existing guidance on this. I vaguely recall having discussed this before. My impressions from that discussion, which would have been around the time of the external cloud-provider split, where:
* Multiple CCMs are not supported
* Nobody seemed to be thinking about multiple CCMs from the same provider, only different providers

Still, I wonder if it's worth reaching out to those folks. Potential outcomes:
* We're heading in a direction which is explicitly and deliberately unsupported for... reasons
* There's some existing prior art we can reference
* They're interested in a solution and are waiting for somebody to work on it

Not sure what the forum to raise this is, tbh. I wonder if @aojea, @danwinship, or @andrewsykim could signpost us appropriately?

pkg/openstack/instancesv2.go
aojea: This providerID topic was discussed recently, it seems there is the assumptions for cloud providers to set it, so new nodes will wait until the providerID is set unless the cloud provider indicates is not implemented ,https://github.com/kubernetes/kubernetes/pull/123713

Otherwise the node is not going to be initialized

pkg/openstack/utils.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+// If Instances.InstanceID or cloudprovider.GetInstanceProviderID is changed, the regexp should be changed too.
+var providerIDRegexp = regexp.MustCompile(`^` + ProviderName + `://([^/]*)/([^/]+)$`)
+
+// instanceIDFromProviderID splits a provider's id and return instanceID.
+// A providerID is build out of '${ProviderName}:///${instance-id}' which contains ':///'.
+// or '${ProviderName}://${region}/${instance-id}' which contains '://'.
+// See cloudprovider.GetInstanceProviderID and Instances.InstanceID.
+func instanceIDFromProviderID(providerID string) (instanceID string, region string, err error) {
```
aojea, line 32: it seems this is just moving code from one place to another, can you make this in a different commit so it simplifies the review?

pkg/openstack/utils.go
```diff
@@ -0,0 +1,53 @@
...
+
+	matches := providerIDRegexp.FindStringSubmatch(providerID)
+	if len(matches) != 3 {
+		return "", "", fmt.Errorf("ProviderID \"%s\" didn't match expected format \"openstack://region/InstanceID\"", providerID)
+	}
+	return matches[2], matches[1], nil
+}
+
+// Return true if instance is not openstack.
+func isNodeUnmanaged(providerID string) bool {
```
aojea, line 47: ok, this is the new change, the notion of "unmanaged node"

pkg/openstack/instancesv2.go
```diff
@@ -79,9 +79,20 @@ func (os *OpenStack) instancesv2() (*InstancesV2, bool) {
 
 // InstanceExists indicates whether a given node exists according to the cloud provider
 func (i *InstancesV2) InstanceExists(ctx context.Context, node *v1.Node) (bool, error) {
+	if i.regionProviderID {
```
aojea, line 82: I'm not familiar with this concept, it seems that this capability makes the controller to skip the nodes considered unmanaged

sergelogvinov:
I agree with you. Multi-region or hybrid cloud setups introduce complexity. Determining the affiliation of uninitialized nodes becomes challenging.

I've conducted extensive research and experimented with various setups, including multi-region or hybrid Kubernetes clusters. For instance https://github.com/sergelogvinov/terraform-talos?tab=readme-ov-file#multi-cloud-compatibility

Currently, I've been working on a branch [OCCM](https://github.com/sergelogvinov/cloud-provider-openstack/tree/multi-regionl-ccm) that supports multi-region OpenStack setups, particularly for `node` and `node-lifecycle` controllers. However, configuring routes and services in such setups can be quite challenging. I believe that CSI-cinder can also be adapted for multi-region deployments. I had initially planned to introduce it after this PR to simplify the review process.

PS. I've developed my own multi-regional implementation of CCM/CSI for Proxmox, [Proxmox CSI Plugin](https://github.com/sergelogvinov/proxmox-csi-plugin) and [Proxmox Cloud Controller Manager](https://github.com/sergelogvinov/proxmox-cloud-controller-manager). After about a year of production experience with my implementation, I'm confident in its viability.

aojea:
definitely a topic to discuss in sig-cloud-provider  , we recently had a long discussion on the provider ID topic that seems to not be well specified https://github.com/kubernetes/kubernetes/pull/123713 so I will also validate all the assumptions so we not make any assumption now that is not correct

pkg/openstack/instancesv2.go
sergelogvinov: My point is that, the regionProviderID is an alpha feature, meaning that all code and logic within this block can be improved in the future, if i miss something.

sergelogvinov:
> definitely a topic to discuss in sig-cloud-provider , we recently had a long discussion on the provider ID topic that seems to not be well specified [kubernetes/kubernetes#123713](https://github.com/kubernetes/kubernetes/pull/123713) so I will also validate all the assumptions so we not make any assumption now that is not correct

this changes affect only `node-lifecycle` controller,  but discussion [kubernetes/kubernetes#123713](https://github.com/kubernetes/kubernetes/pull/123713) was about `node-controller`

PS. With this patch a node without a ProviderID can never be deleted...

mdbooth:
The architecture here is:

* providerID has the form `openstack://<region>/<instance uuid>`
* One CCM per region, sharded by `region` in the providerID
* A single 'primary' CCM is responsible for assigning a providerID to Nodes which don't have one yet

Is that right?

Do the per-region CCMs actually shard the Nodes, or do we just ensure the responses from InstancesV2 are such that the controller will never take any action?

mdbooth:
Have you considered having kubelet set the providerID on Node creation, btw? Here's how CAPO does this:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/1e0b60fd0ffd7fc1e940ada7e85ec8d724381146/templates/cluster-template.yaml#L21-L24

This is used by kubeadm when it creates the service which runs kubelet. It gets the instance-id from the local metadata service.

Here's how we do it in OpenShift... for AWS unfortunately but the principal is the same (I'm convinced we did this for OpenStack too... may have to fix that 🤔):

https://github.com/openshift/machine-config-operator/blob/master/templates/common/aws/files/usr-local-bin-aws-kubelet-providerid.yaml

That unit runs on boot and sets a `KUBELET_PROVIDERID` environment variable which the kubelet unit uses on startup. Again it uses the local (AWS) metadata service to get the instanceID, and AZ in that case.

Unfortunately I don't think OpenStack `region` is available via the metadata service. However, a workaround for this might be to have a separate template for each region where the region was hard-coded and it just filled in instanceID from the metadata service.

mdbooth:
Coming back to this patch specifically, I think I'd be more comfortable proposing a patch in k/k similar to https://github.com/kubernetes/kubernetes/pull/123713 which causes the node-lifecycle controller to have some explicit behaviour (e.g. ignore) for Nodes which don't have a providerID.

I continue to worry that in this PR we are mis-reporting to cloud-provider on the basis that we expect cloud-provider to have a certain behaviour. I think that will be fragile and hard to maintain.

MatthieuFin:
> Have you considered having kubelet set the providerID on Node creation, btw? Here's how CAPO does this:
> 
> https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/1e0b60fd0ffd7fc1e940ada7e85ec8d724381146/templates/cluster-template.yaml#L21-L24
> 
> This is used by kubeadm when it creates the service which runs kubelet. It gets the instance-id from the local metadata service.
> 
> Here's how we do it in OpenShift... for AWS unfortunately but the principal is the same (I'm convinced we did this for OpenStack too... may have to fix that 🤔):
> 
> https://github.com/openshift/machine-config-operator/blob/master/templates/common/aws/files/usr-local-bin-aws-kubelet-providerid.yaml
> 
> That unit runs on boot and sets a `KUBELET_PROVIDERID` environment variable which the kubelet unit uses on startup. Again it uses the local (AWS) metadata service to get the instanceID, and AZ in that case.
> 
> Unfortunately I don't think OpenStack `region` is available via the metadata service. However, a workaround for this might be to have a separate template for each region where the region was hard-coded and it just filled in instanceID from the metadata service.

Hi @mdbooth , I try to used CAPO in multi-cloud environment and you're right I don't encounter this race condition scenario since provider-id is set by kubeadm, I create different template one per cloud with region hardcoded like that:
```yaml
---
apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
kind: KubeadmConfigTemplate
metadata:
  name: pool-region-one
  annotations:
    controlplane.cluster.x-k8s.io/skip-coredns: ""
    controlplane.cluster.x-k8s.io/skip-kube-proxy: ""
spec:
  template:
    spec:
      joinConfiguration:
        nodeRegistration:
          kubeletExtraArgs:
            cloud-provider: external
            provider-id: openstack://region-one/'{{ instance_id }}'
...
```
This work pretty well with multiple OCCM deployed with variables `OS_CCM_REGIONAL="true"` and `OS_V1_INSTANCES="true"` because instance v2 implementation seems to not handle support of `OS_CCM_REGIONAL` behavior.

However, CAPO don't handle region in magic string provider-id in his CRD:
```bash
kubectl get openstackmachine
NAME                       CLUSTER   INSTANCESTATE   READY   PROVIDERID                                          MACHINE                    AGE
control-plane-6r5xs        capoc     ACTIVE          true    openstack:///87db7660-bc3b-48fd-b5c9-0fa2b2816eed   control-plane-6r5xs        3h14m
pool-region-one-66rgq-t5c72   capoc     ACTIVE          true    openstack:///f48f60bd-bae1-4ed9-b58f-aa1d90aa6579   pool-region-one-66rgq-t5c72   154m
```
And he don't create other node because he is stuck on waiting this first one because CRD's provider-id don't containt region name and don't match kubernetes node `spec.providerID` which contains region name.
image

I think that we simply have to add region in provider-id here https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/controllers/openstackmachine_controller.go#L425C1-L425C112 .

I am looking to find a way to implement that correctly in CAPO.

PR #2594, opened 2024-05-13
[occm] add dnsPolicy feature 

### LINKED ISSUE #2592: [occm] doesn't rely on coredns configuration to do dns resolution
/kind bug

**What happened**:
using occm daemonset on a cluster where coredns has a specific domain resolution for internal services (openstack core services).
the dns resolution made by occm seem to not rely on coredns configuration.

the dns resolution done by occm should use the coredns configuration and resolve internal services rather than trying to resolv on the public internet.

**How to reproduce it**:
add into coredns configmap a specific domain resolution configuration:
```
[MYDOMAIN]:53 {
   log
   errors
   ready
   cache 10
   forward . [DNS1] [DNS2]
}
```
reload coredns configuration
test the resolution with the occm pod

i've use the documentation https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/
to debug dns resolution. the conclusion is everything is configured correctly.

**Environment**:
- openstack-cloud-controller-manager version: 1.30.0
- OpenStack version: Ussuri
- Others:

xinity:
this PR adds dnsPolicy configuration to occm which should be changed if deployed in a cluster with a custom coredns configuration like internal dns domain.
fixes #2592

**Release note**:

```release-note
[openstack-cloud-controller-manager] Add dnsPolicy configuration to manifests and charts
```

### DIFF
--- a/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
@@ -107,6 +107,7 @@ spec:
       {{- if .Values.extraInitContainers }}
       initContainers: {{ toYaml .Values.extraInitContainers | nindent 6 }}
       {{- end }}
+      dnsPolicy: {{ .Values.dnsPolicy }}
       hostNetwork: true
       {{- if .Values.priorityClassName }}
       priorityClassName: {{ .Values.priorityClassName }}
--- a/charts/openstack-cloud-controller-manager/values.yaml
+++ b/charts/openstack-cloud-controller-manager/values.yaml
@@ -46,6 +46,8 @@ livenessProbe: {}
 # Set readinessProbe in the same way like livenessProbe
 readinessProbe: {}
 
+dnsPolicy: ClusterFirstWithHostNet
+
 # Set nodeSelector where the controller should run, i.e. controlplane nodes
 nodeSelector:
   node-role.kubernetes.io/control-plane: ""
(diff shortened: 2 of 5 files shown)


charts/openstack-cloud-controller-manager/values.yaml
```diff
@@ -46,6 +46,8 @@ livenessProbe: {}
 # Set readinessProbe in the same way like livenessProbe
 readinessProbe: {}
 
+# set dnsPolicy to 'ClusterFirstWithHostNet' if the cluster is deployed behind a corporate reverse proxy
+dnsPolicy: default
```
pawcykca, line 50: Why didn't you set default value to `ClusterFirstWithHostNet`?

charts/openstack-cloud-controller-manager/templates/daemonset.yaml
```diff
@@ -107,6 +107,8 @@ spec:
       {{- if .Values.extraInitContainers }}
       initContainers: {{ toYaml .Values.extraInitContainers | nindent 6 }}
       {{- end }}
+
```
dulek, line 110: This newline seems to be unnecessary.

charts/openstack-cloud-controller-manager/values.yaml
dulek, line 50: And here we miss a newline.

dulek, line 50: I think I'd default to `ClusterFirstWithHostNet` as it makes more sense. This will require adding a release note to the PR description, that the default behavior is changed.

dulek (changes requested):
Looks good, some remarks inline.

charts/openstack-cloud-controller-manager/values.yaml
xinity: to leave the default option and leave the user to modify it ONLY if needed :)

xinity: i would leave the default behavior and offer the option to be modified ONLY if necessary

pawcykca: But `dnsPolicy` should be set to `ClusterFirstWithHostNet` by default if `hostNetwork` is set to `true`.

PR #2595, opened 2024-05-14
[occm] Multi region openstack cluster

### LINKED ISSUE #1924: [occm] Multi region support

sergelogvinov:

Openstack CCM multi region support, if it has one Identity provider.
The OpenStack cluster includes a single Keystone service and multiple Nova, Cinder, and Neutron services grouped by region.

openstack-regional

fixes #1924

CCM config changes:
The `region` is required param (as was before) and it uses as default region in cluster.
The `regions` can set multiple times, they will merge with `region` param.  So the value of the `region` may or may not exist in the list of defined regions.

```cfg
[Global]
auth-url=https://auth.openstack.example.com/v3/
region=REGION1
# new param 'regions' can be specified multiple times
regions=REGION1
regions=REGION2
regions=REGION3
```

Optionally can be set in cloud.conf, it supports only one auth service (Keystone)
```yaml
clouds:
  kubernetes:
    auth:
      auth_url: https://auth.openstack.example.com/v3
    region_name: "REGION1"
    regions:
      - REGION1
      - REGION2
      - REGION3
```

During the initialization process, OCCM checks for the existence of providerID. If providerID does not exist, it defaults to using `node.name`, as it did previously. Additionally, if the node has the label `topology.kubernetes.io/region`, OCCM will prioritize using this region as the first one to check. This approach ensures that in the event of a region outage, OCCM can continue to function.

In addition, we can assist CCM in locating the node by providing `kubelet` parameters:
* `--provider-id=openstack:///$InstanceID` - InstanceID exists in [metadata](https://docs.openstack.org/nova/latest/admin/metadata-service.html)
* `--provider-id=openstack://$REGION/$InstanceID` - if you can define the region (by default [meta server](https://docs.openstack.org/nova/latest/admin/metadata-service.html) does not have this information)
* `--node-labels=topology.kubernetes.io/region=$REGION` set preferred REGION in label, OCCM will then prioritize searching for the node in this specified region

The OCCM sets `ProviderID`:
* openstack:///$InstanceID - default behavior 
* openstack://$REGION/$InstanceID if env.OS_CCM_REGIONAL is sets to true (https://github.com/kubernetes/cloud-provider-openstack/pull/1909)

OCCM with multi regions can work with/without `env.OS_CCM_REGIONAL=true`

**Release note**:

```release-note
[OCCM] support multi regional cluster with one keystone deployment
```

### DIFF
--- a/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
+++ b/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
@@ -113,7 +113,9 @@ The options in `Global` section are used for openstack-cloud-controller-manager 
 * `password`
   Keystone user password. If you are using [Keystone application credential](https://docs.openstack.org/keystone/latest/user/application_credentials.html), this option is not required.
 * `region`
-  Required. Keystone region name.
+  Required. Keystone region name. The name of region will be set in the `topology.kubernetes.io/region` label of the node.
+* `regions`
+  Optional. Keystone region name, which is used to specify regions for the cloud provider where the instance is running. Can be specified multiple times.The region name may or may not be the same as the region name in the `region` option. They merge together at runtime.
 * `domain-id`
   Keystone user domain ID. If you are using [Keystone application credential](https://docs.openstack.org/keystone/latest/user/application_credentials.html), this option is not required.
 * `domain-name`
@@ -317,7 +319,7 @@ Although the openstack-cloud-controller-manager was initially implemented with N
   call](https://docs.openstack.org/api-ref/load-balancer/v2/?expanded=create-a-load-balancer-detail#creating-a-fully-populated-load-balancer).
   Setting this option to true will create loadbalancers using serial API calls which first create an unpopulated
   loadbalancer, then populate its listeners, pools and members. This is a compatibility option at the expense of
-  increased load on the OpenStack API. Default: false 
+  increased load on the OpenStack API. Default: false
 
 NOTE:
 
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net/http"
 	"runtime"
+	"slices"
 	"strings"
 
 	"github.com/gophercloud/gophercloud/v2"
@@ -53,6 +54,7 @@ type AuthOpts struct {
 	UserDomainID     string                   `gcfg:"user-domain-id" mapstructure:"user-domain-id" name:"os-userDomainID" value:"optional"`
 	UserDomainName   string                   `gcfg:"user-domain-name" mapstructure:"user-domain-name" name:"os-userDomainName" value:"optional"`
 	Region           string                   `name:"os-region"`
+	Regions          []string                 `name:"os-regions" value:"optional"`
 	EndpointType     gophercloud.Availability `gcfg:"os-endpoint-type" mapstructure:"os-endpoint-type" name:"os-endpointType" value:"optional"`
 	CAFile           string                   `gcfg:"ca-file" mapstructure:"ca-file" name:"os-certAuthorityPath" value:"optional"`
 	TLSInsecure      string                   `gcfg:"tls-insecure" mapstructure:"tls-insecure" name:"os-TLSInsecure" value:"optional" matches:"^true|false$"`
@@ -87,6 +89,7 @@ func LogCfg(authOpts AuthOpts) {
 	klog.V(5).Infof("UserDomainID: %s", authOpts.UserDomainID)
 	klog.V(5).Infof("UserDomainName: %s", authOpts.UserDomainName)
 	klog.V(5).Infof("Region: %s", authOpts.Region)
+	klog.V(5).Infof("Regions: %s", authOpts.Regions)
 	klog.V(5).Infof("EndpointType: %s", authOpts.EndpointType)
 	klog.V(5).Infof("CAFile: %s", authOpts.CAFile)
 	klog.V(5).Infof("CertFile: %s", authOpts.CertFile)
@@ -231,6 +234,20 @@ func ReadClouds(authOpts *AuthOpts) error {
 	authOpts.ApplicationCredentialID = replaceEmpty(authOpts.ApplicationCredentialID, cloud.AuthInfo.ApplicationCredentialID)
 	authOpts.ApplicationCredentialName = replaceEmpty(authOpts.ApplicationCredentialName, cloud.AuthInfo.ApplicationCredentialName)
 	authOpts.ApplicationCredentialSecret = replaceEmpty(authOpts.ApplicationCredentialSecret, cloud.AuthInfo.ApplicationCredentialSecret)
+
+	regions := strings.Split(authOpts.Region, ",")
+	if len(regions) > 1 {
+		authOpts.Region = regions[0]
+	}
+
+	for _, r := range cloud.Regions {
+		// Support only single auth section in clouds.yaml
+		if r.Values.AuthInfo == nil && r.Name != authOpts.Region && !slices.Contains(regions, r.Name) {
+			regions = append(regions, r.Name)
+		}
+	}
+
+	authOpts.Regions = regions
 
 	return nil
 }
--- a/pkg/csi/cinder/openstack/openstack.go
+++ b/pkg/csi/cinder/openstack/openstack.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"slices"
 
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/openstack"
@@ -126,7 +127,7 @@ func GetConfigFromFiles(configFilePaths []string) (Config, error) {
 		}
 	}
 
-	for _, global := range cfg.Global {
+	for idx, global := range cfg.Global {
 		// Update the config with data from clouds.yaml if UseClouds is enabled
 		if global.UseClouds {
 			if global.CloudsFile != "" {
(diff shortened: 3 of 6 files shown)


sergelogvinov:
@mdbooth can you take a look on this PR.
Probably I need to add more configuration checks.

Thanks.

pkg/openstack/instancesv2.go
```diff
@@ -51,16 +53,25 @@ func (os *OpenStack) InstancesV2() (cloudprovider.InstancesV2, bool) {
...
+		if err != nil {
+			klog.Errorf("unable to access compute v2 API : %v", err)
+			return nil, false
+		}
+
+		network[region], err = client.NewNetworkV2(os.provider, opt)
+		if err != nil {
+			klog.Errorf("unable to access network v2 API : %v", err)
+			return nil, false
+		}
```
mdbooth, line 74: @pierreprinetti how much work is performed when initialising a new service client? Is it local-only, or do we have to go back to keystone?

I might be inclined to intialise this lazily anyway, tbh.

pkg/openstack/instancesv2.go
```diff
@@ -118,79 +139,128 @@ func (i *InstancesV2) InstanceMetadata(ctx context.Context, node *v1.Node) (*clo
...
 	}
 
 	instanceID, instanceRegion, err := instanceIDFromProviderID(node.Spec.ProviderID)
 	if err != nil {
-		return nil, err
+		return nil, "", err
+	}
+
+	if instanceRegion == "" {
+		return i.getInstanceByID(instanceID, node.Labels[v1.LabelTopologyRegion])
```
mdbooth, line 188: I'd probably be a bit more explicit with where we're looking for stuff here. IIUC there are 2 possible places we can get a specific region from:
- providerID
- LabelTopologyRegion

Both may be unset because the node has not yet been adopted by the node-controller.

providerID may not contain a region either because it was set before we became multi-region, or because it was set by kubelet without a region and it's immutable.

But the end result is that either we know the region or we don't. If we know the region we should look only in that region. If we don't know the region we should look everywhere.

How about logic something like:
```golang
instanceID, instanceRegion, err := instanceIDFromProviderID(node.Spec.ProviderID)
..err omitted...

if instanceRegion == "" {
  instanceRegion = node.Labels[v1.LabelTopologyRegion]
}

var searchRegions []string
if instanceRegion != "" {
  if !slices.Contains(i.regions, instanceRegion) {
    return ...bad region error...
  }
  searchRegions = []string{instanceRegion}
} else {
  searchRegions = ..all the regions, preferred first...
}

for region := range searchRegions {
  mc := ...
  if instanceID != "" {
    getInstanceByID()
  } else {
    getInstanceByName()
  }
  mc.ObserveRequest()
}
```

mdbooth (commented):
I haven't taken a deep look into this, but I much prefer this in principal: it only tells cloud-provider things we know to be true. This makes me much more confident that this will continue to work correctly as cloud-provider evolves.

Would you still run multiple CCMs, or switch to a single active CCM?

MatthieuFin:
Hi @sergelogvinov
I propose an [implementation](https://github.com/kubernetes/cloud-provider-openstack/pull/2551) of multi cloud support for cinder-csi-plugin, which offer multiple openstack clusters support, not only multiple regions, I haven't take look of occm implementation yet, but is it possible to adapt it to support multiple cloud definitions instead of only multiple regions ?

sergelogvinov:
> Hi @sergelogvinov I propose an [implementation](https://github.com/kubernetes/cloud-provider-openstack/pull/2551) of multi cloud support for cinder-csi-plugin, which offer multiple openstack clusters support, not only multiple regions, I haven't take look of occm implementation yet, but is it possible to adapt it to support multiple cloud definitions instead of only multiple regions ?

Thank you for this PR, it is very interesting. Can we have a call/chat in slack #provider-openstack (Serge Logvinov)?

pkg/openstack/instancesv2.go
sergelogvinov: Thank you, this very very good idea.
I've changed the implementation. But one thought - i cannot trust `LabelTopologyRegion`, something can change it and node-lifecycle will remove node (for instance on reboot/upgrade event)... 

So i can use LabelTopologyRegion only as prefered-region. And check this region first. 
Thanks.

docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
```diff
@@ -114,6 +114,8 @@ The options in `Global` section are used for openstack-cloud-controller-manager
   Keystone user password. If you are using [Keystone application credential](https://docs.openstack.org/keystone/latest/user/application_credentials.html), this option is not required.
 * `region`
   Required. Keystone region name.
+* `regions`
+  Optional. Keystone region name, witch is used to specify regions for the cloud provider where the instance is running. Region is default region name. Can be specified multiple times.
```
jichenjc, line 118: s/witch/which/

pkg/openstack/instancesv2.go
jichenjc: similar thought, maybe init them until real usage?

sergelogvinov: I've had one issue in proxmox with lazy initialization. The regions cannot exist, and during the rollout of OCCM, it starts without errors. The kubernetes administrator will think that all configuration is correct. 

So we can check all regions here and crush if needed. What do you think?

pierreprinetti: Apologies for the late response. Building a ProviderClient requires a Keystone roundtrip; building ServiceClients is cheap.

sergelogvinov:
I've rebased the PR. all tests passed and i've tested manually too

Can you take a look please @jichenjc @mdbooth 
It will be great to merge this change into the upcoming release...

Thanks.

sergelogvinov:
Is anything else we can do here? @jichenjc @mdbooth @kayrus 

We had conversation how we need initialize the openstack clients

```golang
	for _, region := range os.regions {
		opt := os.epOpts
		opt.Region = region

		compute[region], err = client.NewComputeV2(os.provider, opt)
		if err != nil {
			klog.Errorf("unable to access compute v2 API : %v", err)
			return nil, false
		}

		network[region], err = client.NewNetworkV2(os.provider, opt)
		if err != nil {
			klog.Errorf("unable to access network v2 API : %v", err)
			return nil, false
		}
```

It seems to be a similar process to the one we followed in `cinder-csi-plugin`.
I believe @MatthieuFin and I can introduce multi OpenStack authentication support after this PR.

```ini
    [Global]
    auth-url="https://auth.cloud.openstackcluster.region-default.local/v3"
    username="region-default-username"
    password="region-default-password"
    region="default"
    tenant-id="region-default-tenant-id"
    tenant-name="region-default-tenant-name"
    domain-name="Default"
    
    [Global "region-one"]
    auth-url="https://auth.cloud.openstackcluster.region-one.local/v3"
    username="region-one-username"
    password="region-one-password"
    region="one"
    tenant-id="region-one-tenant-id"
    tenant-name="region-one-tenant-name"
    domain-name="Default"
```

Thanks.

sergelogvinov:
Thank you @kayrus for refactoring of instance.go (instance-v2).
I’ve moved all my changes to instance.go. And it ready to go.

Whenever you have a moment, could you please take a look at this?

sergelogvinov:
Thank you @kayrus  for reviewing my PR.

I completely agree with your point about hybrid/multi-cluster setups (bare metal + OpenStack + AWS + GCP, etc.). This should first be implemented in https://github.com/kubernetes/cloud-provider first. I hope it will be added someday.

Our case is a bit different:

Imagine an OpenStack setup with only one Keystone endpoint but multiple regions for services like Nova, Neutron, Cinder, and Glance. Each region has one available zone called "nova" (the default installation). I think this type of setup is easier to manage and upgrade. I've seen similar setups used by many cloud providers.

openstack-regional

So, in this case, it’s not fully separated OpenStack clusters. It looks more like one region with many av-zones, similar to how well-known cloud providers organize their systems.

That’s why this PR supports only one Keystone endpoint. Using this endpoint, we can get Nova/Neutron endpoints for each region.

This feature will be introduced as an alpha feature and can be enabled using a CLI flag. Other components like the load balancer, cinder-csi and manila-csi will also need updates. However, we need to focus on supporting the cloud-node and cloud-node-lifecycle controllers first.

PS. I think I forgot to check backward compatibility with non regional OpenStack setups. I will check this soon.
Thanks.

sergelogvinov:
I’ve updated the documentation and verified the setup. It works well both with OS_CCM_REGIONAL=true enabled and without it.

Could you please let me know if the documentation is clear or if there’s anything that needs improvement? 

Thank you! @kayrus

pierreprinetti:
Doesn't this change deserve a release note?

sergelogvinov:
Hi, @MatthieuFin Could you please take a look at this, since you know hybrid cloud?

MatthieuFin:
Hi,
I manage kubernetes clusters with nodes distributed on différent standalone OpenStack clusters (which means différent keystone) so I can't use directly this implementation on my "classical" kubernetes clusters.

To handle node lifecycle with OCCM, I deploy 1 OCCM DaemonSet per OpenStack cluster (with env variable `OS_CCM_REGIONAL=true`), for that I use official helm chart with simply hack [hostNetwork](https://github.com/kubernetes/cloud-provider-openstack/blob/04bd8e1/charts/openstack-cloud-controller-manager/templates/daemonset.yaml#L109) to `false` to be able to have multiple pods on same node.

I'll try to allowed some time to test your implementation and see how I could adapt it to be able to handle differents keystones, but since this is not a blocking point for me, I don't promise anything.

My next needed improvement will probably to be able to select which OpenStack cloud should be use to provide a LoadBalancer svc probably based on annotations or by Class name I didn't take a look about this point for now (currently all my LB svc are provided by only 1 OpenStack cluster with LB feature enable on corresponding OCCM and disable on others)

PR #2596, opened 2024-05-15
Allow automatic recreation on LB ERROR state to be disabled

baurmatt:
This is e.g. helpful when you try to debug the LB in ERROR state.

Allow automatic recreation on LB ERROR state to be disabled. This is e.g. helpful when you try to debug the LB in ERROR state.

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -93,8 +93,9 @@ const (
 	ServiceAnnotationTlsContainerRef = "loadbalancer.openstack.org/default-tls-container-ref"
 	// revive:enable:var-naming
 	// See https://nip.io
-	defaultProxyHostnameSuffix      = "nip.io"
-	ServiceAnnotationLoadBalancerID = "loadbalancer.openstack.org/load-balancer-id"
+	defaultProxyHostnameSuffix                   = "nip.io"
+	ServiceAnnotationLoadBalancerID              = "loadbalancer.openstack.org/load-balancer-id"
+	ServiceAnnotationLoadBalancerRecreateOnError = "loadbalancer.openstack.org/recreate-on-error"
 
 	// Octavia resources name formats
 	servicePrefix  = "kube_service_"
@@ -313,8 +314,11 @@ func (lbaas *LbaasV2) createOctaviaLoadBalancer(ctx context.Context, name, clust
 		svcConf.lbMemberSubnetID = loadbalancer.VipSubnetID
 	}
 
+	// Allow users to disable automatic recreation on Octavia ERROR state
+	recreateOnError := getBoolFromServiceAnnotation(service, ServiceAnnotationLoadBalancerRecreateOnError, true)
+
 	if loadbalancer, err = openstackutil.WaitActiveAndGetLoadBalancer(ctx, lbaas.lb, loadbalancer.ID); err != nil {
-		if loadbalancer != nil && loadbalancer.ProvisioningStatus == errorStatus {
+		if loadbalancer != nil && loadbalancer.ProvisioningStatus == errorStatus && recreateOnError {
 			// If LB landed in ERROR state we should delete it and retry the creation later.
 			if err = lbaas.deleteLoadBalancer(ctx, loadbalancer, service, svcConf, true); err != nil {
 				return nil, fmt.Errorf("loadbalancer %s is in ERROR state and there was an error when removing it: %v", loadbalancer.ID, err)


baurmatt:
@dulek Mentioning you, as you've initially implemented this functionality in https://github.com/kubernetes/cloud-provider-openstack/pull/2341. Thanks btw, it helps in all other case - just debugging got really hard :D

pkg/openstack/loadbalancer.go
```diff
@@ -306,8 +307,11 @@ func (lbaas *LbaasV2) createOctaviaLoadBalancer(name, clusterName string, servic
 		svcConf.lbMemberSubnetID = loadbalancer.VipSubnetID
 	}
 
+	// Allow users to disable automatic recreation on Octavia ERROR state
+	recreateOnError := getBoolFromServiceAnnotation(service, ServiceAnnotationLoadBalancerRecreateOnError, true)
```
dulek, line 311: This is exposing a debugging function to the end users. I think I'd rather make it an option in OCCM configuration, so that administrator can turn it on and investigate what's happening. What do you thinK?

baurmatt: @dulek Thanks for your review! :) In general I agree, but in our Managed Kubernetes setup users wouldn't be able to change the OCCM configuration due to it not being (editable) exposed to the user. Implementing it as an OCCM configuration would also affect all LBs, while my implementation limits the functionality to a single LB.

jichenjc: um.. does this mean every configuration option we have for OCCM will also not available for managed k8s?

not sure other sitaution we have in such scenario and how we handle it?

baurmatt: > um.. does this mean every configuration option we have for OCCM will also not available for managed k8s?

Just to clarify, when I'm talking about OCCM config I mean the `cloud-config` file/secret. In our managed k8s setup, we don't have (persistent) writeable access to it, because the cloudprovider runs on the master nodes which we don't have access to. `cloud-config` is readonly accessible on the worker nodes for `csi-cinder-nodeplugin`.

dulek: @baurmatt: I don't believe the users of the managed K8s should be trying to debug stuff on Octavia side really. Can you provide an example where keeping the Octavia LB in ERROR state aids with debugging? The regular users should not have access to amphora resources (admin-only API) or Nova VMs backing amphoras (these should live in a service tenant). The LB itself does not expose any debugging information. The Nova VM does expose the error, but most of the time it's a `NoValidHost` anyway, so scheduler logs are required to do debugging.

baurmatt: @dulek To the background: I've created a LoadBalancer Service with `loadbalancer.openstack.org/network-id: $uuid` and `loadbalancer.openstack.org/member-subnet-id: $uuid`, which failed because one of the UUIDs was wrong. Thus it was recreated "all the time". This was hard for the OpenStack team to debug because they only had seconds to take a look on the Octavia LB before it was deleted by cloudprovider. Keeping it in ERROR state allowed for easier debugging on my/OpenStack team side.

dulek: Was the error missing from `kubectl describe <svc-name>`? We emit events that should be enough to debug such problems. What was Octavia returning? Just normally 201?

baurmatt: @dulek It only shows that the load balancer went into ERROR state:

```
$ kubectl describe service cloudnative-pg-cluster-primary2
...
Events:
  Type     Reason                  Age                   From                Message
  ----     ------                  ----                  ----                -------
  Warning  SyncLoadBalancerFailed  29m                   service-controller  Error syncing load balancer: failed to ensure load balancer: error creating loadbalancer kube_service_2mqlgjjphg_cloudnative-pg-cluster_cloudnative-pg-cluster-primary2: loadbalancer has gone into ERROR state
  Normal   EnsuringLoadBalancer    24m (x7 over 31m)     service-controller  Ensuring load balancer
  Warning  SyncLoadBalancerFailed  24m (x6 over 29m)     service-controller  Error syncing load balancer: failed to ensure load balancer: load balancer 71f3ac5c-6740-4fca-8d57-a62d30697629 is not ACTIVE, current provisioning status: ERROR
  Normal   EnsuringLoadBalancer    2m34s (x10 over 22m)  service-controller  Ensuring load balancer
  Warning  SyncLoadBalancerFailed  2m34s (x10 over 22m)  service-controller  Error syncing load balancer: failed to ensure load balancer: load balancer 71f3ac5c-6740-4fca-8d57-a62d30697629 is not ACTIVE, current provisioning status: ERROR
```

dulek: Hm, I see, though in this case you end up with an LB in ERROR state. I still don't see how it staying there is helpful to debugging. Maybe seeing the full LB resource helps, as then you can see the wrong ID, but then we can solve that use case by making sure at some more granular log level, we log full request made to Octavia by Gophercloud instead of adding a new option.

I can also see value in CPO validating network and subnet IDs before creating the LB.

baurmatt: Sorry for the late reply, I've been on vacation. It didn't help me directly, because as a user I still wasn't able to get more information.But when I was able to give our cloud operations team the ID, they were able to debug the problem and tell me the reason.

baurmatt:
@dulek @jichenjc How should we proceed from here? :)

/remove-lifecycle stale

PR #2597, opened 2024-05-16
[cinder-csi-plugin] csi-cinder storage capacity

sergelogvinov:
Available capacity of disk storage

To ensure Kubernetes scheduler selects the optimal zone/region based on available disk capacity limits of the account.
Also to have statistics/alerts in cluster for the specific storageClass

This PR doesn't aim to solve the real capacity issue because Kubernetes CSI doesn't have permission to see all the details of the disk infrastructure.

for https://github.com/kubernetes/cloud-provider-openstack/issues/2035, https://github.com/kubernetes/cloud-provider-openstack/pull/2551

```shell
kubectl get csistoragecapacities -ocustom-columns=CLASS:.storageClassName,AVAIL:.capacity,ZONE:.nodeTopology.matchLabels -A
CLASS                    AVAIL     ZONE
csi-cinder-classic-xfs   19990Gi   map[topology.cinder.csi.openstack.org/zone:nova]
```

**Release note**:

required documentation update/helm-chart/migration process
```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -33,6 +33,7 @@ import (
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/cloud-provider-openstack/pkg/csi/cinder/openstack"
 	"k8s.io/cloud-provider-openstack/pkg/util"
 	cpoerrors "k8s.io/cloud-provider-openstack/pkg/util/errors"
@@ -949,7 +950,28 @@ func (cs *controllerServer) ValidateVolumeCapabilities(ctx context.Context, req 
 }
 
 func (cs *controllerServer) GetCapacity(ctx context.Context, req *csi.GetCapacityRequest) (*csi.GetCapacityResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "GetCapacity is not yet implemented")
+	klog.V(4).Infof("GetCapacity: called with args %+v", protosanitizer.StripSecrets(*req))
+
+	topology := req.GetAccessibleTopology()
+	if topology == nil {
+		return &csi.GetCapacityResponse{}, nil
+	}
+
+	region := topology.GetSegments()[corev1.LabelTopologyRegion]
+
+	cloud, cloudExist := cs.Clouds[region]
+	if !cloudExist {
+		return nil, status.Error(codes.InvalidArgument, "[GetCapacity] specified cloud undefined")
+	}
+
+	availableCapacity, err := cloud.GetFreeQuotaStorageSpace()
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "[GetCapacity]: failed with error %v", err)
+	}
+
+	return &csi.GetCapacityResponse{
+		AvailableCapacity: int64(availableCapacity * 1024 * 1024 * 1024),
+	}, nil
 }
 
 func (cs *controllerServer) ControllerGetVolume(ctx context.Context, req *csi.ControllerGetVolumeRequest) (*csi.ControllerGetVolumeResponse, error) {
--- a/pkg/csi/cinder/openstack/openstack_volumes.go
+++ b/pkg/csi/cinder/openstack/openstack_volumes.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/gophercloud/gophercloud/v2/openstack"
+	"github.com/gophercloud/gophercloud/v2/openstack/blockstorage/v3/limits"
 	"github.com/gophercloud/gophercloud/v2/openstack/blockstorage/v3/volumes"
 	"github.com/gophercloud/gophercloud/v2/openstack/compute/v2/volumeattach"
 	"github.com/gophercloud/gophercloud/v2/pagination"
@@ -394,6 +395,22 @@ func (os *OpenStack) GetMaxVolLimit() int64 {
 	}
 
 	return defaultMaxVolAttachLimit
+}
+
+// GetFreeQuotaStorageSpace returns the tenant quota capacity of the block storage, in GB
+func (os *OpenStack) GetFreeQuotaStorageSpace() (int, error) {
+	mc := metrics.NewMetricContext("limits", "get")
+
+	res, err := limits.Get(context.TODO(), os.blockstorage).Extract()
+	if mc.ObserveRequest(err) != nil {
+		return 0, err
+	}
+
+	capacity := res.Absolute.MaxTotalVolumeGigabytes - res.Absolute.TotalGigabytesUsed
+	if capacity < 0 {
+		capacity = 0
+	}
+	return capacity, nil
 }
 
 // diskIsAttached queries if a volume is attached to a compute instance
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v1.31.0
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 2.31.2
+version: 2.31.3
 home: https://github.com/kubernetes/cloud-provider-openstack
 icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
--- a/charts/cinder-csi-plugin/templates/cinder-csi-driver.yaml
+++ b/charts/cinder-csi-plugin/templates/cinder-csi-driver.yaml
@@ -5,6 +5,9 @@ metadata:
 spec:
   attachRequired: true
   podInfoOnMount: true
+  {{- if .Values.csi.provisioner.capacity }}
+  storageCapacity: true
+  {{- end }}
   volumeLifecycleModes:
   - Persistent
   - Ephemeral
--- a/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
+++ b/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
@@ -73,6 +73,10 @@ spec:
             - "--default-fstype=ext4"
             - "--feature-gates=Topology={{ .Values.csi.provisioner.topology }}"
             - "--extra-create-metadata"
+            {{- if .Values.csi.provisioner.capacity }}
+            - "--enable-capacity"
+            - "--capacity-ownerref-level=2"
+            {{- end }}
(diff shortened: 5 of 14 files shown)


pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -396,6 +397,16 @@ func (os *OpenStack) GetMaxVolLimit() int64 {
 	return defaultMaxVolAttachLimit
 }
 
+// GetFreeCapacity returns free capacity of the block storage
```
dulek, line 400: Would be nice to say that result is expected in bytes. Which is also a bit weird to me.

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -396,6 +397,16 @@ func (os *OpenStack) GetMaxVolLimit() int64 {
...
+func (os *OpenStack) GetFreeCapacity() (int64, error) {
+	res, err := volumelimit.Get(os.blockstorage).Extract()
```
dulek, line 402: I'm not 100% sure, but this sounds like an admin API. We can't make admin calls from CSIs, all CPO is supposed to work normally on a regular tenant.

pkg/csi/cinder/openstack/openstack_volumes.go
sergelogvinov: https://docs.openstack.org/api-ref/block-storage/v3/#limits-limits i did not find the role/permission list for csi-cinder. 
I think this feature should be optional.

pkg/csi/cinder/openstack/openstack_volumes.go
sergelogvinov: https://docs.openstack.org/api-ref/block-storage/v3/#limits-limits i did not find the role/permission list for csi-cinder.
I think this feature should be optional.

pkg/csi/cinder/openstack/openstack_volumes.go
sergelogvinov: fixed, changed to Gb

pkg/csi/cinder/openstack/openstack_volumes.go
dulek: Can you confirm it with DevStack and demo tenant?

sergelogvinov: OVH Cloud, openstack account has `computeOperator`, `volumeOperator` permission, here doc: https://help.ovhcloud.com/csm/en-public-cloud-authenticate-api-openstack-service-account?id=kb_article_view&sysparm_article=KB0059364

but i do not know/can get the real openstack permission of it.
If you have a permission list of DevStack  - i'd like to check...

dulek: I've checked that the API is available. What puzzles me now is - how often is `GetCapacity()` called? Would calling this API be a significant blow on the Cinder API in the cloud?

/ok-to-test

sergelogvinov: it calls 1/min (`capacity-poll-interval` flag of csi-provisioner https://github.com/kubernetes-csi/external-provisioner/blob/c7f94435bd29e49edf1af0eda9c4ee2907c4d160/cmd/csi-provisioner/csi-provisioner.go#L111) for each storageClass and accessibleTopology (av zone).

flag `enable-capacity` by default is off, so csi-provisioner does not collect this metrics by default.

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -396,6 +397,21 @@ func (os *OpenStack) GetMaxVolLimit() int64 {
 	return defaultMaxVolAttachLimit
 }
 
+// GetFreeCapacity returns free capacity of the block storage, in GB
```
jichenjc, line 400: at least this is not my understanding 
the free storage usually comes from backend and I don't know whether every storage provider has this 

and the API seems from here https://github.com/openstack/cinder/blob/master/cinder/api/v3/limits.py#L25 ? 
if so ,it's actually the quota, not the real storage ? e.g you can have 100G max NFS pool
but you can set quota to 1T  ?

jichenjc (changes requested):

pkg/csi/cinder/openstack/openstack_volumes.go
sergelogvinov: Yep, this is a quota.
Based on this quota, I cannot allocate more disk space as I am not the owner of the OpenStack cluster.

These metrics provide the scheduler with additional information about which availability zone the pod can be scheduled in.

jichenjc: um... at least the CSI GetFreeSpace is expecting to give `real` free space? 
this might confuse a little bit

and if the quota is bigger than real storage, this is not accurate info?

sergelogvinov: Oh, now I see that you mean. thanks.
I've tried to get the real size of the pool, but it required `rule:admin_api` role for the api method [scheduler_extension:scheduler_stats:get_pools](https://docs.openstack.org/cinder/latest/configuration/block-storage/policy.html)

So, It is better to rename the function to GetFreeQuotaSpace since we cannot get the real size of the pool.

pkg/csi/cinder/controllerserver.go
```diff
@@ -944,7 +945,28 @@ func (cs *controllerServer) ValidateVolumeCapabilities(ctx context.Context, req
...
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "[GetCapacity]: failed with error %v", err)
+		}
+
+		return &csi.GetCapacityResponse{
+			AvailableCapacity: int64(availableCapacity * 1024 * 1024 * 1024),
+		}, nil
+	}
+
+	return &csi.GetCapacityResponse{}, nil
```
kayrus, line 969: https://refactoring.guru/es/replace-nested-conditional-with-guard-clauses

```suggestion
	topology := req.GetAccessibleTopology()
	if topology == nil {
		return &csi.GetCapacityResponse{}, nil
	}
	
	region := topology.GetSegments()[corev1.LabelTopologyRegion]
	cloud, cloudExist := cs.Clouds[region]
	if !cloudExist {
		return nil, status.Error(codes.InvalidArgument, "[GetCapacity] specified cloud undefined")
	}

	availableCapacity, err := cloud.GetFreeQuotaStorageSpace()
	if err != nil {
		return nil, status.Errorf(codes.Internal, "[GetCapacity]: failed with error %v", err)
	}

	return &csi.GetCapacityResponse{
		AvailableCapacity: int64(availableCapacity * 1024 * 1024 * 1024),
	}, nil
```

kayrus (changes requested):
Overall this change looks fine. However I wonder whether we can consider the `maxTotalVolumes` and `totalVolumesUsed` data returned by the limits API. In theory there can be a case, when you have a lot of capacity but running low on amount of volumes.
The `csi.GetCapacityResponse` struct has only `AvailableCapacity`, `MaximumVolumeSize` and `MinimumVolumeSize`: https://pkg.go.dev/github.com/container-storage-interface/spec/lib/go/csi#GetCapacityResponse. I guess it makes sense to add an option toggle the `GetCapacity` + add a check for maxTotalVolumes` vs `totalVolumesUsed` to fake zero capacity response , when the total amount of volumes has been reached.

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -396,6 +397,22 @@ func (os *OpenStack) GetMaxVolLimit() int64 {
...
+	res, err := limits.Get(context.TODO(), os.blockstorage).Extract()
+	if mc.ObserveRequest(err) != nil {
+		return 0, err
+	}
+
+	capacity := res.Absolute.MaxTotalVolumeGigabytes - res.Absolute.TotalGigabytesUsed
+	if capacity < 0 {
+		capacity = 0
+	}
+	return capacity, nil
```
majewsky, line 413: @kayrus asked me to take a look at this, since I'm our team's resident OpenStack quota/usage measurement expert.

This calculation is going to yield incorrect (or at the least, misleading) results when Cinder supports multiple volume types. The `/limits` endpoint lumps quota and usage for all volume types together into a grand total, so if you get `capacity = 10` here, it could mean "6 GiB for volume type A plus 4 GiB for volume type B", and actually creating a 10 GiB volume of either type will fail.

The correct endpoint for the thing that you want to do is https://docs.openstack.org/api-ref/block-storage/v3/index.html#show-quota-usage-for-a-project, modelled in Gophercloud as https://pkg.go.dev/github.com/gophercloud/gophercloud/v2/openstack/blockstorage/v3/quotasets#GetUsage. Unfortunately, the modelling in Gophercloud is not very good, so the relevant `quota_set.gigabytes_$VOLUMETYPE` fields need to be extracted in a custom way. This is how I do it in my own code: https://github.com/sapcc/limes/blob/7153750c217e51e668ba67375e64a2938b97532b/internal/liquids/cinder/usage.go#L39-L52

Hope that helps!

sergelogvinov:
> Overall this change looks fine. However I wonder whether we can consider the `maxTotalVolumes` and `totalVolumesUsed` data returned by the limits API. In theory there can be a case, when you have a lot of capacity but running low on amount of volumes. The `csi.GetCapacityResponse` struct has only `AvailableCapacity`, `MaximumVolumeSize` and `MinimumVolumeSize`: https://pkg.go.dev/github.com/container-storage-interface/spec/lib/go/csi#GetCapacityResponse. I guess it makes sense to add an option toggle the `GetCapacity` (e.g. whether to advertise the `csi.ControllerServiceCapability_RPC_GET_CAPACITY`) + add a check for `maxTotalVolumes` vs `totalVolumesUsed` to fake zero capacity response , when the total amount of volumes has been reached.

This is indeed a very interesting idea. However, I can imagine that receiving an alert about running out of space when there is still space available might cause some confusion, similar to the situation with file systems and inodes. In my opinion, it would be ideal if Kubernetes first supported something like `totalVolumesUsed` before implementing this feature.

pkg/csi/cinder/openstack/openstack_volumes.go
sergelogvinov: Thank you for the interesting thoughts! However, my focus is on addressing customer-side issues, specifically when dealing with account limits for storage space. This limit includes both block volumes and snapshots, and having this information helps the Kubernetes scheduler pods correctly, and also alerting us when we are nearing the usage limit.

The actual storage capacity is managed by the OpenStack team, not the customers. I trust that they have several methods in place to monitor and determine when additional storage capacity is needed.

PR #676, opened 2019-07-01
Adds openstack instance to nodeserver

ramineni:
This commit adds cloud instance to nodeserver to be able to query openstack and adds
error codes as per csi spec.

**What this PR does / why we need it**:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #661 

**Special notes for your reviewer**:

**Release note**:

```release-note
```

PR #2602, opened 2024-05-27
[cinder-csi-plugin] ephemeral volume removal

### LINKED ISSUE #2599: [csi-cinder-plugin] Ephemeral Volume removal process
/kind feature

**What happened**:

CSI Ephemeral Volume has been deprecated  #1493 and it requires OpenStack credentials on each node, adding unnecessary complexity to our support processes.

Can we start to removal process of it?

ref #2551

**How to reproduce it**:

**Environment**:
- OpenStack version:
- Others:

sergelogvinov:
Disable creating new deprecated ephemeral volumes

Remove deprecated ephemeral volumes.

fixes #2599

Hi, @dulek i've disabled only ephemeral volume creation stage.
What do you think, we should remove all code or make it in two steps (releases)?

Thanks.

**Release note**:

```release-note
In-line ephemeral is no longer available.
```

### DIFF
--- a/pkg/csi/cinder/nodeserver.go
+++ b/pkg/csi/cinder/nodeserver.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"strconv"
 	"strings"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
@@ -68,21 +67,13 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 
 	ephemeralVolume := req.GetVolumeContext()["csi.storage.k8s.io/ephemeral"] == "true"
 	if ephemeralVolume {
-		// See https://github.com/kubernetes/cloud-provider-openstack/issues/1493
-		klog.Warningf("CSI inline ephemeral volumes support is deprecated in 1.24 release.")
-		return nodePublishEphemeral(req, ns)
+		// See https://github.com/kubernetes/cloud-provider-openstack/issues/2599
+		return nil, status.Error(codes.Unimplemented, "CSI inline ephemeral volumes support is removed in 1.31 release.")
 	}
 
 	// In case of ephemeral volume staging path not provided
 	if len(source) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume Staging Target Path must be provided")
-	}
-	_, err := ns.Cloud.GetVolume(volumeID)
-	if err != nil {
-		if cpoerrors.IsNotFound(err) {
-			return nil, status.Error(codes.NotFound, "Volume not found")
-		}
-		return nil, status.Errorf(codes.Internal, "GetVolume failed with error %v", err)
 	}
 
 	mountOptions := []string{"bind"}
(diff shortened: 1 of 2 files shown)


jichenjc:
thanks for the PR, could you please update 
- doc needed
- release notes about this PR

jichenjc:
@sergelogvinov from the error seems more clean up needed, maybe sainty test cases?

sergelogvinov:
@jichenjc All tests passed. thanks.

pkg/csi/cinder/nodeserver.go
```diff
@@ -67,22 +66,14 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
 	if len(source) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume Staging Target Path must be provided")
 	}
-	_, err := ns.Cloud.GetVolume(volumeID)
-	if err != nil {
-		if cpoerrors.IsNotFound(err) {
-			return nil, status.Error(codes.NotFound, "Volume not found")
-		}
-		return nil, status.Errorf(codes.Internal, "GetVolume failed with error %v", err)
-	}
```
dulek, line 85: Why do you remove this? This doesn't exactly feel like it's related, with `"csi.storage.k8s.io/ephemeral" == "true"` this code won't be exected.

pkg/csi/cinder/nodeserver.go
```diff
@@ -439,16 +330,7 @@ func (ns *nodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
...
 	}
 
-	_, err := ns.Cloud.GetVolume(volumeID)
-	if err != nil {
-		if cpoerrors.IsNotFound(err) {
-			klog.V(4).Infof("NodeUnstageVolume: Unable to find volume: %v", err)
-			return nil, status.Error(codes.NotFound, "Volume not found")
-		}
-		return nil, status.Errorf(codes.Internal, "GetVolume failed with error %v", err)
-	}
```
dulek, line 449: Same here, I think.

pkg/csi/cinder/nodeserver.go
sergelogvinov: Hi, the main idea is - to remove openstack credentials dependency  from node csi daemonset.

We do not need to check the existence of the Volume in the daemonset by openstack api. When the CSI sidecar calls the NodeStageVolume/NodePublishVolume function, the Volume is already attached to the VM. We just need to find the block device in the /dev/* directory. If any issues occur, the block device mounter will return an error.

I believe, this check `_, err := ns.Cloud.GetVolume(volumeID)` is not necessary here.

What do you think?

dulek: It's been a while ago when this check got in in https://github.com/kubernetes/cloud-provider-openstack/pull/676. @jichenjc, do you remember why it was necessary?

@kayrus, what do you think about this?

jichenjc: seems last update is commit cc73e20cfe8233ff8ad3dbd040866f164225f6f8
but didn't find the commit message realted 

I am guessing this might prevent some concurrent deletion during the actions here ..

jichenjc:
if no objection, let's merge this soon

sergelogvinov:
i've rebased the PR... Thanks.

sergelogvinov:
All tests passed, can you take a look please @zetaab @mdbooth 
Thanks.

zetaab (approved):
/approve
/lgtm

PR #2610, opened 2024-05-31
[cinder-csi-plugin] retry mount operation with rescan incase of failure

### LINKED ISSUE #2588: [csi-cinder-plugin] Pods stuck in "ContainerCreating" because a volume could not be formatted and mounted
**What happened**:
When creating pods with a cinder volume, they may occasionally get stuck in the 'ContainerCreating' state. Upon closer examination, the disk appears with a size of 0B. This issue likely arises because occasionally hypervisors don't send a new volume size to the kernel when a volume is in use. 

However, if we force a rescan of the block device on the node by executing echo "1" > /sys/class/block/[device]/device/rescan, the pod will start

The pod to be started normally

**How to reproduce it**:

Issue it not consistently reproducible

Would it be possible to add the rescan bit during provisioning if it fails, not only on resize? I understand there might be a slight challenge since not all devices support this. However, we can check if the paths exists or if rescan-on-resize is set, could we also set the rescan bit?

Similar to the probeVolume case where "- - -" is set to /scan, we would also set "1" to /rescan as the "- - -" won't cover the case were the size is incorrectly reported

/kind bug

NymanRobin:
If the initial formatting and mounting fails in NodeStageVolume try to rescan the device and retry the operation.

### What this PR does / why we need it:

This prevents failures if the device information is reported wrongly which would otherwise be blocking the mounting.

### Which issue this PR fixes(if applicable):
fixes #2588 

### Special notes for reviewers:

Hi, I thought I would add this as a retry mechanism in case the mounting has already failed to minimize the impact on the current flow, also do not think this rescan should be causing any issues. If there is other ideas or concerns I am all ears, thanks!

```release-notes
None
```

### DIFF
--- a/pkg/csi/cinder/nodeserver.go
+++ b/pkg/csi/cinder/nodeserver.go
@@ -398,7 +398,7 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 			options = append(options, collectMountOptions(fsType, mountFlags)...)
 		}
 		// Mount
-		err = m.Mounter().FormatAndMount(devicePath, stagingTarget, fsType, options)
+		err = ns.formatAndMountRetry(devicePath, stagingTarget, fsType, options)
 		if err != nil {
 			return nil, status.Error(codes.Internal, err.Error())
 		}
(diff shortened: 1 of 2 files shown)


pkg/util/blockdevice/blockdevice_linux.go
```diff
@@ -85,6 +85,16 @@ func checkBlockDeviceSize(devicePath string, deviceMountPath string, newSize int
 	return nil
 }
 
+func triggerRescan(blockDeviceRescanPath string) error {
+	klog.V(4).Infof("Rescanning %q block device geometry", blockDeviceRescanPath)
+	err := os.WriteFile(blockDeviceRescanPath, []byte{'1'}, 0666)
```
jichenjc, line 90: I haven't try this before, but 0666 is convention for such operation?

pkg/csi/cinder/nodeserver.go
```diff
@@ -426,6 +426,25 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	return &csi.NodeStageVolumeResponse{}, nil
 }
 
+// formatAndMountRetry attempts to format and mount a device at the given path.
+// If the initial mount fails, it rescans the device and retries the mount operation.
+func (ns *nodeServer) formatAndMountRetry(devicePath, stagingTarget, fsType string, options []string) error {
+	m := ns.Mount
+	err := m.Mounter().FormatAndMount(devicePath, stagingTarget, fsType, options)
+	if err != nil {
+		klog.Errorf("Initial format and mount failed: %v. Attempting rescan.", err)
```
jichenjc, line 435: if we accept this rescan logic, this can only be Infof log as it's normal case now

NymanRobin: Thanks for pointing this out! I will also change the rescan error to a info level log as there are devices that do not have the interface so no need to log an error in those cases either

pkg/util/blockdevice/blockdevice_linux.go
NymanRobin: It is a good question and I am not sure to be honest if all these permissions are required, I just reused the logic that already was here. However I saw that the original author of this rescan logic was @kayrus so maybe he could make us wiser

NymanRobin:
Is there anymore thoughts or feedback for this?

jichenjc:
@NymanRobin please update your commit msg so we can merge this

NymanRobin:
Awwesome! I did not quite get what update I should do to the commit message could you help me out @jichenjc?
Maybe adding the issue in the subject is the thing missing?

Then the prefix of the subject is a bit of a mystery because I can see four different ways in commits so

cinder-csi:
[cinder-csi-plugin] 
fix:
and without prefix

jichenjc:
I updated your commit message and it should be good now

NymanRobin:
Thanks

pkg/util/blockdevice/blockdevice_linux.go
kayrus: I took this from upstream k8s libs for similar operations.

PR #2613, opened 2024-06-02
[cinder-csi-plugin] Add extraEnv support on cinder-csi-plugin and manila-csi-plugin helm charts

pli01:

Add `extraEnv` support on the cinder-csi-plugin helm chart.

One example use case of extraEnv:  In some of our private customer cloud, cinder-csi-plugin must access the openstack API endpoint with following env variable  `http_proxy, https_proxy, no_proxy`.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v1.30.0
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 2.30.0
+version: 2.30.1-alpha.1
 home: https://github.com/kubernetes/cloud-provider-openstack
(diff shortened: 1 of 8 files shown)


dulek:
/ok-to-test

Looks like this could be useful for the Manila CSI charts too? Could you add it there as well?

pli01:
> Looks like this could be useful for the Manila CSI charts too? Could you add it there as well?

Thanks for the approval.
Yes,  i will try to add here also the modification for Manila CSI charts.

pli01:
Added for manila
@dulek , Is that okay with you ?

dulek:
Looks good, just why do you comment out Manila's values.yaml, but not Cinder's?

pli01:
> Looks good, just why do you comment out Manila's values.yaml, but not Cinder's?

mistake

it s now fixed

PR #2615, opened 2024-06-05
[all] Upgrade Gophercloud to v2

pierreprinetti:

This PR upgrades Gophercloud to v2. While this new version enables `context.Context` for cancellation and tracing, this PR does not tackle that just yet, but rather just enables that follow-up work.

This is currently a bump to a release candidate, just to kick the tires and get some feedback from the CI. I will update this pull request to import `v2.0.0` once it's out. Thank you!

```release-note
NONE
```


pierreprinetti:
/hold cancel
Tests pass 🎉
This is ready for review.

pierreprinetti:
/hold

Everything going according to plan, we are releasing v2.0.0 on Monday. I will update the tag in `go.mod` as soon as it's done, and then we can merge if the maintainers are OK with it.

pierreprinetti:
/hold cancel
Updated to Gophercloud v2.0.0, which has been released this morning.

PTAL!
/cc @zetaab @kayrus

zetaab (approved):
/approve

quite heavy PR, could not find anything wrong. As tests are passing good to go

pkg/openstack/loadbalancer.go
```diff
@@ -1003,6 +1004,39 @@ func (lbaas *LbaasV2) buildBatchUpdateMemberOpts(port corev1.ServicePort, nodes
 	return members, newMembers, nil
 }
 
+func derefOrEmpty[T any](p *T) T {
+	if p != nil {
+		return *p
+	}
+	var t T
+	return t
+}
```
kayrus, line 1013: it makes sense to define this func in `pkg/util`

pierreprinetti: Good point. I have replaced that with a call to `ptr.Deref` in k8s.io/utils

kayrus: Even better. Thanks!

PR #2631, opened 2024-08-19
[octavia-ingress-controller] Pass revision_id when assigning/removing SGs

### LINKED ISSUE #2630: [octavia-ingress-controller] Possible lost update race conditions in adding and removing SGs

/kind feature

(it's a feature'y bug :smile:)

**What happened**:

Neutron doesn't provide a way to add or remove port security groups individually. Instead, you indicate the total set of SGs that should be on the port, meaning to add an SG you must fetch the existing SGs, append the new SG to this set, and then apply this updated set. This opens us up to a TOCTOU race: if a security group is added or removed in the interim, we will lose that update. @dulek has noted this in places.

Per @dulek's notes, we should make use of neutron's `revision-if-match` shim extension to set a revision ID.

**How to reproduce it**:

N/A

N/A

**Environment**:

`master`

stephenfin:

Set revision ID when adding/removing security groups from a port.

Fixes #2630

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer_sg.go
+++ b/pkg/openstack/loadbalancer_sg.go
@@ -86,11 +86,8 @@ func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, svcConf *
 			}
 
 			// Add the SG to the port
-			// TODO(dulek): This isn't an atomic operation. In order to protect from lost update issues we should use
-			//              `revision_number` handling to make sure our update to `security_groups` field wasn't preceded
-			//              by a different one. Same applies to a removal of the SG.
 			newSGs := append(port.SecurityGroups, sg)
-			updateOpts := neutronports.UpdateOpts{SecurityGroups: &newSGs}
+			updateOpts := neutronports.UpdateOpts{SecurityGroups: &newSGs, RevisionNumber: &port.RevisionNumber}
 			mc := metrics.NewMetricContext("port", "update")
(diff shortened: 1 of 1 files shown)


stephenfin:
/retitle [octavia-ingress-controller] Pass revision_id when assigning/removing SGs

pkg/openstack/loadbalancer_sg.go
```diff
@@ -86,11 +86,8 @@ func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, svcConf *
 			}
 
 			// Add the SG to the port
-			// TODO(dulek): This isn't an atomic operation. In order to protect from lost update issues we should use
-			//              `revision_number` handling to make sure our update to `security_groups` field wasn't preceded
-			//              by a different one. Same applies to a removal of the SG.
 			newSGs := append(port.SecurityGroups, sg)
-			updateOpts := neutronports.UpdateOpts{SecurityGroups: &newSGs}
```
dulek, line 93: What's the intended retry logic here? I _guess_ relying on controller backoff is fine, maybe just worth a different error message, but for reference: https://github.com/openshift/cloud-network-config-controller/blob/master/pkg/cloudprovider/openstack.go#L765-L809

stephenfin:
/remove-lifecycle stale
/hold

I'll pick this up again and address comments in the new year.

PR #2634, opened 2024-08-23
[occm] [cinder-csi-plugin] Sanitize AZ name

scrungus:
**What this PR does / why we need it**: AZ names can contain unicode characters, for example "Württemberg". The current behaviour is that this causes the occm pod to crash: 
```
$ crictl logs a358b0388996d (openstack-cloud-controller-manager)
error syncing 'test-hshu2jf73dym-default-worker-6s4mq-s9swb': Node "test-hshu2jf73dym-default-worker-6s4mq-s9swb" is invalid: metadata.labels: Invalid value: "Württemberg Test": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?'), requeuing
```
This PR sanitizes the label value before it is applied. 
We also see the same problem in cinder-csi-plugin: 
```
$ kubectl logs -n openstack-system openstack-cinder-csi-nodeplugin-cl5tc
...
I0822 14:53:22.576401       1 main.go:101] Received NotifyRegistrationStatus call: &RegistrationStatus{PluginRegistered:false,Error:RegisterPlugin error -- plugin registration failed with err: error updating Node object with CSI driver node info: error updating node: timed out waiting for the condition; caused by: failed to patch status "{\"metadata\":{\"annotations\":{\"csi.volume.kubernetes.io/nodeid\":\"{\\\"cinder.csi.openstack.org\\\":\\\"0699b10e-b4bc-4129-9280-5b7c8556418f\\\",\\\"csi.tigera.io\\\":\\\"sst-yde-test-1-yt3243sf6os5-control-plane-m4hpg\\\"}\"},\"labels\":{\"topology.cinder.csi.openstack.org/zone\":\"Düdingen Test\"}}}" for node "sst-yde-test-1-yt3243sf6os5-control-plane-m4hpg": Node "sst-yde-test-1-yt3243sf6os5-control-plane-m4hpg" is invalid: metadata.labels: Invalid value: "Düdingen Test": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?'),}
```

So this change also sanitises the label coming from the metadata service. 

I believe a similar fix is needed for cinder-csi-plugin
**Release note**:

```release-note
Fixes a bug where unicode characters in the openstack availability zone can crash the OCCM or cinder-csi-plugin pods.
```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
@@ -166,4 +167,20 @@ func GetAZFromTopology(topologyKey string, requirement *csi.TopologyRequirement)
 	}
 
 	return zone
+}
+
+func SanitizeLabel(input string) string {
+	// Replace non-alphanumeric characters (except '-', '_', '.') with '-'
+	reg := regexp.MustCompile(`[^-a-zA-Z0-9_.]+`)
+	sanitized := reg.ReplaceAllString(input, "-")
+
+	// Ensure the label starts and ends with an alphanumeric character
+	sanitized = strings.Trim(sanitized, "-_.")
+
+	// Ensure the label is not longer than 63 characters
+	if len(sanitized) > 63 {
+		sanitized = sanitized[:63]
+	}
+
+	return sanitized
 }
--- a/pkg/openstack/instancesv2.go
+++ b/pkg/openstack/instancesv2.go
@@ -27,6 +27,7 @@ import (
 	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/cloud-provider-openstack/pkg/client"
 	"k8s.io/cloud-provider-openstack/pkg/metrics"
+	"k8s.io/cloud-provider-openstack/pkg/util"
 	"k8s.io/cloud-provider-openstack/pkg/util/errors"
 	"k8s.io/klog/v2"
 )
@@ -133,11 +134,13 @@ func (i *InstancesV2) InstanceMetadata(ctx context.Context, node *v1.Node) (*clo
 		return nil, err
 	}
 
+	availabilityZone := util.SanitizeLabel(server.AvailabilityZone)
+
 	return &cloudprovider.InstanceMetadata{
 		ProviderID:    i.makeInstanceID(&server),
 		InstanceType:  instanceType,
 		NodeAddresses: addresses,
-		Zone:          server.AvailabilityZone,
+		Zone:          availabilityZone,
 		Region:        i.region,
 	}, nil
 }
--- a/pkg/util/metadata/metadata.go
+++ b/pkg/util/metadata/metadata.go
@@ -306,7 +306,7 @@ func (m *metadataService) GetAvailabilityZone() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return md.AvailabilityZone, nil
+	return util.SanitizeLabel(md.AvailabilityZone), nil
 }
 
 func CheckMetadataSearchOrder(order string) error {


scrungus:
I am unsure as to why the availability zone information comes from gophercloud in the occm case, and the metadata service in the cinder-csi-plugin. It may make sense to get occm to use the metadata service too, unless there's a specific reason otherwise.

pkg/util/util.go
```diff
@@ -167,3 +168,20 @@ func GetAZFromTopology(topologyKey string, requirement *csi.TopologyRequirement)
...
+	sanitized = strings.Trim(sanitized, "-_.")
+
+	// Ensure the label is not longer than 63 characters
+	if len(sanitized) > 63 {
+		sanitized = sanitized[:63]
+	}
+
+	// Convert to lowercase
+	return strings.ToLower(sanitized), nil
+}
```
dulek, line 187: It doesn't seem this can ever return an error, so why bother with having `(string, error) as a return value?

dulek (changes requested):

dulek:
> I am unsure as to why the availability zone information comes from gophercloud in the occm case, and the metadata service in the cinder-csi-plugin. It may make sense to get occm to use the metadata service too, unless there's a specific reason otherwise.

Doesn't cinder-csi-plugin run on every node, when occm can run on any node of the cluster?

jichenjc:
>>>Doesn't cinder-csi-plugin run on every node, when occm can run on any node of the cluster?

yes, it is CSI need run every node, as it has control and data plane both

pkg/util/util.go
```diff
@@ -167,3 +168,20 @@ func GetAZFromTopology(topologyKey string, requirement *csi.TopologyRequirement)
 
 	return zone
 }
+
+func SanitizeLabel(input string) (string, error) {
```
jichenjc, line 172: I assume this follow existing k8s pattern already
not sure whether there are any way that k8s already provided that we can reuse that function?

anyway, it's good as we met this issue before in our local env (a AZ with blank)

jichenjc:
The service has some problem in start up ..

```
>>>>>>> FAIL: Timeout when waiting for the IP address 172.24.5.107(VIP) accessible
```

but not sure it's related to this PR, not sure worthy to try an empty PR to see whether it's something else cause the problem

pkg/util/util.go
```diff
@@ -167,3 +168,20 @@ func GetAZFromTopology(topologyKey string, requirement *csi.TopologyRequirement)
...
+	// Ensure the label starts and ends with an alphanumeric character
+	sanitized = strings.Trim(sanitized, "-_.")
+
+	// Ensure the label is not longer than 63 characters
+	if len(sanitized) > 63 {
+		sanitized = sanitized[:63]
+	}
+
+	// Convert to lowercase
+	return strings.ToLower(sanitized), nil
```
jichenjc, line 186: this seems not necessary
e.g keep Berlin seems better than berlin

scrungus:
> > > > Doesn't cinder-csi-plugin run on every node, when occm can run on any node of the cluster?
> 
> yes, it is CSI need run every node, as it has control and data plane both

I see, that makes sense then.

pkg/util/util.go
scrungus: @jichenjc I don't think so. There is a lot of validation logic here https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go but unfortunately it doesn't seem like there is any SanitizeLabel type function, only boolean check functions

jichenjc: ok, then use our own implementation is also good

pkg/util/util.go
scrungus: true, updated.

pkg/util/util.go
scrungus: updated

scrungus:
@jichenjc we have a client that we did this work for, they would like to be included in a `Co-Authored-By` in the commit message. Could you please wait for me to update this before merging?

jichenjc:
I hold this, you can feel free to unhold when you are ready

/hold

Issue #2374, opened 2023-09-28
[occm] improve LB logic maintainability


kayrus:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**What happened**:

it's hard to maintain the `pkg/openstack/loadbalancer.go` file, since it contains more than 2.5k lines of code and some functions/methods are hard to read.

**What you expected to happen**:

I'm planning to split `pkg/openstack/loadbalancer.go` into multiple ``pkg/openstack/loadbalancer_*.go` files and simplify some complex func/method logic, e.g. remove a number if/then/elseif/else conditions, which complicate the code reading and maintainability.

**How to reproduce it**:

n/a

**Anything else we need to know?**:

n/a

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version:
- OpenStack version:
- Others:

PR #1384, opened 2021-01-22
[core]: introduce a common openstack client package

kayrus:
**What this PR does / why we need it**:

This PR optimizes the package structure. As an outcome some binary sizes are much smaller.

**Which issue this PR fixes(if applicable)**:
fixes #993

**Special notes for reviewers**:

**Release note**:

```release-note
NONE
```

PR #735, opened 2019-08-23
Deduplicate OpenStack auth functions

kayrus:
**What this PR does / why we need it**:

It unifies all the OpenStack auth options and deduplicates code.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

#640

**Special notes for your reviewer**:

This PR is based on #733. I'll rebase this PR once #733 is merged.

**Release note**:

```release-note
Unify openstack authenticate options
```

PR #2640, opened 2024-09-03
[cinder-csi-plugin] ephemeral volume removal (#2602)

sergelogvinov:

Remove openstack credits from node plugin 

step 2 of #2599

**Release note**:

```release-note
the flag node-service-no-os-client was deprecated
```

**Manual tests**:

- [x] create/delete
- [x] snapshot create/delete
- [x] create from clone


sergelogvinov:
When you have a time, can you take a look on this changes @jichenjc @dulek 

I think it's ready to go, but we need to wait for the official release first.
Thanks.

kayrus:
@sergelogvinov could you please rebase your PR?

sergelogvinov:
@kayrus done.

Thank you!

sergelogvinov:
I've changed the golangci-lint version to 1.61.0, and test `pull-cloud-provider-openstack-check` passed.

kayrus:
@sergelogvinov I submitted a more agnostic PR that allows to run node service without credentials: #2655

pkg/csi/cinder/driver.go
```diff
@@ -32,6 +32,12 @@ import (
 const (
 	driverName  = "cinder.csi.openstack.org"
 	topologyKey = "topology." + driverName + "/zone"
+
+	// MaxVolumesPerNode is the maximum number of volumes that can be attached to a node
+	MaxVolumesPerNode = 256
```
kayrus, line 37: Where does this const come from?

sergelogvinov: It was hardcoded here https://github.com/kubernetes/cloud-provider-openstack/blob/d069b1ef5d34e1400c798424f5d56b11098fb9a7/pkg/csi/cinder/openstack/openstack.go#L144

kayrus: just in case, if the const/var doesn't need to be accessed from a different package, it's better to define its name starting with a lower case (non-exported). https://go.dev/tour/basics/3

sergelogvinov: Yep, `--node-volume-attach-limit` flag customize it

cmd/cinder-csi-plugin/main.go
```diff
@@ -105,23 +122,29 @@ func handle() {
...
+				nodeVolumeAttachLimit = cfg.BlockStorage.NodeVolumeAttachLimit
+			}
+		}
+
 		//Initialize Metadata
-		metadata := metadata.GetMetadataProvider(clouds[cloudNames[0]].GetMetadataOpts().SearchOrder)
+		metadata := metadata.GetMetadataProvider(searchOrder)
 
-		d.SetupNodeService(clouds[cloudNames[0]], mount, metadata, additionalTopologies)
+		d.SetupNodeService(mount, metadata, rescanOnResize, nodeVolumeAttachLimit, additionalTopologies)
```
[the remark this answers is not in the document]
sergelogvinov: switch to `openstack.BlockStorageOpts{}`

pkg/csi/cinder/driver.go
sergelogvinov: MaxVolumesPerNode -> maxVolumesPerNode 👍

pkg/csi/cinder/nodeserver.go
```diff
@@ -296,9 +224,7 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		}
 	}
 
-	// Try expanding the volume if it's created from a snapshot or another volume (see #1539)
-	if vol.SourceVolID != "" || vol.SnapshotID != "" {
-
+	if required, ok := volumeContext[ResizeRequired]; ok && strings.EqualFold(required, "true") {
```
kayrus, line 227: btw, it's enough to just set an empty string to the corresponding `volumeContext` key to define the resize requirement and check it like this:

```suggestion
	if , ok := volumeContext[ResizeRequired]; ok {
```

pkg/csi/cinder/utils.go
```diff
@@ -57,13 +57,17 @@ func NewIdentityServer(d *Driver) *identityServer {
 	}
 }
 
-func NewNodeServer(d *Driver, mount mount.IMount, metadata metadata.IMetadata, cloud openstack.IOpenStack, topologies map[string]string) *nodeServer {
+func NewNodeServer(d *Driver, mount mount.IMount, metadata metadata.IMetadata, opts openstack.BlockStorageOpts, topologies map[string]string) *nodeServer {
+	if opts.NodeVolumeAttachLimit < 0 || opts.NodeVolumeAttachLimit > maxVolumesPerNode {
+		opts.NodeVolumeAttachLimit = maxVolumesPerNode
```
kayrus, line 62: add a log with all opts vars for better visibility

kayrus (changes requested):
Thanks for the PR. Unfortunately, the cleanliness and quality of the code in this repository leave much to be desired. I get the impression that new contributors are simply copying existing patterns, resulting in poorly readable code. Try to avoid this. Unfortunately, we don't have much time to refactor the code. The refactoring of the authentication code alone took me a lot of time (see #735 and #1384 for example or my attempt to rewrite LB code #2374), so we have to work with what we have. The only thing we can do now is to write new functionality with quality and, where possible, remove remnants of the old code. Therefore, I ask you to understand my feedback, and it's likely we'll need several rounds of code review. Thank again ;)

sergelogvinov:
Thank you for your review @kayrus!

In my view, it's better to use CLI flags. All CSI sidecars work the same way, which makes it easier for new members to get started with the project. If we need to change the `rescanOnResize` or `nodeVolumeAttachLimit` settings, we need two configurations (with/without OpenStack credentials). This adds more complexity to manage. People will likely use the one configuration with credentials to speed up the cluster. A CLI flag is very easy to change by adding it to the `daemonset` using helm or kustomize.


I try to follow the rule of keeping code changes to a minimum. However, I think it would be better to refactor cmd/cinder-csi-plugin/main.go. I also believe it’s much better to do this in separate PRs, rather than having one large PR that’s already difficult to review.

pkg/csi/cinder/utils.go
sergelogvinov: AuthOpts does not have access to openstack.BlockStorageOpts opts. Better to add reach logging here https://github.com/kubernetes/cloud-provider-openstack/pull/2652

kayrus:
>  If we need to change the rescanOnResize or nodeVolumeAttachLimit settings, we need two configurations (with/without OpenStack credentials).

The only thing that I don't like, you're defining a new global var, then put its value to a struct.

> I try to follow the rule of keeping code changes to a minimum. However, I think it would be better to refactor cmd/cinder-csi-plugin/main.go

I didn't insist on keeping the code changes to a minimum. What I meant is that refactoring can be a real pain; sometimes, you can spend hours on it and still end up failing or burning out. I suggested adding new features alongside code refactoring. For example, if you notice issues in a part of the code related to a new feature, that’s a sign it should be rewritten. However, if it involves too many dependencies that also need rewriting, it’s a sign to find a compromise. 

A good example is #2655. It wasn’t purely a new feature; it was a security measure. As a result, I didn’t attempt a full rewrite but instead adjusted my changes accordingly. Unfortunately, my changes ended up introducing a bug. I’m not claiming to be a perfect developer, so I hold myself responsible, but if the overall code quality were better, the likelihood of such mistakes would be lower. 

So, feel free to refactor `cmd/cinder-csi-plugin/main.go`.

> I also believe it’s much better to do this in separate PRs, rather than having one large PR that’s already difficult to review.

It depends. Maybe it makes sense to do this before this PR?

pkg/csi/cinder/utils.go
kayrus: This was just an example, that we need to print a config before starting a service

pkg/csi/cinder/nodeserver.go
kayrus: I don't really like a free-form key definition for the [VolumeContext](https://github.com/search?q=%22req.GetVolumeContext%28%29%22&type=code). Does it make sense to use `cinder.csi.openstack.org` in a key name? CSI driver itself uses the `csi.storage.k8s.io/ephemeral` as a key to mark an ephemeral volume. For the topology we already use https://github.com/kubernetes/cloud-provider-openstack/blob/d069b1ef5d34e1400c798424f5d56b11098fb9a7/pkg/csi/cinder/driver.go#L33-L34

sergelogvinov: Ok, I've changed it to `ResizeRequired = driverName + "/resizeRequired"`

sergelogvinov:
I truly appreciate the opportunity to improve this code (plugin), and we have several suggestions and ideas regarding `cmd/cinder-csi-plugin/main.go`. However, I believe it would be more efficient to implement these changes in a separate PR, as we can do it in parallel with others PRs. 

For now, I have removed any flags or changes related to this file. This PR is currently blocking me, and I am unable to move forward with my other PRs.

Thank you for your understanding!

cmd/cinder-csi-plugin/main.go
```diff
@@ -76,7 +88,12 @@ func main() {
 
 	cmd.PersistentFlags().BoolVar(&provideControllerService, "provide-controller-service", true, "If set to true then the CSI driver does provide the controller service (default: true)")
 	cmd.PersistentFlags().BoolVar(&provideNodeService, "provide-node-service", true, "If set to true then the CSI driver does provide the node service (default: true)")
-	cmd.PersistentFlags().BoolVar(&noClient, "node-service-no-os-client", false, "If set to true then the CSI driver node service will not use the OpenStack client (default: false)")
+
+	cmd.PersistentFlags().StringVar(&searchOrder, "search-order", "configDrive,metadataService", "The search order for metadata service")
+
+	// Node specific flags
+	cmd.PersistentFlags().BoolVar(&rescanOnResize, "rescan-on-resize", false, "If set to true then the CSI driver will rescan the device on volume resize")
+	cmd.PersistentFlags().Int64Var(&nodeVolumeAttachLimit, "node-volume-attach-limit", 256, "The maximum number of volumes that can be attached to a node")
```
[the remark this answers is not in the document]
sergelogvinov: rollback changes

pkg/csi/cinder/controllerserver.go
```diff
@@ -1107,12 +1114,18 @@ func getCreateVolumeResponse(vol *volumes.Volume, ignoreVolumeAZ bool, accessibl
 		}
 	}
 
+	volCnx := map[string]string{}
+	if resizeRequired {
+		volCnx[ResizeRequired] = "true"
+	}
+
```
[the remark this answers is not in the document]
sergelogvinov: fixed

pkg/csi/cinder/driver.go
```diff
@@ -32,6 +32,12 @@ import (
 const (
 	driverName  = "cinder.csi.openstack.org"
 	topologyKey = "topology." + driverName + "/zone"
+
+	// maxVolumesPerNode is the maximum number of volumes that can be attached to a node
+	maxVolumesPerNode = 256
+
+	// ResizeRequired parameter, if set to true, will trigger a resize on mount operation
+	ResizeRequired = "resizeRequired"
```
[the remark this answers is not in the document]
sergelogvinov: This should be made public so that other projects can use it. Many CSI plugins follow this practice.

pkg/csi/cinder/utils.go
sergelogvinov: We removed the CLI arguments, so no code was changed. We can improve the extra logging in another PR.

sergelogvinov:
@dulek, @jichenjc, @kayrus, @mdbooth  Is there anything else I can do to help here?

cmd/cinder-csi-plugin/main.go
```diff
@@ -105,23 +122,26 @@ func handle() {
...
+		//Initialize mount
+		mount := mount.GetMountProvider()
+
+		cfg, err := openstack.GetConfigFromFiles(cloudConfig)
+		if err != nil {
+			if !os.IsNotExist(err) {
+				klog.Warningf("Failed to GetConfigFromFiles: %v", err)
 				return
 			}
 		}
```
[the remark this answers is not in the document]
sergelogvinov: Thanks

cmd/cinder-csi-plugin/main.go
```diff
@@ -105,23 +122,26 @@ func handle() {
...
 
-		//Initialize mount
-		mount := mount.GetMountProvider()
+		// if no search order given, use default
+		if len(cfg.Metadata.SearchOrder) == 0 {
+			cfg.Metadata.SearchOrder = fmt.Sprintf("%s,%s", metadata.ConfigDriveID, metadata.MetadataID)
+		}
```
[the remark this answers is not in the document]
sergelogvinov: Yep, I found only one place where it is used.

Moved to the metadata.GetMetadataProvider func.

kayrus (approved):

kayrus:
@zetaab @dulek if you don't have objections, I'll approve this PR tomorrow

pkg/csi/cinder/nodeserver.go
```diff
@@ -448,13 +372,16 @@ func (ns *nodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	if len(volumePath) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "Volume path not provided")
 	}
+	volCapability := req.GetVolumeCapability()
+	if volCapability != nil {
+		if volCapability.GetBlock() != nil {
+			return &csi.NodeExpandVolumeResponse{}, nil
+		}
+	}
```
kayrus, line 380: Could you please add a comment why do we need this and why do we return `&csi.NodeExpandVolumeResponse{}` once we detect that volume is a block device? Don't we need to expand a block device?

```suggestion

	volCapability := req.GetVolumeCapability()
	if volCapability != nil && volCapability.GetBlock() != nil {
		return &csi.NodeExpandVolumeResponse{}, nil
	}
```

pkg/csi/cinder/nodeserver_test.go
```diff
@@ -335,10 +301,19 @@ func TestNodeExpandVolume(t *testing.T) {
...
 	assert := assert.New(t)
 
+	// setup for test
+	tempDir := os.TempDir()
+	defer os.Remove(tempDir)
+	volumePath := filepath.Join(tempDir, FakeTargetPath)
+	err := os.MkdirAll(volumePath, 0750)
+	if err != nil {
+		t.Fatalf("Failed to set up volumepath: %v", err)
+	}
```
kayrus, line 311:
```suggestion
	// setup for test
	tempDir := os.TempDir()
	volumePath := filepath.Join(tempDir, FakeTargetPath)
	err := os.MkdirAll(volumePath, 0750)
	if err != nil {
		t.Fatalf("Failed to set up volumepath: %v", err)
	}
	defer os.RemoveAll(tempDir)
```

kayrus (commented):
I made yet another review round and have a couple of new question.

pkg/csi/cinder/nodeserver_test.go
sergelogvinov: defer os.RemoveAll(tempDir) good catch. defer os.RemoveAll(tempDir) -> defer os.RemoveAll(volumePath)

os.TempDir() - returns `/tmp` folder, it does not create any folders, os.Remove(tempDir) can remove `/tmp` during the tests...

pkg/csi/cinder/nodeserver.go
```diff
@@ -448,13 +372,15 @@ func (ns *nodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	if len(volumePath) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "Volume path not provided")
 	}
+	volCapability := req.GetVolumeCapability()
+	if volCapability != nil && volCapability.GetBlock() != nil {
+		// The Node plugin only resizes mounted volumes, so we need to skip block unmanaged volumes.
+		return &csi.NodeExpandVolumeResponse{}, nil
+	}
```
kayrus, line 379: what makes you think that `GetBlock() != nil` means that the volume is unmanaged? For me it looks like an unrelated bug fix. And to be honest I don't understand the bug description. For me it looks like this change breaks volume resize behavior.
https://github.com/kubernetes-csi/external-resizer/blob/f387899b5a5af4e601a93f1c963aa646f4e54c1b/pkg/resizer/csi_resizer.go#L208-L229 - the `VolumeCapability` is always not nil for block devices.

sergelogvinov: Hi, the resizer only works with the CSI-controller.

The CSI-node runs `resize2fs` or similar tools for formatted (managed) disks. The pod takes care of managing the block device

kayrus: > Hi, the resizer only works with the CSI-controller.

right, the controller generates the `NodeExpandVolume` call, which is then handled by the node-driver. Anyway, I took a look at the AWS CSI controller and it has the same logic. I was confused about the block type. I associated block type with all cinder volumes and fstype with manila shared filesystem storage. Now I see that block type is a raw volume without specified fs.

Anyway, your fix needs to be implemented in a separate PR, since there is a case, when you have a raw volume being resized, and some hypervisors require a block device rescan to notify the guest OS with a new block device size. And since the `RescanBlockDeviceGeometry` func relies on the `volumePath`, there must be a new func that just rescans the block device without checking the FS size.

sergelogvinov: Ok, i've remove it.

kayrus: @sergelogvinov can you open a new issue explaining this use case?

PR #2653, opened 2024-09-17
[cinder-csi-plugin] enable fs resize for any volume for Stage Volume Step

fahedouch:

enable resize fs for all volumes.

I'm in a scenario where the volume source is a block storage device, not a snapshot or another volume. We want to enable filesystem resizing to fully utilize the entire disk in this case.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/nodeserver.go
+++ b/pkg/csi/cinder/nodeserver.go
@@ -251,14 +251,6 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		return nil, status.Error(codes.InvalidArgument, "NodeStageVolume Volume Capability must be provided")
 	}
 
-	vol, err := ns.Cloud.GetVolume(volumeID)
-	if err != nil {
-		if cpoerrors.IsNotFound(err) {
-			return nil, status.Error(codes.NotFound, "Volume not found")
-		}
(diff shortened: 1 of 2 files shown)


kayrus:
> I'm in a scenario where the volume source is a block storage device

the resize is required only when a volume has a FS, and such cases are possible only when a volume was created from a backup or a snapshot. Can you provide a step-by-step instructions on how to reproduce an issue, when a volume already has a FS and requires a resize?

And please sign an CLA.

fahedouch:
> > I'm in a scenario where the volume source is a block storage device
> 
> the resize is required only when a volume has a FS, and such cases are possible only when a volume was created from a backup or a snapshot. Can you provide a step-by-step instructions on how to reproduce an issue, when a volume already has a FS and requires a resize?
> 
> And please sign an CLA.
> 
> P.S. The FS resize without the snapshot/backup conditions happens in the `NodeExpandVolume` [method](https://github.com/kubernetes/cloud-provider-openstack/blob/2d98e8823099fb51a905b8173cf714934e04c18c/pkg/csi/cinder/nodeserver.go#L440).

thanks for your reactivity @kayrus , Indeed creating a fresh fs from disk will get the disk size by default so no need to resize it. let me close this PR for the moment

PR #2696, opened 2024-10-21
[occm] remove SGs from non-existent nodes ports on Update

### LINKED ISSUE #2695: Remove SGs from non-existent nodes ports on Update
Current code assumes that if a Node has been removed from the list on `UpdateLoadBalancer()`, we don't care about it's security group.
We could update `applyNodeSecurityGroupIDForLB()` to also lookup ports that have the SG tag and detach the SG from the ports that aren't belonging to the ports of the current list of Nodes.

https://github.com/kubernetes/cloud-provider-openstack/blob/04bd8e1343d630a0db3e9093dd71f59e477ee7ad/pkg/openstack/loadbalancer_sg.go#L72-L99

EmilienM:

Update `applyNodeSecurityGroupIDForLB()` to also lookup ports that have the SG tag and detach the SG from the ports
that aren't belonging to the ports of the current list of Nodes.

This will ensure ports are cleaned up if a Node is being removed from a Loadbalancer.

fixes #2695

**Release note**:
```release-note
When a Node is removed from a load balancer, we'll now ensure that its ports don't have the security group associated
with the load balancer.
```


pkg/openstack/loadbalancer_sg.go
```diff
@@ -63,7 +63,7 @@ func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, svcConf *
 			continue
 		}
 
-		listOpts := neutronports.ListOpts{DeviceID: serverID}
+		listOpts := neutronports.ListOpts{}
```
EmilienM, line 66: IMO this is an expensive request. I wonder if we can do better. The reason why I did it this way is to limit the number of API calls, but now this list might be really big in some deployments.

An alternative could be to make two queries: one that lists by SG and one that list by server ID.

pkg/openstack/loadbalancer_sg.go
```diff
@@ -75,6 +75,23 @@ func applyNodeSecurityGroupIDForLB(network *gophercloud.ServiceClient, svcConf *
 				continue
 			}
 
+			if port.DeviceID != serverID {
```
MaysaMacedo, line 78: should we have a check similar to line 101?

EmilienM: Good point. I actually think we can move the new block after L101, which will help to leave the loop earlier if the port isn't part of the LB's subnet.

pkg/openstack/loadbalancer_sg.go
kayrus: This will definitely increase the load

EmilienM: an alternative could be another loop where we filter by SGs and update the ports... but it's also some load.

EmilienM:
This hasn't got any traction, and we even have concerns about it so I'll just close it for now.
/close

PR #2715, opened 2024-11-14
[docs] add missing parameters to StorageClass in multi-region-clouds.md

clsv:
Added missing parameters csi.storage.k8s.io/controller-expand-secret-name and csi.storage.k8s.io/controller-expand-secret-namespace to the StorageClass configuration in the Multi-AZ/Region/OpenStack documentation. These parameters were missing, causing issues with PVC expansion. After some time spent troubleshooting, I found these parameters in the [Kubernetes CSI documentation on StorageClass secrets](https://kubernetes-csi.github.io/docs/secrets-and-credentials-storage-class.html#storageclass-secrets). After adding them, PVC expansion works as expected.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/cinder-csi-plugin/multi-region-clouds.md
+++ b/docs/cinder-csi-plugin/multi-region-clouds.md
@@ -17,7 +17,7 @@ stringData:
     [BlockStorage]
     bs-version=v3
     ignore-volume-az=True
-    
+
     [Global]
     auth-url="https://auth.cloud.openstackcluster.region-default.local/v3"
     username="region-default-username"
@@ -26,7 +26,7 @@ stringData:
     tenant-id="region-default-tenant-id"
     tenant-name="region-default-tenant-name"
     domain-name="Default"
-    
+
     [Global "region-one"]
     auth-url="https://auth.cloud.openstackcluster.region-one.local/v3"
     username="region-one-username"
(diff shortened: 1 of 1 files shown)


clsv:
It looks like all tests are okay. :-)

docs/cinder-csi-plugin/multi-region-clouds.md
```diff
@@ -101,6 +101,8 @@ parameters:
   csi.storage.k8s.io/node-stage-secret-namespace: kube-system
   csi.storage.k8s.io/provisioner-secret-name: openstack-config-region-one
   csi.storage.k8s.io/provisioner-secret-namespace: kube-system
+  csi.storage.k8s.io/controller-expand-secret-name: openstack-config-region-one
+  csi.storage.k8s.io/controller-expand-secret-namespace: kube-system
```
kayrus, line 105: For those, who have questions on how does it work.

These secrets (basically they are not secrets, but names of the cloud region from the cloud-config to use) are read by one of the CSI sidecar controllers (e.g. provisioner, resizer, etc), then passed through the CSI gRPC interface to the Cinder CSI controller / node-server method. CSI controller/node-server has a parsed cloud-config with multiple region names. If the passed cloud region name exists in the config map, it uses it.

In future we'll probably move secrets content into the actual `secret-name` to properly follow the CSI spec way.

PR #2688, opened 2024-10-14
[core] improved readability of lb funcs, switch to slices.Contains

kayrus:
**What this PR does / why we need it**:

Fixed nested conditions logic and switched to `slices.Contains`, removed `Contains` func from util package.

**Which issue this PR fixes(if applicable)**:

n/a

**Special notes for reviewers**:

**Release note**:

```release-note
NONE
```

PR #2719, opened 2024-11-18
[occm] fix node internal/external IP order

### LINKED ISSUE #2718: [OCCM] Octavia Regression - Node's ExternalIP is prefered instead of the InternalIP one
/kind bug

**What happened**:
The default behavior of choosing the InternalIP of a Node, as the Octavia Member's IP, is not working. The ExternalIP is preferred by default.

The InternalIP should be used first if exists. If not, the ExternalIP should be return (as explain in the comments).
https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/openstack/loadbalancer.go#L376

**How to reproduce it**:
Have K8S Nodes with an ExternalIP and an InternalIP filled in the Status.

Introduce by #2688 

**Environment**:
OpenStack instances with External and Internal IPs.

kayrus:

This PR fixes a regression introduced in #2688, unit tests ensure that the internal/external IP order works as expected.

fixes #2718 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -387,22 +387,22 @@ func nodeAddressForLB(node *corev1.Node, preferredIPFamily corev1.IPFamily) (str
 	}
 
 	allowedAddrTypes := []corev1.NodeAddressType{corev1.NodeInternalIP, corev1.NodeExternalIP}
-	for _, addr := range addrs {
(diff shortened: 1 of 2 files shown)


Lucasgranet (approved):

kayrus:
@zetaab @dulek ready for review

zetaab (approved):
/approve
/lgtm

PR #2729, opened 2024-11-20
[cinder-csi-plugin] Validate volume capabilities in RPC call `CreateVolume`

nschad:

Cinder currently only supports `VolumeCapability_AccessMode_SINGLE_NODE_WRITER`, therefore we should not allow it to provision volumes with AccessMode `RWX`. This will only lead to "errors" when replica > 1.

Additionally, the [spec](https://github.com/container-storage-interface/spec/blob/c7ab2f379311a9ae234ae42bdfbfebfa9c15e538/spec.md?plain=1#L804C61-L806C28) states (as I understand it) that the `CreateVolume` RPC call must support all specified volume capabilities, which was not the case previously. So this is why I added the check here.

Example of Error Message in K8s

```
 Warning  ProvisioningFailed    2s (x2 over 3s)  cinder.csi.openstack.org_csi-driver-controller-foo  failed to provision volume with StorageClass "bar": rpc error: code = InvalidArgument desc = volume access mode MULTI_NODE_MULTI_WRITER not supported
```

fixes #

**Release note**:

```release-note
[cinder-csi-plugin] Enforce VolumeCapabilities when creating volumes.
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -51,6 +51,15 @@ const (
 	antiAffinityKey       = "cinder.csi.openstack.org/anti-affinity"
 )
 
+func (cs *controllerServer) validateVolumeCapabilities(req []*csi.VolumeCapability) error {
+	for _, volCap := range req {
+		if volCap.GetAccessMode().GetMode() != cs.Driver.vcap[0].GetMode() {
+			return fmt.Errorf("volume access mode %s not supported", volCap.GetAccessMode().GetMode().String())
+		}
+	}
+	return nil
+}
+
 func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
 	klog.V(4).Infof("CreateVolume: called with args %+v", protosanitizer.StripSecrets(*req))
 
@@ -72,6 +81,10 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 
 	if volCapabilities == nil {
 		return nil, status.Error(codes.InvalidArgument, "[CreateVolume] missing Volume capability")
+	}
+
+	if err := cs.validateVolumeCapabilities(volCapabilities); err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
 	// Volume Size - Default is 1 GiB


kayrus:
@nschad thanks for the PR. can you share a detailed step-by-step instructions on how to reproduce a problem?

nschad:
> @nschad thanks for the PR. can you share a detailed step-by-step instructions on how to reproduce a problem?

Sure

1. Create a Deployment with an Volume in `RWX` mode.

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: simple-deployment
spec:
  replicas: 2
  selector:
    matchLabels:
      app: simple-app
  template:
    metadata:
      labels:
        app: simple-app
    spec:
      containers:
      - name: simple-container
        image: nginx
        volumeMounts:
        - mountPath: /mnt/data
          name: simple-volume
      volumes:
      - name: simple-volume
        persistentVolumeClaim:
          claimName: simple-pvc

---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: simple-pvc
spec:
  accessModes:
    - ReadWriteMany  # RWX access mode
  resources:
    requests:
      storage: 5Gi
```

2. Watch the Volume be successfully bounded to Pod 1 and embrace the error you will definitely get when your Pod 2 on Node B can't attach the volume, because cinder doesn't do MULTI_NODE_MULTI_POD

kayrus:
> In my opinion the volume shouldn't be provisioned in the first place with AccessMode `RWX`. That is exactly what this PR does.

I see, but I wonder whether this can be considered as a default behavior for someone else. Do you know what's the behavior of other CSI drivers, e.g. GCE or AWS?

nschad:
> > In my opinion the volume shouldn't be provisioned in the first place with AccessMode `RWX`. That is exactly what this PR does.
> 
> I see, but I wonder whether this can be considered as a default behavior for someone else. Do you know what's the behavior of other CSI drivers, e.g. GCE or AWS?

Yes, for example:

- https://github.com/digitalocean/csi-digitalocean/blob/master/driver/controller.go#L96-L98
- AWS-EBS doing it in ControllerPublishVolume https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/a25f651b6f2e39ba1ef79dbac4d638f36ad62292/pkg/driver/controller.go#L398-L400


> UPD: basically I expect that csi-provisioner handles this and doesn't allow to create a `ReadWriteMany` volume based on the CSI plugin capabilities.

Yeah maybe but it's just not the reality currently. There is the RPC call `ValidateVolumeCapabilities` (also implemented here) but that only gets called AFAIK for pre-provisioned volumes which is for us like 0.01% of the use case.

kayrus:
@nschad in order to proceed, can you sign the CLA?

nschad:
> > @nschad in order to proceed, can you sign the CLA?
> 
> Yep... I haven't forgotten about it. Unfortunately I'm waiting for my company 🙈

CLA issue is finally resolved. PTAL

Also happy to discuss whether this is the way it's meant to be implemented, because there seems to be different options.

PR #2721, opened 2024-11-19
[tests] bump ubuntu to 24.04 jammy

kayrus:
**What this PR does / why we need it**:

Bump test env to ubuntu 24.04 jammy, bump k3s to v1.31.2+k3s1

**Which issue this PR fixes(if applicable)**:
fixes #

**Special notes for reviewers**:

**Release note**:

```release-note
NONE
```

PR #2731, opened 2024-11-28
Revert "[tests] bump ubuntu to 24.04 noble"

kayrus:
Reverts kubernetes/cloud-provider-openstack#2721

Let's roll this back to 22.04 jammy, since devstack starts to complain


stephenfin:
Would it not make more sense to bump the DevStack version? The version we're using currently (2023.2) is rather old and will be reaching EOL shortly. I can take a look at this if you want.

kayrus:
/hold
let's hold for a while and see how further tests are running against the 24.04, looks like they're fine again.

stephenfin:
@kayrus Can we close this now?

stephenfin:
/close

Turns out we weren't even using Ubuntu 24.04 https://github.com/kubernetes/cloud-provider-openstack/pull/2742/commits/93417dc3d297ed3162f78e518c09b1e22e2ec465 so this PR wasn't doing anything. The issue lay elsewhere and is being fixed by #2742.

PR #2692, opened 2024-10-16
[occm] remove instances v1 support

kayrus:
**What this PR does / why we need it**:

This is a code cleanup. The instances v1 cloud controller API support is removed. Instance address logic was moved to a separate `pkg/openstack/instances_addresses.go` file. The `pkg/openstack/instancesv2.go` was renamed to `pkg/openstack/instances.go`. Added context awareness to instances related functions and methods. Added a `SplitTrim` util function to extract comma-separated values from a string without a `strings.TrimSpace` call.

**Which issue this PR fixes(if applicable)**:

n/a

**Special notes for reviewers**:

**Release note**:

```release-note
[occm] remove instances v1 support
```

PR #2732, opened 2024-12-01
[occm]: remove InstancesV1 Zones method

sergelogvinov:

Since InstancesV1 was removed, we need to clean up the zone methods.
Zones() uses only if InstancesV1 interface exists. https://github.com/kubernetes/cloud-provider/blob/master/controllers/node/node_controller.go#L579-L606

All zone information expose by InstancesV2 in InstanceMetadata()
And

https://github.com/kubernetes/cloud-provider/blob/master/cloud.go#L274

```go
// DEPRECATED: Zones is deprecated in favor of retrieving zone/region information from InstancesV2.
// This interface will not be called if InstancesV2 is enabled.
type Zones interface {
```

#2692

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/openstack.go
+++ b/pkg/openstack/openstack.go
@@ -25,13 +25,11 @@ import (
 	"time"
 
 	"github.com/gophercloud/gophercloud/v2"
-	"github.com/gophercloud/gophercloud/v2/openstack/compute/v2/servers"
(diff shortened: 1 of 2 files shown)


sergelogvinov:
Thank you @kayrus for removing instanceV1 interface.

Can you take a look at my removal as well? You might have missed this part of the code.

kayrus:
@sergelogvinov you're right, I missed this, since `Zones()`.
/ok-to-test

PR #2733, opened 2024-12-03
add f5 as supported LB

archerwu9425:
Add `f5` to `supportedLBProvider `
Openstack api support `f5` as LB provider, but OCCM will block at first place as `f5` is not in the supported list.

fixes #

1. In OCCM cloud-config, add below setting:
```
[LoadBalancer]
lb-provider = f5
```
OCCM can start without error

2. Add LB service, openstack with f5 can create LB

**Release note**:

```release-note
Add f5 to supportedLBProvider
```


kayrus:
@archerwu9425 thanks for the PR. Can you also adjust the https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md accordingly?

PR #2758, opened 2025-01-13
[manila-csi-plugin] make auth more tolerant

### LINKED ISSUE #2757: Manila CSI unnecessarily picky about openstack creds
While testing a Manila change on an OpenStack environment with application credentials, I got a successful install however Manila CSI was reporting issues when trying to provision a volume:

```
E1119 16:11:03.280195       1 controller.go:957] error syncing claim "e5eda1ee-29c2-4aef-bfd5-146c0da60439": failed to provision volume with StorageClass "csi-manila-ceph": rpc error: code = InvalidArgument desc = invalid OpenStack secrets: parameter 'os-userName' requires 'os-password'
```

And

```
E1119 16:12:28.492969       1 controller.go:957] error syncing claim "e5eda1ee-29c2-4aef-bfd5-146c0da60439": failed to provision volume with StorageClass "csi-manila-ceph": rpc error: code = InvalidArgument desc = invalid OpenStack secrets: parameter 'os-domainName' requires exactly one of [os-password os-clientCertPath] parameters
```

It turns out, I had unneeded attributes in my clouds.yaml that caused Manila-CSI to fail.

```
openstack:
  auth:
    application_credential_id: xxx
    application_credential_secret: xxx
    auth_url: xxx
    user_domain_name: xxx # Not necessary
    username: xxx # Not necessary
  auth_type: v3applicationcredential
  endpoint_type: public
  identity_api_version: "3"
  region_name: regionOne
```

Removing `user_domain_name` and `username` from my `clouds.yaml` allowed Manila to provision volumes.

Manila-CSI should be able to ignore these extra attributes, like the CLI openstack client does.

mandre:
**What this PR does / why we need it**: Remove dependency on a couple of fields where the logic was wrong. For example, these fields do not necessarily depend on a password being set, as we could be using application credentials.

This prevents manila driver from entering an error state when it finds unnecessary fields in the clouds.yaml. It now simply ignores them.

Fixes #2757

The fix is in `pkg/client/client.go` so presumably affects all binaries?

**Release note**:

```release-note
Make authentication more tolerant to unneeded fields
```

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -38,21 +38,21 @@ import (
 
 type AuthOpts struct {
 	AuthURL          string                   `gcfg:"auth-url" mapstructure:"auth-url" name:"os-authURL" value:"optional" dependsOn:"os-password|os-trustID|os-applicationCredentialSecret|os-clientCertPath"`
-	UserID           string                   `gcfg:"user-id" mapstructure:"user-id" name:"os-userID" value:"optional" dependsOn:"os-password"`
-	Username         string                   `name:"os-userName" value:"optional" dependsOn:"os-password"`
-	Password         string                   `name:"os-password" value:"optional" dependsOn:"os-domainID|os-domainName,os-projectID|os-projectName,os-userID|os-userName"`
-	TenantID         string                   `gcfg:"tenant-id" mapstructure:"project-id" name:"os-projectID" value:"optional" dependsOn:"os-password|os-clientCertPath"`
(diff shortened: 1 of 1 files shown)


kayrus:
I never had issues with application credentials and manila CSI. Are you sure this is not related to a cloud.yaml parser code?

stephenfin:
I stumbled into this today and can confirm it's a real issue (I think it was the same root cause also: I had an errant username field set with application credentials).

We will get good error messages from gophercloud if it's unable to authenticate due to missing fields, so I don't think we lose anything here. I think we should get this in.

/approve
/lgtm

Issue #2764, opened 2025-02-04
[cinder-csi-plugin] Cinder-csi-plugin does not respect pagination from Cinder API on list volumes call


kon-angelo:
We are using cinder-csi-plugin to create and manage volumes.

In the "controller"-server the `external-attacher` is running with `--max-entries` disabled (aka equal to `0`).

The issue is as follows:

The external-attacher is spamming this message for **most** volumes:
```
"\"VolumeAttachment attached status and actual state do not match.
Adding back to VolumeAttachment queue for forced reprocessing\"
VolumeAttachment=\"csi-4667c27a88346ef604b35263eae2f71bbfebf173ee80a8040d1c9d4fa176e1e2\" 
volumeHandle=\"6c0da8b7-5d88-4c95-9200-1c4fb294fb2b\" attachedStatus=true found=false"
```
This in turn makes the attacher issuing force "re-processing" of existing attachments, hammering both KAPI and the Cinder API.

Turns out that the cinder-csi is making a list call without using `AllPages`. I am not sure if this is a peculiarity of our openstack environment, but the call in https://github.com/kubernetes/cloud-provider-openstack/blob/1665392bf2d248747d8d744540bf660c39235d78/pkg/csi/cinder/openstack/openstack_volumes.go#L87

**does not** list all volumes, even when `limit == 0`. Instead it returns a paginated response and the next token. 

The following logs outline this better:

```
csi-driver-controller-6686c76dd-86mm8 openstack-csi-driver I0204 19:48:47.522252      11 controllerserver.go:384] ListVolumes: called with &csi.ListVolumesRequest{MaxEntries:0, StartingToken:"", XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0} request
csi-driver-controller-6686c76dd-86mm8 openstack-csi-driver I0204 19:48:51.277343      11 controllerserver.go:448] ListVolumes: retrieved 1000 entries and "69795aaf-9357-4212-ab4e-e50dc7ca5752" next token from cloud ""
csi-driver-controller-6686c76dd-86mm8 openstack-csi-driver I0204 19:48:51.277384      11 controllerserver.go:503] ListVolumes: completed with 1000 entries and "" next token
```

In
https://github.com/kubernetes/cloud-provider-openstack/blob/1665392bf2d248747d8d744540bf660c39235d78/pkg/csi/cinder/controllerserver.go#L500
[...]

PR #2766, opened 2025-02-05
[cinder-csi-plugin] Refactor list volumes

kon-angelo:
Fixes the linked issue and in addition does a small refactoring simplifying the code for the list volumes. In particular it changes the behavior slightly in 2 ways:
- if the list operation returns any page token even with max-entries == 0, we will respect the token and return it in the response.
- in the case of multiple cloud envs, we will no longer do the discovery for "the first cloud with volumes", instead we return the next cloud env. The discovery costs additional Cinder API calls which are more costly than simply issuing an additional local grpc call from the external-attacher.

Regarding point 2: the external-attacher will continue sending requests to list volumes while the response contains a token:

https://github.com/kubernetes-csi/external-attacher/blob/e5ae92e87bb6c3d99061ca515285c5c9e76e0414/pkg/attacher/lister.go#L63-L65

With this change, we will expect the csi-attacher to issue a request per cloud in our configuration. 

fixes https://github.com/kubernetes/cloud-provider-openstack/issues/2764

**Release note**:

```release-note
Fix an issue with the `ListVolumes` not respecting pagination tokens from Cinder API with `req.maxEntries` is set to 0.
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -32,18 +32,24 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
-
-	"k8s.io/klog/v2"
-
 	sharedcsi "k8s.io/cloud-provider-openstack/pkg/csi"
 	"k8s.io/cloud-provider-openstack/pkg/csi/cinder/openstack"
 	"k8s.io/cloud-provider-openstack/pkg/util"
 	cpoerrors "k8s.io/cloud-provider-openstack/pkg/util/errors"
+	"k8s.io/klog/v2"
 )
 
 type controllerServer struct {
 	Driver *Driver
 	Clouds map[string]openstack.IOpenStack
+}
+
+type cloudsStartingToken struct {
+	// CloudName is the cloud that we have to list next.
+	CloudName string `json:"cloud"`
+	// Token is the pagination token returned by the last Cinder list volume operation. If empty, we will list all volumes
+	// from the beginning.
+	Token string `json:"token"`
 }
 
 const (
@@ -403,11 +409,6 @@ func (cs *controllerServer) ControllerUnpublishVolume(ctx context.Context, req *
 	return &csi.ControllerUnpublishVolumeResponse{}, nil
 }
 
-type CloudsStartingToken struct {
-	CloudName string `json:"cloud"`
-	Token     string `json:"token"`
-}
-
 func (cs *controllerServer) extractNodeIDs(attachments []volumes.Attachment) []string {
 	nodeIDs := make([]string, len(attachments))
 	for i, attachment := range attachments {
@@ -441,10 +442,7 @@ func (cs *controllerServer) ListVolumes(ctx context.Context, req *csi.ListVolume
 	maxEntries := int(req.MaxEntries)
 
 	var err error
-	var cloudsToken = CloudsStartingToken{
-		CloudName: "",
-		Token:     "",
-	}
+	var cloudsToken = cloudsStartingToken{}
 
 	cloudsNames := maps.Keys(cs.Clouds)
 	sort.Strings(cloudsNames)
@@ -466,7 +464,9 @@ func (cs *controllerServer) ListVolumes(ctx context.Context, req *csi.ListVolume
 		}
 		idx++
 	}
-	volumeList, nextPageToken, err := cs.Clouds[cloudsNames[idx]].ListVolumes(ctx, maxEntries, startingToken)
+
+	var volumeList []volumes.Volume
+	volumeList, cloudsToken.Token, err = cs.Clouds[cloudsNames[idx]].ListVolumes(ctx, maxEntries, startingToken)
 	if err != nil {
 		klog.Errorf("Failed to ListVolumes: %v", err)
 		if cpoerrors.IsInvalidError(err) {
@@ -475,12 +475,11 @@ func (cs *controllerServer) ListVolumes(ctx context.Context, req *csi.ListVolume
 		return nil, status.Errorf(codes.Internal, "ListVolumes failed with error %v", err)
 	}
 	volumeEntries := cs.createVolumeEntries(volumeList)
-	klog.V(4).Infof("ListVolumes: retrieved %d entries and %q next token from cloud %q", len(volumeEntries), nextPageToken, cloudsNames[idx])
+	klog.V(4).Infof("ListVolumes: retrieved %d entries and %q next token from cloud %q", len(volumeEntries), cloudsToken.Token, cloudsNames[idx])
 
-	cloudsToken.Token = nextPageToken
 	switch {
 	// if we have not finished listing all volumes from this cloud, we will continue on next call.
-	case nextPageToken != "":
+	case cloudsToken.Token != "":
 		// if we listed all volumes from this cloud but more clouds exist, return a token of the next cloud.
 	case idx+1 < len(cloudsNames):
 		cloudsToken.CloudName = cloudsNames[idx+1]
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -32,6 +32,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
+
 	sharedcsi "k8s.io/cloud-provider-openstack/pkg/csi"
 	"k8s.io/cloud-provider-openstack/pkg/csi/cinder/openstack"
 	"k8s.io/cloud-provider-openstack/pkg/util"
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -18,8 +18,8 @@ package cinder
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
+	"slices"
 	"sort"
 	"strconv"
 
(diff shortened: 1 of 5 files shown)


kayrus (commented):
Thanks for the PR. It took me a while to understand this logic, since I didn't review the initial #2551 and I don't have environment with a multi-region setup. See my comments below.

pkg/csi/cinder/controllerserver.go
```diff
@@ -458,104 +457,50 @@ func (cs *controllerServer) ListVolumes(ctx context.Context, req *csi.ListVolume
...
-			if cloudsNames[i] == currentCloudName {
-				currentCloudName = cloudsNames[i+1]
-				break
-			}
-		}
-	}
-
-	startIdx := 0
+	idx := 0
 	for _, cloudName := range cloudsNames {
```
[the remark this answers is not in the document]
MatthieuFin: Hi @kayrus I'm currently on vacation but I can give you some hint from my memories, usage of "secrets" with `volCloud := req.GetSecrets()["cloud"]` method is more a hack than a feature as explain on my [PR](https://github.com/kubernetes/cloud-provider-openstack/pull/2551), I got inspired by discussion in this [issue](https://github.com/kubernetes-csi/external-provisioner/issues/87) where usage of secrets fields in gRPC csi spec to store current cloud.

So now that clear, we can simply use "secrets" field in gRPC call for `ListSnapshots` because csi spec provide a 'secrets" field in gRPC csi spec [here](https://github.com/container-storage-interface/spec/blob/master/spec.md#listsnapshots) that is fully convenient. This secret is filled by informations provided in StorageClass (I guess we list Snapshot of a SC or snapshot of a defined volume but never globally all snapshots)

But unfortunatelly (I don't remember why, but I guess that there is an explanation) in `ListVolume` call we implement gRPC call spec [ListVolumesRequest](https://github.com/container-storage-interface/spec/blob/master/spec.md#listvolumes) where there is no `secrets` field but only `max_entries` and `starting_token` so as we don't have possibility to know which cloud is concerned (and that make sens we wanna list "k8s" volumes which concern all our clouds) we have to loop over each clouds, and I use "token" in JSON format (if I remember right) to store processed cloud and keep a "state" if ListVolume use pagination, in that way if a client provide a `max_entries != 0` the token permit us to know which cloud we have to call to retrieve next volumes.

To be honest that not my proudest implementation (ListVolume function), I have maybe made some error, but I didn"t notice it for now, I'll take a look on this issue on my side when i'll back to work.

MatthieuFin:
I quickly read @kon-angelo's `ListVolumes` implementation and it's much simpler and readable, I focused on providing the requested max-entries but I agree with your approach, it seems more maintainable this way.

pkg/csi/cinder/controllerserver.go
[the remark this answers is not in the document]
kon-angelo: @MatthieuFin I think this is correct. There is no way to get the `cloud` info as there is no secrets field in the request - you simply need to list everything from your current configuration.

@kayrus The reason to loop over the `cloud` names is  because we are creating a stable index of all the available `clouds`. Henceforth, we can know from the token which is the next cloud we need to list volumes for. I also don't see (at the moment at least) another way to know this info.

pkg/csi/cinder/controllerserver.go
```diff
@@ -32,19 +32,27 @@ import (
...
 	cpoerrors "k8s.io/cloud-provider-openstack/pkg/util/errors"
-	"k8s.io/klog/v2"
 )
 
 type controllerServer struct {
 	Driver *Driver
 	Clouds map[string]openstack.IOpenStack
 }
 
+type cloudsStartingToken struct {
```
kayrus, line 48: I also wonder whether it's better to use `<token>:<cloud>` format and use `SplitN` along with `Join` to avoid useless json encoding/decoding.

pkg/csi/cinder/controllerserver.go
```diff
@@ -33,11 +33,12 @@ import (
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
+	"k8s.io/klog/v2"
```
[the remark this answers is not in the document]
kayrus: kindly ping

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -82,8 +83,18 @@ func (os *OpenStack) ListVolumes(ctx context.Context, limit int, startingToken s
 	var nextPageToken string
 	var vols []volumes.Volume
 
-	opts := volumes.ListOpts{Limit: limit, Marker: startingToken}
 	mc := metrics.NewMetricContext("volume", "list")
+	opts := volumes.ListOpts{Limit: limit, Marker: startingToken}
+	if limit == 0 {
+		page, err := volumes.List(os.blockstorage, opts).AllPages(ctx)
```
kayrus, line 89: in this case there is no need to provide opts
```suggestion
		page, err := volumes.List(os.blockstorage, nil).AllPages(ctx)
```

pkg/csi/cinder/controllerserver.go
```diff
@@ -32,7 +32,6 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
-
```
kayrus, line 35: please leave it as is. this will help to avoid issues in backporting.

pkg/csi/cinder/openstack/openstack_volumes.go
kon-angelo: Should be addressed now

pkg/csi/cinder/controllerserver.go
kon-angelo: I will provide a commit that changes the token to the colon-separated one. We don't care about any kind of backwards-compatibility here so it should be straightforward to change

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -82,8 +82,19 @@ func (os *OpenStack) ListVolumes(ctx context.Context, limit int, startingToken s
...
 	mc := metrics.NewMetricContext("volume", "list")
+	if limit == 0 {
+		page, err := volumes.List(os.blockstorage, nil).AllPages(ctx)
+		if err != nil {
+			return nil, "", err
+		}
+		vols, err = volumes.ExtractVolumes(page)
+		if mc.ObserveRequest(err) != nil {
+			return vols, "", err
+		}
```
kayrus, line 94:
```suggestion
		}
	return vols, "", nil
```

pkg/csi/cinder/utils_test.go
```diff
@@ -146,3 +146,23 @@ func TestLogGRPC(t *testing.T) {
...
 	}
 }
+func TestSplitToken(t *testing.T) {
+	tests := []struct {
+		input string
+		token string
+		cloud string
+	}{
+		{input: "foo:bar", token: "foo", cloud: "bar"},
+		{input: "foo", token: "foo", cloud: ""},
```
kayrus, line 156: just in case
```suggestion
		{input: "foo", token: "foo", cloud: ""},
		{input: "foo:", token: "foo", cloud: ""},
```

pkg/csi/cinder/utils_test.go
```diff
@@ -146,3 +146,23 @@ func TestLogGRPC(t *testing.T) {
...
+		{input: ":bar", token: "", cloud: "bar"},
+		{input: "foo:bar:baz", token: "foo", cloud: "bar:baz"},
```
kayrus, line 158:
```suggestion
		{input: "foo:bar:baz", token: "foo", cloud: "bar:baz"},
		{input: "", token: "", cloud: ""},
```

pkg/csi/cinder/controllerserver.go
```diff
@@ -439,123 +433,56 @@ func (cs *controllerServer) ListVolumes(ctx context.Context, req *csi.ListVolume
...
-			if sendEmptyToken {
-				data = []byte("")
-			}
-			klog.V(4).Infof("ListVolumes: completed with %d entries and %q next token", len(cloudsVentries), string(data))
-			return &csi.ListVolumesResponse{
-				Entries:   cloudsVentries,
-				NextToken: string(data),
-			}, nil
+	var volumeList []volumes.Volume
+	volumeList, token, err = cs.Clouds[cloudsNames[idx]].ListVolumes(ctx, maxEntries, token)
```
kayrus, line 455: we can simply use `cloudName` here
```suggestion
	volumeList, token, err = cs.Clouds[cloudName].ListVolumes(ctx, maxEntries, token)
```

pkg/csi/cinder/controllerserver.go
```diff
@@ -439,123 +433,56 @@ func (cs *controllerServer) ListVolumes(ctx context.Context, req *csi.ListVolume
...
+	if err != nil {
+		klog.Errorf("Failed to ListVolumes: %v", err)
+		if cpoerrors.IsInvalidError(err) {
+			return nil, status.Errorf(codes.Aborted, "[ListVolumes] Invalid request: %v", err)
 		}
+		return nil, status.Errorf(codes.Internal, "ListVolumes failed with error %v", err)
+	}
+	volumeEntries := cs.createVolumeEntries(volumeList)
+	klog.V(4).Infof("ListVolumes: retrieved %d entries and %q next token from cloud %q", len(volumeEntries), token, cloudsNames[idx])
```
kayrus, line 464: and here too
```suggestion
	klog.V(4).Infof("ListVolumes: retrieved %d entries and %q next token from cloud %q", len(volumeEntries), token, cloudName)
```

kayrus (commented):
Thanks, @kon-angelo! Almost perfect. I just left some nit comments, and I hope we're ready to go.

PR #2852, opened 2025-03-02
Add missing release namespace for manila charts

mnaser:
The manila charts are missing the namespace and when templating
things with Helm, it will omit it and be problematic.  The cinder
chart has the namespaces set everywhere and its better to be
explicit so this adds it to the manila chart.

fixes #

**Release note**:

```release-note
NONE
```


gouthampacha:
/ok-to-test

@mnaser needs a version bump on https://github.com/kubernetes/cloud-provider-openstack/blob/master/charts/manila-csi-plugin/Chart.yaml#L5

PR #2855, opened 2025-03-05
Align Cinder-plugin Helm Chart Labels with recommended ones by K8s.

jonitich:
**What this PR does / why we need it**: Aligning labels to Kubernetes standard ones. https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels.
Rendered file: [diff.txt](https://github.com/user-attachments/files/19090280/diff.txt)

fixes #

**Release note**:

```release-note
NONE
```


mnaser (changes requested):
The concern I have with this is that this would have to be a major release bump, due to the fact that mismatching labels will make it so things cannot deploy anymore.

jonitich:
> The concern I have with this is that this would have to be a major release bump, due to the fact that mismatching labels will make it so things cannot deploy anymore.

hi @mnaser, yes indeed, changing those labels will make a rollout restart of the pods due to pod labels' immutability.

mnaser:
> > The concern I have with this is that this would have to be a major release bump, due to the fact that mismatching labels will make it so things cannot deploy anymore.
> 
> hi @mnaser, yes indeed, changing those labels will make a rollout restart of the pods due to pod labels' immutability.

I don't think it'll actually cause a rolling restart but it will just simply fail due to the fact that the pod template labels are immutable on the deployment level AFAIK.

PR #2860, opened 2025-03-14
[cinder-csi-plugin]: ResourceExhausted code for CreateVolume

hown3d:
If OpenStack quotas for volumes are exceeded, the driver currently returns the gRPC call with code `Internal`. This is not propagated to the PVC as an event and is not traceable very easily.
Cinder returns with a HTTP 413 response code if the quota for volumes is exceeded:
https://github.com/openstack/cinder/blob/c4f61c11ef9b590606a2a276bdf256622516181f/cinder/quota_utils.py#L130-L140

CSI spec defines that on a quota limit error, the driver should return a `ResourceExhausted` grpc error code:
https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume-errors

**Release note**:

```release-note
Updated the error code to ResourceExhausted for CreateVolume calls if the volume quota is exceeded.
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -18,6 +18,7 @@ package cinder
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"slices"
 	"sort"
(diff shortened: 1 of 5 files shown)


zetaab (approved):
/approve

Boston01:
@zetaab  i have a pending merge request for a while can some give use help for this  [Create one storage cluster for wide spreaded kubernetes cluster in many Azs](https://github.com/kubernetes/cloud-provider-openstack/pull/2843)

hown3d:
@zetaab can we get this in?

PR #2862, opened 2025-03-24
[cinder-csi-plugin] Don't report topology capability when --with-topology=False

### LINKED ISSUE #2861: [cinder-csi-plugin] Setting `--with-topology=false` on controller causes provisioner errors
/kind bug

**What happened**:

In https://github.com/kubernetes/cloud-provider-openstack/pull/2743 we added support for a `--with-topology` boolean option that allowed users to disable the topology feature when it doesn't make sense. This was based on an existing option in the Manila CSI driver and like that, the option applied to both the Controller and Node services. The bug concerns the latter: while we are currently [conditionally reporting topology information for the nodes](https://github.com/kubernetes/cloud-provider-openstack/blob/53a4505c3caa96365b12cfc7230a84e429744b7c/pkg/csi/cinder/nodeserver.go#L300-L302), we [unconditionally report the plugin capability](https://github.com/kubernetes/cloud-provider-openstack/blob/53a4505c3caa96365b12cfc7230a84e429744b7c/pkg/csi/cinder/identityserver.go#L65-L71). This causes failures in the provisioner like so:

```
Warning  ProvisioningFailed    39s (x7 over 102s)    cinder.csi.openstack.org_openstack-cinder-csi-driver-controller-796d68c75b-8gjtq_3dfb94be-e558-4f59-8800-335fb70315d5  failed to provision volume with StorageClass "standard-csi": error generating accessibility requirements: no topology key found on CSINode ostest-h469z-worker-0-vn4q8
```

That is due [to logic in `kubernetes-csi/external-provisioner`](https://github.com/kubernetes-csi/external-provisioner/blob/17e2429e9ff7b1399d2c31ad75d66c902892bd1d/pkg/controller/topology.go#L177). That logic only runs when [the driver reports the topology plugin capability](https://github.com/kubernetes-csi/external-provisioner/blob/17e2429e9ff7b1399d2c31ad75d66c902892bd1d/pkg/controller/controller.go#L685).

PVCs should be provisioned when `--with-topology=false`.

**How to reproduce it**:

1. Set `--with-topology=false` for the `cinder-csi-plugin` executable providing the `node` service (`--provide-controller-service=true`, which is the default).
2. Attempt to provision a PVC. 

Fix incoming.

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: `master`
- OpenStack version: (not relevant)
- Others: (not relevant)

stephenfin:

The `PluginCapability_Service_VOLUME_ACCESSIBILITY_CONSTRAINTS` capability flag determines whether the provisioner attempts to look up topology information from the node or not [1][1] [2][2]. If we report it but don't return topology information from the node, we end up with failures to provision [3][3]. Fix the issue by optionally reporting the capability, like Manila already does.

In addition to the main fix, we also duplicate some helpful logs from Cinder to Manila, plus we fix a minor issue in the startup logging for both. More information on all these changes is provided in [the commits](https://github.com/kubernetes/cloud-provider-openstack/pull/2862/commits).

[1]: https://github.com/kubernetes-csi/external-provisioner/blob/17e2429e9f/pkg/controller/controller.go#L685-L700
[2]: https://github.com/kubernetes-csi/external-provisioner/blob/17e2429e9f/pkg/controller/controller.go#L994-L996
[3]: https://github.com/kubernetes-csi/external-provisioner/blob/17e2429e9f/pkg/controller/topology.go#L177

fixes #2861

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/driver.go
+++ b/pkg/csi/cinder/driver.go
@@ -101,7 +101,7 @@ func NewDriver(o *DriverOpts) *Driver {
 	klog.Info("Driver: ", d.name)
 	klog.Info("Driver version: ", d.fqVersion)
 	klog.Info("CSI Spec version: ", specVersion)
-	klog.Infof("Topology awareness: %T", d.withTopology)
+	klog.Infof("Topology awareness: %t", d.withTopology)
 
 	d.AddControllerServiceCapabilities(
 		[]csi.ControllerServiceCapability_RPC_Type{
--- a/pkg/csi/manila/identityserver.go
+++ b/pkg/csi/manila/identityserver.go
@@ -22,6 +22,7 @@ import (
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"k8s.io/klog/v2"
 )
 
 type identityServer struct {
@@ -30,8 +31,14 @@ type identityServer struct {
 }
 
 func (ids *identityServer) GetPluginInfo(ctx context.Context, req *csi.GetPluginInfoRequest) (*csi.GetPluginInfoResponse, error) {
+	klog.V(5).Infof("Using default GetPluginInfo")
+
 	if ids.d.name == "" {
 		return nil, status.Error(codes.Unavailable, "Driver name not configured")
+	}
+
+	if ids.d.fqVersion == "" {
+		return nil, status.Error(codes.Unavailable, "Driver is missing version")
 	}
 
 	return &csi.GetPluginInfoResponse{
@@ -51,6 +58,7 @@ func (ids *identityServer) Probe(ctx context.Context, req *csi.ProbeRequest) (*c
 }
 
 func (ids *identityServer) GetPluginCapabilities(ctx context.Context, req *csi.GetPluginCapabilitiesRequest) (*csi.GetPluginCapabilitiesResponse, error) {
+	klog.V(5).Infof("GetPluginCapabilities called with req %+v", req)
 	caps := []*csi.PluginCapability{
 		{
 			Type: &csi.PluginCapability_Service_{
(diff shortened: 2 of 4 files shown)


pkg/csi/cinder/driver.go
```diff
@@ -101,7 +101,7 @@ func NewDriver(o *DriverOpts) *Driver {
 	klog.Info("Driver: ", d.name)
 	klog.Info("Driver version: ", d.fqVersion)
 	klog.Info("CSI Spec version: ", specVersion)
-	klog.Infof("Topology awareness: %T", d.withTopology)
+	klog.Infof("Topology awareness: %t", d.withTopology)
```
mdbooth, line 104: That's a fun bug!

pkg/csi/manila/identityserver.go
```diff
@@ -30,10 +31,16 @@ type identityServer struct {
...
 func (ids *identityServer) GetPluginInfo(ctx context.Context, req *csi.GetPluginInfoRequest) (*csi.GetPluginInfoResponse, error) {
+	klog.V(5).Infof("Using default GetPluginInfo")
+
 	if ids.d.name == "" {
 		return nil, status.Error(codes.Unavailable, "Driver name not configured")
 	}
 
+	if ids.d.fqVersion == "" {
+		return nil, status.Error(codes.Unavailable, "Driver is missing version")
+	}
```
mdbooth, line 42: What happened previously when this was not set? Did it fail, or did it just log incorrectly? If the latter you'll want to at least add a release note for this.

stephenfin: Assuming I am reading things correctly, it would have failed since the field is required:

https://github.com/container-storage-interface/spec/blob/98819c45a37a67e0cd466bd02b813faf91af4e45/lib/go/csi/csi.pb.go#L693-L694

pkg/csi/cinder/driver.go
stephenfin: Worse: I wrote it initially :facepalm:

mandre (commented):
/lgtm

PR #2865, opened 2025-03-24
[cinder-csi-plugin] Rationalise all things topology

stephenfin:

This is a patch that came out of my work on https://github.com/kubernetes/cloud-provider-openstack/issues/2861. Currently, generation of the topology request for the Cinder Volume is logically separated from the generation of the topology request for the CSI Volume. This makes things harder to understand. Move some things around to resolve this and make everything a little easier to grok. This is kept separate from https://github.com/kubernetes/cloud-provider-openstack/pull/2862 since I don't expect we'll want to backport this.

n/a

I have a decoder of the logic embedded in the commit message. Hopefully this helps with review, but please don't assume the decoder itself is correct! :sweat_smile: I have also added additional unit tests that should capture any changes in logic.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -1082,7 +1082,7 @@ func getCreateVolumeResponse(vol *volumes.Volume, volCtx map[string]string, igno
 		if accessibleTopologyReq != nil {
 			accessibleTopology = accessibleTopologyReq.GetPreferred()
 		}
-	} else {
+	} else if vol.AvailabilityZone != "" {
 		accessibleTopology = []*csi.Topology{
 			{
 				Segments: map[string]string{topologyKey: vol.AvailabilityZone},
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -86,21 +86,21 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	// Volume Type
 	volType := volParams["type"]
 
+	// Volume AZ
+
+	accessibleTopologyReq := req.GetAccessibilityRequirements()
+	ignoreVolumeAZ := cloud.GetBlockStorageOpts().IgnoreVolumeAZ
+
+	// First check if volAvailability is already specified, if not get preferred from Topology
+	// Required, in case vol AZ is different from node AZ
 	var volAvailability string
 	if cs.Driver.withTopology {
-		// First check if volAvailability is already specified, if not get preferred from Topology
-		// Required, incase vol AZ is different from node AZ
-		volAvailability = volParams["availability"]
-		if volAvailability == "" {
-			accessibleTopologyReq := req.GetAccessibilityRequirements()
-			// Check from Topology
-			if accessibleTopologyReq != nil {
-				volAvailability = sharedcsi.GetAZFromTopology(topologyKey, accessibleTopologyReq)
-			}
+		if volParams["availability"] != "" {
+			volAvailability = volParams["availability"]
+		} else if accessibleTopologyReq != nil {
+			volAvailability = sharedcsi.GetAZFromTopology(topologyKey, accessibleTopologyReq)
 		}
 	}
-
-	ignoreVolumeAZ := cloud.GetBlockStorageOpts().IgnoreVolumeAZ
 
 	// get the PVC annotation
 	pvcAnnotations := sharedcsi.GetPVCAnnotations(cs.Driver.pvcLister, volParams)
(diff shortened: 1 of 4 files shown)


pkg/csi/cinder/controllerserver.go
```diff
@@ -1033,7 +1038,24 @@ func (cs *controllerServer) ControllerExpandVolume(ctx context.Context, req *csi
...
 }
 
-func getCreateVolumeResponse(vol *volumes.Volume, volCtx map[string]string, ignoreVolumeAZ bool, accessibleTopologyReq *csi.TopologyRequirement) *csi.CreateVolumeResponse {
+func getTopology(vol *volumes.Volume, topologyReq *csi.TopologyRequirement, ignoreVolumeAZ bool) []*csi.Topology {
+	var accessibleTopology []*csi.Topology
+	if ignoreVolumeAZ {
+		if topologyReq != nil {
+			accessibleTopology = topologyReq.GetPreferred()
+		}
+	} else {
```
mdbooth: I think you've lost your previous change which refused to create a segment with an empty AZ.

pkg/csi/cinder/controllerserver.go
```diff
@@ -1038,18 +1036,20 @@ func (cs *controllerServer) ControllerExpandVolume(ctx context.Context, req *csi
...
-		if topologyReq != nil {
-			accessibleTopology = topologyReq.GetPreferred()
-		}
-	} else {
-		// NOTE(stephenfin): We retrieve the AZ from the created volume rather than
-		// using the value we provided in our request since these can differ due to
-		// Cinder's '[DEFAULT] allow_availability_zone_fallback' option.
-		accessibleTopology = []*csi.Topology{
-			{Segments: map[string]string{topologyKey: vol.AvailabilityZone}},
+	if withTopology {
```
mdbooth, line 1041: nit: I think a shortcut with reduced indentation would be more readable

```go
if !withTopology {
    return nil
}
```

pkg/csi/cinder/controllerserver.go
```diff
@@ -1038,18 +1036,20 @@ func (cs *controllerServer) ControllerExpandVolume(ctx context.Context, req *csi
...
+			if topologyReq != nil {
+				accessibleTopology = topologyReq.GetPreferred()
+			}
+		} else {
+			// NOTE(stephenfin): We retrieve the AZ from the created volume rather than
+			// using the value we provided in our request since these can differ due to
+			// Cinder's '[DEFAULT] allow_availability_zone_fallback' option.
+			accessibleTopology = []*csi.Topology{
+				{Segments: map[string]string{topologyKey: vol.AvailabilityZone}},
+			}
```
mdbooth, line 1052: IIUC, these changes originate from the topology defined on the CSINode. With this change we're (optionally) ignoring topology requested by the Node and only observing topology directly requested on the PV via the StorageClass (or PVC?).

This makes sense to me.

Didn't you also discuss allowing `--with-topology=false` to disable setting topology on the CSINode altogether? I was expecting to see that in this PR.

mdbooth (commented):
Thanks for this series. Thanks also for splitting it into separate commits with excellent commit messages. That made it much easier to review.

I had some nits which you're welcome to ignore, but I think an issue in the fifth commit might be worth fixing.

I wonder if it's worth splitting the last commit, which actually changes the behaviour of `--with-topology` into a separate PR. Whether it's worth it might depend on the next point.

I also wonder if we need to guard this behaviour change somehow: has `--with-topology` been included in a release yet? If not I think we're good, but if it has I don't think we can make a breaking change like this without guardrails.

stephenfin:
> I wonder if it's worth splitting the last commit, which actually changes the behaviour of `--with-topology` into a separate PR. Whether it's worth it might depend on the next point.
> 
> I also wonder if we need to guard this behaviour change somehow: has `--with-topology` been included in a release yet? If not I think we're good, but if it has I don't think we can make a breaking change like this without guardrails.

Good point. I've dropped this. We would want a release note for the change.

pkg/csi/cinder/controllerserver_test.go
```diff
@@ -260,6 +260,80 @@ func TestCreateVolumeWithIgnoreVolumeAZ(t *testing.T) {
...
+			expectedVolumeAZ: "",
+			// FIXME(stephenfin): This should be unset
+			expectedCSITopology: "nova",
+		},
+		{
+			name: "availability parameter",
+			volumeParams: map[string]string{
+				"availability": "cinder",
+			},
+			// FIXME(stephenfin): This should be "cinder" per the parameter
```
mandre, line 285: Should this test, and the change for fixing the issues it highlights should be part of a different PR? Or perhaps provide the fix directly... I wonder what's easiest.

edit: ok, I see you use this new test in the next commit to validate it's not breaking behavior. So perhaps providing the fix into the same PR makes sense?

stephenfin: I originally had the fix rolled in here, but Matt suggested it probably didn't belong in here as it was a behaviour change.

mandre: Fair enough. Do I understand correctly that the plan is to create a new PR with the fix once this one merges?

stephenfin: Yes, that's my plan

PR #2869, opened 2025-03-26
[docs] add more uses cases to node-selector service annotation

kayrus:

Add more context about the `node-selector` service annotation and how to use it in combination with the `node.kubernetes.io/exclude-from-external-load-balancers` node label.

fixes #

**Release note**:

```release-note
NONE
```


kayrus:
@stephenfin @zetaab 
the `node.kubernetes.io/exclude-from-external-load-balancers=false` label is supported starting from k8s 1.34, the PR is ready to be merged.

PR #2870, opened 2025-03-27
[occm] fix lbaas logs

kayrus:

To avoid `loadbalancer.go:1478] "Target node label %s=%s is set to LoadBalancer service %s" env="production" default/openstack-loadbalancer="(MISSING)"`

fixes #

Introduced in #2601 

**Release note**:

```release-note
NONE
```


kayrus:
@stephenfin can you approve as well?

PR #2742, opened 2024-12-09
tests: Bump DevStack to Epoxy (2025.1)

stephenfin:
**What this PR does / why we need it**:

Bump the version of DevStack used in CI from Bobcat (2023.2), which is now EOL, to Epoxy (2025.1).

**Which issue this PR fixes(if applicable)**:

(none)

**Special notes for reviewers**:

(none)

**Release note**:

```release-note
NONE
```

PR #2872, opened 2025-04-04
[release-1.31] [cinder-csi-plugin] Don't report topology capability when --with-topology=False

stephenfin:
This is a manual backport of #2862 (`master`) and #2871 (`release-1.32`). As noted in the first commit, the merge conflict was trivial and due to the absence of #2734, which added a log into the Manila CSI Driver.

```release-note
NONE
```

### DIFF
--- a/pkg/csi/cinder/driver.go
+++ b/pkg/csi/cinder/driver.go
@@ -88,7 +88,7 @@ func NewDriver(o *DriverOpts) *Driver {
 	klog.Info("Driver: ", d.name)
 	klog.Info("Driver version: ", d.fqVersion)
 	klog.Info("CSI Spec version: ", specVersion)
-	klog.Infof("Topology awareness: %T", d.withTopology)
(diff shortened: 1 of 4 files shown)


stephenfin:
These failures are [due to the version of golangci-lint we use](https://github.com/golangci/golangci-lint/issues/3718). I've proposed https://github.com/kubernetes/cloud-provider-openstack/pull/2874 to fix this. We'll need to merge that first.

stephenfin:
> These failures are [due to the version of golangci-lint we use](https://github.com/golangci/golangci-lint/issues/3718). I've proposed #2874 to fix this. We'll need to merge that first.

That's merged now, so I've rebased and force pushed this.

stephenfin:
Jobs are failing because the `stable/2023.2` branch was recently deleted upstream. Looks like that's also borked on `master`. This is going to be fun :see_no_evil:

stephenfin:
Waiting on https://github.com/kubernetes/cloud-provider-openstack/pull/2742

PR #2883, opened 2025-04-29
[manila-csi-plugin] Add support for authentication via clouds.yaml

### LINKED ISSUE #2886: Fix authentication issue via clouds.yaml in Manila CSI driver
/kind bug

**What happened**:
The Manila CSI driver currently does not support authentication via `clouds.yaml`. This limits the ability to authenticate using the standard OpenStack configuration file that is commonly used across OpenStack services.

The Manila CSI driver should support authentication through `clouds.yaml`

dkokkino:

1. Fixes a bug where `AuthOps` and `Region` were incorrectly required when using `clouds.yaml` for authentication.
2. Resolves an issue where the validator could not properly handle boolean values retrieved from secrets.
3. Adds a new test case to cover the boolean validation fix.
4. Introduces support for authenticating the Manila CSI driver via `clouds.yaml`

fixes #2886

**Release note**:

```release-note
Now support configuration from a clouds.yaml file for the Manila CSI driver.
```

### DIFF
--- a/pkg/csi/manila/validator/validator.go
+++ b/pkg/csi/manila/validator/validator.go
@@ -19,6 +19,7 @@ package validator
 import (
 	"fmt"
 	"reflect"
+	"strconv"
 )
 
 // Validator validates input data and stores it in the output struct.
@@ -113,10 +114,24 @@ func (v *Validator) Populate(data map[string]string, out interface{}) error {
 				return fmt.Errorf("parameter '%s' cannot be empty", fName)
 			}
 
-			// The value is present, populate the field and continue with the next one
+			field := vOut.Field(int(fIdx))
 
-			vOut.Field(int(fIdx)).SetString(value)
-			continue
+			switch field.Kind() {
+			case reflect.Bool:
+				boolValue, err := strconv.ParseBool(value)
+				if err != nil {
+					return fmt.Errorf("invalid boolean value for parameter '%s': %v", fName, err)
+				}
+				field.SetBool(boolValue)
+				continue
+
+			case reflect.String:
+				field.SetString(value)
+				continue
+
+			default:
+				return fmt.Errorf("unsupported field type for parameter '%s'", fName)
+			}
 		}
 
 		// Value not present, determine whether this field is required
--- a/pkg/csi/manila/validator/validator_test.go
+++ b/pkg/csi/manila/validator/validator_test.go
@@ -16,7 +16,11 @@ limitations under the License.
 
 package validator
 
-import "testing"
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/v2"
+)
 
 func TestValueRequired(t *testing.T) {
 	type s1 struct {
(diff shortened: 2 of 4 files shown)


pkg/csi/manila/validator/validator.go
```diff
@@ -113,10 +114,23 @@ func (v *Validator) Populate(data map[string]string, out interface{}) error {
...
+				boolValue, err := strconv.ParseBool(value)
+				if err != nil {
+					return fmt.Errorf("invalid boolean value for parameter '%s': %v", fName, err)
+				}
+				field.SetBool(boolValue)
+				continue
+			}
+
+			// Handle string fields
+			if field.Kind() == reflect.String {
```
stephenfin, line 130: Could we use a switch statement here? Also, do we want an error case for other values? I suspect it'll be mostly dead code unless we extend the `AuthOpts` struct but it's no harm (you should test what happens if you try to convert something with `EndpointType` set, since that's of type `gophercloud.Availability`).

pkg/csi/manila/validator/validator_test.go
```diff
@@ -224,3 +224,39 @@ func TestFieldNames(t *testing.T) {
...
+		t.Errorf("expected A to be false, got: %v", obj.A)
+	}
+
+	// Test with a invalid boolean string "foo"
+	obj = &s{}
+	err = v.Populate(map[string]string{"a": "foo"}, obj)
+	if err == nil {
+		t.Errorf("Populate with 'foo': expected an error, but got nil")
+	}
+
```
stephenfin, line 261: nit: extra newline

stephenfin (changes requested):
Two small nits, but this looks pretty good to me. You can likely change this to ready-for-review now. I would suggest filing a bug describing the three issues you're fixing here and adding a release note indicating that we now support configuration from a `clouds.yaml` file for the Manila CSI driver.

pkg/csi/manila/validator/validator.go
dkokkino: Okay i'll make the change to a switch case and add the error case for other values

I also tested setting a field of type `gophercloud.Availability` (like EndpointType), and confirmed that assigning a `string` value via reflection works as expected without special handling. Since `Availability` is just a named `string` type, `reflect.SetString()` handles it correctly. I'll include that test.

gouthampacha:
/lgtm

thank you @dkokkino

stephenfin:
As I've noted elsewhere, failures in CI are due to the removal of `stable/2023.2` branch. We need https://github.com/kubernetes/cloud-provider-openstack/pull/2742 or a similar PR to address this.

stephenfin:
Thanks for the rework here.

/lgtm
/approve
/hold while we fix CI

stephenfin:

CI is green again.

stephenfin:
/unhold

Manila CI jobs are green :partying_face:

PR #2888, opened 2025-05-12
tests: Bump DevStack to Caracal (2024.1)

stephenfin:

Bump the version of DevStack used in CI from Bobcat (2023.2), which is now EOL, to Caracal (2024.1). A future change will bump this further to Dalmatian (2024.2) and then Epoxy (2025.1).

(none)

(none)

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/ci-csi-cinder-e2e.sh
+++ b/tests/ci-csi-cinder-e2e.sh
@@ -41,6 +41,24 @@ cleanup() {
 }
 trap cleanup EXIT
 
+# Install Ansible from Trixie (Debian Testing) to allow us to talk to a host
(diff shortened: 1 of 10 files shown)


stephenfin:
For some reason the Ansible timeout kicked in rather than the Ginkgo one. I've increased the gap between the two in case we have some very slow tests (which I guess we do).

kayrus:
@stephenfin  have you applied all the patches we have in gophercloud? the timeouts could be caused by 500/400 response codes in API.

stephenfin:
/close

Let's focus on #2742

PR #2901, opened 2025-06-03
release 1.33.0

### LINKED ISSUE #2884: Cinder CSI plugin doesn't work with latest OpenStack
It seems that we're testing with `stable/2023.2` at the moment, and as of this issue being filed the latest is `stable/2025.1`.

https://github.com/kubernetes/cloud-provider-openstack/blob/1d06879a8af0292acf29bc5f258df422bf2489b5/tests/playbooks/roles/install-devstack/defaults/main.yaml#L4

The latest version of DevStack ships with the `volumev3` endpoint missing:

https://github.com/openstack/devstack/commit/d7c3c0accc89e4e99915c24fa7c3bff2e90a715e

This probably works fine right now for most users, but eventually, it'll break, `gophercloud` has addressed this in the newer versions:

https://github.com/gophercloud/gophercloud/pull/3209

This seems to have landed inside v2.7.0:

https://github.com/gophercloud/gophercloud/commit/908ce172717ef96ddea87a969e405e2190a18371

However, we're still at `gophercloud` 2.5.0 at the moment.

zetaab:

fixes #2884

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: v1.32.0
+appVersion: v1.33.0
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 2.33.0-alpha.2
+version: 2.33.0
 home: https://github.com/kubernetes/cloud-provider-openstack
 icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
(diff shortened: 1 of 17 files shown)


zetaab:
tests are broken currently, trying to fix these https://github.com/kubernetes/cloud-provider-openstack/pull/2903

zetaab:
@kayrus & @stephenfin I think we should anyway do release 1.33 with old openstack version in CI. So this PR needs lgtm. After that we need figure out what is wrong with newer openstack versions in CI

PR #2905, opened 2025-06-05
tests: Temporarily unblock gate with use of EOL Bobcat tags

stephenfin:

We currently have no less than three PRs in-flight attempting to bump the DevStack version used in the gate (and a possible forth one need for Epoxy):

* https://github.com/kubernetes/cloud-provider-openstack/pull/2888
* https://github.com/kubernetes/cloud-provider-openstack/pull/2742
* https://github.com/kubernetes/cloud-provider-openstack/pull/2903

While we continue work on these and investigate the performance regressions seen in Dalmatian (and possibly Caracal), unblock the gate by relying on the `2023.2-eol` tags rather than the (now deleted) `stable/2023.2` branches.

Note that `python-barbicanclient` does not have a `2023.2-eol` tag so we can't use install from Git. Since we're not hacking on though there's no good reason to be installing from Git, so we remove both this and `python-octaviaclient` from the list.

None.

This should be considered a temporary bandaid. I'm actively investigating these test timeouts locally now that the Gophercloud CI bump is completed (and I think @zetaab is doing the same).

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/playbooks/roles/install-devstack/defaults/main.yaml
+++ b/tests/playbooks/roles/install-devstack/defaults/main.yaml
@@ -1,7 +1,7 @@
 ---
 user: "stack"
 workdir: "/home/{{ user }}/devstack"
-branch: "stable/2023.2"
+branch: "2023.2-eol"
 enable_services:
   - nova
   - glance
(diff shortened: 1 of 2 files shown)


stephenfin:
@kayrus @zetaab Wanna get this in now to unblock the gate and continue working on our fixes?

PR #2907, opened 2025-06-09
[occm] Fix incorrect SA name in auth-delegate clusterRoleBiding in OCCM helm chart

HaoruiPeng:

OCCM helm chart binds clusterrole 'system:auth-delegator' to service account `openstack-cloud-controller-manager` for service monitor, the value is read from ` occm.name `. 
However the correct SA name created by the chart is 'cloud-controller-manager', value read from `.Values.serviceAccountName`.

The wrong SA name caused that the service monitor fails the authentication to create tokenreviews:
Error message:
```
Failed to make webhook authenticator request: tokenreviews.authentication.k8s.io is forbidden: User "system:serviceaccount:openstack-system:cloud-controller-manager" cannot create resource "tokenreviews" in API group "authentication.k8s.io" at the cluster scope
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/openstack-cloud-controller-manager/templates/clusterrolebinding-sm.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/clusterrolebinding-sm.yaml
@@ -10,7 +10,7 @@ metadata:
     {{- end }}
 subjects:
 - kind: User
-  name: system:serviceaccount:{{ .Release.Namespace }}:{{ include "occm.name" . }}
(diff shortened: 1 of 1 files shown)


kayrus:
Hi @HaoruiPeng , could you please fix lint warnings?

HaoruiPeng:
Hi @kayrus, lint warning fixed :)

PR #2915, opened 2025-06-26
[manila-csi-plugin] support muilple share rules

### LINKED ISSUE #2725: [manila-csi-plugin] support muilple share rules

/kind feature

**What happened**:

Multiple share rules cannot be registered.

I would like to register multiple share rules separated by commas as follows.
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: csi-manila-nfs
provisioner: nfs.manila.csi.openstack.org
allowVolumeExpansion: true
parameters:
  # Manila share type
  type: default

  csi.storage.k8s.io/provisioner-secret-name: csi-manila-secrets
  csi.storage.k8s.io/provisioner-secret-namespace: default
  csi.storage.k8s.io/controller-expand-secret-name: csi-manila-secrets
  csi.storage.k8s.io/controller-expand-secret-namespace: default
  csi.storage.k8s.io/node-stage-secret-name: csi-manila-secrets
  csi.storage.k8s.io/node-stage-secret-namespace: default
  csi.storage.k8s.io/node-publish-secret-name: csi-manila-secrets
  csi.storage.k8s.io/node-publish-secret-namespace: default
  nfs-shareClient: "10.1.10.0/24,10.2.20.0/24"
```

**How to reproduce it**:

```
Warning  ProvisioningFailed    3s (x6 over 53s)   manila.csi.openstack.org_openstack-manila-csi-controllerplugin-0_7699907b-feed-47cb-a5be-0dd45489522b  failed to provision volume with StorageClass "csi-manila-nfs": rpc error: code = Internal desc = failed to grant access to volume pvc-4cd02bad-b321-4aa0-85ba-64f451d1ac6b: Bad request with: [POST https://xxxxx/v2/8f21cfb9acbd4b9b91f3c39542bf68a4/shares/7617dbea-dca0-4412-afd8-cbaf36283cf0/action], error message: {"badRequest": {"code": 400, "message": "Only one '/' permitted in '10.1.10.0/24,10.2.20.0/24'"}}
```

**Environment**:
- OpenStack version:
- Others:

silvacarloss:
Manila currently supports only setting a single access rule, making it difficult to have multiple access rules to a shared file system, which is a common scenario in a cloud where we need to have multiple clients being able to mount the shares simultaneously. This PR introduces support for multiple share access rules by accepting a list of IPs/Cephx users instead of a single string.

fixes #2725 

Author of the change stated that they didn't have a CephFS environment, so could not test it with CephFS.
Another pull request was proposed for this issue [1], but due to maintainers not being able to update and rebase the original author's PR, we opened a new PR.
[1] https://github.com/kubernetes/cloud-provider-openstack/pull/2727

**Release note**:

```release-note
[manila-csi-plugin] Added support for configuring multiple share access rules to a shared filesystem.
```

### DIFF
--- a/pkg/csi/manila/shareadapters/cephfs.go
+++ b/pkg/csi/manila/shareadapters/cephfs.go
@@ -19,6 +19,7 @@ package shareadapters
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/gophercloud/gophercloud/v2"
@@ -32,82 +33,81 @@ type Cephfs struct{}
 
 var _ ShareAdapter = &Cephfs{}
 
-func (Cephfs) GetOrGrantAccess(ctx context.Context, args *GrantAccessArgs) (accessRight *shares.AccessRight, err error) {
+func (Cephfs) GetOrGrantAccesses(ctx context.Context, args *GrantAccessArgs) ([]shares.AccessRight, error) {
 	// First, check if the access right exists or needs to be created
 
-	var rights []shares.AccessRight
-
-	accessTo := args.Options.CephfsClientID
-	if accessTo == "" {
-		accessTo = args.Share.Name
-	}
-
-	rights, err = args.ManilaClient.GetAccessRights(ctx, args.Share.ID)
+	rights, err := args.ManilaClient.GetAccessRights(ctx, args.Share.ID)
 	if err != nil {
 		if _, ok := err.(gophercloud.ErrResourceNotFound); !ok {
 			return nil, fmt.Errorf("failed to list access rights: %v", err)
 		}
-	} else {
+	}
+
+	accessToList := []string{args.Share.Name}
+	if args.Options.CephfsClientID != "" {
+		accessToList = strings.Split(args.Options.CephfsClientID, ",")
+	}
+
+	created := false
+	for _, at := range accessToList {
 		// Try to find the access right
-
+		found := false
 		for _, r := range rights {
-			if r.AccessTo == accessTo && r.AccessType == "cephx" && r.AccessLevel == "rw" {
+			if r.AccessTo == at && r.AccessType == "cephx" && r.AccessLevel == "rw" {
 				klog.V(4).Infof("cephx access right for share %s already exists", args.Share.Name)
-
-				accessRight = &r
+				found = true
 				break
 			}
 		}
-	}
 
-	if accessRight == nil {
 		// Not found, create it
-
-		accessRight, err = args.ManilaClient.GrantAccess(ctx, args.Share.ID, shares.GrantAccessOpts{
-			AccessType:  "cephx",
-			AccessLevel: "rw",
-			AccessTo:    accessTo,
-		})
-
-		if err != nil {
-			return
+		if !found {
+			result, err := args.ManilaClient.GrantAccess(ctx, args.Share.ID, shares.GrantAccessOpts{
+				AccessType:  "cephx",
+				AccessLevel: "rw",
+				AccessTo:    at,
+			})
+			if err != nil {
+				return nil, fmt.Errorf("failed to grant access right: %v", err)
+			}
+			if result.AccessKey == "" {
+				// Wait till a ceph key is assigned to the access right
+				backoff := wait.Backoff{
+					Duration: time.Second * 5,
+					Factor:   1.2,
+					Steps:    10,
+				}
+				wait.ExponentialBackoff(backoff, func() (bool, error) {
+					rights, err := args.ManilaClient.GetAccessRights(ctx, args.Share.ID)
+					if err != nil {
+						return false, fmt.Errorf("error get access rights for share %s: %v", args.Share.ID, err)
+					}
+					if len(rights) == 0 {
+						return false, fmt.Errorf("cannot find the access right we've just created")
+					}
+					for _, r := range rights {
+						if r.AccessTo == at && r.AccessKey != "" {
+							return true, nil
+						}
+					}
+					klog.V(4).Infof("Access key for %s is not set yet, retrying...", at)
+					return false, nil
+				})
+			}
+			created = true
 		}
 	}
 
-	if accessRight.AccessKey != "" {
-		// The access right is ready
-		return
-	}
-
-	// Wait till a ceph key is assigned to the access right
-
-	backoff := wait.Backoff{
-		Duration: time.Second * 5,
-		Factor:   1.2,
-		Steps:    10,
-	}
-
-	return accessRight, wait.ExponentialBackoff(backoff, func() (bool, error) {
-		rights, err := args.ManilaClient.GetAccessRights(ctx, args.Share.ID)
+	// Search again because access rights have changed
+	if created {
+		rights, err = args.ManilaClient.GetAccessRights(ctx, args.Share.ID)
 		if err != nil {
-			return false, err
-		}
-
-		var accessRight *shares.AccessRight
-
-		for i := range rights {
-			if rights[i].AccessTo == accessTo {
-				accessRight = &rights[i]
-				break
+			if _, ok := err.(gophercloud.ErrResourceNotFound); !ok {
+				return nil, fmt.Errorf("failed to list access rights: %v", err)
 			}
 		}
-
-		if accessRight == nil {
-			return false, fmt.Errorf("cannot find the access right we've just created")
-		}
-
-		return accessRight.AccessKey != "", nil
-	})
+	}
+	return rights, nil
 }
 
 func (Cephfs) BuildVolumeContext(args *VolumeContextArgs) (volumeContext map[string]string, err error) {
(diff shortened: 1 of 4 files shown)


kayrus:
/ok-to-test

I'll make a review a bit later

pkg/csi/manila/shareadapters/cephfs.go
```diff
@@ -32,82 +33,84 @@ type Cephfs struct{}
...
 		}
-
-		if accessRight == nil {
-			return false, fmt.Errorf("cannot find the access right we've just created")
-		}
-
-		return accessRight.AccessKey != "", nil
-	})
+	}
+	return rights, nil
```
gouthampacha, line 113: I'll leave the comment here, but this logic seems incorrect to me. 

The idea of allowing multiple CephX users access would mean that we only ever need one client user per node.
This method is currently creating/retrieving and returning a list of access rules, where previously, it would have created/identified one CephX access rule. From a single node's perspective, this is a bunch of unnecessary steps.. 

The node plugin would need to identify the right access key and forward it to the Ceph node plugin for mounting.. The ceph node plugin expects exactly one user name and access key - so manila-csi needs a way to isolate what identity means for a node. We probably need node specific parameters to tell us what the user name should be.

Does that make sense?

pkg/csi/manila/shareadapters/nfs.go
```diff
@@ -43,22 +43,43 @@ func (NFS) GetOrGrantAccess(ctx context.Context, args *GrantAccessArgs) (*shares
...
+		if !found {
+			_, err = args.ManilaClient.GrantAccess(ctx, args.Share.ID, shares.GrantAccessOpts{
+				AccessType:  "ip",
+				AccessLevel: "rw",
+				AccessTo:    at,
+			})
+			if err != nil {
+				return nil, fmt.Errorf("failed to grant access right: %v", err)
+			}
+			created = true
```
gouthampacha, line 69: Same thing here. When verifying if we have access, We need to identify the IP address of the node and check if it exists (or, just assume that it'll work if we handle the access creation within the controller).

This logic would be perfect for NFS because there is no "access key" (or secret) necessary to mount a share in this case.

In the proposed change right niw, each node checking for/creating access rules is a bit wasteful and racy.

gouthampacha (commented):
Hi @silvacarloss 

Thanks for updating this change. Please see comments inline

docs/manila-csi-plugin/using-manila-csi-plugin.md
```diff
@@ -71,7 +71,7 @@ Parameter | Required | Description
 ----------|----------|------------
 `shareID` | if `shareName` is not given | The UUID of the share
 `shareName` | if `shareID` is not given | The name of the share
-`shareAccessID` | _yes_ | The UUID of the access rule for the share
+`shareAccessIDs` | _yes_ | The UUID of the access rule for the share
```
gouthampacha, line 74: nit:

```suggestion
`shareAccessIDs` | _yes_ | Comma separated UUIDs of access rules for the share
```

gouthampacha: also this is a backwards incompatible change - there's a chance that it might be annoying to someone to patch pvcs to update this field on existing PVCs when the driver is updated... 

it feels like it'll be straightforward to just accept both values, and prefer one of them with simple updates:

```go
type NodeVolumeContext struct {
    ShareID        string `name:"shareID" value:"optionalIf:shareName=." precludes:"shareName"`
    ShareName      string `name:"shareName" value:"optionalIf:shareID=." precludes:"shareID"`
    ShareAccessIDs string `name:"shareAccessIDs" value:"optionalIf:ShareAccessID=." precludes:"ShareAccessID"`
    ShareAccessID  string `name:"shareAccessID" value:"optionalIf:ShareAccessIDs=." precludes:"ShareAccessIDs"` // Keep this for backwards compatibility
    // ...
}

// In your logic:
func getAccessIDs(ctx NodeVolumeContext) []string {
    if ctx.ShareAccessIDs != "" {
        // Split by comma if multiple
        return strings.Split(ctx.ShareAccessIDs, ",")
    } else if ctx.ShareAccessID != "" {
        // Backwards compatibility: treat as single-element list
        return []string{ctx.ShareAccessID}
    }
    return nil
}
```

pkg/csi/manila/shareadapters/nfs.go
```diff
@@ -43,22 +43,43 @@ func (NFS) GetOrGrantAccess(ctx context.Context, args *GrantAccessArgs) (*shares
...
-	// Not found, create it
-
-	return args.ManilaClient.GrantAccess(ctx, args.Share.ID, shares.GrantAccessOpts{
-		AccessType:  "ip",
-		AccessLevel: "rw",
-		AccessTo:    args.Options.NFSShareClient,
-	})
+	// Search again because access rights have changed
+	if created {
+		rights, err = args.ManilaClient.GetAccessRights(ctx, args.Share.ID)
```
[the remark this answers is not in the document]
mandre: And we're [setting the client micro-version to 2.37](https://github.com/kubernetes/cloud-provider-openstack/blob/41a6359836200039ed847e4a326e4663d14d80dc/pkg/csi/manila/manilaclient/builder.go#L33), which explains why it's working.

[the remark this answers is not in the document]
gouthampacha: Agree on bumping the microversion with a different change; it's desirable to not depend on an old API version - it'll certainly simplify things, and we'll get to use newer features including the ability to paginate and such.

pkg/csi/manila/shareadapters/cephfs.go
```diff
@@ -32,82 +33,92 @@ type Cephfs struct{}
...
-
-	backoff := wait.Backoff{
-		Duration: time.Second * 5,
-		Factor:   1.2,
-		Steps:    10,
-	}
-
-	return accessRight, wait.ExponentialBackoff(backoff, func() (bool, error) {
-		rights, err := args.ManilaClient.GetAccessRights(ctx, args.Share.ID)
+	// Search again because access rights have changed
```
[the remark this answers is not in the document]
mandre: The call to `args.ManilaClient.GrantAccess()` returns the created access right, which we can append to the `rights` list. If we do that, we do not have to call `args.ManilaClient.GetAccessRights()` after we've modified the access rights.

But this is a nit and can totally be part of a follow-up patch.

pkg/csi/manila/shareadapters/shareadapter.go
```diff
@@ -43,10 +43,11 @@ type SecretArgs struct {
 }
 
 type ShareAdapter interface {
-	// GetOrGrantAccess first tries to retrieve an access right for args.Share.
-	// An access right is created for the share in case it doesn't exist yet.
+	// GetOrGrantAccess first tries to retrieve the list of access rights for args.Share.
```
mandre, line 46: The function name is now `GetOrGrantAccesses`.

silvacarloss: Oops, fixed it now :)

silvacarloss:
Hello, as we mentioned a couple of weeks ago we were testing further with CephFS. We can confirm that all tests passed.

gouthampacha:
/lgtm 

thank you @silvacarloss

docs/manila-csi-plugin/using-manila-csi-plugin.md
```diff
@@ -71,7 +71,7 @@ Parameter | Required | Description
 ----------|----------|------------
 `shareID` | if `shareName` is not given | The UUID of the share
 `shareName` | if `shareID` is not given | The name of the share
-`shareAccessID` | _yes_ | The UUID of the access rule for the share
+`shareAccessIDs` | _yes_ | Comma separated UUIDs of access rules for the share
```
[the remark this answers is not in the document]
silvacarloss: Hey, good point and thanks for the feedback! So the share adapter file in this PR is handling that situation in this function: `getAccessIDs`. So we first check if there is a value for `shareAccessIDs` and then we attempt with `shareAccessID`. If you prefer, we can keep the `shareAccessID` in this part of the documentation and add a note mentioning that multiple access IDS field is encouraged, as we did in the shareoptions.go. Is there some other place I'm missing this though?

silvacarloss:
> see also the #3000

@kayrus thank you for the heads-up and for the review. Can we keep #3000 as a separate issue and deal with it in another pull request? The current pull request is addressing an RFE we heard from a couple of deployments, so would be nice if we don't need to hold because of the access rule state issue. We can fix the other issue as soon as we have the cycles to.

silvacarloss:
/retest-required

Apparently hitting an image pull issue

silvacarloss:
Rebased on top of master to pick up the image pull fix.

gouthampacha:
@kayrus: Can you help merge this?

stephenfin:
This has been here a long time. All bar one of the comments are resolved and I don't believe there's anything else needed in this PR. I certainly wouldn't block on #3000.

@silvacarloss could you please address @mandre's comments regarding a plural form of the `CephfsClientID` struct field when you can. If you have the context to able to work on a fix for #3000 also it would be hugely appreciated.

/retitle [manila-csi-plugin] Support multiple share rules
/approve
/lgtm
/unhold

PR #2948, opened 2025-07-29
[occm] Fix floating IP allocation for shared internal load balancers

### LINKED ISSUE #2891: [occm] Sharing an internal load balancer should not try to allocate a floating IP
/kind bug

**What happened**:

We need to share a mysql port between cluster internally (internal LB without floating IP) and decided to piggyback onto the existing internal LB.

Sharing the LB works fine, but the occm tries to attach a floating IP to the LB, causing spurious warnings.

IMO sharing the LB should not try to change the LB config any further than adding the port.

The LB is shared without causing warnings and trying to attach a floating IP to an internal LB

**How to reproduce it**:

Deploy a LB service with the sharing annotation as [documented](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md#sharing-load-balancer-with-multiple-services) targeting a LB with the annotation `service.beta.kubernetes.io/openstack-internal-load-balancer`

**Environment**:

Note: we are using clusters on a SaaS so we do not have direct access to the CCM, if ther eis a version where it is already fixed, please let uns know so we can request an upgrade

- OpenStack version:
+---------------+---------+
| Service Type  | Version |
+---------------+---------+
| block-storage | 3.0     |
| orchestration | 1.0     |
| image         | 2.1     |
| stack-manager | None    |
| compute       | 2.1     |
| dns           | 2.0     |
| network       | 2.0     |
| load-balancer | 2.25    |
| identity      | 3.14    |
| key-manager   | 1.0     |
+---------------+---------+

skartikey:
Fixes an issue where the OpenStack Cloud Controller Manager incorrectly attempts to allocate floating IPs when sharing internal load balancers between services. This causes spurious warnings and potential failures when using the load balancer sharing feature with internal services.

**What happened:**
When sharing an internal load balancer (marked with `service.beta.kubernetes.io/openstack-internal-load-balancer: "true"`), the controller would still try to attach a floating IP to the shared load balancer, even though internal load balancers should not have floating IPs by design.

**What this PR does:**
- Adds a check in `ensureFloatingIP` to skip floating IP operations for shared internal load balancers
- Returns the VIP address directly for internal load balancers when the service is not the load balancer owner
- Prevents unnecessary floating IP allocation attempts while preserving existing behavior for all other scenarios

fixes #2891

This change only affects the code path for services that are both internal (`svcConf.internal = true`) and not the load balancer owner (`isLBOwner = false`). All existing behavior for external load balancers and load balancer owners remains unchanged.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -670,6 +670,12 @@ func (lbaas *LbaasV2) ensureFloatingIP(ctx context.Context, clusterName string, 
 		return lb.VipAddress, nil
 	}
 
+	// For shared internal load balancers, don't attempt to create/manage floating IPs
+	if svcConf.internal && !isLBOwner {
+		klog.V(4).Infof("Skipping floating IP management for shared internal load balancer service %s", serviceName)
+		return lb.VipAddress, nil
+	}
+
 	// first attempt: if we've found a FIP attached to LBs VIP port, we'll be using that.
 
 	// we cannot add a FIP to a shared LB when we're a secondary Service or we risk adding it to an internal
--- a/pkg/openstack/loadbalancer_test.go
+++ b/pkg/openstack/loadbalancer_test.go
@@ -2675,3 +2675,150 @@ func Test_getProxyProtocolFromServiceAnnotation(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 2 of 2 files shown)


hkominos:
Has there been any update on this?

PR #2951, opened 2025-07-29
[manila-csi-plugin] Support for `cloud-config` secret

### LINKED ISSUE #2532: [manila-csi-plugin] Support for `cloud-config` secret
/kind feature

**What happened**: All the other OpenStack-related plugins support a `cloud-config` secret that contains credentials for authentication with the keystone. Conversely, Manila CSI Plugin wants its own format for auth secrets, making it difficult to integrate it with existing Kubernetes-on-Openstack environments (e.g., the Charmed Kubernetes Distribution.

**What you expected to happen**: it would be easier to mount the same `cloud-config` secret in all the OpenStack plugins ecosystem, instead of having a different integration path just for Manila. Is it something feasible?

musaprg:

This PR adds support for the OpenStack cloud-config configuration format in the Manila CSI plugin, making its credential format compatible with the Cinder CSI plugin.

There is a discrepancy in the credential format between the Cinder CSI and Manila CSI plugins. The Cinder CSI plugin uses the cloud-config file `cloud.conf` as a credential, whereas the Manila CSI plugin uses individual keys stored in a Kubernetes Secret resource. Due to this difference, users have been unable to share the same Secret resources across multiple plugins using a consistent format.

This PR updates the Manila CSI plugin to accept the cloud-config format and falls back to the existing behavior if the `cloud.conf` data is missing.

fixes #2532

**Release note**:

```release-note
Support cloud.conf as a credential of manila-csi-plugin
```

### DIFF
--- a/docs/manila-csi-plugin/using-manila-csi-plugin.md
+++ b/docs/manila-csi-plugin/using-manila-csi-plugin.md
@@ -80,7 +80,30 @@ _Note that the Node Plugin of CSI Manila doesn't care about the origin of a shar
 
 ### Secrets, authentication
 
-Authentication with OpenStack may be done using either user or trustee credentials.
+Authentication with OpenStack may be done using either user or trustee credentials, and can be configured using two formats:
+
+1. **Individual keys format** - Individual secret keys written in the k8s Secret resource.
+2. **Cloud-config format** - Standard OpenStack configuration format
+
+#### Individual keys format
+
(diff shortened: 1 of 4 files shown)


musaprg:
Sorry for bothering you. Let me mark this as a draft because we’ve started discussions about this in https://github.com/kubernetes/cloud-provider-openstack/issues/2532#issuecomment-3133393603
I’ll reopen or close this PR after the direction is decided.

PR #2958, opened 2025-08-01
tests: create volumev3 service type

kayrus:

Devstack starting from epoxy renamed cinder volume types

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/playbooks/roles/install-devstack/tasks/main.yml
+++ b/tests/playbooks/roles/install-devstack/tasks/main.yml
@@ -115,3 +115,13 @@
         cmd: |
           set -ex
           sudo -u {{ user }} -H ./stack.sh
+
+    - name: Create volumev3 service type
+      shell:
+        executable: /bin/bash
+        chdir: "{{ workdir }}"
+        cmd: |
+          set -ex
+          source openrc admin admin
+          openstack service create --name cinder volumev3
+          openstack endpoint create --region $OS_REGION_NAME volumev3 public `openstack endpoint list --service block-storage -f value -c URL`


tests/playbooks/roles/install-devstack/tasks/main.yml
```diff
@@ -115,3 +115,13 @@
...
           set -ex
           sudo -u {{ user }} -H ./stack.sh
+
+    - name: Create volumev3 service type
+      shell:
+        executable: /bin/bash
+        chdir: "{{ workdir }}"
+        cmd: |
+          set -ex
+          source openrc admin admin
```
stephenfin, line 125:
```suggestion
          export OS_CLOUD=devstack-admin
```

tests/playbooks/roles/install-devstack/tasks/main.yml
```diff
@@ -115,3 +115,13 @@
...
+          openstack service create --name cinder volumev3
+          openstack endpoint create --region $OS_REGION_NAME volumev3 public `openstack endpoint list --service block-storage -f value -c URL`
```
stephenfin, line 127: Region is such a built-in assumption that I would just hardcode it, personally

```suggestion
          openstack endpoint create --region RegionOne volumev3 public $(openstack endpoint list --service block-storage -f value -c URL)
```

stephenfin (commented):
My comments are nits at best. What we have is fine. I think this should be stable-only though (1.33 and earlier).

stephenfin:
> My comments are nits at best. What we have is fine. I think this should be stable-only though (1.33 and earlier).

If we do go for stable-only, we either need to fold this into the existing Epoxy + Ubuntu 24.04 backports or harden the script since `volumev3` will already exist there and Keystone has a unique index on endpoint name + service type.

PR #2964, opened 2025-08-04
[occm] Support keystone token for openstack auth

archerwu9425:
Support keystone token for openstack auth in occm

fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
+++ b/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md
@@ -156,6 +156,8 @@ The options in `Global` section are used for openstack-cloud-controller-manager 
   The secret of an application credential to authenticate with.
 * `tls-insecure`
   If set to `true`, then the server’s certificate will not be verified. Default is `false`.
+* `token`
+  Keystone token.
 
 ###  Networking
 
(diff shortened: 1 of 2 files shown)


kayrus:
@archerwu9425 Thanks for the PR. What is the use case for supporting token-based authentication? Since tokens are short-lived, they are not intended for use in long-running services like OCCM.

archerwu9425:
> @archerwu9425 Thanks for the PR. What is the use case for supporting token-based authentication? Since tokens are short-lived, they are not intended for use in long-running services like OCCM.

@kayrus keystone token not have to be short-lived, the life time can be configured. In our case, we have configured keystone token as the only way for openstack auth, make its lifetime to medium like 15 days to avoid using long-live credential like application credential. 

Also I'm wonder if we should find a way to support using short-lived token for OCCM and support dynamic load config, so we can rotate the config without restart the OCCM pod. Short-lived token is used more and more in production env for security consideration, and for cloud provider like aws, we can use instance profile to avoid setting any credential in cluster. Wondering if we have plans for running occm with short-lived token.

jiangfwa:
@kayrus we met the same issues about using short-lived token for openstack API auth. From security perspective, having a long-lived credential in kubernetes cluster is a bit risky. Should we make the openstack-cloud-controller-manager to support token-based authentication?

kayrus:
@archerwu9425 could you please resolve conflicts with docs?

archerwu9425:
@kayrus conflict resolved. Thanks

PR #2965, opened 2025-08-04
[cinder-csi-plugin]: do not allow DetachVolume when status is detaching

Boston01:

This PR updates the `DetachVolume` logic to check if a disk is already in the `detaching` state before attempting to detach it again. If the disk is in the `detaching` state, the function now returns early with an appropriate message, preventing repeated detach attempts and unnecessary error loops. This helps avoid situations where `WaitDiskDetached` fails and the detach operation is retried in a loop.

fixes #

- Ensures that volumes in the `detaching` state are not subject to repeated detach attempts.
- Improves error handling and reduces unnecessary API calls during detach failures.

**Release note**:

```release-note
Prevent repeated detach attempts for volumes already in 'detaching' state, improving stability and error handling during volume detach operations.
```

### DIFF
--- a/pkg/csi/cinder/openstack/openstack_volumes.go
+++ b/pkg/csi/cinder/openstack/openstack_volumes.go
@@ -43,6 +43,7 @@ import (
 const (
 	VolumeAvailableStatus    = "available"
 	VolumeInUseStatus        = "in-use"
+	VolumeDetachingStatus    = "detaching"
 	operationFinishInitDelay = 1 * time.Second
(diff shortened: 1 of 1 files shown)


Boston01:
Hi @zetaab, it's been a while i hope you're doing great.
Can you have a look on this merge request ?

stephenfin:
Eek, that timeout looks like the thing we were seeing in https://github.com/kubernetes/cloud-provider-openstack/pull/2742 and should have fixed. will investigate later in the week.

PR #2985, opened 2025-09-01
[cinder-csi-plugin] Update ignore-volume-az docs

### LINKED ISSUE #2185: [cinder-csi-plugin] `ignore-volume-az` appears to be either useless or incorrectly documented
/kind bug

(though this is potentially more question than bug)

**What happened**:

The `ignore-volume-az` option is documented like so:

> `ignore-volume-az` Optional. When `Topology` feature enabled, by default, PV volume node affinity is populated with volume accessible topology, which is volume AZ. But, some of the openstack users do not have compute zones named exactly the same as volume zones. This might cause pods to go in pending state as no nodes available in volume AZ. Enabling `ignore-volume-az=true`, ignores volumeAZ and schedules on any of the available node AZ. Default `false`. Check `cross_az_attach` in [nova configuration](https://docs.openstack.org/nova/latest/configuration/config.html) for further information.

Thought it is implied here, it isn't explicitly stated that this only affects the created `PersistentVolume`. When this is set to `false`, each PV will have a `spec.nodeAffinity` field like so:

```
$ kubectl get pv foo -o yaml
...
  nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchExpressions:
        - key: topology.cinder.csi.openstack.org/zone
          operator: In
          values:
          - nova
```

By setting it to `true`, we do not populate this field. However, this does not affect the `CSINode` or `Node` CRDs. When the `Topology` feature is enabled, each `CSINode` will have `spec.drivers[cinder.csi.openstack.org].topologyKeys` value populated with `[topology.cinder.csi.openstack.org/zone]`, while the corresponding `Node` will have a `topology.cinder.csi.openstack.org/zone: $az` metadata label. `$az` here is the Compute availability zone retrieved from the OpenStack instance metadata service, which kubelet in turn retrieves via the `NodeGetInfo` gRPC API. The value `$az` is used when making a request for a volume to Cinder unless overridden by the `availability` parameter on the `StorageClass`. All this means that if there is a mismatch between Nova and Cinder AZs, setting `ignore-volume-az` won't help since the volume request sent to Cinder will use an invalid (to Cinder) AZ.

Given the above, I can only think of one use case where the option would be useful: where you had set the `availability` parameter on your storage class(es), allowing you to e.g. have a storage class per AZ. In this case, you would want to ensure the AZ requested for the Cinder volume was not transformed into a nodeAffinity for the resulting PVs. However, if this is what you're doing, why are you even enabling the `Topology` feature in the first place? Disabling this won't change the request to Cinder if this value is provided. As such, I don't know who this config option is intended for and neither #1735 nor #1739 help me.

I'm not sure :smile:

**How to reproduce it**:

1. Deploy OpenStack using different Cinder and Nova AZs.
2. Enable the `Topology` feature for Cinder CSI and set `[BlockStorage] ignore-volume-az = True`
3. Attempt to create a Pod with a PVC attached.

This will fail unless `[DEFAULT] allow_availability_zone_fallback` is set to `true` in `cinder.conf`. Even in this instance, we're still requesting invalid data from cinder: cinder is just able to ignore it and fallback to `[DEFAULT] default_availability_zone` (while the PVC won't have a `nodeAffinity` value corresponding to the _real_ AZ used by Cinder because of `[BlockStorage] ignore-volume-az = True`

N/A

**Environment**:
- OpenStack version: Wallaby
- Others: -

stephenfin:

This option is only useful if you are manually configuring volume AZs using the `topology` Storage Class parameter. Update docs to reflect this and suggest (likely better) alternatives for working around AZ mismatches.

fixes #2185

https://github.com/kubernetes/cloud-provider-openstack/pull/2865 makes the topology code easier (IMO) to grok, if anyone doubts my testing notes from https://github.com/kubernetes/cloud-provider-openstack/issues/2185#issuecomment-1488353753.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/cinder-csi-plugin/features.md
+++ b/docs/cinder-csi-plugin/features.md
@@ -41,7 +41,7 @@ For usage, refer [sample app](./examples.md#use-topology)
 
 ## Block Volume
 
-Cinder volumes to be exposed inside containers as a block device instead of as a mounted file system. The corresponding CSI feature (CSIBlockVolume) is GA since Kubernetes 1.18.
+Cinder volumes to be exposed inside containers as a block device instead of as a mounted file system. The corresponding CSI feature (CSIBlockVolume) is GA since Kubernetes v1.18.
 
 Prerequisites to use the feature:
 * Make sure the volumeMode is `Block` in Persistence Volume Claim Spec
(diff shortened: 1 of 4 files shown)


rptaylor:
@stephenfin thanks for this clarification!
Do you know how this relates to the `--feature-gates=Topology=false` setting that is applied by the Helm chart?
https://github.com/kubernetes/cloud-provider-openstack/blob/master/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml#L72

Mentioned here: https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/cinder-csi-plugin/features.md#topology

If you have `--feature-gates=Topology=false` is it still necessary to also set `--with-topology=false` for the node and controller plugins (`.Values.csi.plugin.controllerPlugin.extraArgs`, `.Values.csi.plugin.nodePlugin.extraArgs`) ?

stephenfin:
> @stephenfin thanks for this clarification! Do you know how this relates to the `--feature-gates=Topology=false` setting that is applied by the Helm chart? [`master`/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml#L72](https://github.com/kubernetes/cloud-provider-openstack/blob/master/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml?rgh-link-date=2025-09-10T18%3A28%3A53Z#L72)
> 
> Mentioned here: [`master`/docs/cinder-csi-plugin/features.md#topology](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/cinder-csi-plugin/features.md?rgh-link-date=2025-09-10T18%3A28%3A53Z#topology)
> 
> If you have `--feature-gates=Topology=false` is it still necessary to also set `--with-topology=false` for the node and controller plugins (`.Values.csi.plugin.controllerPlugin.extraArgs`, `.Values.csi.plugin.nodePlugin.extraArgs`) ?

No, it's not. Setting that to `false` disables it at the provisioning stage. However, note that feature flags are not kept around forever and this will eventually be removed from the provisioner. This was the rationale for adding the `--with-topology` flag. When it is, these charts will break. We likely need to update them sooner rather than later.

PR #2684, opened 2024-10-02
[occm] Support Octavia load balancing method via annotation

Lucasgranet:
**What this PR does / why we need it**:
Adds the ability to change the LB Algorithm of an Octavia using an annotation.

This PR brings a new annotation:
```yaml
loadbalancer.openstack.org/lb-method: "" # Default value: the one defined in the OCCM configuration file
```

**Special notes for reviewers**:
N/A

**Release note**:

```release-note
[occm] Octavia - ability to change the load balancing method using an annotation.
```

PR #2988, opened 2025-09-10
occm: Set lbMethod annotation to svcConfig

### LINKED ISSUE #2987: [occm] lb_method doesnt get updated when setting annotation
/kind bug

**What happened**:
When updating a loadbalancer service using the following annotation

```
  annotations:
    loadbalancer.openstack.org/lb-method: "SOURCE_IP"
```
The algorithm of the loadbalancer does not change from ROUND_ROBIN to SOURCE_IP.

The loadbalancer algorithm should be set to "SOURCE_IP"

**How to reproduce it**:

Create a loadbalancer, internal or external, and then set the above annotation
to change the loadbalancer algorithm.

The above bug has been tested with setting the LB algorithm to all the possible
algorithms (SOURCE_IP, LEAST_CONNECTIONS) and it didnt work.
However, after manually changing the LB algorithm via Horizon Openstack and then
using the above annotation to set it to any possible algorithm the lb algorithm instead
of changing to the desired algorithm it changes it to ROUND_ROBIN. Therefore, I think
the problem is caused by the following line not working as intended:

https://github.com/kubernetes/cloud-provider-openstack/blob/v1.33.1/pkg/openstack/loadbalancer.go#L907

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: 1.33.1
- OpenStack version: -

Thank you very much, for your time and for maintaing the openstack cloud provider :)

nikParasyr:

Currently the lbMethod is not set into the svcConfig which leads to the default lbMethod being used

This was initially introduced in #https://github.com/kubernetes/cloud-provider-openstack/pull/2684/files but was later removed.

fixes #2987 

**Release note**:

```release-note
Fix setting LbMethod on occm
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1540,6 +1540,7 @@ func (lbaas *LbaasV2) checkService(ctx context.Context, service *corev1.Service,
 func (lbaas *LbaasV2) makeSvcConf(ctx context.Context, serviceName string, service *corev1.Service, svcConf *serviceConfig) error {
 	svcConf.connLimit = getIntFromServiceAnnotation(service, ServiceAnnotationLoadBalancerConnLimit, -1)
 	svcConf.lbID = getStringFromServiceAnnotation(service, ServiceAnnotationLoadBalancerID, "")
+	svcConf.poolLbMethod = getStringFromServiceAnnotation(service, ServiceAnnotationLoadBalancerLbMethod, "")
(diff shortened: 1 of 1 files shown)


kayrus:
/lgtm
@nikParasyr , thanks

kayrus:
golang download timeout:

```
 TASK [install-golang : Download the Go tarball] ********************************
[ERROR]: Task failed: Module failed: Connection failure: The read operation timed out
```

PR #2997, opened 2025-09-22
[manila-csi-plugin] Fix unresolved name for release version on worker node setup

silvacarloss:
While setting up an environment and attempting to use the CI scripts to create a couple of worker nodes, an error was raised due to the `release.stdout` value not being resolved to anything.

While setting up the master nodes, we do use `k3s_release` to define the version to be used in the URL.

This change fixes the worker node set up script to use the value of `k3s_release`.


stephenfin (commented):
This is the same as what we do for the master nodes. It makes sense to do it for workers also.

/approve
/lgtm

PR #3007, opened 2025-10-07
[k8s-keystone-auth] Remove use of github.com/go-chi/chi/v5

stephenfin:

Per $subject, this is overkill for our tiny use case. Tests are added to ensure we preserve behavior (though visual inspection would say as much).

(none)

**Release note**:

```release-note
NONE
```


zetaab (approved):
/approve
/lgtm

PR #3020, opened 2025-11-11
[occm] Add HTTP health checks

rybnico:

I realised that we had many failed requests to load balancers on our K8s platforms using our OpenStack platform when we did maintenance that involved replacing nodes in a rolling fashion. I opened an [issue](https://github.com/kubernetes/kubernetes/issues/133052) in the Kubernetes project and I was told that the correct solution would be to have HTTP health checks on the Services to make sure not only the port is reachable, but also the service behind the port.

As the OCCM only supports HTTP health checks for load balancers with a TrafficPolicy of "Local" on the HealthCheckNodePort, I implemented HTTP monitors using additional annotations.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
+++ b/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
@@ -194,6 +194,30 @@ Request Body:
 
   NOTE: Health monitors for the `ovn` provider are only supported on OpenStack Wallaby and later.
 
+- `loadbalancer.openstack.org/health-monitor-http-ports`
+
+  A list of the ports (specified by name or number) that should get an HTTP(S) health check, separated by commas.
+
+- `loadbalancer.openstack.org/health-monitor-http-types`
+
+  One or more check types can be selected for HTTP(S) health checks. These can be set to either `HTTP` or `HTTPS`. If multiple HTTP health check ports are defined, you can set either one value for all checks or multiple values separated by commas for individual checks, using the same index as in `health-monitor-https-ports`. Defaults to `HTTP`.
+
+- `loadbalancer.openstack.org/health-monitor-http-method`
+
+  The HTTP method that should be used for HTTP health checks. Valid values are `CONNECT`, `DELETE`, `GET`, `HEAD`, `OPTIONS`, `PATCH`, `POST`, `PUT` and `TRACE`. Defaults to `GET`
+
+- `loadbalancer.openstack.org/health-monitor-http-version`
+
+  The HTTP version that should be used for HTTP health checks. Valid values are `1.0` and `1.1`. Defaults to `1.0`.
+
+- `loadbalancer.openstack.org/health-monitor-expected-codes`
+
+  A list of HTTP status codes that should be considered healthy by the HTTP health checks. Valid values include a single value, such as `200`, a list, such as `200, 202` or a range, such as `200-204`. Defaults to `200`.
+
+- `loadbalancer.openstack.org/health-monitor-url-path`
+
+  The URL that should be used by the HTTP health checks. The value must start with a `/`. Defaults to `/`.
+
 - `loadbalancer.openstack.org/health-monitor-delay`
 
   Defines the health monitor delay in seconds for the loadbalancer pools.
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -87,6 +87,12 @@ const (
 	ServiceAnnotationLoadBalancerHealthMonitorTimeout        = "loadbalancer.openstack.org/health-monitor-timeout"
 	ServiceAnnotationLoadBalancerHealthMonitorMaxRetries     = "loadbalancer.openstack.org/health-monitor-max-retries"
 	ServiceAnnotationLoadBalancerHealthMonitorMaxRetriesDown = "loadbalancer.openstack.org/health-monitor-max-retries-down"
+	ServiceAnnotationLoadBalancerHealthMonitorHTTPPorts      = "loadbalancer.openstack.org/health-monitor-http-ports"
+	ServiceAnnotationLoadBalancerHealthMonitorHTTPTypes      = "loadbalancer.openstack.org/health-monitor-http-types"
+	ServiceAnnotationLoadBalancerHealthMonitorHTTPMethod     = "loadbalancer.openstack.org/health-monitor-http-method"
+	ServiceAnnotationLoadBalancerHealthMonitorHTTPVersion    = "loadbalancer.openstack.org/health-monitor-http-version"
+	ServiceAnnotationLoadBalancerHealthMonitorExpectedCodes  = "loadbalancer.openstack.org/health-monitor-expected-codes"
+	ServiceAnnotationLoadBalancerHealthMonitorURLPath        = "loadbalancer.openstack.org/health-monitor-url-path"
 	ServiceAnnotationLoadBalancerLoadbalancerHostname        = "loadbalancer.openstack.org/hostname"
 	ServiceAnnotationLoadBalancerAddress                     = "loadbalancer.openstack.org/load-balancer-address"
 	// revive:disable:var-naming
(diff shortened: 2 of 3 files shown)


kayrus:
> As the OCCM only supports HTTP health checks for load balancers with a TrafficPolicy of "Local" on the HealthCheckNodePort, I implemented HTTP monitors using additional annotations.

I never faced such issue before. With the `TrafficPolicy: Cluster` k8s automatically marks a pod as unavailable, and traffic is redirected by the node's kube-proxy to another node. If the node is marked for maintenance it's payload should be gracefully transferred to another node. When the payload is going to be shut down, it must stop receive new requests and gracefully handle already incoming connections. Thus I believe there is something wrong with your setup or your applications.

BTW, with the HTTP healthcheck you'll get the same problem, since loadbalancers usually have a latency with a healthcheck, and this latency is much more higher than k8s API understands that the pod is going to shut down, see https://github.com/kubernetes/kubernetes/issues/133165 for more details. See also #2148 and try to use the node selector for the node maintenance https://github.com/kubernetes/cloud-provider-openstack/pull/2601

See also https://github.com/kubernetes/cloud-provider-openstack/pull/2869 docs update, which should be available since k8s 1.34

rybnico:
> I never faced such issue before. With the `TrafficPolicy: Cluster` k8s automatically marks a pod as unavailable, and traffic is redirected by the node's kube-proxy to another node. If the node is marked for maintenance it's payload should be gracefully transferred to another node. When the payload is going to be shut down, it must stop receive new requests and gracefully handle already incoming connections. Thus I believe there is something wrong with your setup or your applications.

I encountered this issue when adding new nodes. New nodes may be added to the LoadBalancer before the network is fully set up. The service node ports open before Kubernetes networking can forward requests to the service pods, so the TCP health check succeeds but the application itself cannot be reached.

> BTW, with the HTTP healthcheck you'll get the same problem, since loadbalancers usually have a latency with a healthcheck, and this latency is much more higher than k8s API understands that the pod is going to shut down, see [kubernetes/kubernetes#133165](https://github.com/kubernetes/kubernetes/issues/133165) for more details. See also #2148 and try to use the node selector for the node maintenance #2601

The LoadBalancer should not add a new backend until at least one successful HTTP health check has been performed. This should help with the scenario I described above.

> See also #2869 docs update, which should be available since k8s 1.34

I can't think of a solution for auto-labelling nodes for Cluster-API or KOPS-based deployments without manual action or a custom controller.

Aside from the issues we encountered with our load balancers when maintaining the clusters, I believe that support for HTTP health checks would be beneficial in general.

kayrus:
> The LoadBalancer should not add a new backend until at least one successful HTTP health check has been performed. This should help with the scenario I described above.

Unfortunately this is not true for OpenStack Octavia. A new pool member receives the traffic immediately independently on the healthcheck results. This is true for OpenStack Amphora and for F5 Octavia backends/drivers. Other backends/drivers should follow the same approach. If not, please provide an evidence.

> Aside from the issues we encountered with our load balancers when maintaining the clusters, I believe that support for HTTP health checks would be beneficial in general.

I don't think so, considering the facts above and the overall high latency for healthchecks from the LB side. Believe me, I'm trying to find a solution for the `TrafficPolicy: Local` downtime, when a new pool member is added, for several years. And the only solution to do this properly: fix octavia default behavior, then fix corresponding drivers/backends. Until this is implemented, please use service annotations and node labels to exclude a node from the lb pool members.

rybnico:
> A new pool member receives the traffic immediately independently on the healthcheck results. This is true for OpenStack Amphora and for F5 Octavia backends/drivers. Other backends/drivers should follow the same approach. If not, please provide an evidence.

You're absolutely right. This doesn't seem logical to me, and other load balancer solutions that I know of always check new backends before sending traffic to them. Sorry for the misinformation, and thank you for clearing things up.

I found a theoretical solution to this problem: if you create a new member with `AdminStateUp` set to `false` and then set it to `true` after creation, the operating status doesn't become `ONLINE` if the health check fails. This solution is a bit cumbersome to implement (new members have to be created with `AdminStateUp` set to `false` and updated later), but I think it would solve this issue and others relating to TCP/UDP health checks too.

I would be happy to give it a try if you're interested in an implementation.

kayrus:
> I found a theoretical solution to this problem: if you create a new member with `AdminStateUp` set to `false` and then set it to `true` after creation, the operating status doesn't become `ONLINE` if the health check fails. This solution is a bit cumbersome to implement (new members have to be created with `AdminStateUp` set to `false` and updated later), but I think it would solve this issue and others relating to TCP/UDP health checks too.

I'm aware about this, but in this case we need to have two reconciliation cycles:
1) add an offline member
2) mark it online

And we must have a delay between these cycles, what you can not really determine. Besides current k8s upstream cloud-controller logic doesn't have a two cycles logic.

kayrus:
Just for the reference to my research about this topic:

haproxy adds a pool member and considers it's alive: https://github.com/haproxy/haproxy/blob/697f7d1142a26352940532d0481cc17f9225d0d1/src/server.c#L2836

To change this behavior the haproxy config must contain `init-state fully-down`. But this must be indicated in the Octavia API. The resulting [haproxy template](https://github.com/openstack/octavia/blob/be37addac0a73afe7422e43e4dd0f2e283805567/octavia/common/jinja/haproxy/combined_listeners/templates/macros.j2) patch should look like:

```diff
diff --git a/octavia/common/jinja/haproxy/combined_listeners/templates/macros.j2 b/octavia/common/jinja/haproxy/combined_listeners/templates/macros.j2
index 5a81c37..45c6b91 100644
--- a/octavia/common/jinja/haproxy/combined_listeners/templates/macros.j2
+++ b/octavia/common/jinja/haproxy/combined_listeners/templates/macros.j2
@@ macro member_macro(constants, lib_consts, pool, member) %}
     {% if pool.health_monitor and pool.health_monitor.enabled %}
         {% if member.monitor_address %}
@@
     {% endif %}
     {% if pool.alpn_protocols is defined %}
         {% set alpn_opt = " alpn %s"|format(pool.alpn_protocols) %}
     {% else %}
         {% set alpn_opt = "" %}
     {% endif %}
+    {% if member.monitor_init_state_down %}
+        {% set init_state_opt = " init-state fully-down" %}
+    {% else %}
+        {% set init_state_opt = "" %}
+    {% endif %}
-    {{ "server %s %s:%d weight %s%s%s%s%s%s%s%s%s%s%s%s%s%s%s"|e|format(
-        member.id, member.address, member.protocol_port, member.weight,
-        hm_opt, persistence_opt, proxy_protocol_opt, member_backup_opt,
-        member_enabled_opt, def_opt_prefix, def_crt_opt, ca_opt, crl_opt,
-        def_verify_opt, def_sni_opt, ciphers_opt, tls_versions_opt,
-        alpn_opt)|trim() }}
+    {{ "server %s %s:%d weight %s%s%s%s%s%s%s%s%s%s%s%s%s%s%s%s"|e|format(
+        member.id, member.address, member.protocol_port, member.weight,
+        hm_opt, persistence_opt, proxy_protocol_opt, member_backup_opt,
+        member_enabled_opt, def_opt_prefix, def_crt_opt, ca_opt, crl_opt,
+        def_verify_opt, def_sni_opt, ciphers_opt, tls_versions_opt,
+        alpn_opt, init_state_opt)|trim() }}
 {% endmacro %}
```

Once the upstream octavia codebase get this support, backend vendors should also add this support, and we can implement support for this option in OCCM.

kayrus:
@rybnico I submitted https://bugs.launchpad.net/octavia/+bug/2131245

rybnico:
Thank you for the detailed information, @kayrus!

I just wanted to ask if you had created a PR/bug in the Octavia project. Then I saw your reply.

I will keep an eye on the Octavia bug. Please tell me if there's anything I can help with.

kayrus:
> I just wanted to ask if you had created a PR/bug in the Octavia project.

No, I hope that someone from Octavia team can handle this.

> Please tell me if there's anything I can help with.

you can subscribe to this bug to increase the bug severity

PR #3039, opened 2026-01-13
[occm] Fix serviceMonitor labelSelector and remove deprecated bearerTokenFile auth

### LINKED ISSUE #3144: [occm] Helm chart ServiceMonitor does not work out-of-the-box
/kind bug

**What happened**:
The occm Helm chart exposes Prometheus metrics and documents them in `docs/metrics.md`, but the default `ServiceMonitor` does not successfully scrape those metrics, especially when deployed with prometheus.

we observed some issues with the generated resources, 
- The `ServiceMonitor` selector does not match the `Service` created by the chart.
- The `ServiceMonitor` still relies on the deprecated `bearerTokenFile` authentication.
- As a result, OCCM metrics are not collected well.

without working metrics, the occm reconciliation failures or repeated OpenStack API errors can go unnoticed until users begin experiencing faileds.

Using the `ServiceMonitor` should allow Prometheus to scrape OCCM metrics without additional manual configuration.

OCCM operators should be able to consume the metrics described in `docs/metrics.md` and build alerts for controller reconciliation failures and openstack api errors.

**How to reproduce it**:
```text
1. deploy `openstack-cloud-controller-manager` using the upstream Helm chart.
2. enable `ServiceMonitor`.
3. deploy prometheus operator.
4. observe that the `ServiceMonitor` does not successfully scrape the OCCM metrics,
because the generated resources do not correctly reference the `Service` and still use deprecated authentication.

```

This issue was identified while integrating OCCM monitoring into production deployments.

The proposed fix is available in **PR #3039**, which is,
- updates the ServiceMonitor selector to match the generated Service,
- removes the deprecated `bearerTokenFile` usage,
- replaces the generated Service and ServiceMonitor manifests with resources following the current recommendations from `docs/metrics.md`.

The goal is to make OCCM metrics available without any additional configurations!

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: master branch
- OpenStack version:
- Others: prometheus operator, helm chart deployment

claudia-lola:
The labels on the Service did not match the ServiceMonitor selector. Prometheus was unable to discover the target. This change sets the labels so the ServiceMonitor selects the correct Service.
`bearerTokenFile` is deprecated and no longer works. ServiceMonitor was updated to use a supported authentication method.
`authorization-always-allow-paths="/metrics"`  was added to ensure metrics can be scraped.
fixes #3144 

**Release note**:

```release-note
Fix serviceMonitor labelSelector and remove deprecated bearerTokenFile auth
```


ednxzu:
manually tested this in my cluster, and it works fine

claudia-lola:
Hi @kayrus @stephenfin and @zetaab what is the estimate for a review of this PR please?

jihyun-huh:
Hi, @claudia-lola 
I opened issue #3144  to describe the operational problem this PR addresses.
I think it provides additional context around to reviewers why this change is needed.

jihyun-huh:
@gouthampacha Hi, i saw you are looking into this repo, can i friendly ask you for adding ok-to-test tag?

charts/openstack-cloud-controller-manager/templates/daemonset.yaml
```diff
@@ -58,6 +58,7 @@ spec:
             - --cloud-provider=openstack
             - --use-service-account-credentials=false
             - --controllers={{- trimAll "," (include "occm.enabledControllers" . ) -}}
+            - --authorization-always-allow-paths="/metrics"
```
jihyun-huh: One small question before asking an approval, 
could `--authorization-always-allow-paths=/metrics` be placed inside the `serviceMonitor.enabled` block? 

When `ServiceMonitor` is disabled, I don't think the chart needs to change the authorization policy, even the endpoint remains bound to localhost(0.0.0.0).

if i miss something in this PR's history, plz tell me

```suggestion
{{- if .Values.serviceMonitor.enabled }}
- --authorization-always-allow-paths=/metrics
- --bind-address=0.0.0.0
{{- else }}
- --bind-address=127.0.0.1
{{- end }}
```

charts/openstack-cloud-controller-manager/templates/servicemonitor.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: {{ include "occm.name" . }}
-  labels: 
+  labels:
```
jihyun-huh: could you clean this difference?

jihyun-huh (commented):
Hi, @claudia-lola I added some small reviews.
apart from that, the approach looks reasonable to me.

jihyun-huh:
Thanks @claudia-lola, 
Ive reviewed the latest update and verified the rendered chart locally with `serviceMonitor.enabled=true`.

Hi, @kayrus I don’t see any remaining blockers from my side. Looks good to merge.

jihyun-huh:
Hi @claudia-lola , one small follow-up after #3080 was merged, the OCCM chart version on master is now `2.36.1`, which is the same version currently used in this branch.
https://github.com/kubernetes/cloud-provider-openstack/blob/d4b45cea305f9a79905ec712a1979e7c6230b456/charts/openstack-cloud-controller-manager/Chart.yaml#L7
Could you please rebase on the latest master and update the chart version to `2.36.2`?

cc. @gouthampacha @kayrus

jihyun-huh:
@gouthampacha, would you be able to take a reviewer pass and add /lgtm, if this looks good?

cc. @kayrus for approval after the reviewer pass. This fixes the out-of-the-box ServiceMonitor issue described in #3144.

gouthampacha:
/lgtm

looks good, thanks @claudia-lola

PR #3043, opened 2026-01-19
feat(openstack-cloud-controller-manager): add option to change deployment type

cwrau:

This PR allows the user to choose between DaemonSet and Deployment.

**Release note**:

```release-note
feat(openstack-cloud-controller-manager): allow user to choose between DaemonSet and Deployment
```

### DIFF
--- a/charts/openstack-cloud-controller-manager/values.yaml
+++ b/charts/openstack-cloud-controller-manager/values.yaml
@@ -2,6 +2,14 @@
 #
 # Define deployment mode for the controller and provide cloud credentials in cloudConfig at the end of the file
 #
+# Controller kind: DaemonSet or Deployment
+# Use DaemonSet to run on all control-plane nodes (default)
+# Use Deployment to run a specific number of replicas
+controllerKind: DaemonSet
+
+# Number of replicas (only used when controllerKind is Deployment)
+controllerReplicas: 2
+
 ## Annotations to apply to all resources
 commonAnnotations: {}
 # commonAnnotations:
@@ -48,9 +56,13 @@ readinessProbe: {}
 
 dnsPolicy: ClusterFirst
 
-# Set nodeSelector where the controller should run, i.e. controlplane nodes
+# Set nodeSelector where the controller should run, i.e. controlplane nodes (used for DaemonSet)
 nodeSelector:
   node-role.kubernetes.io/control-plane: ""
+
+# Set nodeSelector for Deployment (used when controllerKind is Deployment)
+# Defaults to empty (no node restriction)
+deploymentNodeSelector: {}
 
 # Set tolerations for nodes where the controller should run, i.e. node
 # should be uninitialized, controlplane...
--- a/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/daemonset.yaml
@@ -1,5 +1,5 @@
 apiVersion: apps/v1
-kind: DaemonSet
+kind: {{ .Values.controllerKind }}
 metadata:
   name: {{ include "occm.name" . }}
   labels: {{- include "common.labels.standard" . | nindent 4 }}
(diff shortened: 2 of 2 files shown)


kayrus:
@cwrau thanks for the PR. Could you please describe your use case for the daemonset deployment style? I'd set the default style to be a deployment unless daemonset is set. Also please fix chart issues found by lint.

cwrau:
> @cwrau thanks for the PR. Could you please describe your use case for the daemonset deployment style? I'd set the default style to be a deployment unless daemonset is set. Also please fix chart issues found by lint.

I just wanted to keep the existing concept, so users don't have to do anything to keep the current setup.

I can set the default to deployment, but I imagine that that would kinda be a breaking change.

kayrus:
@cwrau I never used helm-charts from this repo, and I just realized that it was always a DaemonSet, this is a surprise for me. All OCCM deployments that I worked with were always a Deployment. And I wonder what was the reason to be initially a DaemonSet.

cwrau:
Still waiting for review...

stephenfin:
My gut suggests this is too large a "knob", and that if we've a good reason to use a `Deployment` rather than a `DaemonSet`, we should just make that change for everyone. We have just merged https://github.com/kubernetes/cloud-provider-openstack/issues/3194 to change how we release Helm Charts, so that we no longer release from `master` except when releasing a new release minor release (which is really a "major" version, e.g. `v1.37.0`). With this change, I think we are okay to make this change and document it in a release note?

Regarding the actual change, it seems [both are fine](https://kubernetes.io/blog/2025/02/14/cloud-controller-manager-chicken-egg-problem/#example-cloud-controller-manager-not-scheduling-due-to-uninitialized-taint). I guess a Deployment is less wasteful in larger clusters since there's no need to run CCM on a few dozen (or more) nodes.

stephenfin:
Also note that if you rebase this branch then the lint failure will be removed. We no longer expect people to bump the chart `version` on change to `master`

charts/openstack-cloud-controller-manager/values.yaml
```diff
@@ -55,10 +63,14 @@ hostNetwork: true
 # Set to ClusterFirst if hostNetwork is disabled
 dnsPolicy: ClusterFirstWithHostNet
 
-# Set nodeSelector where the controller should run, i.e. controlplane nodes
+# Set nodeSelector where the controller should run, i.e. controlplane nodes (used for DaemonSet)
 nodeSelector:
```
kayrus, line 67: can we use a single `nodeSelector` values for both daemonset and deployment?

cwrau: It's been a while, but I don't think so, because one can't remove a value from the helm default values. And with a deployment I would need an empty nodeSelector (at least that's part of the use case for this change). 

Let me check that again

cwrau: Yeah, just checked, without this split field gitops users (at least for flux) can't clear the nodeSelector, as that would only work by setting it to `null` but setting fields to `null` in yaml is dropped by the apiServer and so never reaches helm inside of flux.

stephenfin:
/ok-to-test

Just clear that, but I'd still like to figure out an answer to my earlier question https://github.com/kubernetes/cloud-provider-openstack/pull/3043#issuecomment-5605354179

cwrau:
> /ok-to-test
> 
> Just clear that, but I'd still like to figure out an answer to my earlier question https://github.com/kubernetes/cloud-provider-openstack/pull/3043#issuecomment-5605354179

Is that question for me? Then I don't understand the question 😅

stephenfin:
> > /ok-to-test
> > Just clear that, but I'd still like to figure out an answer to my earlier question [(earlier comment)](https://github.com/kubernetes/cloud-provider-openstack/pull/3043#issuecomment-5605354179)
> 
> Is that question for me? Then I don't understand the question 😅

Probably more @kayrus than you, but to restate the question: do we need an option, or can we can just update the chart to use `Deployment` instead of `DaemonSet`?

cwrau:
> > > /ok-to-test
> > > Just clear that, but I'd still like to figure out an answer to my earlier question [(earlier comment)](https://github.com/kubernetes/cloud-provider-openstack/pull/3043#issuecomment-5605354179)
> > 
> > Is that question for me? Then I don't understand the question 😅
> 
> Probably more @kayrus than you, but to restate the question: do we need an option, or can we can just update the chart to use `Deployment` instead of `DaemonSet`?

Mh, good question. I just assumed that a daemonset was the preferred way to deploy this. 

Also, making it an option instead of changing it isn't a breaking change. 

We already do a flux postRenderer to change the kind to Deployment, so as far as I can tell there is no technical requirement for it to be a daemonset 🤔

PR #3063, opened 2026-01-23
feat(openstack-cloud-controller-manager): add splitMode

cwrau:

This allows the user to deploy the controller-manager split between the
management cluster and the workload cluster.

That way the cluster is not dependent "on itself".

The management mode only deploys the controller and the workload mode only
deploys the necessary RBAC roles and bindings.

**Release note**:

```release-note
[openstack-cloud-controller-manager] Allow deploying the controller-manager split between management and workload clusters.
```

### DIFF
--- a/charts/openstack-cloud-controller-manager/Chart.lock
+++ b/charts/openstack-cloud-controller-manager/Chart.lock
@@ -1,0 +1,6 @@
+dependencies:
+- name: common
+  repository: https://charts.bitnami.com/bitnami
+  version: 2.14.1
+digest: sha256:ac04d66e6abb3cfc89f4c10e49d79770c352c510744bc51e3595937721b7f41a
+generated: "2024-01-10T10:12:58.796725699+01:00"
--- a/charts/openstack-cloud-controller-manager/Chart.lock
+++ b/charts/openstack-cloud-controller-manager/Chart.lock
@@ -1,6 +1,0 @@
-dependencies:
-- name: common
-  repository: https://charts.bitnami.com/bitnami
-  version: 2.14.1
-digest: sha256:ac04d66e6abb3cfc89f4c10e49d79770c352c510744bc51e3595937721b7f41a
-generated: "2024-01-10T10:12:58.796725699+01:00"
--- a/charts/openstack-cloud-controller-manager/templates/controller.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/controller.yaml
@@ -1,5 +1,10 @@
+{{- if or (not .Values.splitMode.enabled) (eq .Values.splitMode.type "management") }}
 apiVersion: apps/v1
+{{- if .Values.splitMode.enabled }}
+kind: Deployment
+{{- else }}
 kind: DaemonSet
+{{- end }}
 metadata:
   name: {{ include "occm.name" . }}
   labels: {{- include "common.labels.standard" . | nindent 4 }}
@@ -9,11 +14,19 @@ metadata:
     {{- toYaml . | nindent 4 }}
     {{- end }}
 spec:
+  {{- if .Values.splitMode.enabled }}
+  replicas: {{ .Values.splitMode.replicas }}
+  {{- end }}
   selector:
     matchLabels:
       {{- include "occm.controllermanager.matchLabels" . | nindent 6 }}
+  {{- if .Values.splitMode.enabled }}
+  strategy:
+    type: RollingUpdate
+  {{- else }}
   updateStrategy:
     type: RollingUpdate
+  {{- end }}
   template:
     metadata:
       annotations:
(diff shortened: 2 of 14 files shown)


cwrau:
This has shown itself to be useful when theres trouble with the nodes resulting in dozens or hundreds of `NotReady` nodes which in turn break other stuff like cilium, as their pods get scheduled there and will never run.

And the ccm can't clean those nodes up, as it doesn't run yet.

With this one can run the ccm outside the cluster which makes sure that it always works™️

cwrau:
Ok, I thought I had to bump the chart version as I had to for https://github.com/kubernetes/cloud-provider-openstack/pull/3067, but that also didn't help and I don't know how I would fix that dependency issue

cwrau:
Hello, is there something I can do to get this merged?

cwrau:
@dulek / @kayruz

Hello, not to be annoying, but is there something I can do to help getting this merged? 🙏

kayrus:
Hi @cwrau. Thanks for the PR. Please remove the ZIP file from the PR and fix lint issues.

cwrau:
> Hi @cwrau. Thanks for the PR. Please remove the ZIP file from the PR

Oh, I assuemed the zip was wanted, as there is no .gitignore.

>  and fix lint issues.

There aren't linting issues, the linter doesn't run because it fails building the dependencies, it seems CI must be adjusted? In our helm charts repo we set CT_CHARTS_REPOS to the repos so ct can download them itself:

```yaml
      - name: Extract helm repos
        id: helm-repos
        run: |
          [[ "$RUNNER_DEBUG" == 1 ]] && set -x
          (
            echo -n "repos="
            yq -r '.dependencies[] | .repository' "charts/$CHART/Chart.yaml" | sort -u | grep https:// | awk '{printf (NR>1 ? "," : "") NR "=" $1}'
            echo
          ) | tee -a "$GITHUB_OUTPUT"
...
      - name: Lint chart
        env:
          CT_CHART_REPOS: ${{ steps.helm-repos.outputs.repos }}
          CT_CHARTS: charts/${{ matrix.chart }}
        run: |
          [[ "$RUNNER_DEBUG" == 1 ]] && set -x
          if ! [[ -f "charts/$CHART/values.yaml" ]]; then
            echo "No values.yaml found for $CHART, skipping 'ct lint'" >&2
            helm lint "charts/$CHART"
          else
            ct lint --check-version-increment=false
          fi
```

charts/openstack-cloud-controller-manager/Chart.lock
```diff
@@ -0,0 +1,6 @@
+dependencies:
```
kayrus: could you remove the `*.lock` file and add `*.lock` and `*.zip` into the `.gitignore`?

cwrau: Sure 👍

kayrus:
@cwrau please bump the chart version

cwrau:
> @cwrau please bump the chart version

Done 👌

charts/openstack-cloud-controller-manager/templates/controller.yaml
```diff
@@ -71,10 +99,23 @@ spec:
...
           ports:
           - containerPort: 10258
+            {{- if not .Values.splitMode.enabled }}
             hostPort: 10258
+            {{- end }}
             name: http
             protocol: TCP
           {{- end }}
+          {{- if .Values.splitMode.enabled }}
+          volumeMounts: {{- toYaml .Values.splitMode.extraVolumeMounts | nindent 12 }}
```
234u34k: Could we avoid rendering the empty list as a value here? With split mode enabled and the default `splitMode.extraVolumeMounts: []`, this emits `[]` and then appends the required kubeconfig mount, producing invalid YAML:

```yaml
volumeMounts:
  []
  - mountPath: /etc/kubernetes
```

The `volumes` block at R163 has the same problem with the default `splitMode.extraVolumes: []`.

I reproduced it on the current head with:

```sh
helm dependency build charts/openstack-cloud-controller-manager
helm template review charts/openstack-cloud-controller-manager \
  --set splitMode.enabled=true \
  --set splitMode.kubeconfig.secretName=kubeconfig
```

Helm fails with a YAML parse error in `templates/controller.yaml`.

Could we keep `volumeMounts:` and `volumes:` as block sequences and append the extra entries only when their values are non-empty? A management split-mode render test using the default empty lists would cover both cases.

cwrau: Added a guard around the extraVolumeMounts 👍

cwrau:
The tests seem to fail because of a quota error?

kayrus:

@cwrau no, it seems there are some race conditions.

cwrau:
Heyho, just wanted to check up on this, as CI seems to have worked?

PR #2655, opened 2024-09-17
[cinder-csi]: allow node service to run without openstack client

kayrus:
**What this PR does / why we need it**:

This PR allows to run cinder-csi-plugin node service without requiring the openstack credentials.

**Which issue this PR fixes(if applicable)**:
fixes #2599
supersedes #2640

**Special notes for reviewers**:

This PR will be used to backport to 1.31

**Release note**:

This PR is intended to improve the cinder CSI driver security.

```release-note
[cinder-csi]: allow node service to run without openstack client
```

PR #3067, opened 2026-01-23
feat(cinder-csi-plugin): add splitMode

cwrau:

This allows the user to deploy the controller-manager split between the
management cluster and the workload cluster.

That way one can keep all openstack credentials out of the workload cluster.

**Release note**:

```release-note
[cinder-csi-controller-manager] Allow deploying the controller-manager split between management and workload clusters.
```

### DIFF
--- a/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
+++ b/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
@@ -1,3 +1,4 @@
+{{- if or (not .Values.splitMode.enabled) (eq .Values.splitMode.type "management") }}
 kind: Deployment
 apiVersion: apps/v1
 metadata:
@@ -10,7 +11,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
     {{- end }}
 spec:
-  replicas: {{ .Values.csi.plugin.controllerPlugin.replicas }}
+  replicas: {{ and .Values.splitMode.enabled (eq .Values.splitMode.type "management") | ternary .Values.splitMode.replicas .Values.csi.plugin.controllerPlugin.replicas }}
   strategy:
     type: {{ .Values.csi.plugin.controllerPlugin.strategy.type }}
 {{- if eq .Values.csi.plugin.controllerPlugin.strategy.type "RollingUpdate" }}
@@ -28,7 +29,11 @@ spec:
       annotations:
         {{- include "cinder-csi.controllerplugin.podAnnotations" . | nindent 8 }}
     spec:
-      serviceAccount: csi-cinder-controller-sa
+      {{- if .Values.splitMode.enabled }}
+      automountServiceAccountToken: false
+      {{- else }}
+      serviceAccountName: csi-cinder-controller-sa
+      {{- end }}
       securityContext:
         {{- toYaml .Values.csi.plugin.controllerPlugin.podSecurityContext | nindent 8 }}
       containers:
@@ -43,6 +48,9 @@ spec:
             - "--timeout={{ .Values.timeout }}"
             - "--leader-election=true"
             - "--default-fstype=ext4"
+            {{- if and .Values.splitMode.enabled (eq .Values.splitMode.type "management") }}
+            - "--kubeconfig=/etc/kubernetes/kubeconfig"
+            {{- end }}
             {{- if .Values.csi.attacher.extraArgs }}
             {{- with .Values.csi.attacher.extraArgs }}
             {{- tpl . $ | trim | nindent 12 }}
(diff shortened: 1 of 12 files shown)


cwrau:
Hello, anything I can do to help this getting reviewed?

cwrau:
Hello, is there something I can do to get this merged?

cwrau:
Still waiting for review...

charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
```diff
@@ -219,7 +256,12 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /csi
-          {{- with .Values.csi.plugin.volumeMounts }}
+          {{- if .Values.secret.enabled }}
```
234u34k: Could this preserve the existing `hostMount` path as well? With the chart defaults (`secret.enabled=false`, `secret.hostMount=true`), the rendered pod still declares the `cloud-config` hostPath volume and passes `--cloud-config=/etc/config/cloud.conf`, but this condition means the plugin container no longer mounts that volume.

Minimal reproduction from the repository root at this head:

1. Run `helm template repro ./charts/cinder-csi-plugin --show-only templates/controllerplugin-deployment.yaml`.
2. In the rendered `cinder-csi-plugin` container, observe that `CLOUD_CONFIG` points to `/etc/config/cloud.conf`, while its `volumeMounts` contains no `cloud-config` entry.
3. In the same pod specification, observe that `volumes` still declares `cloud-config` as a hostPath for `/etc/config`.

`helm lint ./charts/cinder-csi-plugin` also succeeds, but the controller cannot open the referenced configuration and returns before starting the CSI service. The node template has the same missing mount and consequently loses its configured BlockStorage settings. An unchanged installation or upgrade therefore regresses even though split mode is disabled.

Suggest preserving the `cloud-config` mount for the `hostMount` case in both templates and adding a default-values render test that asserts every referenced `cloud-config` volume has a matching container mount.

234u34k (changes requested):
Blocking default-values regression; reproduction and requested test are documented inline.

charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
cwrau: Added 👍

234u34k (approved):
Confirmed on `c52e1c93`. The `hostMount` path is restored in both templates and the checks pass. This addresses my requested change, thank you

chenwng:
I think it would be great to describe this feature and its use case in `docs/cinder-csi-plugin/using-cinder-csi-plugin.md`

kayrus:
relates to #2655

charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
mandre: There are still hardcoded `/etc/config` paths that should be changed with `{{ .Values.secret.path }}`.

```
❯ rg "/etc/config/" charts/cinder-csi-plugin
charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
234:              value: /etc/config/{{ .Values.secret.filename }}

charts/cinder-csi-plugin/templates/nodeplugin-daemonset.yaml
113:              value: /etc/config/{{ .Values.secret.filename }}
```

```suggestion
              value: {{ .Values.secret.path }}/{{ .Values.secret.filename }}
```

charts/cinder-csi-plugin/values.yaml
```diff
@@ -236,3 +234,24 @@ imagePullSecrets: []
...
 # add annotations to all pods, deployment, daemonset and podmonitor
 commonAnnotations: {}
+
+# External mode for split-cluster deployments (e.g., Cluster API)
+# When disabled, deploys both controller and node plugin with standard RBAC
+splitMode:
+  enabled: false
+  # type: "management" - Deploy controller with external kubeconfig (management cluster)
+  # type: "workload" - Deploy node plugin with RBAC for external controller (workload cluster)
+  type: ""
```
mandre: I'd suggest either setting a default value that is not empty, or adding a check that fails early if `splitMode.enabled == true` and `type == ''` because otherwise nothing gets deployed and that might be a little confusing for users.

charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
```diff
@@ -239,11 +281,19 @@ spec:
           hostPath:
             path: /etc/config
         {{- end }}
-        {{- with .Values.csi.plugin.volumes }}
+        {{- with .Values.splitMode.enabled | ternary .Values.splitMode.volumes .Values.csi.plugin.volumes }}
```
mandre, line 284: The ternary makes it that splitMode completely ignores `.Values.csi.plugin.volumes` in favor of `.Values.splitMode.volumes`. That might be confusing. Consider merging rather than replacing.

cwrau: That won't work sadly, as one cannot remove an entry from the volumes and with the splitMode you don't want the `/etc/cacert`

charts/cinder-csi-plugin/templates/nodeplugin-daemonset.yaml
mandre: There's one more hardcoded `/etc/config` path that you missed.

PR #3071, opened 2026-02-04
[occm]: support providers that do not require nodeports

oblazek:
Add a new loadBalancer option `ProviderRequiresNodeports` which by default is true to keep the existing behavior but when set to false allows user to not have to allocate nodeports in case there is support by the loadbalancer provider.

This allows usage of services of type LoadBalancer with `allocateLoadBalancerNodePorts=false` in case provider like ours supports that. In our case we send traffic to our k8s nodes (in openstack) using ipip tunneling so there is no need to allocate nodeports.

```release-note
occm: support providers that do not need nodeports
```

### DIFF
--- a/pkg/ingress/config/config.go
+++ b/pkg/ingress/config/config.go
@@ -62,4 +62,9 @@ type octaviaConfig struct {
 	// the load balancer instead of the bulk update API call.
 	// Default is false.
 	ProviderRequiresSerialAPICalls bool `mapstructure:"provider-requires-serial-api-calls"`
+
+	// (Optional) If the provider requires nodeports for accessing the nodes or allows usage
+	// of `allocateLoadBalancerNodePorts=false`.
+	// Default is true.
+	ProviderRequiresNodeports bool `mapstructure:"provider-requires-nodeports"`
 }
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -1033,10 +1033,18 @@ func (lbaas *LbaasV2) buildBatchUpdateMemberOpts(ctx context.Context, port corev
 			memberSubnetID = nil
 		}
 
-		if port.NodePort != 0 { // It's 0 when AllocateLoadBalancerNodePorts=False
+		protoPort := port.NodePort
+		if !lbaas.opts.ProviderRequiresNodeports {
+			// overwrite protoPort as NodePort might be 0
+			protoPort = port.Port
+		}
+
+		// protoPort set when provider doesn't require nodePorts (and e.g. AllocateLoadBalancerNodePorts=false)
+		// or when NodePort is set
+		if protoPort > 0 {
 			member := v2pools.BatchUpdateMemberOpts{
 				Address:      addr,
-				ProtocolPort: int(port.NodePort),
+				ProtocolPort: int(protoPort),
 				Name:         &node.Name,
 				SubnetID:     memberSubnetID,
 			}
--- a/pkg/openstack/loadbalancer_test.go
+++ b/pkg/openstack/loadbalancer_test.go
@@ -3,10 +3,11 @@ package openstack
 import (
 	"context"
 	"fmt"
-	"k8s.io/utils/ptr"
 	"reflect"
 	"sort"
 	"testing"
+
+	"k8s.io/utils/ptr"
 
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/openstack/loadbalancer/v2/listeners"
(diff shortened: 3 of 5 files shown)


oblazek:
added/fixed tests so that the defaults are overwritten when needed

pkg/openstack/loadbalancer.go
nrobert13, line 1020: wouldn't you also need to use the pod addr instead of the node one in case no nodeport is required?

oblazek: Well usually not, as OCCM does not watch endpoints. That's the purpose of CNI (like cilium) imo. In our case we do ipip tunneling from our externalLB (the one that receives configuration from OCCM in the end) to k8s nodes where cilium listens.

oblazek: so in our case you have: `extLB -> node1IP: clientIP -> svcIP` (outer IP header: inner IP header)

oblazek: so that's why there is no need for nodeport

nrobert13: I see, we have a slightly different use-case. we use native routing with cilium, so the pod network is directly reachable from external LB, therefore no additional ( nodePort ) hop would be required, so the LB member address could be the pod IP.

oblazek: Interesting, that's a lot of updates your externalLB needs to process. Anyway regarding the pod CIDR, how would you pass that to the OCCM? AFAICS it does not watch pods/endpoints.

nrobert13: not necessarily a lot more updates if you use `externalTrafficPolicy: local`, which means the traffic is only sent to the nodes where the upstream pods are running. so the amount of updates would be higher only if the new pods are being rescheduled on the same nodes ( so no update of the LB required with `nodePort`) , otherwise it would be the same amount....
I'm not sure about the best way to have the podIPs in the OCCM, would it feasible to watch the Endpoints/EndpointSlices?

oblazek: Yeah that totally would, we have a similar controller that does that (but that's meant for other LB - not the external one). But I suppose OCCM maintainers don't want to do that.

pkg/openstack/loadbalancer.go
```diff
@@ -1033,10 +1033,21 @@ func (lbaas *LbaasV2) buildBatchUpdateMemberOpts(ctx context.Context, port corev
...
 		}
 
-		if port.NodePort != 0 { // It's 0 when AllocateLoadBalancerNodePorts=False
+		var protoPort int32
+		if !lbaas.opts.ProviderRequiresNodeports {
+			protoPort = port.Port
+		} else {
+			if port.NodePort != 0 { // It's 0 when AllocateLoadBalancerNodePorts=False
+				protoPort = port.NodePort
+			}
```
stephenfin: Given the new `> 0` check below, could we always assign this?

```suggestion
			protoPort = port.NodePort
```

pkg/openstack/loadbalancer_test.go
```diff
@@ -1925,6 +1932,21 @@ func TestBuildBatchUpdateMemberOpts(t *testing.T) {
 			expectedLen:             0,
 			expectedNewMembersCount: 0,
 		},
+		{
+			name:  "NodesPortequalszero, providerRequiresNodeports=false",
```
stephenfin: femtonit:

```suggestion
			name:  "NodesPort equals zero, providerRequiresNodeports=false",
```

pkg/ingress/config/config.go
```diff
@@ -62,4 +62,9 @@ type octaviaConfig struct {
 	// the load balancer instead of the bulk update API call.
 	// Default is false.
 	ProviderRequiresSerialAPICalls bool `mapstructure:"provider-requires-serial-api-calls"`
+
+	// (Optional) If the provider requires nodeports for accessing the nodes or allows usage
+	// of `allocateLoadBalancerNodePorts=false`
```
stephenfin: nit:

```suggestion
	// of `allocateLoadBalancerNodePorts=false`.
```

Also, would it be possible to (briefly) capture the reasons you'd want to set this to `false`, given you've already touched on the use case in [the discussion below](https://github.com/kubernetes/cloud-provider-openstack/pull/3071/changes#r3110366758).

stephenfin (changes requested):
Change looks good. I would just like to see a little more context on the "why" in the docs, since this requires specific configuration to use correctly.

I also have some small nits that would be good to address before merge.

nrobert13:
@stephenfin would you be able to check our slightly different use-case described [here](https://github.com/kubernetes/cloud-provider-openstack/issues/3074)?

pkg/openstack/loadbalancer.go
oblazek: yeah totally!

pkg/ingress/config/config.go
oblazek: done

pkg/openstack/loadbalancer_test.go
oblazek: thanks, fixed

PR #3075, opened 2026-02-26
core: refactor lbaas util functions

kayrus:

This is an experimental code refactoring using Claude AI as a development assistant. Each change was discussed and reviewed by me during the development process, it took about 6h in total. The refactoring reduces code duplication, fixes several bugs (missing metrics, potential panics, double observations), and optimizes API call patterns while preserving all existing functionality. All tests pass without modification. @zetaab @dulek @stephenfin I'd like you to review this as well. What do you think about such AI-assisted PRs?

**Below is AI summary**:

## Key improvements

1. **Reduced code duplication (-118 lines, ~15% reduction)** - Extracted common patterns into reusable generic helpers:
   - `executeAndWaitActive` / `executeExtractAndWaitActive` - Centralizes metrics tracking, error handling, and LB wait logic
   - `list` / `listWithUniqueResult` - Standardizes pagination with early-exit optimization
   - `getSingleResource` - Consistent single-resource retrieval with metrics

2. **Fixed bugs and inconsistencies:**
   - Fixed missing metrics observation in 5 functions (`GetPools`, `GetL7policies`, `GetL7Rules`, `GetListenersByLoadBalancerID`, `getOctaviaVersion` extraction error)
   - Fixed double metrics observation bug in `DeleteListener`
   - Eliminated redundant check in `GetPoolByName`

3. **Performance optimization:**
   - `GetPoolByListener` now uses direct listener Get instead of listing all pools (major API call reduction)
   - `listWithUniqueResult` implements early-exit when multiple results detected on first page

4. **Improved consistency:**
   - Centralized NotFound error handling for all delete operations
   - Standardized error message format across all operations
   - Unified function documentation style (following AWS SDK Go conventions)
   - Consistent metrics tracking across all operations

5. **Better organization:**
   - Moved `UpdatePool` and `DeletePool` to correct section (Pool Operations, not Pool Member Operations)
   - Added clear section separators
   - Fixed function naming (`GetMembersbyPool` → `GetPoolMembers`)

## Testing

- All existing unit tests pass without modification
- No functional changes - pure refactoring
- Verified with `go test ./pkg/util/openstack/...`

## Files changed

- `pkg/util/openstack/loadbalancer.go` - Main refactoring

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/util/openstack/loadbalancer.go
+++ b/pkg/util/openstack/loadbalancer.go
@@ -272,7 +272,7 @@ func WaitActiveAndGetLoadBalancer(ctx context.Context, client *gophercloud.Servi
 	}
 
 	var loadbalancer *loadbalancers.LoadBalancer
-	err := wait.ExponentialBackoff(backoff, func() (bool, error) {
+	err := wait.ExponentialBackoffWithContext(ctx, backoff, func(ctx context.Context) (bool, error) {
 		mc := metrics.NewMetricContext("loadbalancer", "get")
 		var err error
 		loadbalancer, err = loadbalancers.Get(ctx, client, loadbalancerID).Extract()
@@ -340,7 +340,7 @@ func waitLoadBalancerDeleted(ctx context.Context, client *gophercloud.ServiceCli
 		Factor:   waitLoadbalancerFactor,
 		Steps:    waitLoadbalancerDeleteSteps,
 	}
-	err := wait.ExponentialBackoff(backoff, func() (bool, error) {
+	err := wait.ExponentialBackoffWithContext(ctx, backoff, func(ctx context.Context) (bool, error) {
 		mc := metrics.NewMetricContext("loadbalancer", "get")
 		_, err := loadbalancers.Get(ctx, client, loadbalancerID).Extract()
 		if mc.ObserveRequest(err) != nil && cpoerrors.IsNotFound(err) {
--- a/pkg/util/openstack/loadbalancer.go
+++ b/pkg/util/openstack/loadbalancer.go
@@ -163,7 +163,7 @@ func executeAndWaitActive(ctx context.Context, client *gophercloud.ServiceClient
 	lbID, resourceType, operation string, fn func() error) error {
 
 	_, err := executeExtractAndWaitActive(ctx, client, lbID, resourceType, operation,
-		func() (any, error) {
+		func() (*any, error) {
 			return nil, fn()
 		})
 	return err
@@ -174,32 +174,23 @@ func executeAndWaitActive(ctx context.Context, client *gophercloud.ServiceClient
 // Used for create operations that return the created resource.
 // For delete operations, NotFound errors are logged but not returned as errors.
 func executeExtractAndWaitActive[T any](ctx context.Context, client *gophercloud.ServiceClient,
-	lbID, resourceType, operation string, fn func() (T, error)) (T, error) {
+	lbID, resourceType, operation string, fn func() (*T, error)) (*T, error) {
 
 	mc := metrics.NewMetricContext(resourceType, operation)
 	result, err := fn()
-
-	// For delete operations, treat NotFound as success (already deleted)
-	if operation == "delete" && err != nil && cpoerrors.IsNotFound(err) {
-		klog.V(2).Infof("%s was already deleted", resourceType)
-		_ = mc.ObserveRequest(nil)
-		// Still wait for load balancer to be active for consistency
-		if _, waitErr := WaitActiveAndGetLoadBalancer(ctx, client, lbID); waitErr != nil {
-			return result, fmt.Errorf("failed to wait for load balancer %s ACTIVE after %s %s: %v",
-				lbID, operation, resourceType, waitErr)
-		}
-		return result, nil
-	}
-
-	// Observe the request result (error or success)
 	if mc.ObserveRequest(err) != nil {
-		return result, fmt.Errorf("failed to %s %s on load balancer %s: %v", operation, resourceType, lbID, err)
+		// For delete operations, treat NotFound as success (already deleted)
+		if operation == "delete" && cpoerrors.IsNotFound(err) {
+			klog.V(2).Infof("%s was already deleted", resourceType)
+		} else {
+			return result, fmt.Errorf("failed to %s %s on load balancer %s: %v", operation, resourceType, lbID, err)
+		}
 	}
-
 	if _, err := WaitActiveAndGetLoadBalancer(ctx, client, lbID); err != nil {
 		return result, fmt.Errorf("failed to wait for load balancer %s ACTIVE after %s %s: %v",
 			lbID, operation, resourceType, err)
 	}
+
 	return result, nil
 }
 
--- a/pkg/util/openstack/loadbalancer.go
+++ b/pkg/util/openstack/loadbalancer.go
@@ -255,8 +255,7 @@ func getSingleResource[T any](ctx context.Context, resourceType, operation strin
 	mc := metrics.NewMetricContext(resourceType, operation)
 	result, err := fn()
 	if mc.ObserveRequest(err) != nil {
-		var zero T
-		return zero, fmt.Errorf("failed to %s %s: %v", operation, resourceType, err)
+		return result, fmt.Errorf("failed to %s %s: %v", operation, resourceType, err)
 	}
 	return result, nil
 }
--- a/pkg/util/openstack/loadbalancer.go
+++ b/pkg/util/openstack/loadbalancer.go
@@ -152,6 +152,127 @@ func IsOctaviaFeatureSupported(ctx context.Context, client *gophercloud.ServiceC
 	return false
 }
 
+// ==============================================================================
+// Internal Helper Functions
+// ==============================================================================
+
+// executeAndWaitActive executes an operation with metrics tracking and waits for
+// the load balancer to return to ACTIVE state. Used for operations that don't
+// return a result (delete, update without returning updated object).
+func executeAndWaitActive(ctx context.Context, client *gophercloud.ServiceClient,
+	lbID, resourceType, operation string, fn func() error) error {
+
+	_, err := executeExtractAndWaitActive(ctx, client, lbID, resourceType, operation,
+		func() (*struct{}, error) {
+			if err := fn(); err != nil {
+				return nil, err
+			}
+			return &struct{}{}, nil
+		})
+	return err
+}
+
+// executeExtractAndWaitActive executes an operation with metrics tracking, extracts
+// the result, and waits for the load balancer to return to ACTIVE state.
+// Used for create operations that return the created resource.
+// For delete operations, NotFound errors are logged but not returned as errors.
+func executeExtractAndWaitActive[T any](ctx context.Context, client *gophercloud.ServiceClient,
+	lbID, resourceType, operation string, fn func() (T, error)) (T, error) {
+
+	var zero T
+	mc := metrics.NewMetricContext(resourceType, operation)
+	result, err := fn()
+
+	// For delete operations, treat NotFound as success (already deleted)
+	if err != nil && operation == "delete" && cpoerrors.IsNotFound(err) {
+		klog.V(2).Infof("%s was already deleted", resourceType)
+		_ = mc.ObserveRequest(nil)
+		// Still wait for load balancer to be active for consistency
+		if _, waitErr := WaitActiveAndGetLoadBalancer(ctx, client, lbID); waitErr != nil {
+			return zero, fmt.Errorf("failed to wait for load balancer %s ACTIVE after %s %s: %v",
+				lbID, operation, resourceType, waitErr)
+		}
+		return zero, nil
+	}
+
+	// Observe the request result (error or success)
+	if mc.ObserveRequest(err) != nil {
+		return zero, fmt.Errorf("failed to %s %s: %v", operation, resourceType, err)
+	}
+
+	if _, waitErr := WaitActiveAndGetLoadBalancer(ctx, client, lbID); waitErr != nil {
+		return zero, fmt.Errorf("failed to wait for load balancer %s ACTIVE after %s %s: %v",
+			lbID, operation, resourceType, waitErr)
+	}
+	return result, nil
+}
+
+// list performs pagination and returns all results with metrics tracking.
+func list[T any](ctx context.Context, resourceType, operation string,
+	pager pagination.Pager, extractFn func(pagination.Page) ([]T, error)) ([]T, error) {
+
+	mc := metrics.NewMetricContext(resourceType, operation)
+	allPages, err := pager.AllPages(ctx)
+	if mc.ObserveRequest(err) != nil {
+		return nil, err
+	}
+
+	results, err := extractFn(allPages)
+	if err != nil {
+		return nil, err
+	}
+
+	return results, nil
+}
+
+// listWithUniqueResult performs pagination and expects exactly one result.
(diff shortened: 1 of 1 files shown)


gouthampacha:
Hi @kayrus: 

I don't have +2 powers here, so I'm just exercising my review judgement as a long time OpenStack maintainer and sharing my 0.02 cents of opinion. 

I think the intent to address bugs is good, but, i'd like to suggest a different approach. You should break this down into multiple commits - get Claude to do it. I know we don't apply the same commit standards here as we do in OpenStack or core Kubernetes projects; but, for the sake of maintainability, seeing single-intent commits is preferred.

stephenfin:
> Hi @kayrus:
> 
> I don't have +2 powers here, so I'm just exercising my review judgement as a long time OpenStack maintainer and sharing my 0.02 cents of opinion.
> 
> I think the intent to address bugs is good, but, i'd like to suggest a different approach. You should break this down into multiple commits - get Claude to do it. I know we don't apply the same commit standards here as we do in OpenStack or core Kubernetes projects; but, for the sake of maintainability, seeing single-intent commits is preferred.

I [asked for the same thing](https://kubernetes.slack.com/archives/C0LSA3T7C/p1772116007348569?thread_ts=1772115647.466899&cid=C0LSA3T7C) :sweat_smile: Must be an OpenStack thing...

valeryia-hurynovich:
I see that the bug fixes and refactoring are deeply intertwined. The bugs (missing metrics observations) are fixed implicitly via the new helpers, so there's no standalone commit of fixed metrics possible without either the helpers or inline fixes. However it is possible to split this logically via commits if it's important to split this anyway. I can propose to split these fixes into 5 or 6 logical commits:
1. Bug fixes (standalone)
2. Renaming of functions
3. Extract generic helpers
4. Refactor existing functions to use the helpers
5. Performance
6. Code organization

valeryia-hurynovich:
@kayrus I was thinking about that this PR has some general independent parts:

1. renaming of `GetMembersbyPool` to `GetPoolMembers`
2. The `GetPoolByListener` performance optimization
3. Code organization as it written down in explanation of this PR

Splitting those out while keeping the bug-fix/helper-extraction/refactor as one atomic commit would reduce the noise in review without fragmenting what can't actually be separated, so I think that split this into some general commits will be feasible and it can be not a broken glass if just to separate some things, it's not a big deal here
I'd propose to split it like:

1. Rename `GetMembersbyPool` to `GetPoolMembers` (affects 4 files)
2. Performance `GetPoolByListener`
3. Code organization: section separators, doc comments, constant annotations
4. Core: bug fixes + helper extraction + refactor together (the atomic part that can't be split)

@gouthampacha would like to ask just how much reasonable to split it into clear commits?
Otherwise, I consider it's not like a big deal to split this if it's worth to be

kayrus:
I left only the code improvement with new helpers. @zetaab @gouthampacha @stephenfin please give this PR a try one more time.

pkg/util/openstack/loadbalancer.go
jihyun-huh, line 1: Hi @kayrus, sorry to interrupt
I have been following this PR for a long time since it was opened, and I read the latest version again after today's update.

The current scope feels much clearer than the previous version, especially now it is focused on the helper extraction. I also tested on local and it worked fine. 

I especially appreciate the `GetPoolByListener` function change. now it is looking up a listener first then using default pool ID without listing every pool, which seems like a nice simplification.

and also while reading it, I had one small question about it.

pkg/util/openstack/loadbalancer.go
```diff
@@ -152,6 +152,127 @@ func IsOctaviaFeatureSupported(ctx context.Context, client *gophercloud.ServiceC
...
+// ==============================================================================
+
+// executeAndWaitActive executes an operation with metrics tracking and waits for
+// the load balancer to return to ACTIVE state. Used for operations that don't
+// return a result (delete, update without returning updated object).
+func executeAndWaitActive(ctx context.Context, client *gophercloud.ServiceClient,
+	lbID, resourceType, operation string, fn func() error) error {
+
+	_, err := executeExtractAndWaitActive(ctx, client, lbID, resourceType, operation,
+		func() (*struct{}, error) {
```
jihyun-huh: just curious.. not blocking
the function `executeExtractAndWaitActive`  using `*struct{}`, and i understand this keeps the metrics or error handling and ACTIVE wait logic in one place.

was your motivation is avoiding duplication here, or did you also consider keeping the returning and empty returning helper paths separate for readability?

pkg/util/openstack/loadbalancer.go
```diff
@@ -320,421 +403,233 @@ func DeleteLoadbalancer(ctx context.Context, client *gophercloud.ServiceClient,
...
-	return policy, nil
+	return executeExtractAndWaitActive(ctx, client, lbID, "loadbalancer_l7policy", "create",
+		func() (*l7policies.L7Policy, error) {
+			return l7policies.Create(ctx, client, opts).Extract()
+		})
 }
 
-// DeleteL7policy deletes a l7 policy.
+// DeleteL7policy deletes an L7 policy and waits for the load balancer to become active.
 func DeleteL7policy(ctx context.Context, client *gophercloud.ServiceClient, policyID string, lbID string) error {
```
yahor-kurachkin, line 517: I noticed that DeleteL7policy behavior changed slightly after the refactoring. The old implementation returned an error on 404, while the new one uses executeExtractAndWaitActive which treats 404 as success. Was the old behavior intentional, or just an inconsistency that got fixed as a side effect of the refactor?

pkg/util/openstack/loadbalancer.go
```diff
@@ -152,6 +152,127 @@ func IsOctaviaFeatureSupported(ctx context.Context, client *gophercloud.ServiceC
...
+		}
+		return zero, nil
+	}
+
+	// Observe the request result (error or success)
+	if mc.ObserveRequest(err) != nil {
+		return zero, fmt.Errorf("failed to %s %s: %v", operation, resourceType, err)
+	}
+
+	if _, waitErr := WaitActiveAndGetLoadBalancer(ctx, client, lbID); waitErr != nil {
```
gouthampacha: We should capture and return this result, i.e., the updated load balancer object for correctness

pkg/util/openstack/loadbalancer.go
```diff
@@ -152,6 +152,127 @@ func IsOctaviaFeatureSupported(ctx context.Context, client *gophercloud.ServiceC
...
+		if _, waitErr := WaitActiveAndGetLoadBalancer(ctx, client, lbID); waitErr != nil {
+			return zero, fmt.Errorf("failed to wait for load balancer %s ACTIVE after %s %s: %v",
+				lbID, operation, resourceType, waitErr)
+		}
+		return zero, nil
+	}
+
+	// Observe the request result (error or success)
+	if mc.ObserveRequest(err) != nil {
+		return zero, fmt.Errorf("failed to %s %s: %v", operation, resourceType, err)
```
gouthampacha: maybe logging the `lbID` would be useful

gouthampacha (approved):
/lgtm 

This works for me, but, a couple of comments inline that could be addressed if you make another patch

pkg/util/openstack/loadbalancer.go
kayrus: The `executeAndWaitActive` func is a wrapper around `executeExtractAndWaitActive` that returns only errors. This is usable for `Delete` and some `Update` and `Create` actions, where the response it not used. For sure I can use `executeExtractAndWaitActive` for each call with another custom `fn()` signature, but adding an additional `executeAndWaitActive` wrapper clearly indicates that this call doesn't return a response.

P.S. I updated the func signature content to be much simpler.

pkg/util/openstack/loadbalancer.go
kayrus: done

pkg/util/openstack/loadbalancer.go
kayrus: fair point, done

pkg/util/openstack/loadbalancer.go
kayrus: > The old implementation returned an error on 404

this was a hidden bug, now fixed :)

yahor-kurachkin (approved):
Nice work! LGTM

pkg/util/openstack/loadbalancer.go
```diff
@@ -152,6 +152,118 @@ func IsOctaviaFeatureSupported(ctx context.Context, client *gophercloud.ServiceC
...
+// Used for create operations that return the created resource.
+// For delete operations, NotFound errors are logged but not returned as errors.
+func executeExtractAndWaitActive[T any](ctx context.Context, client *gophercloud.ServiceClient,
+	lbID, resourceType, operation string, fn func() (T, error)) (T, error) {
+
+	mc := metrics.NewMetricContext(resourceType, operation)
+	result, err := fn()
+
+	// For delete operations, treat NotFound as success (already deleted)
+	if operation == "delete" && err != nil && cpoerrors.IsNotFound(err) {
```
valeryia-hurynovich: I'm wondering that if it will be already fully cascade-deleted this call will fail and surface as a spurious wait error

valeryia-hurynovich: this comment is no longer valid

valeryia-hurynovich (approved):
LGTM

kayrus:
@jihyun-huh @gouthampacha @stephenfin @zetaab a few improvements were made (by me, not AI :) ). please review one more time.

pkg/util/openstack/loadbalancer.go
```diff
@@ -152,6 +152,109 @@ func IsOctaviaFeatureSupported(ctx context.Context, client *gophercloud.ServiceC
...
+		result = &items[0]
+
+		return false, nil
+	})
+
+	if mc.ObserveRequest(err) != nil && cpoerrors.IsNotFound(err) {
+		return nil, cpoerrors.ErrNotFound
+	}
+
+	return result, nil
```
kayrus: I thinking to replace this block with one line

```suggestion
	return result, mc.ObserveRequest(err)
```

jihyun-huh: Yes that looks much simpler, thanks for your comment! 
and it seems also preserves `non-NotFound` errors instead of returning `nil`,

so.. one thing I noticed is while checking `ObserveRequest: Observe` currently returns `nil` when the metrics object is not configured.

https://github.com/kayrus/cloud-provider-openstack/blob/c453fb7c9bbb2091aee763d9d2809cc81fdf31c6/pkg/metrics/metrics.go#L48-L50
```go
func (mc *MetricContext) Observe(om *OpenstackMetrics, err error) error {
	if om == nil {
		// mc.RequestMetrics not set, ignore this request
		return nil
	}
```

im wondering, would it be safer for this branch to return the original err instead? 
or otherwise the one-line return could hide the operation error when metrics are disabled.

this is separated with my earlier question above about keeping the non result wrapper, I agree with that and make sense!

kayrus: `APIRequestMetrics` is defined in https://github.com/kubernetes/cloud-provider-openstack/blob/c453fb7c9bbb2091aee763d9d2809cc81fdf31c6/pkg/metrics/metrics_api.go#L27, thus it is less probable that it can be nil. I'd fix this in a separate PR.

pkg/util/openstack/loadbalancer.go
gouthampacha: not done?

gouthampacha:
/lgtm

minor issue with returning a stale object in `executeExtractAndWaitActive` remains

pkg/util/openstack/loadbalancer.go
kayrus: why not? see line 190.

kayrus:
> minor issue with returning a stale object in `executeExtractAndWaitActive` remains

it is fixed, see line 190

anguslees (approved):
/lgtm

Next time, however - a PR like this is very hard to review.  You have mixed unrelated no-op code cleanups across multiple areas with real bug fixes.  Please don't do this again, since if we have to roll back an unexpected behaviour change, then we're also rolling back all these other improvements.

Using AI for this type of cleanup is a great idea.  Easier code gen means even more reason to optimise for review however, not the reverse.

kayrus:
@anguslees I understand the concern about mixing changes. The refactoring was necessary because the new helper function needed to be integrated across multiple functions, I could have introduced the helper in one PR, then applied it to each function in dozens of follow-up PRs. However, that approach wouldn't scale and would look bizarre.

pkg/util/openstack/loadbalancer.go
```diff
@@ -152,6 +152,105 @@ func IsOctaviaFeatureSupported(ctx context.Context, client *gophercloud.ServiceC
...
+
+// getSingleResource executes a single resource retrieval operation with metrics tracking.
+// Used for Get operations that retrieve a resource by ID.
+func getSingleResource[T any](ctx context.Context, resourceType, operation string,
+	fn func() (T, error)) (T, error) {
+
+	mc := metrics.NewMetricContext(resourceType, operation)
+	result, err := fn()
+	if mc.ObserveRequest(err) != nil {
+		return result, fmt.Errorf("failed to %s %s: %v", operation, resourceType, err)
```
zetaab, line 249: instead of %v, we should use %w

pkg/util/openstack/loadbalancer.go
```diff
@@ -152,6 +152,105 @@ func IsOctaviaFeatureSupported(ctx context.Context, client *gophercloud.ServiceC
...
+	lbID, resourceType, operation string, fn func() (*T, error)) (*T, error) {
+
+	mc := metrics.NewMetricContext(resourceType, operation)
+	result, err := fn()
+	if mc.ObserveRequest(err) != nil {
+		// For delete operations, treat NotFound as success (already deleted)
+		if operation == "delete" && cpoerrors.IsNotFound(err) {
+			klog.V(2).Infof("%s was already deleted", resourceType)
+		} else {
+			return result, fmt.Errorf("failed to %s %s on load balancer %s: %v", operation, resourceType, lbID, err)
```
zetaab, line 186: instead of %v, we should use %w

pkg/util/openstack/loadbalancer.go
```diff
@@ -152,6 +152,105 @@ func IsOctaviaFeatureSupported(ctx context.Context, client *gophercloud.ServiceC
...
+		}
+	}
+	if _, err := WaitActiveAndGetLoadBalancer(ctx, client, lbID); err != nil {
+		return result, fmt.Errorf("failed to wait for load balancer %s ACTIVE after %s %s: %v",
```
zetaab, line 190: instead of %v, we should use %w

pkg/util/openstack/loadbalancer.go
```diff
@@ -305,436 +360,204 @@ func waitLoadbalancerDeleted(ctx context.Context, client *gophercloud.ServiceCli
...
-		if len(listenerPools) > 1 {
-			return false, cpoerrors.ErrMultipleResults
-		}
-		return true, nil
-	})
-	if mc.ObserveRequest(err) != nil {
-		if cpoerrors.IsNotFound(err) {
-			return nil, cpoerrors.ErrNotFound
-		}
+	listener, err := listWithUniqueResult(ctx, "loadbalancer_listener", "list", listeners.List(client, listOpts), listeners.ExtractListeners)
```
zetaab, line 441: this is actually now hardfail if not found, we are not checking ErrNotFound error at all

zetaab (changes requested):

Issue #2767, opened 2025-02-10
failed to  list *v1.VolumeSnapshotContent: the server could not find the requested resource (get volumesnapshotcontents.snapshot.storage.k8s.io)


rptaylor:
A default deployment of cinder CSI results in constant errors in logs:

```
csi-snapshotter W0210 21:36:23.865665       1 reflector.go:424] github.com/kubernetes-csi/external-snapshotter/client/v6/informers/externalversions/factory.go:117: failed to
 list *v1.VolumeSnapshotContent: the server could not find the requested resource (get volumesnapshotcontents.snapshot.storage.k8s.io)                                       
csi-snapshotter E0210 21:36:23.865719       1 reflector.go:140] github.com/kubernetes-csi/external-snapshotter/client/v6/informers/externalversions/factory.go:117: Failed to
 watch *v1.VolumeSnapshotContent: failed to list *v1.VolumeSnapshotContent: the server could not find the requested resource (get volumesnapshotcontents.snapshot.storage.k8s
.io)         
```

IIUC, some other component (CRDs and/or https://github.com/kubernetes-csi/external-snapshotter ?) need to be installed for this to work.
If the cinder-CSI chart does not install them by default, shouldn't the csi-snapshotter container be disabled by default (via helm value) to avoid wasting resources and producing spurious errors?

JihadMotii-REISys:
were you able to resolve this issue @rptaylor ? Thanks in advance!

rptaylor:
> were you able to resolve this issue [@rptaylor](https://github.com/rptaylor) ? Thanks in advance!

No I think some changes in the helm chart would be needed.

rptaylor:
/remove-lifecycle stale

If anyone can confirm what is supposed to happen maybe we could fix it.
Why does the Helm chart try to use CRDs that it doesn't install? How is that expected to work?

Dentling:
any updates on this?

I'm seeing the same errors.

petterroea:
Also seeing same errors, but using 2.5 year old release. Looking into whether or not errors are caused by PEBCAK on my side, or fixable by upgrading

rptaylor:
/remove-lifecycle stale  
Still relevant AFAIK

rptaylor:
It looks like https://github.com/kubernetes/cloud-provider-openstack/pull/3078  could solve this, but no maintainers or reviewers have reviewed it or approved tests to run on it. 

/remove-lifecycle stale

PR #3078, opened 2026-03-01
[cinder-csi-plugin] Make csi-snapshotter sidecar optional

234u34k:
The csi-snapshotter sidecar is always deployed regardless of whether VolumeSnapshot CRDs are installed. On clusters without external-snapshotter, this produces continuous error logs.

Add `csi.snapshotter.enabled` (default `true` for backward compatibility) to gate the snapshotter container and its RBAC resources. Users without CRDs can set it to `false`.

Ref #2767

```release-note
cinder-csi: Add csi.snapshotter.enabled toggle to allow disabling the csi-snapshotter sidecar when VolumeSnapshot CRDs are not installed.
```

### DIFF
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v1.36.0
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 2.36.1
+version: 2.36.2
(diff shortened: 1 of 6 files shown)


234u34k:
/remove-lifecycle rotten

@kayrus @zetaab, this is rebased onto current master and remains relevant to #2767. The chart now preserves the existing snapshotter behavior for upgrades using `--reuse-values`, and chart lint renders both enabled and disabled configurations. Could one of you run `/ok-to-test` and review it when you have time?

Issue #2682, opened 2024-10-01
[occm] Shared loadbalancer should not be renamed.


kratisto:
Hello, 

Is this a BUG REPORT or FEATURE REQUEST?:
/kind bug

While testing shared loadbalancers, i think i discovered a regression since this [PR](https://github.com/kubernetes/cloud-provider-openstack/pull/2552). 

A shared loadbalancer is renamed due to [this if](https://github.com/kubernetes/cloud-provider-openstack/blob/d069b1ef5d34e1400c798424f5d56b11098fb9a7/pkg/openstack/loadbalancer.go#L1694). What happens here is : 

We enter the `ensureOctaviaLoadBalancer` function as `clusterb` with a shared loadbalancer named : `kube_service_clustera_kns_fakeservice`

At the end we can see the cluster has been renamed : `kube_service_clusterb_kns_fakeservice`.

And we have the event : 
`Warning LoadBalancerRename Loadbalancer fakeloadbalancerid has a name of kube_service_clustera_kns_fakeservice with incorrect cluster-name component. Renaming it to kube_service_clusterb_kns_fakeservice.`

The tags has been changed too : 
`["kube_service_clustera_kns_fakeservice","clustera","kube_service_clusterb_kns_fakeservice"]`
become `["kube_service_clusterb_kns_fakeservice","clustera","kube_service_clusterb_kns_fakeservice"]`

I created a test that reproduce the [regression](https://gist.github.com/kratisto/b9918274314bf45582a9454dbf9d24ed). 

Didn't test if it's the case , but i suspect that if we launch a third time as `clusterc` , the shared loadbalancer will be renamed again. 
So the ownership will be for the last runner of `ensureOctaviaLoadBalancer`

mxm-tr:
Yep, this prevents LBs from being shared between multiple clusters

PR #3079, opened 2026-03-01
[occm] Skip LB rename for shared load balancers from other clusters

234u34k:
The cluster-name rename logic from #2552 does not account for shared load balancers. When cluster B references a shared LB created by cluster A, the OCCM sees a mismatched cluster name and renames the LB, breaking cluster A.

Before renaming, verify that the LB's namespace and service name components match the current Service. This distinguishes a legitimate cluster-name change from a cross-cluster shared LB access.

Ref #2682

```release-note
occm: Fixed shared load balancers being renamed when referenced by a Service from a different cluster. A new `enable-lb-rename` option (default `true`) disables renaming entirely, required when sharing a load balancer across clusters between Services with the same namespace and name.
```

### DIFF
--- a/pkg/openstack/openstack.go
+++ b/pkg/openstack/openstack.go
@@ -118,7 +118,7 @@ type LoadBalancerOpts struct {
 	EnableIngressHostname          bool                `gcfg:"enable-ingress-hostname"`            // Used with proxy protocol by adding a dns suffix to the load balancer IP address. Default false.
 	IngressHostnameSuffix          string              `gcfg:"ingress-hostname-suffix"`            // Used with proxy protocol by adding a dns suffix to the load balancer IP address. Default nip.io.
 	MaxSharedLB                    int                 `gcfg:"max-shared-lb"`                      //  Number of Services in maximum can share a single load balancer. Default 2
-	EnableLBRename                 bool                `gcfg:"enable-lb-rename"`                   // if false, load balancers are never renamed after a cluster-name change. Default true
+	EnableLBRename                 bool                `gcfg:"enable-lb-rename"`                   // If false, load balancers are never renamed after a cluster-name change. Default true
 	ContainerStore                 string              `gcfg:"container-store"`                    // Used to specify the store of the tls-container-ref
 	ProviderRequiresSerialAPICalls bool                `gcfg:"provider-requires-serial-api-calls"` // default false, the provider supports the "bulk update" API call
 	// revive:disable:var-naming
--- a/pkg/openstack/openstack.go
+++ b/pkg/openstack/openstack.go
@@ -118,6 +118,7 @@ type LoadBalancerOpts struct {
 	EnableIngressHostname          bool                `gcfg:"enable-ingress-hostname"`            // Used with proxy protocol by adding a dns suffix to the load balancer IP address. Default false.
 	IngressHostnameSuffix          string              `gcfg:"ingress-hostname-suffix"`            // Used with proxy protocol by adding a dns suffix to the load balancer IP address. Default nip.io.
 	MaxSharedLB                    int                 `gcfg:"max-shared-lb"`                      //  Number of Services in maximum can share a single load balancer. Default 2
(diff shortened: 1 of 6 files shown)


jihyun-huh:
Hi @234u34k, thanks for this PR
I was looking at this with the issue #2682, since the issue seems to have gone rotten without much follow-up.

even if i have no permission to review this, the direction makes sense to me.

one thing i want to double-check, the original #2682 example seems to use same namespace and service name across clusters. for example, cluster-B reconciling a shared LB named `kube_service_clustera_kns_fakeservice` for `kns/fakeservice`, in this case, would the new `expectedOldName` still match and allow the rename?

If so,, should the rename path also takes the explicit shared LB into account. for example when `loadbalancer.openstack.org/load-balancer-id` is set and the existing LB name does not match the current Service’s expected LB name?

may be im missing part of the ownership logic here, but think a regression test can cover the #2682 examples.

cc. @kayrus @stephenfin @gouthampacha

234u34k:
Confirmed. `oldClusterName` is derived from the LB name itself, so when both
clusters run a Service with the same namespace and name (the #2682 example),
`expectedOldName` reconstructs the existing name and the rename still fires.
My guard only covered the case where the tuples differ.

Gating on the `load-balancer-id` annotation doesn't work either: OCCM persists
that annotation on every Service it manages, and after a cluster-name change
the stale LB is only ever found through it, so the legitimate #2552 rename
flows through the same branch. And for the identical namespace/name case there
is nothing in Octavia that distinguishes the two. The LB name, its tags and its
children all look exactly like a legitimately stale LB of this cluster.

Updated the PR accordingly:

1. extracted the ownership check into `lbNameMatchesService` and added
   regression tests covering the #2682 examples (same and different
   namespace/name across clusters, the legitimate #2552 rename, truncated
   names)
2. added an `enable-lb-rename` option (default true, #2552 behavior unchanged)
   so deployments sharing load balancers across clusters with identical
   Service tuples can disable renaming, since that case is not decidable from
   the OpenStack side

Also added a V(4) log when a rename is skipped and a docs entry for the option.

@kayrus @stephenfin could one of you take a look and issue /ok-to-test? CI has
not run on this PR yet.

pkg/openstack/openstack.go
```diff
@@ -118,6 +118,7 @@ type LoadBalancerOpts struct {
 	EnableIngressHostname          bool                `gcfg:"enable-ingress-hostname"`            // Used with proxy protocol by adding a dns suffix to the load balancer IP address. Default false.
 	IngressHostnameSuffix          string              `gcfg:"ingress-hostname-suffix"`            // Used with proxy protocol by adding a dns suffix to the load balancer IP address. Default nip.io.
 	MaxSharedLB                    int                 `gcfg:"max-shared-lb"`                      //  Number of Services in maximum can share a single load balancer. Default 2
+	EnableLBRename                 bool                `gcfg:"enable-lb-rename"`                   // if false, load balancers are never renamed after a cluster-name change. Default true
```
jihyun-huh: very small suggestion..

```suggestion
	EnableLBRename                 bool                `gcfg:"enable-lb-rename"`                   // If false, load balancers are never renamed after a cluster-name change. Default true
```

jihyun-huh (commented):
Thanks for updating this!
Ive read the new changes and they look good to me.

one small thing I'd consider, is adding a test covering `enable-lb-rename=false` ,
so the new configuration path is explicitly covered. 

I also tested the scenarios locally, including `enable-lb-rename=false`, and it worked well as expected.

pkg/openstack/openstack.go
234u34k: I've added 8fe627a, which I believe will address both pieces of feedback. Appreciate your review. This branch's test configuration now sets `enable-lb-rename=false` and verifies it overrides the default. I also applied the comment capitalization fix. `go test -tags=unit ./pkg/openstack` passes locally. PTAL.

234u34k:
@stephenfin could I ask for a review from you on this at your convenience? Believe it's ready now

234u34k:
@stephenfin @anguslees, following up on the review request. The existing feedback is addressed, and all test and check jobs pass. Could either of you take a look when you have time please?

PR #3084, opened 2026-03-17
[cinder-csi-plugin] Add mkfs-options StorageClass parameter

### LINKED ISSUE #3068: [cinder-csi-plugin] Add 'mkfsOptions' parameter
/kind feature

I would like to create a PR to add an 'mkfsOptions' parameter to Cinder CSI, using the Ceph CSI project's implementation as a blueprint. However, before preparing the PR, I would like to know the maintainers' opinion on this.

**What happened**:

We use the Cinder CSI plugin with Ceph-backed block storage. We have observed that the provisioning of volumes is exceptionally slow, especially if they are large (for example, a 100 GB volume took ~7 minutes). This is caused by the 'discard' step of mkfs (mkfs.ext4 in this case; the same behaviour can be observed when using XFS).

As disabling the discard feature at the infrastructure layer is undesirable, the ideal solution would be to specify the 'nodiscard' option when creating a new file system.

The following comment, which appears in the Ceph CSI's source code, explains why this feature exists in its code base:
> rbd images are thin-provisioned and return zeros for unwritten areas.  A freshly created image will not benefit from discard and we also want to avoid as much unnecessary zeroing as possible.  

Adding a 'mkfsOptions' parameter would make this enhancement flexible and usable for other purposes, too. It would also not change existing behaviour.

**How to reproduce it**:

Create a large volume on a block device backed by Ceph in an infrastructure that passes discards to Ceph. Observe long provisioning times (several minutes for 100 GB devices).

In Ceph CSI, the issue [https://github.com/ceph/ceph-csi/issues/621](https://github.com/ceph/ceph-csi/issues/621) resulted in PR [Disable mkfs.ext4/xfs discard for RBD RXO volumes](https://github.com/ceph/ceph-csi/issues/621), which added hard-coded parameters to 'mkfs'.
Later, 'mkfsOptions' were introduced as a generic solution for issue [Change limit INODES rbd device](https://github.com/ceph/ceph-csi/issues/374).

rybnico:

Add support for a new `mkfsOptions` StorageClass parameter that allows cluster admins to pass custom mkfs flags when formatting volumes. This is particularly useful for Ceph-backed storage where passing `-E nodiscard` to mkfs.ext4 can dramatically speed up provisioning (e.g. from ~7 minutes to seconds for a 100 GB volume).

The parameter value is comma-separated, with each token further split on whitespace to produce individual mkfs command-line arguments. Example: `mkfsOptions: "-E nodiscard,-O ^metadata_csum"`

Implementation:
- Extract mkfsOptions from StorageClass params in CreateVolume and pass via VolumeContext (both new-volume and idempotent early-return paths)
- Parse mkfsOptions in NodeStageVolume using parseMkfsOptions() helper
- Use FormatAndMountSensitiveWithFormatOptions() from k8s.io/mount-utils to pass format options separately from mount options
- No behavior change when mkfsOptions is not set

fixes #3068 

Inspired by the Ceph CSI implementation (ceph/ceph-csi#621).

**Release note**:

```release-note
NONE
```


kayrus:
reading the https://github.com/kubernetes/utils/blob/28399d86e0b55e612b83ad94dc6bac8e5de2a5ac/mount/mount_linux.go#L396 source code it seems that this is not a `mkfs` option, but a `mount` option. Does it make sense to rename it?

rybnico:
> reading the https://github.com/kubernetes/utils/blob/28399d86e0b55e612b83ad94dc6bac8e5de2a5ac/mount/mount_linux.go#L396 source code it seems that this is not a `mkfs` option, but a `mount` option. Does it make sense to rename it?

No, it's an option for `mkfs` when formatting the device for the first time, not a mount option. We use the function `FormatAndMountSensitiveWithFormatOptions` from `mount_linux.go` and pass the `mkfsOptions` as the argument `formatOptions`. Since it's called `mkfsOptions` in Ceph CSI and other CSI controllers, I think it's the correct name for the field.

kayrus:
> No, it's an option for `mkfs` when formatting the device for the first time, not a mount option.

could you point me to the code that passes `sensitiveOptions` to `mkfs`? I've been looking but can't find it. from what I can see, this list is only used for mount options.

rybnico:
> > No, it's an option for `mkfs` when formatting the device for the first time, not a mount option.
> 
> could you point me to the code that passes `sensitiveOptions` to `mkfs`? I've been looking but can't find it. from what I can see, this list is only used for mount options.

Of course - but we don't pass them as argument `sensitiveOptions` but rather `formatOptions`.

We create `formatOptions` from `mkfsOptions` [here](https://github.com/kubernetes/cloud-provider-openstack/pull/3084/changes#diff-a4637ad9e75532ac89924fa2d458e5b6337a3ecc0dac70655f0f43598009ae9eR224)
We pass this as a new argument `formatOptions` to the `formatAndMountRetry` function [here](https://github.com/kubernetes/cloud-provider-openstack/pull/3084/changes#diff-a4637ad9e75532ac89924fa2d458e5b6337a3ecc0dac70655f0f43598009ae9eR231)

In `formatAndMountRetry`, the `formatOptions` are passed to `FormatAndMountSensitiveWithFormatOptions` [here](https://github.com/kubernetes/cloud-provider-openstack/pull/3084/changes#diff-a4637ad9e75532ac89924fa2d458e5b6337a3ecc0dac70655f0f43598009ae9eR261) and [here](https://github.com/kubernetes/cloud-provider-openstack/pull/3084/changes#diff-a4637ad9e75532ac89924fa2d458e5b6337a3ecc0dac70655f0f43598009ae9eR271).

In [mount-utils](https://github.com/kubernetes/mount-utils/blob/e371f8a5454f40f8e721e1e9bc4c955f05095b8e/mount.go#L207), the options is finally passed to [formatAndMountSensitive](https://github.com/kubernetes/mount-utils/blob/e371f8a5454f40f8e721e1e9bc4c955f05095b8e/mount_linux.go#L553) and added to the `args` in the call to [mounter.format](https://github.com/kubernetes/mount-utils/blob/e371f8a5454f40f8e721e1e9bc4c955f05095b8e/mount_linux.go#L609-L610)

pkg/csi/cinder/nodeserver.go
```diff
@@ -429,6 +437,20 @@ func getDevicePath(volumeID string, m mount.IMount) (string, error) {
 	return devicePath, nil
 }
 
+// parseMkfsOptions splits a comma-separated mkfsOptions string into individual
+// command-line arguments. Each comma-separated token is further split on
+// whitespace. Empty tokens are filtered out.
+func parseMkfsOptions(mkfsOpts string) []string {
```
kayrus: you can use https://github.com/kubernetes/cloud-provider-openstack/blob/9c801053ed5f45155a85d350d32b25f4cbfc0a77/pkg/util/util.go#L175-L182

kayrus:
Thanks, got it. I mixed up `sensitiveOptions` and `formatOptions`. They were hidden in github UI, and I missed the `formatOptions`:

Screenshot_20260413_122003

Next time I'll try to use a proper code viewer :)

docs/cinder-csi-plugin/using-cinder-csi-plugin.md
```diff
@@ -283,6 +283,7 @@ helm install --namespace kube-system --name cinder-csi ./charts/cinder-csi-plugi
 |-------------------------   |-----------------------|-----------------|-----------------|
 | StorageClass `parameters`  | `availability`          | `nova`          | String. Volume Availability Zone |
 | StorageClass `parameters`  | `type`                  | Empty String    | String. Name/ID of Volume type. Corresponding volume type should exist in cinder     |
+| StorageClass `parameters`  | `mkfsOptions`           | Empty String    | String. Comma-separated mkfs options passed to mkfs when formatting the volume. Useful for passing `-E nodiscard` on Ceph-backed storage to speed up provisioning. |
```
kayrus: it makes sense to use kebab-case style here

pkg/csi/cinder/nodeserver.go
rybnico: Good catch! I removed the `parseMkfsOptions` function and use `SplitTrim` instead.

docs/cinder-csi-plugin/using-cinder-csi-plugin.md
rybnico: Yes, you're right. It isn't consistent across CSI implementations. However, looking at the other parameters in Cinder CSI, kebab-style is a better fit.
I also removed the overly specific comment about Ceph/nodiscard.

kayrus:
>  pkg/csi/cinder/nodeserver.go:29:1: File is not properly formatted (gofmt)
>	"google.golang.org/grpc/status"
> ^
> 1 issues:
> * gofmt: 1
> exit status 1
> make: *** [Makefile:84: check] Error 1


can you fix formatting? and rename the PR title

rybnico:
> > pkg/csi/cinder/nodeserver.go:29:1: File is not properly formatted (gofmt)
> > "google.golang.org/grpc/status"
> > ^
> > 1 issues:
> > 
> > * gofmt: 1
> >   exit status 1
> >   make: *** [Makefile:84: check] Error 1
> 
> can you fix formatting? and rename the PR title

I fixed it. I don't know why it didn't get caught locally. I also changed the PR title.

pkg/csi/cinder/controllerserver_test.go
```diff
@@ -200,6 +200,94 @@ func TestCreateVolumeWithParam(t *testing.T) {
...
+		Name: FakeVolName,
+		VolumeCapabilities: []*csi.VolumeCapability{
+			{
+				AccessMode: &csi.VolumeCapability_AccessMode{
+					Mode: csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
+				},
+			},
+		},
+		Parameters: map[string]string{
+			"mkfs-options": "-E nodiscard",
```
kayrus, line 269: I think it also makes sense to add tests for multiple format options, separated with comma.

rybnico: This was a very good suggestion because it made me aware of a bug that occurred when using multiple arguments. Since it doesn't make sense to split the arguments by comma, I now use the `strings.Fields` function to split them by space.
I added tests for different/multiple arguments in `mkfs-options` and I made mocking commands in `pkg/util/mount/mount_mock.go` possible to check if the correct arguments are used in the `mkfs` call.

I appreciate your thorough review!

kayrus: hm. it seems we need something smarter to avoid cases like this: https://go.dev/play/p/fp3YAphLD8m

```
Fields are: ["-L" "\"my" "data" "disk\"" "/dev/sdb1"]
```

kayrus: see https://github.com/kubernetes-sigs/kustomize/pull/5943

rybnico: I noticed that, too. However, I was hesitant to add a dependency on shlex.
As far as I know, the label argument is the only one that takes a quoted string as a parameter. And the label argument doesn't make sense in our context anyway.
If you'd like, though, I can improve the parsing. Would you prefer to use shlex or copy kustomization's shlex implementation? 
Another option is to keep it as it is and add a note to the documentation stating that quoted strings are not supported.
What do you prefer?

kayrus: > Another option is to keep it as it is and add a note to the documentation stating that quoted strings are not supported.

This is a highly sensitive case that could allow CLI injection with root privileges. This PR must not be merged until all such vulnerabilities are fully mitigated.

> copy kustomization's shlex implementation

I'd prefer this approach, but it doesn't scale well. It might make sense to open an issue in the kustomization repository and ask whether this logic could be extracted into a separate Kubernetes package.

rybnico: I did think about CLI injection. A few points:

- A `StorageClass` can only be created by cluster-admins, who already have more direct ways to run arbitrary commands on nodes (privileged DaemonSets, hostPath mounts, etc.). This path doesn't meaningfully expand the attack surface.
- The command runs via `os/exec.Command`, which execves directly without invoking a shell. Characters like `;`, `|`, `&&`, and `$()` are passed as literal argv bytes to mkfs and are never shell-interpreted.
- Other CSI drivers don't validate mkfs options either. See the Ceph-CSI example: https://github.com/ceph/ceph-csi/blob/13b59fce63f3f9e15ae35ff9d37a5c4d8f6fa98a/internal/rbd/nodeserver.go#L928

Is it really worth using shlex and/or asking the kustomize project to extract their function? That's quite a bit of an upstream process for a ~60-line helper. Ultimately, it doesn't improve security and isn't necessary since there are no other "quotable" arguments.

rybnico: @kayrus Could you take another look and let me know how we should proceed? Thanks in advance!

kayrus: @zetaab @stephenfin can you take a look at this PR?

rybnico: @kayrus @zetaab @stephenfin Would you mind having another look and telling me if the argument parsing needs additional work for this PR to get merged? Thanks!

zetaab (approved):
/lgtm

pkg/csi/cinder/controllerserver_test.go
zetaab: I was just thinking could it work in similar way than pod command in kubernetes ([]string)? By that way people could easily understand the syntax if its more complex than just this in test.

rybnico: First, thanks for your review!
Using a string array for the options would be great, but the [parameters field](https://github.com/kubernetes/kubernetes/blob/b859f5a5c1b5ece16624f5290b835fe63898cd37/pkg/apis/storage/types.go#L50) in the `StorageClass` is of type `map[string]string`.

rybnico: @kayrus @stephenfin Any chance of getting the last approval to get this merged?

mohdazaruddin:
Any plan for merging this ? We are also having the same issues with ceph backed storage and waiting for this release

pkg/csi/cinder/nodeserver_test.go
```diff
@@ -236,6 +240,134 @@ func TestNodeStageVolumeBlock(t *testing.T) {
...
+			mkfsOptions:       "-E nodiscard",
+			expectedArguments: []string{"-E", "nodiscard"},
+		},
+		{
+			name:              "multiple options",
+			mkfsOptions:       "-E nodiscard -j",
+			expectedArguments: []string{"-E", "nodiscard", "-j"},
+		},
+		{
+			// Demonstrates the bug in util.SplitTrim: it also splits on
```
mandre: Looks like a stale comment, the current implementation uses `strings.Fields()`.

How about something like:
```go
// Verifies that commas within a single option value are preserved
```

pkg/csi/cinder/controllerserver_test.go
mandre: As @rybnico explains above, I also think this does not pose a new security threat, as `StorageClass` are admin-managed. This also seems to match the trust model of other CSI drivers.

mandre (commented):
LGTM.

We would need to rebase the PR before we can merge to fix merge conflicts.

pkg/csi/cinder/nodeserver_test.go
rybnico: Good catch; I've replaced the stale comment with your suggestion.

rybnico:
> LGTM.
> 
> We would need to rebase the PR before we can merge to fix merge conflicts.

Thank you for your review. I've rebased the PR.

mandre (commented):
/approve

mandre (commented):
/lgtm

PR #3091, opened 2026-04-02
[occm] Sync Helm chart with raw manifests

fvrk4n:

The OCCM Helm chart had drifted from the raw manifests in three areas:

1. **Scheduling**: The Helm chart used `nodeSelector` with an exact empty string match (`node-role.kubernetes.io/control-plane: ""`), while the raw manifest uses `nodeAffinity` with `operator: Exists`. This fails on clusters where the label value is not empty (e.g. RKE2 sets it to `true`). Replaced `nodeSelector` with `affinity` in both `values.yaml` and `daemonset.yaml`.

2. **Tolerations**: The raw manifest defines four tolerations but the Helm chart only had two. Added the missing `CriticalAddonsOnly=true:NoExecute` (without this, pods fail to schedule on clusters that taint control plane nodes with CriticalAddonsOnly) and `node-role.kubernetes.io/master:NoSchedule` (backward compatibility with older clusters using the legacy label).

3. **RBAC**: The raw manifest defines both `system:cloud-controller-manager` and `system:cloud-node-controller` RBAC resources. The Helm chart only created `system:cloud-controller-manager`. Added the missing ClusterRole and ClusterRoleBinding for `system:cloud-node-controller`.

fixes #

Tested end to end on an RKE2 cluster with 3 control plane and 3 worker nodes. Control plane nodes had `CriticalAddonsOnly=true:NoExecute` taint and `node-role.kubernetes.io/control-plane: true` label (not empty string). DaemonSet scheduled 3/3 pods only on control plane nodes, all running with clean logs.

Note: The raw manifest binds `system:cloud-node-controller` to a `cloud-node-controller` service account that no pod uses. The Helm chart intentionally binds it to `.Values.serviceAccountName` (`cloud-controller-manager`) so the DaemonSet pods actually receive both sets of permissions.

**Release note**:
```release-note
[openstack-cloud-controller-manager] Sync Helm chart with raw manifests: replace nodeSelector with nodeAffinity, add missing CriticalAddonsOnly and master tolerations, add cloud-node-controller RBAC resources.
```


fvrk4n:
@kayrus please review it

stephenfin:
This looks good to me. My only concern is the removal of the `nodeSelector` variable. Would it make sense to either (a) continue to respect this if set but to unset it by default or (b) to emit an error if it's set? If (b), something like the below should work?

```
  {{- if .Values.nodeSelector }}
  {{ fail "nodeSelector has been replaced by affinity in this release. See UPGRADING.md for migration steps." }}
  {{- end }}
```

Note that I had a similar issue in https://github.com/kubernetes/cloud-provider-openstack/pull/3181 but that was slightly different since the variable there was unusable in virtually all scenarios, whereas this one is only unusable, as you noted, "on clusters where the label value is not empty".

PR #3097, opened 2026-04-15
[occm] Make hostNetwork configurable in the Helm chart

zwindler:

Makes `hostNetwork` configurable in the cloud controller manager Helm chart by adding a `hostNetwork` value (default: `true`).

Currently `hostNetwork: true` is hardcoded in the DaemonSet template. This makes sense for the default use case where OCCM manages nodes and routes, but for users who only run a subset of controllers (e.g. just the `service` controller for Octavia LB), `hostNetwork` is an unnecessary privilege escalation. The controller only needs to talk to the Kubernetes API and the OpenStack API, both reachable from pod networking.

This PR also changes the default `dnsPolicy` to `ClusterFirstWithHostNet`, as [recommended by the Kubernetes docs](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy) for pods running with `hostNetwork`:

> `ClusterFirstWithHostNet`: For Pods running with hostNetwork, you should explicitly set its DNS policy to `ClusterFirstWithHostNet`. Otherwise, Pods running with hostNetwork and `"ClusterFirst"` will fallback to the behavior of the `"Default"` policy.

The current chart uses `dnsPolicy: ClusterFirst` with `hostNetwork: true`, which falls back to host DNS. This has no practical impact today since OCCM doesn't resolve cluster-internal service names, but it's *technically* incorrect. Users who set `hostNetwork: false` should also set `dnsPolicy: ClusterFirst` (documented in `values.yaml`).

N/A

The default remains `hostNetwork: true`, so this is a no-op for existing deployments, apart from the `dnsPolicy` default change.

**Release note**:
```release-note
[openstack-cloud-controller-manager] Make hostNetwork configurable in the Helm chart (default: true) and default dnsPolicy to ClusterFirstWithHostNet
```


zwindler:
Hi 👋 — gentle ping on this one. It's a small, backwards-compatible change (the default stays `hostNetwork: true`, so it's a no-op for existing deployments), and CI lint + CLA are green.

Whenever someone has a spare moment, would a maintainer be willing to add `/ok-to-test` so the full suite can run, and take a look when convenient? No rush at all — I know review bandwidth is precious, and I really appreciate the time.

Happy to split the `hostNetwork` toggle and the `dnsPolicy` fix into two separate PRs if that makes review easier — just let me know. 🙏

jihyun-huh:
Hi, thank you for your PR and i also need this option in my team
i tested locally and it looks good

@kayrus @jichenjc Could you review this with ok-to-test plz?

charts/openstack-cloud-controller-manager/templates/daemonset.yaml
```diff
@@ -109,8 +109,8 @@ spec:
       {{- if .Values.extraInitContainers }}
       initContainers: {{ toYaml .Values.extraInitContainers | nindent 6 }}
       {{- end }}
-      dnsPolicy: {{ .Values.dnsPolicy }}
-      hostNetwork: true
+      hostNetwork: {{ .Values.hostNetwork }}
+      dnsPolicy: {{ if and .Values.hostNetwork (eq .Values.dnsPolicy "ClusterFirst") }}ClusterFirstWithHostNet{{ else }}{{ .Values.dnsPolicy }}{{ end }}
```
stephenfin, line 113: Do we need this logic here? I agree that we should change the default, but we can do that in `values.yaml`

stephenfin (changes requested):
One comment

charts/openstack-cloud-controller-manager/templates/daemonset.yaml
zwindler: You are right, this is probably over-engineered

I'll fix the PR on monday, thanks for the review

zwindler: Done. The template now renders `dnsPolicy: {{ .Values.dnsPolicy }}` as before, and the default in `values.yaml` is `ClusterFirstWithHostNet` (with a comment pointing users who disable `hostNetwork` to `ClusterFirst`). Also bumped the chart to 2.36.4 since master moved to 2.36.3 in the meantime.

stephenfin (commented):
/approve
/lgtm

PR #3100, opened 2026-04-28
Pin GitHub Actions to git SHAs

m-bull:
Tagged releases of GitHub Actions used in workflows are no longer allowed in `kubernetes/cloud-provider-openstack`, and all actions must be pinned to a full-length commit SHA.

I haven't updated the action versions, these are pinned to the SHAs represented by the tags that were already in place. Happy to bump action versions if requested as these are getting a bit old!

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -10,23 +10,23 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


stephenfin (commented):
/lgtm
/approve
/retest-required

We likely need to update these actions, but that shouldn't block this.

Issue #2624, opened 2024-07-05
The kubernetes loadbalancer will be overwritten on OpenStack if two LoadBalancer service are using the same name.


bryan-song-007:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
>
  /kind bug
> /kind feature

**What happened**:
1. Create two K8S clusters running on the same OpenStack (cluster A and cluster B);
2. Create one deployment on each of the clusters: kubectl create deployment test-web --replicas 3 --image nginx --port 80
3. Expose the deployments in cluster A: 
    kubectl expose deployment test-web --name openstack-octavia-lb --type LoadBalancer --target-port 80 --port 6666, 
    a new load balancer will be created on OpenStack, this is expected.
5. Expose the deployments in cluster B using the same service name with the service in cluster A:
    kubectl expose deployment test-web --name openstack-octavia-lb --type LoadBalancer --target-port 80 --port 8888
6. OpenStack will not create a new load balancer but update the existing load balancer in step 3, and set the port to 8888, this is not expected, I guess that is because the tag of the load balancer will be the same from the cluster A and cluster B.
7. Delete the load balancer service in cluster A, and the load balancer on OpenStack will be deleted as well, so cluster B will have no load balancer to use.

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: openstack-cloud-controller-manager:v1.24.0
- OpenStack version: OpenStack Zed and Antelope

bryan-song-007:
Simlar issue can be found here: https://github.com/kubernetes/cloud-provider-openstack/issues/2241#issuecomment-1618921550 .

Not a bug, close it.

Issue #2571, opened 2024-04-10
Overlapping load balancers when running multiple clusters in the same tenant


canori01:
/kind bug

What happened:
I'm running a couple of RKE clusters with OpenStack integration in separate tenants each.
When I perform the Istio installation with one ingress gateway and kind: LoadBalancer on the k8s level, I'm getting a load balancer instance with the name "kube_service_kubernetes_istio-system_istio-ingressgateway" and IP from an external network, as expected.

However, when I perform the same Istio installation in different OpenStack tenant on different RKE cluster, I'm not able to create a load balancer instance under the same name, as when I do it so, the occm is trying to re-register the instance, currently running (same name, different UUID)

What you expected to happen:
Load balancers with the same name will be running in parallel.

How to reproduce it:
I assume it's a general issue with lb instances naming, so any service with the same name, configured in different RKE clusters, exposes services by using occm integration to OpenStack. In my case, the Istion instance is the blame.

Anything else we need to know?:

Environment:

OpenStack-cloud-controller-manager(or other related binary) version: Kubernetes v1.27.10
OpenStack version: Yoga

dulek:
The RKE clusters have to set the `--cluster-name` option to a unique cluster identifier when running OCCM. This way the "kubernetes" part will be replaced with that value, and your LBs will not conflict.

As we've faced the same problem in OpenShift, I've recently wrote some code helping with a transition from default `--cluster-name=kubernetes` to the unique value. See #2552. If you'd like to discuss this approach further, please use #2241. I'm going to close this one.

canori01:
That works. Thanks!

PR #3103, opened 2026-04-30
[occm] Tag load balancers with cluster identity to prevent name collisions

### LINKED ISSUE #3102: [occm] Cross-cluster load balancer name collision when multiple Kubernetes clusters share an OpenStack project
/kind bug

**What happened**:

OCCM identifies an existing Octavia load balancer for a Service by name on
the first reconcile (`getLoadbalancerByName` in `pkg/openstack/loadbalancer.go`).
The name is built from `kube_service_<cluster-name>_<namespace>_<service>`,
where `<cluster-name>` defaults to `kubernetes`. When two Kubernetes clusters
share the same OpenStack project and use the same `--cluster-name` (which is
the default for many distributions: kubeadm, RKE2, etc.), Services with
identical namespace and name produce identical load balancer names.

Octavia does not require load balancer names to be unique inside a project,
so OCCM in cluster B happily picks up cluster A's load balancer, sets the
`loadbalancer.openstack.org/load-balancer-id` annotation on its own Service,
and starts driving cluster A's load balancer (rewriting listeners,
members, FIP, etc.). Cluster A then loses its load balancer.

This is the same root cause discussed in the closed issues #2241, #2571 and
#2624. The accepted upstream guidance is "use a unique `--cluster-name`",
which is correct but does not *defend* against the failure mode at all -
two operators independently bootstrapping clusters in the same tenant will
keep hitting it.

OCCM should never adopt a load balancer that is owned by a different
Kubernetes cluster, even when names collide. A unique `--cluster-name`
should be a recommendation, not the only safety mechanism.

1. Create two Kubernetes clusters (cluster A and cluster B) in the same
   OpenStack project. Both run OCCM with the default
   `--cluster-name=kubernetes` (or any matching value).
2. On cluster A: `kubectl create deployment web --image nginx --port 80 && kubectl expose deployment web --type LoadBalancer --port 6666 --target-port 80`.
   An Octavia LB named `kube_service_kubernetes_default_web` is created
   in OpenStack.
3. On cluster B: same commands, exposing the service on port 8888.
4. Observe that no second load balancer is created. Instead OCCM in
   cluster B locates cluster A's LB by name, annotates its own Service
   with the same `load-balancer-id`, and rewrites the LB to point at
   cluster B's nodes on port 8888. Cluster A's Service is now broken.

I'd like to propose adding a stable cluster identifier (the UID of the
`kube-system` namespace) as a load balancer tag of the form
`kube_cluster_id_<uid>`. The lookup would treat a load balancer with a
foreign `kube_cluster_id_*` tag as not-found instead of adopting it, and
fall back to the legacy behaviour for load balancers that don't carry any
`kube_cluster_id_*` tag (existing deployments and externally-created LBs).

This is a strictly additive, backward-compatible change that defends
against the failure mode without forcing operators to coordinate
`--cluster-name` values. I have a working implementation and will open a
PR shortly that links this issue.

**Environment**:
- openstack-cloud-controller-manager version: master (reproduced against
  v1.33.0 as well)
- OpenStack version: any with Octavia tags support (>= API v2.5 / Stein)
- Others: N/A

enginrect:

OCCM identifies an existing Octavia load balancer for a Service by name on
the first reconcile (via `getLoadbalancerByName`). The name format
`kube_service_<cluster-name>_<namespace>_<service>` defaults to a
`<cluster-name>` of `kubernetes`, so two Kubernetes clusters in the same
OpenStack project that happen to use the default cluster-name and have
Services with identical namespace/name produce identical load balancer
names. Octavia does not enforce uniqueness of names, so OCCM in cluster B
ends up adopting and overwriting cluster A's load balancer. This has been
reported repeatedly (see #2241, #2571, #2624) and the standing guidance
"set a unique `--cluster-name`" is correct but does not actually defend
against the failure mode.

This PR adds a stable Kubernetes cluster identifier - the UID of the
`kube-system` namespace - as a load balancer tag of the form
`kube_cluster_id_<uid>`. Lookup behaviour:

- LBs that carry the matching `kube_cluster_id_<our-uid>` tag are kept.
- LBs that carry no `kube_cluster_id_*` tag fall back to the legacy
  behaviour (preserves existing deployments and externally-created LBs).
- LBs that carry only foreign `kube_cluster_id_*` tags are treated as
  NotFound, with a warning. OCCM will then create its own load balancer
  rather than overwriting one that belongs to another cluster.

The cluster UID is read once at controller-manager start-up. If the
lookup fails (RBAC denial, missing namespace, etc.) the safeguard is
disabled and OCCM falls back to the legacy name-based behaviour, so the
change is strictly additive. Pre-existing load balancers also gain the
`kube_cluster_id_*` tag during the next reconciliation.

fixes #3102

- Backward compatibility:
  - Load balancers without any `kube_cluster_id_*` tag keep the previous
    behaviour. They are tagged on the next successful reconcile.
  - Load balancers looked up via the existing
    `loadbalancer.openstack.org/load-balancer-id` annotation (i.e. on
    every reconcile after the first one) go through `GetLoadbalancerByID`,
    which is unaffected.
- New RBAC: `get` on `namespaces` is added to both the manifest
  ClusterRole (`manifests/controller-manager/cloud-controller-manager-roles.yaml`)
  and the helm chart (`charts/openstack-cloud-controller-manager/templates/clusterrole.yaml`).
  If the verb is unavailable the safeguard simply degrades to the legacy
  behaviour with a warning log; OCCM does not refuse to start.
- Octavia API >= v2.5 (Stein) is required for the tag feature. This is
  already gated by `svcConf.supportLBTags` and behaves as before on older
  clouds.
- New unit tests:
  - `TestFilterLoadBalancersByClusterID` covers the matching, legacy,
    foreign-only, and mixed cases.
  - `TestFetchClusterUID` covers happy path and graceful degradation
    (missing namespace, forbidden) with a fake clientset.

How to verify manually:

```sh
go test ./pkg/openstack/...
```

A reproduction of the original failure mode (two clusters in the same
project, same `--cluster-name`, same Service ns/name) is described in
#3102.

**Release note**:
```release-note
[openstack-cloud-controller-manager] Octavia load balancers are now tagged with `kube_cluster_id_<kube-system-uid>` so OCCM no longer adopts a load balancer owned by another cluster in the same OpenStack project. The cloud-controller-manager ClusterRole gains `get` on `namespaces`.
```

### DIFF
--- a/charts/openstack-cloud-controller-manager/templates/clusterrole.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/clusterrole.yaml
@@ -96,3 +96,9 @@ rules:
   - list
   - get
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - get
--- a/charts/openstack-cloud-controller-manager/templates/clusterrole.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/clusterrole.yaml
@@ -86,3 +86,9 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - get
--- a/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
+++ b/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
@@ -11,6 +11,7 @@
     - [Restrict Access For LoadBalancer Service](#restrict-access-for-loadbalancer-service)
     - [Use PROXY protocol to preserve client IP](#use-proxy-protocol-to-preserve-client-ip)
     - [Sharing load balancer with multiple Services](#sharing-load-balancer-with-multiple-services)
+    - [Running multiple Kubernetes clusters in one OpenStack project](#running-multiple-kubernetes-clusters-in-one-openstack-project)
     - [IPv4 / IPv6 dual-stack services](#ipv4--ipv6-dual-stack-services)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
@@ -640,6 +641,55 @@ $ openstack loadbalancer listener list --loadbalancer 2b224530-9414-4302-8163-5a
 ```
 
 The load balancer will be deleted after `service-2` is deleted.
+
+### Running multiple Kubernetes clusters in one OpenStack project
+
+OCCM names each load balancer it creates as `kube_service_<cluster-name>_<namespace>_<service>`,
+where `<cluster-name>` comes from the kube-controller-manager `--cluster-name` flag and
+defaults to `kubernetes`. When two Kubernetes clusters share the same OpenStack project
+and use the same `--cluster-name`, namespace and service name, the resulting load balancer
+names collide. OpenStack does not require load balancer names to be unique, so previously
+the second cluster could "adopt" the load balancer that the first cluster owned and start
+overwriting its configuration. The recommended way to avoid this remains to set a unique
+`--cluster-name` on every Kubernetes cluster (see issues
+[#2241](https://github.com/kubernetes/cloud-provider-openstack/issues/2241),
+[#2571](https://github.com/kubernetes/cloud-provider-openstack/issues/2571),
+[#2624](https://github.com/kubernetes/cloud-provider-openstack/issues/2624)).
+
+To make OCCM resilient even when that recommendation isn't followed, OCCM also tags
+every load balancer it creates with the UID of the cluster's `kube-system` namespace,
+which is a stable and unique identifier per Kubernetes cluster. The tag has the form
+`kube_cluster_id_<uid>`. When OCCM looks up a load balancer by name on the first
+reconcile of a Service, it ignores load balancers that carry a `kube_cluster_id_*` tag
+for a *different* cluster and creates a new one instead. Load balancers that don't
+carry any `kube_cluster_id_*` tag (legacy load balancers, or load balancers created by
+external tooling) keep their previous behaviour for backward compatibility, and gain
+the cluster-id tag on the next reconcile.
+
+OCCM reads the `kube-system` namespace UID once at start-up; this requires the
+`get` verb on the `namespaces` resource (already part of the standard cloud-controller
+RBAC). If the lookup fails (for example because of a custom RBAC restriction) OCCM logs
+a warning and continues without the safeguard, falling back to the legacy name-based
+behaviour.
+
+After successful reconcile, the load balancer tags will look similar to:
+
+```shell
+$ openstack loadbalancer show <lb-id> -c name -c tags
++-------+--------------------------------------------------------------------------+
+| Field | Value                                                                    |
++-------+--------------------------------------------------------------------------+
+| name  | kube_service_kubernetes_default_service-1                                |
+| tags  | kube_service_kubernetes_default_service-1,                               |
+|       | kube_cluster_id_11111111-2222-3333-4444-555555555555                     |
++-------+--------------------------------------------------------------------------+
+```
+
+> NOTE: This safeguard only protects new (or freshly reconciled) load balancers. It
+> does not retroactively resolve an already-stolen load balancer between two running
+> clusters. If you suspect cross-cluster collisions, set unique `--cluster-name` on
+> each cluster and let OCCM rename the existing resources (see
+> [PR #2552](https://github.com/kubernetes/cloud-provider-openstack/pull/2552)).
 
 ### IPv4 / IPv6 dual-stack services
 Since Kubernetes 1.20, Kubernetes clusters can run in dual-stack mode,
--- a/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
+++ b/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
@@ -651,10 +651,7 @@ and use the same `--cluster-name`, namespace and service name, the resulting loa
 names collide. OpenStack does not require load balancer names to be unique, so previously
 the second cluster could "adopt" the load balancer that the first cluster owned and start
 overwriting its configuration. The recommended way to avoid this remains to set a unique
-`--cluster-name` on every Kubernetes cluster (see issues
-[#2241](https://github.com/kubernetes/cloud-provider-openstack/issues/2241),
-[#2571](https://github.com/kubernetes/cloud-provider-openstack/issues/2571),
-[#2624](https://github.com/kubernetes/cloud-provider-openstack/issues/2624)).
+`--cluster-name` on every Kubernetes cluster.
 
 To make OCCM resilient even when that recommendation isn't followed, OCCM also tags
 every load balancer it creates with the UID of the cluster's `kube-system` namespace,
--- a/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
+++ b/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
@@ -663,6 +663,12 @@ carry any `kube_cluster_id_*` tag (legacy load balancers, or load balancers crea
 external tooling) keep their previous behaviour for backward compatibility, and gain
 the cluster-id tag on the next reconcile.
 
+A load balancer that carries this cluster's `kube_cluster_id_*` tag *and* a tag of
+another cluster at the same time is ambiguous: OCCM cannot tell which cluster owns it,
+so it neither adopts nor deletes it. The reconcile fails with an error and OCCM records
+a `LoadBalancerClusterIDConflict` Warning event on the Service. Remove the stale
+`kube_cluster_id_*` tags from the load balancer to resolve it.
+
 OCCM reads the `kube-system` namespace UID once at start-up; this requires the
 `get` verb on the `namespaces` resource (already part of the standard cloud-controller
 RBAC). If the lookup fails (for example because of a custom RBAC restriction) OCCM logs
--- a/manifests/controller-manager/cloud-controller-manager-roles.yaml
+++ b/manifests/controller-manager/cloud-controller-manager-roles.yaml
@@ -93,6 +93,12 @@ items:
     - list
     - get
     - watch
+  - apiGroups:
+    - ""
+    resources:
+    - namespaces
+    verbs:
+    - get
 - apiVersion: rbac.authorization.k8s.io/v1
   kind: ClusterRole
   metadata:
--- a/manifests/controller-manager/cloud-controller-manager-roles.yaml
+++ b/manifests/controller-manager/cloud-controller-manager-roles.yaml
@@ -99,6 +99,12 @@ items:
     - get
     - list
     - watch
+  - apiGroups:
+    - ""
+    resources:
+    - namespaces
+    verbs:
+    - get
 - apiVersion: rbac.authorization.k8s.io/v1
   kind: ClusterRole
   metadata:
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -121,14 +121,19 @@ const (
 	clusterIDTagPrefix = "kube_cluster_id_"
 )
 
-// clusterIDTag formats the Octavia load balancer tag carrying the cluster
-// identifier. It returns the empty string when uid is empty so callers can
-// safely append the result without a separate nil-check.
-func clusterIDTag(uid string) string {
+// withClusterIDTag appends the Octavia load balancer tag carrying the cluster
+// identifier to tags. It is a no-op when uid is empty (the cluster identity
+// could not be determined) or when the tag is already present, so callers
+// don't need any guard logic of their own.
+func withClusterIDTag(tags []string, uid string) []string {
 	if uid == "" {
-		return ""
+		return tags
 	}
-	return clusterIDTagPrefix + uid
+	tag := clusterIDTagPrefix + uid
+	if slices.Contains(tags, tag) {
+		return tags
+	}
+	return append(tags, tag)
 }
 
 // LbaasV2 is a LoadBalancer implementation based on Octavia
@@ -293,7 +298,7 @@ func filterLoadBalancersByClusterID(lbs []loadbalancers.LoadBalancer, clusterUID
 	if clusterUID == "" || len(lbs) == 0 {
 		return lbs, false
 	}
-	wantTag := clusterIDTag(clusterUID)
+	wantTag := clusterIDTagPrefix + clusterUID
 	var owned []loadbalancers.LoadBalancer
 	taggedAny := false
 	for _, lb := range lbs {
@@ -363,10 +368,7 @@ func (lbaas *LbaasV2) createOctaviaLoadBalancer(ctx context.Context, name, clust
 	}
 
 	if svcConf.supportLBTags {
-		createOpts.Tags = withLBNameTag(svcConf.lbName, svcConf.lbTags)
-		if tag := clusterIDTag(lbaas.clusterUID); tag != "" && !slices.Contains(createOpts.Tags, tag) {
-			createOpts.Tags = append(createOpts.Tags, tag)
-		}
+		createOpts.Tags = withClusterIDTag(withLBNameTag(svcConf.lbName, svcConf.lbTags), lbaas.clusterUID)
 	}
 
 	if svcConf.flavorID != "" {
@@ -1981,10 +1983,7 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 	// Ensure the LB name tag, the cluster-identity tag, plus any tags from the
 	// Service annotation in a single update.
 	if svcConf.supportLBTags {
-		lbTags := withLBNameTag(lbName, svcConf.lbTags)
-		if tag := clusterIDTag(lbaas.clusterUID); tag != "" && !slices.Contains(lbTags, tag) {
-			lbTags = append(lbTags, tag)
-		}
+		lbTags := withClusterIDTag(withLBNameTag(lbName, svcConf.lbTags), lbaas.clusterUID)
 		klog.V(4).Infof("Desired load balancer tags: %v (LB name plus annotation %s)", lbTags, ServiceAnnotationLoadBalancerTags)
 		if tags, changed := cpoutil.Merge(loadbalancer.Tags, lbTags); changed {
 			klog.InfoS("Updating load balancer tags", "lbID", loadbalancer.ID, "tags", tags)
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -119,12 +119,6 @@ const (
 	// OCCM to disambiguate load balancers when multiple Kubernetes clusters
 	// share the same OpenStack project and a service name happens to collide.
 	clusterIDTagPrefix = "kube_cluster_id_"
-
-	// eventLBStolen is emitted when getLoadbalancerByName finds a load
-	// balancer with a matching name that belongs to a different Kubernetes
-	// cluster (different cluster-id tag). The lookup is treated as NotFound
-	// so OCCM creates a new load balancer instead of stealing an existing one.
-	eventLBStolen = "LoadBalancerNameCollision"
 )
 
 // clusterIDTag formats the Octavia load balancer tag carrying the cluster
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -121,21 +121,6 @@ const (
 	clusterIDTagPrefix = "kube_cluster_id_"
 )
 
-// withClusterIDTag appends the Octavia load balancer tag carrying the cluster
-// identifier to tags. It is a no-op when uid is empty (the cluster identity
-// could not be determined) or when the tag is already present, so callers
-// don't need any guard logic of their own.
-func withClusterIDTag(tags []string, uid string) []string {
-	if uid == "" {
-		return tags
-	}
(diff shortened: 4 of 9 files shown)


enginrect:
Hi @kayrus @stephenfin @zetaab — first-time contributor here. This PR addresses the long-standing cross-cluster LB collision issue (refs #2241, #2571, #2624) with an additive, backward-compatible `kube_cluster_id_<kube-system-uid>` tag on Octavia load balancers. Lookups now reject LBs tagged for a different cluster, fall back to legacy behaviour for untagged LBs, and tag pre-existing LBs on the next reconcile. The safeguard degrades gracefully (warning log + legacy behaviour) if the new ``get`` on `namespaces` RBAC is not granted, so the change is strictly additive.

Could one of you take a look and add `/ok-to-test` when convenient? The failing "Lint Charts" check is unrelated to this PR — it is a pre-existing repository-policy issue on master where the workflow uses unpinned action tags, and it currently fails on every PR.

Thanks!

enginrect:
Quick update: I've rebased on top of master (which now has the SHA-pinned actions from #3100) and bumped the helm chart patch version to 2.35.1 — `Lint Charts` and EasyCLA are both green now. The PR is still blocked on `needs-ok-to-test`. @kayrus @stephenfin @zetaab — would one of you mind adding `/ok-to-test` when convenient? Happy to address review feedback as it comes in.

pkg/openstack/openstack.go
```diff
@@ -160,6 +167,11 @@ type OpenStack struct {
 	kclient               kubernetes.Interface
 	nodeInformer          coreinformers.NodeInformer
 	nodeInformerHasSynced func() bool
+	// clusterUID is the kube-system namespace UID, used as a stable cluster
+	// identifier on OpenStack load balancer tags. May be empty if the lookup
+	// failed or RBAC does not allow it; in that case OCCM falls back to the
+	// legacy name-based load balancer identification.
+	clusterUID string
```
kayrus: `clusterUID` can be defined only in `LoadBalancer` struct, please remove it from here.

pkg/openstack/openstack.go
```diff
@@ -372,7 +405,7 @@ func (os *OpenStack) LoadBalancer() (cloudprovider.LoadBalancer, bool) {
 
 	klog.V(1).Info("Claiming to support LoadBalancer")
 
-	return &LbaasV2{LoadBalancer{secret, network, lb, os.lbOpts, os.kclient, os.eventRecorder}}, true
+	return &LbaasV2{LoadBalancer{secret, network, lb, os.lbOpts, os.kclient, os.eventRecorder, os.clusterUID}}, true
```
kayrus: you can set `clusterUID` value here with:

```go
clusterID := fetchClusterUID(os.kclient)
```

pkg/openstack/loadbalancer.go
kayrus: ~you can filter by tags using `ListOpts.Tags`. `filterLoadBalancersByClusterID` doesn't make sense.~
UPD: please ignore this comment

pkg/openstack/loadbalancer.go
```diff
@@ -105,8 +105,31 @@ const (
...
+	// namespace UID). Together with the existing servicePrefix tag it allows
+	// OCCM to disambiguate load balancers when multiple Kubernetes clusters
+	// share the same OpenStack project and a service name happens to collide.
+	clusterIDTagPrefix = "kube_cluster_id_"
+
+	// eventLBStolen is emitted when getLoadbalancerByName finds a load
+	// balancer with a matching name that belongs to a different Kubernetes
+	// cluster (different cluster-id tag). The lookup is treated as NotFound
+	// so OCCM creates a new load balancer instead of stealing an existing one.
+	eventLBStolen = "LoadBalancerNameCollision"
```
kayrus: is it used somewhere?

pkg/openstack/openstack.go
enginrect: Good catch — thanks! You're absolutely right, that was a cohesion miss on my part. `clusterUID` is only consumed by the LB code path so it shouldn't pollute the global `OpenStack` struct. Removed in 56756ea; the field now lives only on `LoadBalancer` where it belongs.

pkg/openstack/openstack.go
enginrect: Thank you, that's a much cleaner pattern. Fetching lazily inside `LoadBalancer()` keeps the `kube-system` namespace lookup out of the global `Initialize()` path and limits the change to the LB construction site. As a nice side effect, clusters that disable LB now skip the lookup entirely. Done in 56756ea.

pkg/openstack/loadbalancer.go
enginrect: Great question — and the honest answer is: no, it's not used anywhere, sorry about the leftover. The original intent was to emit a Warning event via `eventRecorder` from `getLoadbalancerByName` when we drop a foreign-tagged LB, but plumbing the event recorder and `*v1.Service` into that free-standing function felt out of scope for this PR (the existing warning log already covers operator visibility), so I dropped the emission and forgot to remove the constant. Cleaned up in 56756ea.

enginrect:
Thanks for the review @kayrus! Pushed the changes in 56756ea:

- Dropped the duplicate `clusterUID` field from the `OpenStack` struct
  (cohesion fix; thanks for catching this).
- Moved the `kube-system` UID lookup into the `LoadBalancer()` factory
  as a lazy fetch, so it lives in the LB code path only.
- Removed the unused `eventLBStolen` constant — explained in the inline
  thread.

`go test ./pkg/openstack/...` is green locally and the existing checks
(Lint Charts, EasyCLA) are still passing.

enginrect:
The check job was tripping on SA1019 because the new `TestFetchClusterUID` was using the deprecated `fake.NewSimpleClientset`. Pushed bdabd79 swapping it for `fake.NewClientset` (identical signature, same behaviour for our tests).

Verified locally:
- `go test ./pkg/openstack/...` → pass
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.3.1 run --timeout=20m ./...` → 0 issues

enginrect:
Rebased on top of current master (which now includes `release 1.36.0` and `Bump github.com/moby/spdystream`).

The only conflict was in `charts/openstack-cloud-controller-manager/Chart.yaml`: master bumped the chart version to `2.36.0` for the 1.36.0 release, while our earlier commit had bumped it from `2.35.0` to `2.35.1`. Resolved by setting the chart version to `2.36.1` (patch bump on top of the new baseline). No conflicts in any Go source files — the feature code, tests, and docs rebased cleanly.

Verified locally before pushing:

```
$ go build ./...                     # OK
$ go vet ./pkg/openstack/...         # OK
$ go test ./pkg/openstack/... -count=1
ok  k8s.io/cloud-provider-openstack/pkg/openstack    0.462s

$ go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.3.1 run --timeout=20m ./...
0 issues.
```

All 8 changed files are unchanged in scope from before the rebase — no new code was introduced beyond the version number update. Conflict marker scan (`<<<<<<`, `>>>>>>`, `=======`) across all changed files: **0 matches**.

@kayrus — all checks were green before the rebase, and local verification is clean. Would appreciate `/lgtm` when you have a moment. Thanks!

JonnyHaystack:
I'm affected by this issue, so I'd like to see this PR merged (assuming it does as described), but I'm not sure what the AI/LLM policy is on this repo and I noticed the replies from @enginrect in this PR are quite obviously LLM generated. Looking at their profile their entire bio is clearly written by LLM too, so one begins to wonder if there's even a human actively controlling this account.

I'd have hoped people would state up front if they're using AI for a contribution to an open-source project, and to what extent they are doing so... This appears on the surface like an agent is handling a lot of this without human interaction, which I would personally say warrants extra scrutiny.

enginrect:
@JonnyHaystack 
Thanks for raising this concern.

I want to clarify that this PR comes from a real issue I personally experienced while operating Kubernetes on top of OpenStack using OCCM. The problem described in the issue affected my actual environment, and the patch reflects my own intent and understanding of the behavior I wanted to fix.

Firstly, **(True)I did use Cursor assistance for some parts of the coding and for drafting responses.** This is my first contribution here, so I may also have been overly detailed or overly polished in my replies. Since I am Korean and English is not my first language, I also used AI assistance to make sure my communication was clear and accurate.

I am still fully responsible for the change, the design decision, the final code, and the discussion in this PR. This is not an autonomous AI-generated contribution.

If maintainers(of this repo) have a specific policy, disclosure requirement, or extra review/security process around AI-assisted contributions, I will follow it and rework on this.

JonnyHaystack:
Thanks for making that clear :slightly_smiling_face: hoping to see this merged!

enginrect:
@kayrus 
I've rebased onto current master again. 
BTW, the conflict was with #3149 (LB/listener/pool tags annotations, merged yesterday). Resolved it, so both features coexist.

All checks are green. PTAL when you have time.

pkg/openstack/loadbalancer.go
```diff
@@ -1894,9 +1978,13 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 	// save address into the annotation
 	lbaas.updateServiceAnnotation(service, ServiceAnnotationLoadBalancerAddress, addr)
 
-	// Ensure the LB name tag plus any tags from the Service annotation in a single update.
+	// Ensure the LB name tag, the cluster-identity tag, plus any tags from the
+	// Service annotation in a single update.
 	if svcConf.supportLBTags {
 		lbTags := withLBNameTag(lbName, svcConf.lbTags)
+		if tag := clusterIDTag(lbaas.clusterUID); tag != "" && !slices.Contains(lbTags, tag) {
```
kayrus: can you embed if logic directly inside the `clusterIDTag` func?

enginrect: Done in 1bc41189

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
```diff
@@ -641,6 +642,55 @@ $ openstack loadbalancer listener list --loadbalancer 2b224530-9414-4302-8163-5a
...
+where `<cluster-name>` comes from the kube-controller-manager `--cluster-name` flag and
+defaults to `kubernetes`. When two Kubernetes clusters share the same OpenStack project
+and use the same `--cluster-name`, namespace and service name, the resulting load balancer
+names collide. OpenStack does not require load balancer names to be unique, so previously
+the second cluster could "adopt" the load balancer that the first cluster owned and start
+overwriting its configuration. The recommended way to avoid this remains to set a unique
+`--cluster-name` on every Kubernetes cluster (see issues
+[#2241](https://github.com/kubernetes/cloud-provider-openstack/issues/2241),
+[#2571](https://github.com/kubernetes/cloud-provider-openstack/issues/2571),
+[#2624](https://github.com/kubernetes/cloud-provider-openstack/issues/2624)).
```
kayrus: nit: avoid linking to issues in documentation. I found only one exception in the repo: this [line](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md?plain=1#L360) references an OCCM issue because it describes a workaround for an unfixed bug rather than standard behavior. For typical use cases and resolved issues, documentation should stand alone without issue references.

pkg/openstack/loadbalancer.go
```diff
@@ -112,8 +112,30 @@ const (
...
+	// OCCM to disambiguate load balancers when multiple Kubernetes clusters
+	// share the same OpenStack project and a service name happens to collide.
+	clusterIDTagPrefix = "kube_cluster_id_"
 )
 
+// withClusterIDTag appends the Octavia load balancer tag carrying the cluster
+// identifier to tags. It is a no-op when uid is empty (the cluster identity
+// could not be determined) or when the tag is already present, so callers
+// don't need any guard logic of their own.
+func withClusterIDTag(tags []string, uid string) []string {
```
kayrus: The uid is the primary input - it's what drives the logic. Let's put it first.
```suggestion
func withClusterIDTag(uid string, tags []string) []string {
```
And please put this function just after the `withLBNameTag` func.

pkg/openstack/loadbalancer.go
kayrus: I see my earlier comment, but I've forgotten my reasoning for changing my mind. Let's add a comment explaining why we're not using API-level tag filtering here. Primarily because we need to log warnings for unmatched tags and for transition period.

That said, I think we should still consider server-side filtering to reduce load on the Octavia API. Do you have any suggestions or concerns about this approach?

pkg/openstack/loadbalancer.go
```diff
@@ -163,8 +185,17 @@ type listenerKey struct {
...
+//
+// When clusterUID is non-empty, the returned load balancer must either carry
+// the matching clusterIDTagPrefix tag for that UID, or carry no clusterIDTagPrefix
+// tag at all (legacy load balancer that pre-dates the tag). Load balancers
+// whose name matches but that carry a different cluster-id tag belong to
+// another Kubernetes cluster sharing the OpenStack project; they are ignored
+// and the lookup returns ErrNotFound, which causes OCCM to create a new load
+// balancer instead of accidentally adopting (and overwriting) one that is
+// owned by a different cluster.
+func getLoadbalancerByName(ctx context.Context, client *gophercloud.ServiceClient, name string, legacyName string, clusterUID string) (*loadbalancers.LoadBalancer, error) {
```
kayrus: I think it makes more sense to convert this function to a `LoadBalancer` method. That would simplify calls like `getLoadbalancerByName(ctx, lbaas.lb, name, legacyName, lbaas.clusterUID)` to just `lbaas.getLoadbalancerByName(ctx, name, legacyName)`. Also move this method close to other `LoadBalancer` methods.

kayrus (commented):
Thanks for the patience. See my comments.

UPD: update release note in the initial PR description with less text

kayrus:
Just a heads-up: our current e2e test coverage is poor, and I'd like to merge #3153 before this PR. Could you write e2e tests covering the clusterID tag functionality based on #3153 and submit them as a follow-up PR? Let me know if this works for you.

Add following e2e tests:
* loadbalancer exists, but doesn't have a `kube_cluster_id_XXX` tag -> OCCM reconciliation must add a missing tag
* loadbalancer exists, but has multiple `kube_cluster_id_XXX` tags -> OCCM must not reconcile a loadbalancer and raise an error in events
* prevent setting the user-define reserved tag prefixes to loadbalancer
* and add any other tests you think are worth covering

pkg/openstack/loadbalancer.go
kayrus, line 193: add a `clusterIDTagPrefix` prefix as an exception for reserved tags

enginrect: I think server-side filtering is hard to adopt at this point, mainly because of the legacy LBs. The Octavia tag filter can only express "has tag X", so I don't think a single call can return "LBs tagged with our cluster id, plus LBs with no cluster-id tag at all". To keep the legacy case working we would need a second unfiltered list call as a fallback, and in the common case that doubles the API requests instead of reducing them. Another concern is that we would stop seeing foreign-tagged LBs entirely, so the warning log for the collision case couldn't be emitted.

That said, I think once the tag has been deployed widely enough that the legacy fallback can be dropped, switching to `ListOpts.Tags` would be a good follow-up optimization. I added a comment in the code explaining this reasoning.

pkg/openstack/loadbalancer.go
enginrect: Done in 1bc41189

pkg/openstack/loadbalancer.go
enginrect: Done in b7c487e8

pkg/openstack/loadbalancer.go
enginrect: Done in b7c487e8

pkg/openstack/loadbalancer.go
enginrect: Done in b7c487e8

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
enginrect: Done in 2e2348c7

enginrect:
@kayrus sry, i've missed your comment with #3153.
BTW, works for me, I'll wait for #3153. I've already prototyped the e2e tests locally on top of #3153 and they pass a dry-run.

One thing to clarify: by "follow-up PR" do you mean a separate new PR after this one is merged, or should I rebase this PR on #3153 and add the e2e tests here?

enginrect:
@kayrus 

I've rebased onto master, conflicts were trivial.

Also did case 2 from your e2e list - an LB tagged for this cluster and another one at the same time is now refused with a Warning event, instead of being adopted.

#3153 hasn't moved in a while, so I went ahead with the behavior change here rather than waiting. The e2e tests can still follow once #3153 lands. Is that ok with you?

pkg/openstack/loadbalancer.go
```diff
@@ -1805,7 +1939,7 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 		}
 	} else {
 		legacyName := lbaas.getLoadBalancerLegacyName(service)
-		loadbalancer, err = getLoadbalancerByName(ctx, lbaas.lb, lbName, legacyName)
+		loadbalancer, err = lbaas.getLoadbalancerByName(ctx, service, lbName, legacyName)
```
kayrus: you can merge these two calls into a single method. what do you think?

```go
		legacyName := lbaas.getLoadBalancerLegacyName(service)
		loadbalancer, err = lbaas.getLoadbalancerByName(ctx, service, lbName, legacyName)
```

->

```go
		loadbalancer, err = lbaas.getLoadbalancerByName(ctx, service, lbName)
```

```go
func (lbaas *LbaasV2) getLoadbalancerByName(ctx context.Context, service *corev1.Service, name string) (*loadbalancers.LoadBalancer, error) {
		legacyName := lbaas.getLoadBalancerLegacyName(service)
...
```

enginrect: Done. My bad, I missed that.

Kept the `len(legacyName) > 0` guard though - an empty name would list every LB.

pkg/openstack/loadbalancer.go
```diff
@@ -422,6 +485,78 @@ func (lbaas *LbaasV2) getLoadBalancerLegacyName(service *corev1.Service) string
...
+			opts := loadbalancers.ListOpts{
+				Name: legacyName,
+			}
+			allLoadbalancers, err = openstackutil.GetLoadBalancers(ctx, lbaas.lb, opts)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			return nil, cpoerrors.ErrNotFound
+		}
```
kayrus, line 531: this block can look better with
```suggestion
		if len(legacyName) == 0 {
			return nil, cpoerrors.ErrNotFound
		}
		// Backoff to get load balnacer by legacy name.
		opts := loadbalancers.ListOpts{
			Name: legacyName,
		}
		allLoadbalancers, err = openstackutil.GetLoadBalancers(ctx, lbaas.lb, opts)
		if err != nil {
			return nil, err
		}
```

enginrect: Done.

pkg/openstack/loadbalancer.go
```diff
@@ -244,6 +211,103 @@ func withLBNameTag(lbName, annotation string) []string {
...
+//     all load balancers are dropped. The third return value is true in that
+//     case to let the caller emit a more specific log line / event.
+//
+// Several cluster-id tags none of which matches clusterUID make a load balancer
+// foreign, not ambiguous: this cluster has no claim to it, so it is dropped
+// like any other foreign load balancer and a new one is created instead.
+func filterLoadBalancersByClusterID(lbs []loadbalancers.LoadBalancer, clusterUID string) ([]loadbalancers.LoadBalancer, *loadbalancers.LoadBalancer, bool) {
+	if clusterUID == "" || len(lbs) == 0 {
+		return lbs, nil, false
+	}
```
stephenfin: I think this should be the responsibility of the (sole) caller, not this function. It's fair for function that filters load balancers by cluster ID to insist that it receive a set of load balancers and a cluster ID, IMO. Doing this would simplify our return types also.

pkg/openstack/loadbalancer.go
```diff
@@ -422,6 +485,77 @@ func (lbaas *LbaasV2) getLoadBalancerLegacyName(service *corev1.Service) string
 	return cloudprovider.DefaultLoadBalancerName(service)
 }
 
+// getLoadbalancerByName gets the load balancer which is in valid status by the
```
stephenfin: nit: moving this has made the diff considerably trickier to review 😞

charts/openstack-cloud-controller-manager/templates/clusterrole.yaml
```diff
@@ -86,3 +86,9 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - get
```
kayrus: does it make sense to add:

```yaml
  resourceNames:
    - kube-system
```

since the code gets the exact `kube-system` namespace uuid.

manifests/controller-manager/cloud-controller-manager-roles.yaml
```diff
@@ -99,6 +99,12 @@ items:
     - get
     - list
     - watch
+  - apiGroups:
+    - ""
+    resources:
+    - namespaces
+    verbs:
+    - get
```
kayrus: does it make sense to add:

```yaml
  resourceNames:
    - kube-system
```

since the code gets the exact `kube-system` namespace uuid.

PR #3107, opened 2026-05-05
Add -h / --help option to hack scripts

stephenfin:

Add `-h` / `--help` options to two release-related scripts to help users understand what they are for.

(none)

(none)

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/release-image-digests.sh
+++ b/hack/release-image-digests.sh
@@ -14,11 +14,47 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# example:
-# ./release-image-digests.sh registry.k8s.io/images/k8s-staging-provider-os/images.yaml [v1.35.0] [v1.34.1]
+usage() {
+  cat <<EOF
+Usage: $0 <yaml_file> [<tag>...]
+
+Populate an image promoter manifest with content-digest-to-tag mappings for
+staged container images.
+
+The Kubernetes image promoter promotes images from staging to production using
+a manifest that maps content digests (sha256:...) to tags. This script
+automates the step of looking up those digests from the GCR staging registry
+(gcr.io/k8s-staging-provider-os) after images have been pushed, and writing
+the dmap entries into the manifest YAML.
+
+Run this after cutting a release to update the promoter manifest before opening
+the promotion PR.
+
+Arguments:
+  yaml_file   Path to the image promoter manifest, e.g.
+              registry.k8s.io/images/k8s-staging-provider-os/images.yaml
+  tag         One or more release tags to process, e.g. v1.35.0 v1.34.1.
+              If omitted, all tags already present in the manifest are used.
+
+Examples:
+  $0 registry.k8s.io/images/k8s-staging-provider-os/images.yaml v1.35.0
+  $0 registry.k8s.io/images/k8s-staging-provider-os/images.yaml v1.35.0 v1.34.1
+  $0 registry.k8s.io/images/k8s-staging-provider-os/images.yaml
+EOF
+}
+
+ARGS=$(getopt -o h --long help -n "$0" -- "$@") || { usage >&2; exit 1; }
+eval set -- "$ARGS"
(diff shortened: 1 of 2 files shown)


anguslees (approved):

hack/release-image-digests.sh
```diff
@@ -14,11 +14,47 @@
...
+              If omitted, all tags already present in the manifest are used.
+
+Examples:
+  $0 registry.k8s.io/images/k8s-staging-provider-os/images.yaml v1.35.0
+  $0 registry.k8s.io/images/k8s-staging-provider-os/images.yaml v1.35.0 v1.34.1
+  $0 registry.k8s.io/images/k8s-staging-provider-os/images.yaml
+EOF
+}
+
+ARGS=$(getopt -o h --long help -n "$0" -- "$@") || { usage >&2; exit 1; }
```
winiciusallan: IIRC, OSX machines ships with BSD `getopt`, which does not support the `--long` flag[1]. I tested in a old mac machine that I have here and I've gotten.

```bash
$ bash hack/release-image-digests.sh --help
ERROR: File h does not exist
```

Maybe we can note this somewhere that Mac users can install the [gnu-getopt](https://formulae.brew.sh/formula/gnu-getopt) or use the builtin `getopts` (with s), but without the long flag.

[1] https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/getopt.3.html

hack/verify-image-digests.sh
```diff
@@ -14,14 +14,47 @@
...
+  esac
+done
+
+# if only one arg remains it's the yaml file; two args means match + yaml file
+if [ $# -eq 1 ]; then
+  MATCH='.*'
+  YAML_FILE=$1
+else
+  MATCH=${1:-'.*'}
+  YAML_FILE=${2:?$(usage)}
```
winiciusallan: When not providing the second argument, the message seems very odd. It's just a big oneline with the usage message 😆.

```bash
x bash hack/verify-image-digests.sh
hack/verify-image-digests.sh: line 56: 2: Usage: hack/verify-image-digests.sh [<match>] <yaml_file> Verify that content digests in an image promoter manifest match what is currently in the GCR staging registry and the production registry. For each image/tag pair recorded in the manifest, this script fetches the digest from both gcr.io/k8s-staging-provider-os and registry.k8s.io/provider-os and compares them against the digest stored in the dmap. Any mismatch is reported as an error. Arguments: match A regex to filter tags (e.g. 'v1.33.*'). Defaults to '.*' (all tags). yaml_file Path to the image promoter manifest, e.g. registry.k8s.io/images/k8s-staging-provider-os/images.yaml Examples: hack/verify-image-digests.sh 'v1.33.*' registry.k8s.io/images/k8s-staging-provider-os/images.yaml hack/verify-image-digests.sh registry.k8s.io/images/k8s-staging-provider-os/images.yaml
```

What do you think of?

```suggestion
    MATCH=${1:-'.*'}
    if [ -z "${2-}" ]; then
        usage >&2
        exit 1
    fi
    YAML_FILE=${2}
```

winiciusallan (commented):
A few comments that is worth noting. Besides that, thanks for the simple, but useful work!

hack/release-image-digests.sh
stephenfin: I've changed this to a simple loop.

mandre (commented):
/lgtm

mandre:
Ah, conflicts with https://github.com/kubernetes/cloud-provider-openstack/pull/3116. That will teach you do do well focused PRs :stuck_out_tongue:

stephenfin:
> Ah, conflicts with #3116. That will teach you do do well focused PRs 😛

Fixed.

stephenfin:
I should probably skip the functional jobs for `hack/` changes also?

https://github.com/kubernetes/test-infra/pull/37816

PR #3109, opened 2026-05-06
Add dependabot configuration

stephenfin:

Add dependabot configuration for both `master` and stable release branches. This is taken from CAPO and modified to suit us. We target `release-1.36` and `release-1.34` since `release-1.33` and below are no longer supported.

(n/a)

(none)

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,195 @@
+# Please see the documentation for all configuration options: https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+## main branch config starts here
+# github-actions
+- directory: "/"
+  package-ecosystem: "github-actions"
+  schedule:
+    interval: "monthly"
+    day: "monday"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,228 @@
...
+        patterns: ["*"]
+    cooldown:
+      default-days: 7
+    open-pull-requests-limit: 10
+    labels:
+      - "ok-to-test"
+      - "release-note-none"
+  # Go directories
+  - directories:
+      - "/"
```
mandre: This would be more consistent.
```suggestion
  - directory: "/"
```

Also for each stable branches below.

mandre (commented):
/approve
LGTM, just one minor nit.

.github/dependabot.yml
stephenfin: Indeed. Done.

mandre (commented):
/lgtm

PR #3110, opened 2026-05-07
[magnum-auto-healer] Add TLS cert verification

stephenfin:

When the protocol is HTTPS, we were creating a HTTP transport with `InsecureSkipVerify: true` which causes Go's TLS client to skip all certificate verification. This leaves the connection open to man-in-the-middle attacks.

I suspect things were done this way since the function connects directly to a node's raw IP address, but Kubernetes API server TLS certificates are typically issued for hostnames or DNS names (e.g., `api.cluster.example.com`) rather than raw IPs. Certificate hostname verification would fail unless the certificate includes the node IP as a Subject Alternative Name (SAN), which isn't guaranteed. Additionally, the Kubernetes cluster CA cert that signed the API server's certificate isn't typically included in theOS trust store and must be loaded explicitly. Neither of these issues are insurmountable though, and we should be secure-by-default. 

Fix things by adding a configurable CA file and an explicit opt-in `insecure-skip-verify` flag, defaulting to the in-cluster Kubernetes CA.

(none)

Please note the AI was used to assist in generation of this patch. This is noted in the relevant commit messages.

**Release note**:

```release-note
[magnum-auto-healer] Default to certificate verification with HTTPS

`magnum-auto-healer` now defaults to verifying certificates when using HTTPS
transport. When used, the API server cert will need to use node IPs as SANs.
A new `insecure-skip-verify` option is provided to allow users to opt into the
previous insecure behavior, which can be helpful during upgrades.
```

### DIFF
--- a/docs/magnum-auto-healer/using-magnum-auto-healer.md
+++ b/docs/magnum-auto-healer/using-magnum-auto-healer.md
@@ -183,6 +183,23 @@ spec:
 EOF
 ```
 
+#### Endpoint health check parameters
+
+The `Endpoint` type health check supports the following parameters:
+
+| Parameter | Type | Default | Description |
+|---|---|---|---|
+| `protocol` | string | `HTTPS` | URL scheme to use. `HTTP` or `HTTPS` (case-insensitive). |
+| `port` | int | `6443` | Port to connect to on the node. |
+| `endpoints` | []string | `["/healthz"]` | List of URL paths to check. |
+| `ok-codes` | []int | `[200]` | HTTP response codes considered healthy. |
+| `unhealthy-duration` | duration | `300s` | How long a node must be continuously unhealthy before repair is triggered. |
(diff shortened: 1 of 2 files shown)


kayrus:
@stephenfin does it make sense to reuse `pkg/client/client.go` code here?

stephenfin:
> @stephenfin does it make sense to reuse `pkg/client/client.go` code here?

There's some overlap but `pkg/client/client.go` seems tailored towards creating an OpenStack client, with an `auth_url` and the likes, and I don't think we want to require a `clouds.yaml` to use this plugin? I could drag out the building of the `TLSClientConfig` field to helper function but very little of that would actually be used (we don't need equivalents [of `CertFile` and `KeyFile`](https://github.com/kubernetes/cloud-provider-openstack/blob/46904da2e07ff10b1ec626150fe0b4c96fdbf162/pkg/client/client.go#L278-L285), for example) so again, I'm not sure how much this would get us. Have I missed something?

stephenfin:
/close

I ended up duplicating this myself :facepalm: I prefer the option naming there.

https://github.com/kubernetes/cloud-provider-openstack/pull/3170

PR #3115, opened 2026-05-19
[cinder-csi-plugin] reject non-multiattach publish on another node

gnom4ik:

This fixes non-multiattach Cinder volume publish behavior when the volume is already attached to a different node.

Before this change, `AttachVolume` treated an existing attachment as idempotent only when it was already attached to the requested instance. For a non-multiattach volume attached to another instance, the driver still attempted a new Nova volume attach. In some OpenStack environments this can leave the volume with the old `attached` entry and an additional `attaching` entry for the new node.

This change makes the driver return `FailedPrecondition` before calling Nova attach when a non-multiattach volume already has an attachment to another node. `ControllerPublishVolume` now preserves that gRPC status code instead of wrapping it as `Internal`.

None

Tested against a real OpenStack-backed Kubernetes cluster (`training`) with a patched `cinder-csi-plugin` controller image.

Verification scenario:
1. Created a PVC backed by Cinder.
2. Started a pod using the PVC on one worker node and confirmed the volume was attached.
3. Manually created a second `VolumeAttachment` for the same PV targeting another worker node.
4. Confirmed the second publish failed with `FailedPrecondition`.
5. Confirmed OpenStack kept only the original attachment and did not create a second `attaching` attachment.
6. Deleted the test pod/PVC and confirmed the volume detached and was cleaned up.

Local tests:
- `go test ./pkg/csi/cinder/openstack`
- `go test ./pkg/csi/cinder -run 'TestControllerPublishVolume|TestControllerPublishVolumePreservesFailedPrecondition'`

Note: `go test ./pkg/csi/cinder/...` was not fully runnable on macOS arm64 because an existing node expansion test depends on block-device functionality that is not implemented for Darwin.

**Release note**:

```release-note
[cinder-csi-plugin] Return FailedPrecondition when publishing a non-multiattach Cinder volume that is already attached to another node.
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -332,6 +332,9 @@ func (cs *controllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 	_, err = cloud.AttachVolume(ctx, instanceID, volumeID)
 	if err != nil {
 		klog.Errorf("Failed to AttachVolume: %v", err)
+		if s, ok := status.FromError(err); ok {
+			return nil, status.Error(s.Code(), fmt.Sprintf("[ControllerPublishVolume] Attach Volume failed with error %v", err))
+		}
 		return nil, status.Errorf(codes.Internal, "[ControllerPublishVolume] Attach Volume failed with error %v", err)
 
 	}
--- a/pkg/csi/cinder/controllerserver_test.go
+++ b/pkg/csi/cinder/controllerserver_test.go
@@ -564,6 +564,28 @@ func TestControllerPublishVolume(t *testing.T) {
 	assert.Equal(expectedRes, actualRes)
 }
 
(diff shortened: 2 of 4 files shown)


gnom4ik:
This PR has local unit test coverage and was also verified against a real OpenStack-backed Kubernetes cluster.

Tested:
- go test ./pkg/csi/cinder/openstack
- go test ./pkg/csi/cinder -run 'TestControllerPublishVolume|TestControllerPublishVolumePreservesFailedPrecondition'

Manual verification:
- Created a Cinder-backed PVC.
- Attached it to one worker through a pod.
- Created a second VolumeAttachment for the same PV targeting another worker.
- Confirmed ControllerPublishVolume returned FailedPrecondition.
- Confirmed OpenStack kept only the original attachment and did not create a second attaching attachment.

Could a Kubernetes org member please run /ok-to-test?

PR #3117, opened 2026-05-19
Add zizmor security scanner

stephenfin:

Run zizmor on pushes to master and pull requests that modify workflow files. Results are uploaded as SARIF to GitHub's security-events API for code scanning integration.

(none)

This was previously done for Gophercloud in https://github.com/gophercloud/gophercloud/pull/3659

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -16,6 +16,7 @@ jobs:
         uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744
         with:
           fetch-depth: 0
+          persist-credentials: false
 
       - name: Configure Git
         run: |
--- a/.github/workflows/zizmor.yaml
+++ b/.github/workflows/zizmor.yaml
@@ -1,0 +1,28 @@
+name: zizmor
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - '.github/**'
+  pull_request:
+    paths:
+      - '.github/**'
+
+permissions: {}
+
+jobs:
+  zizmor:
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+      contents: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@b1d7e1fb5de872772f31590499237e7cce841e8e # v0.5.3
--- a/.github/workflows/zizmor.yaml
+++ b/.github/workflows/zizmor.yaml
@@ -20,9 +20,9 @@ jobs:
       contents: read
     steps:
       - name: Checkout repository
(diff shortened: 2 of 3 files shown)


.github/workflows/zizmor.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+jobs:
+  zizmor:
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+      contents: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
winiciusallan: while we're here, what do you thinking of bumping this action to [v7.0.1](https://github.com/actions/checkout/releases/tag/v7.0.1)?

.github/workflows/zizmor.yaml
```diff
@@ -0,0 +1,28 @@
...
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@b1d7e1fb5de872772f31590499237e7cce841e8e # v0.5.3
```
winiciusallan: and the zizmor-action to [v0.6.3](https://github.com/zizmorcore/zizmor-action/releases/tag/v0.6.3) :)

stephenfin: Done

.github/workflows/zizmor.yaml
stephenfin: Done. Note that I have [another PR to add dependabot](https://github.com/kubernetes/cloud-provider-openstack/pull/3109). It needs refreshing but it will make this automatic in the future.

winiciusallan (approved):

winiciusallan:
wow... the error looks like a race condition when starting the server and calling `Stop()`. The test polls the connection with the server, but the server is assigned only after `net.Listen(...)` :thinking:.

https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/server_test.go#L29-L47
https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/csi/cinder/server.go#L86-L95

stephenfin:

Will investigate the race separately.

.github/workflows/release.yaml
```diff
@@ -16,6 +16,7 @@ jobs:
         uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744
         with:
           fetch-depth: 0
+          persist-credentials: false
```
mandre, line 19: Are we sure this isn't going to break the `helm/chart-releaser-action` below if it needs creds to push to the `gh-pages` branch?

stephenfin: I checked and no, it doesn't. Setting this means subsequent `git` calls won't have the token available, but `helm/chart-releaser-action` expects the explicit `CR_TOKEN` which is present regardless. I think we're doing everything correctly here.

mandre (commented):
/lgtm
/approve

PR #3124, opened 2026-06-05
[cinder-csi-plugin] Wait for volume availability before attach

hemna:

`ControllerPublishVolume` now waits for the volume to reach `available` or `in-use` status before calling the Cinder attachment API.

Previously, if the CO called `ControllerPublishVolume` immediately after `CreateVolume`, the volume could still be in `creating` state on the backend. This caused Cinder to reject the attachment with a 409 Conflict:

> Volume X status must be available or downloading to reserve, but the current status is creating.

This forced the CO (external-attacher) to retry blindly, generating unnecessary API calls to Cinder and delaying volume attachment. The issue is most pronounced with storage backends where volume creation is asynchronous and takes several seconds (e.g., when volumes are backed by network-attached storage).

**How the fix works:**

A new `WaitVolumeTargetStatusWithContext` method uses `wait.PollUntilContextCancel` which:
1. Polls the volume status every 3 seconds
2. Returns immediately if the volume is already `available` or `in-use`
3. Returns an error if the volume enters an error state
4. Respects the gRPC context deadline (no fixed step count — bounded by the CO's RPC timeout)

If the context expires before the volume is ready, the driver returns `FAILED_PRECONDITION` which tells the CO to retry with exponential backoff.

fixes #

The `volumeReadyPollInterval` is set to 3 seconds, which balances responsiveness against API load on the Cinder service. The total wait time is bounded by the gRPC context deadline (typically 15-30s depending on the external-attacher configuration), not a fixed backoff.

The existing `WaitVolumeTargetStatus` (used in `ControllerExpandVolume`) is left unchanged to avoid disrupting existing behavior.

**Release note**:
```release-note
[cinder-csi-plugin] ControllerPublishVolume now waits for volumes to become available before calling the Cinder attach API, eliminating 409 Conflict errors when volumes are still being provisioned.
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -34,6 +34,7 @@ import (
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
+	"k8s.io/apimachinery/pkg/util/wait"
 	sharedcsi "k8s.io/cloud-provider-openstack/pkg/csi"
 	"k8s.io/cloud-provider-openstack/pkg/csi/cinder/openstack"
 	"k8s.io/cloud-provider-openstack/pkg/util"
@@ -313,12 +314,42 @@ func (cs *controllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 		return nil, status.Error(codes.InvalidArgument, "[ControllerPublishVolume] Volume capability must be provided")
 	}
 
-	_, err := cloud.GetVolume(ctx, volumeID)
+	// Wait for the volume to reach a state valid for attachment.
+	// Cinder rejects attachment requests for volumes not in "available"
+	// (or "in-use" for multiattach) state. Without this wait, the driver
+	// would immediately hit a 409 Conflict if CreateVolume has just been
+	// called and the volume is still being provisioned on the backend.
+	vol, err := cs.waitForVolumeAttachable(ctx, cloud, volumeID)
 	if err != nil {
 		if cpoerrors.IsNotFound(err) {
 			return nil, status.Errorf(codes.NotFound, "[ControllerPublishVolume] Volume %s not found", volumeID)
 		}
-		return nil, status.Errorf(codes.Internal, "[ControllerPublishVolume] get volume failed with error %v", err)
+		klog.Errorf("Failed waiting for volume %s to become attachable: %v", volumeID, err)
+		return nil, status.Errorf(codes.FailedPrecondition, "[ControllerPublishVolume] Volume %s is not available for attach: %v", volumeID, err)
+	}
+
+	// Check if the volume is already attached to this instance (idempotent).
+	// CSI spec requires that a repeated ControllerPublishVolume for the same
+	// volume+node returns 0 OK (see CSI v1.12 spec §ControllerPublishVolume).
+	// This uses the volume object returned by waitForVolumeAttachable to
+	// avoid a redundant volumes.Get API call.
+	for _, att := range vol.Attachments {
+		if instanceID == att.ServerID {
+			klog.V(4).Infof("Volume %s is already attached to instance %s", volumeID, instanceID)
+			devicePath, err := cloud.GetAttachmentDiskPath(ctx, instanceID, volumeID)
+			if err != nil {
+				klog.Errorf("Failed to GetAttachmentDiskPath: %v", err)
+				return nil, status.Errorf(codes.Internal, "[ControllerPublishVolume] failed to get device path of attached volume: %v", err)
+			}
+			pvInfo := map[string]string{"DevicePath": devicePath}
+			return &csi.ControllerPublishVolumeResponse{PublishContext: pvInfo}, nil
+		}
+	}
+
+	// If the volume is in-use and not multiattach-capable, it is attached to
+	// a different node (same-node was handled above). Fail immediately.
+	if vol.Status == openstack.VolumeInUseStatus && !vol.Multiattach {
+		return nil, status.Errorf(codes.FailedPrecondition, "[ControllerPublishVolume] Volume %s is already attached to a different node and does not support multiattach", volumeID)
 	}
 
 	_, err = cloud.GetInstanceByID(ctx, instanceID)
@@ -357,6 +388,45 @@ func (cs *controllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 	return &csi.ControllerPublishVolumeResponse{
 		PublishContext: pvInfo,
 	}, nil
+}
+
+// waitForVolumeAttachable polls the volume status until it leaves transitional
+// states (e.g. "creating", "downloading") and reaches "available" or "in-use".
+// For "in-use" volumes, the caller is responsible for checking whether the
+// attachment is an idempotent same-node retry (CSI spec requires 0 OK) or a
+// conflict with a different node.
+// The wait is bounded by the context deadline (typically the gRPC RPC timeout).
+func (cs *controllerServer) waitForVolumeAttachable(ctx context.Context, cloud openstack.IOpenStack, volumeID string) (*volumes.Volume, error) {
+	bsOpts := cloud.GetBlockStorageOpts()
+
+	var lastVol *volumes.Volume
+	err := wait.PollUntilContextCancel(ctx, bsOpts.VolumeStatusPollInterval(), true, func(ctx context.Context) (bool, error) {
+		vol, err := cloud.GetVolume(ctx, volumeID)
+		if err != nil {
+			return false, err
+		}
+		lastVol = vol
+
+		if vol.Status == openstack.VolumeAvailableStatus {
+			return true, nil
+		}
+		if vol.Status == openstack.VolumeInUseStatus {
+			// Return the volume as-is. The caller will inspect
+			// attachments for idempotency before rejecting.
+			return true, nil
+		}
+
+		if slices.Contains(openstack.VolumeErrorStates[:], vol.Status) {
+			return false, fmt.Errorf("volume %s is in error state: %s", volumeID, vol.Status)
+		}
+		return false, nil
+	})
+
+	if err != nil && ctx.Err() != nil {
+		return lastVol, fmt.Errorf("timeout waiting for volume %s to become available: %w", volumeID, err)
+	}
+
+	return lastVol, err
 }
 
 func (cs *controllerServer) ControllerUnpublishVolume(ctx context.Context, req *csi.ControllerUnpublishVolumeRequest) (*csi.ControllerUnpublishVolumeResponse, error) {
--- a/pkg/csi/cinder/openstack/openstack.go
+++ b/pkg/csi/cinder/openstack/openstack.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"os"
 	"sync"
+	"time"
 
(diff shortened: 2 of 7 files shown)


hemna:
The CI failure seems unrelated to my patch.  k8s is killing the CI job.

pkg/csi/cinder/controllerserver.go
```diff
@@ -51,6 +52,10 @@ const (
 	cinderCSIClusterIDKey = "cinder.csi.openstack.org/cluster"
 	affinityKey           = "cinder.csi.openstack.org/affinity"
 	antiAffinityKey       = "cinder.csi.openstack.org/anti-affinity"
+
+	// volumeReadyPollInterval is the interval at which the volume status is
+	// polled when waiting for a volume to become available before attaching.
+	volumeReadyPollInterval = 3 * time.Second
```
kayrus: does it make sense to make this variable configurable?

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -298,6 +298,37 @@ func (os *OpenStack) WaitVolumeTargetStatus(ctx context.Context, volumeID string
 	return waitErr
 }
 
+// WaitVolumeTargetStatusWithContext waits for volume to be in target state,
+// respecting the context deadline. This is useful when the caller has a
+// context with a deadline (e.g., gRPC request context) and wants the wait
+// to be bounded by that deadline rather than a fixed number of steps.
+// The interval parameter controls how often the volume status is polled.
+func (os *OpenStack) WaitVolumeTargetStatusWithContext(ctx context.Context, volumeID string, tStatus []string, interval time.Duration) error {
```
kayrus: this method can be extended to return volume object, or even better, adjust the `GetInstanceByID` method to wait for the volume status. Currently the logic calls `WaitVolumeTargetStatusWithContext`, then `GetInstanceByID`, which does not make sense.

pkg/csi/cinder/controllerserver.go
```diff
@@ -321,6 +326,18 @@ func (cs *controllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 		return nil, status.Errorf(codes.Internal, "[ControllerPublishVolume] get volume failed with error %v", err)
 	}
 
+	// Wait for the volume to be available before attempting attach.
+	// Cinder rejects attachment requests for volumes not in "available" (or
+	// "in-use" for multiattach) state. Without this wait, the driver would
+	// immediately hit a 409 Conflict if CreateVolume has just been called and
+	// the volume is still being provisioned on the backend.
+	targetStatus := []string{openstack.VolumeAvailableStatus, openstack.VolumeInUseStatus}
```
kayrus: these target statuses should depend on multiattach feature turned on or off. in some cases, when cinder has issues and the volume is already in use, the CSI plugin will try to attach an in-use volume and get 409.

kayrus (commented):
thanks for the PR, see my notes

pkg/csi/cinder/controllerserver.go
hemna: added volume-status-poll-delay (in seconds) to BlockStorageOpts, defaulting to 3 seconds via VolumeStatusPollInterval().

pkg/csi/cinder/openstack/openstack_volumes.go
hemna: Adjusted to only available is accepted for non-multiattach volumes. If a non-multiattach volume is already in-use, it fails immediately rather than waiting. Multiattach volumes accept both available and in-use.

pkg/csi/cinder/controllerserver.go
```diff
@@ -359,6 +368,49 @@ func (cs *controllerServer) ControllerPublishVolume(ctx context.Context, req *cs
...
+			return true, nil
+		}
+
+		// If a non-multiattach volume is already in-use, it cannot be
+		// attached again - fail immediately rather than waiting.
+		if !vol.Multiattach && vol.Status == openstack.VolumeInUseStatus {
+			return false, fmt.Errorf("volume %s is already in-use and does not support multiattach", volumeID)
+		}
+
+		for _, eState := range openstack.VolumeErrorStates {
```
kayrus: it makes sense to use the https://pkg.go.dev/slices#Contains

pkg/csi/cinder/controllerserver.go
```diff
@@ -359,6 +368,49 @@ func (cs *controllerServer) ControllerPublishVolume(ctx context.Context, req *cs
...
+		}
+		if vol.Multiattach && vol.Status == openstack.VolumeInUseStatus {
+			return true, nil
+		}
+
+		// If a non-multiattach volume is already in-use, it cannot be
+		// attached again - fail immediately rather than waiting.
+		if !vol.Multiattach && vol.Status == openstack.VolumeInUseStatus {
+			return false, fmt.Errorf("volume %s is already in-use and does not support multiattach", volumeID)
+		}
```
kayrus: I think this approach would be cleaner
```suggestion
		if vol.Status == openstack.VolumeInUseStatus {
		    if vol.Multiattach {
			  return true, nil
			}
			// If a non-multiattach volume is already in-use, it cannot be
		    // attached again - fail immediately rather than waiting
			return false, fmt.Errorf("volume %s is already in-use and does not support multiattach", volumeID)
		}
```

pkg/csi/cinder/controllerserver.go
```diff
@@ -313,14 +314,22 @@ func (cs *controllerServer) ControllerPublishVolume(ctx context.Context, req *cs
...
 			return nil, status.Errorf(codes.NotFound, "[ControllerPublishVolume] Volume %s not found", volumeID)
 		}
-		return nil, status.Errorf(codes.Internal, "[ControllerPublishVolume] get volume failed with error %v", err)
+		klog.Errorf("Failed waiting for volume %s to become attachable: %v", volumeID, err)
+		return nil, status.Errorf(codes.FailedPrecondition, "[ControllerPublishVolume] Volume %s is not available for attach: %v", volumeID, err)
 	}
 
+	_ = volume // volume is available for future use (e.g. multiattach validation)
+
 	_, err = cloud.GetInstanceByID(ctx, instanceID)
```
kayrus, line 355: `volumes.Get` API call still happens here in `GetInstanceByID` after the `waitForVolumeAttachable`. Does it make sense to merge these two methods?

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -283,7 +283,7 @@ func (os *OpenStack) WaitVolumeTargetStatus(ctx context.Context, volumeID string
 				return true, nil
 			}
 		}
-		for _, eState := range volumeErrorStates {
+		for _, eState := range VolumeErrorStates {
```
kayrus: it makes sense to use the https://pkg.go.dev/slices#Contains

pkg/csi/cinder/openstack/openstack.go
```diff
@@ -91,6 +92,18 @@ type BlockStorageOpts struct {
 	RescanOnResize           bool  `gcfg:"rescan-on-resize"`
 	IgnoreVolumeAZ           bool  `gcfg:"ignore-volume-az"`
 	IgnoreVolumeMicroversion bool  `gcfg:"ignore-volume-microversion"`
+	VolumeStatusPollDelay    int   `gcfg:"volume-status-poll-delay"`
```
kayrus: the documentation must be updated accordingly with a supported time format example

hemna:
> `volumes.Get` API call still happens here in `GetInstanceByID` after the `waitForVolumeAttachable`. Does it make sense to merge these two methods?

`GetInstanceByID` does not call `volumes.Get` — it calls `servers.Get` (Nova compute API) to verify the target instance exists. These are checking two different resources: the volume readiness vs. the compute node existence.

The actual redundant `volumes.Get` call is inside `AttachVolume` (openstack_volumes.go line 209), which re-fetches the volume to check if it's already attached (idempotency) and to read the `Multiattach` flag for compute API microversion selection. Eliminating that redundancy would require changing `AttachVolume`'s signature to accept a `*volumes.Volume` parameter, which changes the `IOpenStack` interface and all its implementations (mock, fakecloud, etc.) — that seems out of scope for this bugfix PR but could be a follow-up optimization.

kayrus:
/lgtm

@zetaab @stephenfin can you review as well?

pkg/csi/cinder/controllerserver.go
```diff
@@ -359,6 +382,46 @@ func (cs *controllerServer) ControllerPublishVolume(ctx context.Context, req *cs
...
+		if vol.Status == openstack.VolumeAvailableStatus {
+			return true, nil
+		}
+		if vol.Status == openstack.VolumeInUseStatus {
+			if vol.Multiattach {
+				return true, nil
+			}
+			// If a non-multiattach volume is already in-use, it cannot be
+			// attached again - fail immediately rather than waiting.
+			return false, fmt.Errorf("volume %s is already in-use and does not support multiattach", volumeID)
```
234u34k: Could this path account for an idempotent retry to the same node before rejecting an `in-use`, non-multiattach volume? If the initial attachment succeeds but the RPC response is lost, repeating the same `ControllerPublishVolume` request after Cinder reports `in-use` returns `FailedPrecondition` here, before the caller can inspect `vol.Attachments`.

A minimal reproduction is:

1. Attach a non-multiattach volume to node A using `ControllerPublishVolume` with a single-node capability.
2. Repeat the identical request for node A after the volume status becomes `in-use`.
3. `waitForVolumeAttachable` returns an error at this line, so the same-node attachment check is not reached.

CSI v1.12 states that `ControllerPublishVolume` must be idempotent and that an already-published volume must receive `0 OK` when the node, volume capability, and readonly value match:

https://github.com/container-storage-interface/spec/blob/v1.12.0/spec.md#L1331-L1333

Would it be possible to distinguish that same-node retry from an attachment to a different node before returning this error? Regression cases for node A returning success and node B returning `FailedPrecondition` would cover both paths.

234u34k (commented):
I found one idempotency edge case in the updated `ControllerPublishVolume` path. Details and suggested regression coverage are inline.

hemna:
Good catch — you're right that this broke CSI v1.12 idempotency for the same-node retry case.

Fixed in the latest force-push. The change:

1. `waitForVolumeAttachable` no longer rejects non-multiattach in-use volumes. It returns the volume to the caller whenever status reaches `available` or `in-use`, regardless of multiattach.

2. In `ControllerPublishVolume`, the idempotency check (`for _, att := range vol.Attachments`) now runs *before* any in-use rejection. If the volume is already attached to the requesting node → return success (0 OK). Only after that do we check if the volume is in-use on a *different* node for non-multiattach volumes → `FailedPrecondition`.

3. Added regression tests for both paths:
   - `TestControllerPublishVolumeIdempotentSameNode` — volume in-use, attached to same node → success
   - `TestControllerPublishVolumeInUseDifferentNode` — volume in-use, attached to different node, non-multiattach → FailedPrecondition

234u34k:
Confirmed on `a3bcb73f`. The same-node check now runs before the different-node `FailedPrecondition`, and the tests cover both cases. This addresses my comment. Thanks for the fix.

kayrus (commented):
/lgtm
/approve
/hold

feel free to /unhold when you think this PR is ready

PR #3125, opened 2026-06-08
[manila-csi-plugin] Retry proxied probe on Unavailable

### LINKED ISSUE #3111: [manila-csi-plugin] Node plugin fatally exits on startup when proxied CSI driver socket is not yet ready
**What happened**:

The Manila CSI node plugin crashes immediately on startup if its dependency, the proxied CSI driver socket (e.g., the NFS CSI plugin at `/var/lib/kubelet/plugins/csi-nfsplugin/csi.sock`), is not yet available. There is no retry logic for transient connection errors, causing a fatal exit instead of a graceful retry.

The Manila CSI node plugin should retry connecting to the proxied CSI driver endpoint when it encounters transient connection errors (`codes.Unavailable`), similar to how it already retries on timeout errors (`codes.DeadlineExceeded`).

**Root Cause Analysis**:

In [`pkg/csi/manila/driver.go` `initProxiedDriver()`](https://github.com/kubernetes/cloud-provider-openstack/blob/46904da2e07ff10b1ec626150fe0b4c96fdbf162/pkg/csi/manila/driver.go#L265-L294), the driver creates a 15-second context and calls `ProbeForever` to probe the proxied CSI plugin socket. However, `ProbeForever` (from [`csi-lib-utils/rpc/common.go`](https://github.com/kubernetes-csi/csi-lib-utils/blob/4d601c285c808f9bbe76956ac3fe491d4a774b07/rpc/common.go#L134-L170)) only retries when the gRPC status code is `DeadlineExceeded`. When the socket returns "connection refused" or "no such file or directory", the gRPC status code is `Unavailable`, and `ProbeForever` immediately returns an error:

```go
// csi-lib-utils/rpc/common.go
if st.Code() != codes.DeadlineExceeded {
    return fmt.Errorf("CSI driver probe failed: %s", err)
}
```

This error propagates up through `initProxiedDriver()` → `SetupNodeService()` → `main.go` which calls `klog.Fatalf`, terminating the process. The 15-second retry window is never actually used because only timeouts trigger a retry, not connection refusals.

The full error chain looks like:

```
F0511 02:55:43.317504  1 main.go:112] Driver node service initialization failed:
  failed to initialize proxied CSI driver: probe failed: CSI driver probe failed:
  rpc error: code = Unavailable desc = connection error: desc = "transport: Error
  while dialing: dial unix /var/lib/kubelet/plugins/csi-nfsplugin/csi.sock: connect:
  connection refused"
```

**How to reproduce it**:

1. Deploy the Manila CSI driver with a proxied driver (e.g., NFS CSI plugin) on a Kubernetes/OpenShift cluster
2. Reboot a node (or delete the proxied driver pod so its socket disappears)
3. Both DaemonSet pods restart concurrently on the node
4. The Manila CSI node plugin attempts to connect to the NFS socket before it's ready
5. The Manila plugin fatally exits within ~1 second

A possible fix could be to add retry logic for `codes.Unavailable` errors in `initProxiedDriver()`. The fix can be applied locally in cloud-provider-openstack without changes to `csi-lib-utils`, by wrapping the `ProbeForever` call in a retry loop. For example:

```go
func (d *Driver) initProxiedDriver() (csiNodeCapabilitySet, error) {
	conn, err := d.csiClientBuilder.NewConnection(d.fwdEndpoint)
	if err != nil {
		return nil, fmt.Errorf("connecting to %s endpoint failed: %v", d.fwdEndpoint, err)
	}
	defer conn.Close()

	ctx, cancel := context.WithTimeout(context.Background(), time.Second*15)
	defer cancel()

	identityClient := d.csiClientBuilder.NewIdentityServiceClient(conn)

	// Retry probing on transient connection errors (e.g., Unavailable)
	// since the proxied CSI driver may not be ready yet after a node restart.
	for {
		err = identityClient.ProbeForever(ctx, conn, time.Second*5)
		if err == nil {
			break
		}
		st, ok := status.FromError(err)
		if ok && st.Code() == codes.Unavailable {
			klog.Infof("Proxied CSI driver not yet available, retrying: %v", err)
			select {
			case <-ctx.Done():
				return nil, fmt.Errorf("timed out waiting for proxied CSI driver: %v", err)
			case <-time.After(time.Second):
				continue
			}
		}
		return nil, fmt.Errorf("probe failed: %v", err)
	}

	// ... rest of the function unchanged
```

Alternatively, an upstream fix to `csi-lib-utils` `ProbeForever` could be proposed to also retry on `codes.Unavailable`, since this is the expected transient error when a co-located CSI driver hasn't started yet.

carterpewpew:
The Manila CSI node plugin fatally exits on startup when its proxied CSI driver socket (e.g. NFS CSI plugin) is not yet available. After a node reboot, both DaemonSets restart concurrently and the Manila driver crashes because upstream ProbeForever only retries on DeadlineExceeded, not on Unavailable (connection refused). This causes a fatal exit within ~1 second instead of retrying within the existing 15-second timeout window. This PR wraps the ProbeForever call in a retry loop that specifically handles codes.Unavailable, bounded by the existing context timeout.

fixes #3111 

1. Deploy Manila CSI with a proxied NFS driver
2. Reboot a node so both DaemonSets restart concurrently
3. Observe that the Manila node plugin retries instead of fatally exiting

**Release note**:

```release-note
[manila-csi-plugin]  Fix fatal exit on startup when proxied CSI driver socket is not yet ready by retrying on transient Unavailable errors.
```

### DIFF
--- a/pkg/csi/manila/driver.go
+++ b/pkg/csi/manila/driver.go
@@ -29,6 +29,8 @@ import (
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/kubernetes-csi/csi-lib-utils/protosanitizer"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	v1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/cloud-provider-openstack/pkg/csi/manila/csiclient"
 	"k8s.io/cloud-provider-openstack/pkg/csi/manila/manilaclient"
@@ -274,8 +276,21 @@ func (d *Driver) initProxiedDriver() (csiNodeCapabilitySet, error) {
 
 	identityClient := d.csiClientBuilder.NewIdentityServiceClient(conn)
 
-	if err = identityClient.ProbeForever(ctx, conn, time.Second*5); err != nil {
-		return nil, fmt.Errorf("probe failed: %v", err)
+	for {
+		if err = identityClient.ProbeForever(ctx, conn, time.Second*5); err == nil {
+			break
+		}
+		if status.Code(err) != codes.Unavailable {
+			return nil, fmt.Errorf("probe failed: %v", err)
+		}
+		klog.Warningf("proxied CSI driver probe returned Unavailable for %s, retrying: %v", d.fwdEndpoint, err)
+		select {
+		case <-ctx.Done():
+		case <-time.After(time.Second):
+		}
+		if ctx.Err() != nil {
+			return nil, fmt.Errorf("timed out probing proxied CSI driver %s: %v", d.fwdEndpoint, err)
+		}
 	}
 
 	pluginInfo, err := identityClient.GetPluginInfo(ctx)
--- a/pkg/csi/manila/driver_test.go
+++ b/pkg/csi/manila/driver_test.go
@@ -1,0 +1,136 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package manila
+
+import (
+	"context"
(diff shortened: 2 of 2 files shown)


gouthampacha:
/lgtm
/approve

apologies for not getting back to this sooner. This looks good to me

kayrus:
@gouthampacha to be honest I don't like this change. If the pod is crashing, it should be restarted by k8s.

gouthampacha:
That's right, but, the retry is tempered - there's a 15s startup timeout, if the proxied driver pod isn't up by then, the crash will be handled by kubelet as usual. This pattern just avoids the CrashLoopBackOff escalations to 5 minutes, which might be annoying to build tools around. Wdyt?

PR #3126, opened 2026-06-09
[cinder-csi-plugin] Fix volume from snapshot AZ selection

### LINKED ISSUE #3088: [cinder-csi-plugin] VolumeSnapshots not topology aware
/kind bug

**What happened**:
This issue has already been reported in Issue [1945](https://github.com/kubernetes/cloud-provider-openstack/issues/1945) but never got enough attention. But the problem still exists in current versions of the cinder-csi-plugin. So I copycat the original description:

When running a Kubernetes clusters spanning over multiple availability zones and storage is per zone snapshots restore fails if volume isn't created in the same zone as the snapshot exists in. I also noticed that the snapshot and snapshot content doesn't contain any information related to topology (besides the original volume id).

I expect the volume to be created in the same zone as the snapshot exists in every time. Otherwise this feature is more or less useless for backup / restore scenarios.
In order to do this I assume the snapshot somehow needs a topology reference.

**How to reproduce it**:
0. I assume cinder-csi and external-snapshotter are installed. Openstack has three availability zones with zone agnostic storage.
 
1. create StorageClass and VolumeSnapshotClass
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
  name: csi-cinder-sc-delete-wait
reclaimPolicy: Delete
volumeBindingMode: WaitForFirstConsumer
allowVolumeExpansion: true
``` 
```
apiVersion: snapshot.storage.k8s.io/v1
deletionPolicy: Delete
driver: cinder.csi.openstack.org
kind: VolumeSnapshotClass
metadata:
  name: cinder-csi-snapclass-force-create
parameters:
  force-create: "true"
```

2. create namespace and pod with pvc
```
---
apiVersion: v1
kind: Namespace
metadata:
  name: test
  labels:
    pod-security.kubernetes.io/enforce: "restricted"
    pod-security.kubernetes.io/enforce-version: "latest"
    pod-security.kubernetes.io/audit: "restricted"
    pod-security.kubernetes.io/warn: "restricted"
---
# First PVC (source)
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: test-pvc-source
  namespace: test
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 2Gi
  storageClassName: csi-cinder-sc-delete-wait
---
apiVersion: v1
kind: Pod
metadata:
  name: test-pod-source
  namespace: test
spec:
  securityContext:
    runAsNonRoot: true
    runAsUser: 1000
    runAsGroup: 1000
    fsGroup: 1000
    seccompProfile:
      type: RuntimeDefault
  automountServiceAccountToken: false
  containers:
    - name: app
      image: busybox:1.36
      command: ["sh", "-c", "while true; do date >> /data/test.log; sleep 5; done"]
      securityContext:
        allowPrivilegeEscalation: false
        readOnlyRootFilesystem: true
        capabilities:
          drop:
            - ALL
      volumeMounts:
        - name: data
          mountPath: /data
  volumes:
    - name: data
      persistentVolumeClaim:
        claimName: test-pvc-source
```

3. create snapshot of persistent volume
```
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshot
metadata:
  name: test-pvc-snap
  namespace: test
spec:
  volumeSnapshotClassName: cinder-csi-snapclass-force-create
  source:
    persistentVolumeClaimName: test-pvc-source
```

4. create second pod with pvc that references the volume snapshot as data source aka restore
```
# Second PVC restored from the snapshot
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: test-pvc-restore
  namespace: test
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 2Gi
  storageClassName: csi-cinder-sc-delete-wait
  dataSource:
    name: test-pvc-snap
    kind: VolumeSnapshot
    apiGroup: snapshot.storage.k8s.io
---
apiVersion: v1
kind: Pod
metadata:
  name: test-pod-restore
  namespace: test
spec:
  securityContext:
    runAsNonRoot: true
    runAsUser: 1000
    runAsGroup: 1000
    fsGroup: 1000
    seccompProfile:
      type: RuntimeDefault
  automountServiceAccountToken: false
  containers:
    - name: app
      image: busybox:1.36
      command: ["sh", "-c", "tail -F /data/test.log || ls -l /data"]
      securityContext:
        allowPrivilegeEscalation: false
        readOnlyRootFilesystem: true
        capabilities:
          drop:
            - ALL
      volumeMounts:
        - name: data
          mountPath: /data
  volumes:
    - name: data
      persistentVolumeClaim:
        claimName: test-pvc-restore
```

5. Observe that pod likely cannot be scheduled because scheduler wants to create the volume in a different az than the snapshot resides
```
62s         Normal    WaitForFirstConsumer   persistentvolumeclaim/test-pvc-restore         waiting for first consumer to be created before binding
13s         Normal    ExternalProvisioning   persistentvolumeclaim/test-pvc-restore         Waiting for a volume to be created either by the external provisioner 'cinder.csi.openstack.org' or manually by the system administrator. If volume creation is delayed, please verify that the provisioner is running and correctly registered.
31s         Normal    Provisioning           persistentvolumeclaim/test-pvc-restore         External provisioner is provisioning volume for claim "test/test-pvc-restore"
30s         Warning   ProvisioningFailed     persistentvolumeclaim/test-pvc-restore         failed to provision volume with StorageClass "csi-cinder-sc-delete-wait": rpc error: code = Internal desc = CreateVolume failed with error Expected HTTP response code [202] when accessing [POST https://volumev3.nbg.nsc.noris.cloud/v3/volumes], but got 400 instead: {"badRequest": {"code": 400, "message": "Invalid input received: Volume must be in the same availability zone as the snapshot"}}
```

Normal provisioning of volumes works all the time, when using `volumeBindingMode: WaitForFirstConsumer` but snapshot restore only works when the scheduler randomly chooses the correct availability zone. 

**Environment**:
    - cinder-csi-plugin:v1.34.1
    - csi-provisioner:v5.3.0
    - csi-attacher:v4.10.0
- OpenStack version: 2025.2
- Others: external-snapshotter: 8.4.0

carterpewpew:

When restoring a volume from a snapshot in a multi-AZ cluster with volumeBindingMode: WaitForFirstConsumer, the topology-derived availability zone can conflict with the snapshot's AZ. Cinder requires the new volume to be in the same AZ as the snapshot, so it rejects the request with HTTP 400: "Volume must be in the same availability zone as the snapshot". This happens because the scheduler picks a node (and therefore an AZ) before the CSI driver gets called, and that AZ gets passed straight to Cinder's volume create without considering what AZ the snapshot lives in. The fix clears the topology derived AZ when creating a volume from a snapshot, letting Cinder auto-select the snapshot's AZ. Explicit availability set via StorageClass parameters is not affected.

fixes #3088

Only the topology-derived AZ path is changed. If the user explicitly sets availability in StorageClass params, that value is passed through unchanged.

Two tests added:
1. TestCreateVolumeFromSnapshotWithTopology - verifies the AZ is cleared when restoring from snapshot with topology
2. TestCreateVolumeFromSnapshotWithExplicitAvailability - verifies explicit availability param is preserved even with snapshot + topology

**Release note**:

```release-note
[cinder-csi-plugin] Fix snapshot restore in multi-AZ clusters with topology-aware provisioning by clearing the topology-derived availability zone and letting Cinder select the snapshot's zone.
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -94,10 +94,12 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	// First check if volAvailability is already specified, if not get preferred from Topology
 	// Required, in case vol AZ is different from node AZ
 	var volAvailability string
+	volAvailabilityFromTopology := false
 	if volParams["availability"] != "" {
 		volAvailability = volParams["availability"]
 	} else if cs.Driver.withTopology && accessibleTopologyReq != nil {
 		volAvailability = sharedcsi.GetAZFromTopology(topologyKey, accessibleTopologyReq)
+		volAvailabilityFromTopology = true
 	}
 
 	// get the PVC annotation
@@ -156,6 +158,10 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		// If the snapshot exists but is not yet available, fail.
 		if err == nil && snap.Status != "available" {
 			return nil, status.Errorf(codes.Unavailable, "VolumeContentSource Snapshot %s is not yet available. status: %s", snapshotID, snap.Status)
+		}
+		if err == nil && volAvailabilityFromTopology {
+			klog.V(4).Infof("CreateVolume: clearing topology-derived availability zone %q for snapshot %s to let Cinder select the correct zone", volAvailability, snapshotID)
+			volAvailability = ""
 		}
 
 		// In case a snapshot is not found
--- a/pkg/csi/cinder/controllerserver_test.go
+++ b/pkg/csi/cinder/controllerserver_test.go
@@ -417,6 +417,103 @@ func TestCreateVolumeFromSnapshot(t *testing.T) {
 	assert.Equal(FakeSnapshotID, actualRes.Volume.ContentSource.GetSnapshot().SnapshotId)
 }
 
+func TestCreateVolumeFromSnapshotWithTopology(t *testing.T) {
+	fakeCs, osmock := fakeControllerServer()
+
+	properties := map[string]string{cinderCSIClusterIDKey: FakeCluster}
+	// Expect empty AZ: the driver should clear the topology-derived AZ
+	// and let Cinder place the volume in the snapshot's AZ.
+	osmock.On("CreateVolume", FakeVolName, mock.AnythingOfType("int"), FakeVolType, "", FakeSnapshotID, "", "", properties).Return(&FakeVolFromSnapshot, nil)
+	osmock.On("GetVolumesByName", FakeVolName).Return(FakeVolListEmpty, nil)
+	osmock.On("GetBlockStorageOpts").Return(openstack.BlockStorageOpts{})
+
+	assert := assert.New(t)
+
+	src := &csi.VolumeContentSource{
+		Type: &csi.VolumeContentSource_Snapshot{
+			Snapshot: &csi.VolumeContentSource_SnapshotSource{
+				SnapshotId: FakeSnapshotID,
+			},
+		},
+	}
+
+	fakeReq := &csi.CreateVolumeRequest{
+		Name: FakeVolName,
+		VolumeCapabilities: []*csi.VolumeCapability{
+			{
+				AccessMode: &csi.VolumeCapability_AccessMode{
+					Mode: csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
+				},
+			},
+		},
+		VolumeContentSource: src,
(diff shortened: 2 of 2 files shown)


linwalth:
This is relevant to our interests.

stephenfin:
I think this likely makes sense. Can you rebase and force push this? I'll happily review it once done.

carterpewpew:
Apologies for the delay, Rebased onto current master. @stephenfin  PTAL

PR #3129, opened 2026-06-15
[cinder-csi-plugin] Validate volume status before attach

hemna:

`AttachVolume` now validates that the volume is in an acceptable state before calling the Nova volume attach API.

- For regular (single-attach) volumes: status must be `available`
- For multi-attach capable volumes: status must be `available` or `in-use`
- `migration_status` must not be `migrating`

Previously, the driver would call `volumeattach.Create` regardless of volume status, relying on Nova/Cinder to reject the request. This produced opaque 409 Conflict errors that made troubleshooting difficult and generated unnecessary API calls for volumes in transitional states (e.g., `creating`, `downloading`, `detaching`).

Additionally, a volume can have status `available` while actively being migrated (`migration_status: migrating`). Attempting to attach in this state would fail with an opaque Cinder error. The driver now checks this explicitly.

With this change the driver returns immediate, descriptive errors when the volume is not attachable, allowing the CO (external-attacher) to retry with exponential backoff via standard CSI error handling.

**How it works:**

A custom `volumeForAttach` struct is used to extract both the standard volume fields and `migration_status` from the Cinder API response in a single API call. This is necessary because gophercloud's `Volume` struct does not map the `migration_status` field.

The validation is placed inside `AttachVolume` (after the existing idempotency check for already-attached volumes). This means:
- No additional API calls are introduced (replaces the existing `GetVolume` call)
- No new interface methods are required
- No changes to mock/fakecloud implementations are needed

**Which issue this PR fixes (if applicable)**:
Related to #3124

This is a simpler alternative to the polling/waiting approach in #3124. Rather than adding a wait-loop with poll intervals and context deadline handling, this PR fails fast with a clear error and lets the CO handle retries — which it already does via exponential backoff.

The `DetachVolume` function already performs similar status validation (lines 307-318 of `openstack_volumes.go`), so this brings `AttachVolume` to parity.

The validation order is:
1. Already attached to same instance? → return early (idempotent)
2. `migration_status == "migrating"`? → reject
3. Volume status acceptable for attach type (single vs multi-attach)? → reject if not

**Release note**:
```release-note
[cinder-csi-plugin] AttachVolume now validates volume status and migration_status before calling the Nova attach API, returning clear errors for volumes not in an attachable state.
```

### DIFF
--- a/pkg/csi/cinder/openstack/openstack_volumes.go
+++ b/pkg/csi/cinder/openstack/openstack_volumes.go
@@ -202,23 +202,50 @@ func (os *OpenStack) GetVolume(ctx context.Context, volumeID string) (*volumes.V
 	return vol, nil
 }
 
+// volumeForAttach contains the volume fields needed for attach validation,
+// including migration_status which gophercloud's Volume struct does not map.
+type volumeForAttach struct {
+	ID              string               `json:"id"`
+	Status          string               `json:"status"`
+	Multiattach     bool                 `json:"multiattach"`
+	Attachments     []volumes.Attachment `json:"attachments"`
+	MigrationStatus *string              `json:"migration_status"`
+}
+
 // AttachVolume attaches given cinder volume to the compute
 func (os *OpenStack) AttachVolume(ctx context.Context, instanceID, volumeID string) (string, error) {
 	computeServiceClient := os.compute
 
-	volume, err := os.GetVolume(ctx, volumeID)
-	if err != nil {
+	// Fetch the volume with migration_status using a single API call.
+	// We use a custom struct because gophercloud's Volume does not include migration_status.
+	mc := metrics.NewMetricContext("volume", "get")
+	var volResult volumeForAttach
+	err := volumes.Get(ctx, os.blockstorage, volumeID).ExtractInto(&volResult)
+	if mc.ObserveRequest(err) != nil {
 		return "", err
 	}
 
-	for _, att := range volume.Attachments {
+	for _, att := range volResult.Attachments {
 		if instanceID == att.ServerID {
 			klog.V(4).Infof("Disk %s is already attached to instance %s", volumeID, instanceID)
-			return volume.ID, nil
+			return volResult.ID, nil
 		}
 	}
 
-	if volume.Multiattach {
+	// Check migration_status: if the volume is migrating, it cannot be attached.
+	// Per OpenStack docs, "starting", "migrating", and "completing" all indicate
+	// a migration is in progress.
+	if volResult.MigrationStatus != nil {
+		switch *volResult.MigrationStatus {
+		case "starting", "migrating", "completing":
+			return "", fmt.Errorf("volume %s has migration_status %q, volume must not be migrating before attach", volumeID, *volResult.MigrationStatus)
+		}
+	}
+
+	if volResult.Multiattach {
+		if volResult.Status != VolumeAvailableStatus && volResult.Status != VolumeInUseStatus {
+			return "", fmt.Errorf("volume %s is in %s status, volume must be available or in-use for multi-attach capable volumes", volumeID, volResult.Status)
+		}
 		// For multiattach volumes, supported compute api version is 2.60
 		// Init a local thread safe copy of the compute ServiceClient
 		computeServiceClient, err = openstack.NewComputeV2(os.compute.ProviderClient, os.epOpts)
@@ -226,18 +253,22 @@ func (os *OpenStack) AttachVolume(ctx context.Context, instanceID, volumeID stri
 			return "", err
 		}
 		computeServiceClient.Microversion = "2.60"
+	} else {
+		if volResult.Status != VolumeAvailableStatus {
+			return "", fmt.Errorf("volume %s is in %s status, volume must be available", volumeID, volResult.Status)
+		}
 	}
 
-	mc := metrics.NewMetricContext("volume", "attach")
+	mcAttach := metrics.NewMetricContext("volume", "attach")
 	_, err = volumeattach.Create(ctx, computeServiceClient, instanceID, &volumeattach.CreateOpts{
-		VolumeID: volume.ID,
+		VolumeID: volResult.ID,
 	}).Extract()
 
-	if mc.ObserveRequest(err) != nil {
+	if mcAttach.ObserveRequest(err) != nil {
 		return "", fmt.Errorf("failed to attach %s volume to %s compute: %v", volumeID, instanceID, err)
 	}
 
-	return volume.ID, nil
+	return volResult.ID, nil
 }
 
 // WaitDiskAttached waits for attached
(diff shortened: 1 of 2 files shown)


pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -202,23 +202,52 @@ func (os *OpenStack) GetVolume(ctx context.Context, volumeID string) (*volumes.V
...
 		}
 	}
 
-	if volume.Multiattach {
+	// Check migration_status: if the volume is migrating, it cannot be attached.
+	if volResult.MigrationStatus != nil && *volResult.MigrationStatus == "migrating" {
+		return "", fmt.Errorf("volume %s has migration_status %q, volume must not be migrating before attach", volumeID, *volResult.MigrationStatus)
+	}
+
+	if volResult.Multiattach {
```
yahor-kurachkin, line 245: Just a minor observation. `if volResult.Multiattach` here and then a few lines further down there's a second `if volResult.Multiattach`. Would it make sense to merge them into a single `if/else` block?

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -202,23 +202,52 @@ func (os *OpenStack) GetVolume(ctx context.Context, volumeID string) (*volumes.V
...
 		if instanceID == att.ServerID {
 			klog.V(4).Infof("Disk %s is already attached to instance %s", volumeID, instanceID)
-			return volume.ID, nil
+			return volResult.ID, nil
 		}
 	}
 
-	if volume.Multiattach {
+	// Check migration_status: if the volume is migrating, it cannot be attached.
+	if volResult.MigrationStatus != nil && *volResult.MigrationStatus == "migrating" {
```
yahor-kurachkin: I think it would be worth checking for not only `"migrating"`, but also `"starting"` and `"completing"` and possibly other statuses. According to the OpenStack [documentation](https://docs.openstack.org/cinder/latest/contributor/migration.html#what-can-be-tracked-during-volume-migration), `"migrating"` isn't the only status that can be returned while migration is in progress.

pkg/csi/cinder/openstack/openstack_volumes_test.go
```diff
@@ -0,0 +1,445 @@
...
+	if err != nil {
+		t.Fatalf("expected no error for multiattach available volume, got: %v", err)
+	}
+	if result != volumeID {
+		t.Errorf("expected volume ID %q, got %q", volumeID, result)
+	}
+}
+
+func TestAttachVolume_MultiAttach_InUseStatus(t *testing.T) {
+	volumeID := "vol-123"
```
yahor-kurachkin: I think it would make sense to extract these strings into constants. For example, `"vol-123"` appears 10 times in the tests.

pkg/csi/cinder/openstack/openstack_volumes.go
hemna: Done. Merged both `if volResult.Multiattach` blocks into a single if/else that handles status validation and microversion setup together.

pkg/csi/cinder/openstack/openstack_volumes.go
hemna: Good catch. Updated to reject all in-progress migration statuses (`starting`, `migrating`, `completing`) per the [OpenStack migration docs](https://docs.openstack.org/cinder/latest/contributor/migration.html). Used a switch statement for clarity. The test is now table-driven covering all three.

pkg/csi/cinder/openstack/openstack_volumes_test.go
hemna: Done. Extracted `testVolumeID`, `testInstanceID`, and `testOtherInstance` as package-level constants. All inline literals replaced.

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -202,42 +202,73 @@ func (os *OpenStack) GetVolume(ctx context.Context, volumeID string) (*volumes.V
...
 }
 
+// volumeForAttach contains the volume fields needed for attach validation,
+// including migration_status which gophercloud's Volume struct does not map.
+type volumeForAttach struct {
+	ID              string               `json:"id"`
+	Status          string               `json:"status"`
+	Multiattach     bool                 `json:"multiattach"`
+	Attachments     []volumes.Attachment `json:"attachments"`
+	MigrationStatus *string              `json:"migration_status"`
```
kayrus, line 212: I don't think that `migration_status`makes sense here since the CSI plugin typically runs with project admin (not cloud admin) privileges, and this field might not be accessible. If this is needed, please first add `MigrationStatus *string` to gophercloud's Volume struct via a PR upstream, then we can use it here without defining a private volume struct.

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -202,42 +202,73 @@ func (os *OpenStack) GetVolume(ctx context.Context, volumeID string) (*volumes.V
...
 		}
 	}
 
-	if volume.Multiattach {
+	// Check migration_status: if the volume is migrating, it cannot be attached.
+	// Per OpenStack docs, "starting", "migrating", and "completing" all indicate
+	// a migration is in progress.
+	if volResult.MigrationStatus != nil {
+		switch *volResult.MigrationStatus {
+		case "starting", "migrating", "completing":
```
kayrus, line 240: I'd add a slice of desired statuses into a global var along with `slices.Contains`

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -202,42 +202,73 @@ func (os *OpenStack) GetVolume(ctx context.Context, volumeID string) (*volumes.V
...
+			return "", fmt.Errorf("volume %s has migration_status %q, volume must not be migrating before attach", volumeID, *volResult.MigrationStatus)
+		}
+	}
+
+	if volResult.Multiattach {
+		if volResult.Status != VolumeAvailableStatus && volResult.Status != VolumeInUseStatus {
```
kayrus, line 246: I'd use `slices.Contains` here too

pkg/csi/cinder/openstack/openstack_volumes.go
```diff
@@ -202,42 +202,73 @@ func (os *OpenStack) GetVolume(ctx context.Context, volumeID string) (*volumes.V
...
 		computeServiceClient, err = openstack.NewComputeV2(os.compute.ProviderClient, os.epOpts)
 		if err != nil {
 			return "", err
 		}
 		computeServiceClient.Microversion = "2.60"
+	} else {
+		if volResult.Status != VolumeAvailableStatus {
+			return "", fmt.Errorf("volume %s is in %s status, volume must be available", volumeID, volResult.Status)
+		}
 	}
```
kayrus, line 260: I'd rework this part with a possible early return on err to avoid complex if/else/if structure

PR #3134, opened 2026-06-19
[manila-csi-plugin] Use GET-based share access rules API when available

### LINKED ISSUE #2277: [manila-csi-plugin]: use new access rules GET method when possible

/kind feature

**What happened**:

Some OpenStack deployments can have middlewares that limit an amount of POST/PUT requests, which are categorized as write requests. One of the wrongly categorized request is a deprecated [list access rules](https://docs.openstack.org/api-ref/shared-file-system/?expanded=list-access-rules-deprecated-detail#list-access-rules-deprecated) action.

In order to prevent excessive requests rate limiting, CSI Manila plugin should support a [newer](https://docs.openstack.org/api-ref/shared-file-system/?expanded=describe-share-access-rule-detail#share-access-rules-since-api-v2-45) Manila API call.

**How to reproduce it**:

n/a

See also https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1582

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: n/a
- OpenStack version: n/a
- Others: n/a

cc @Carthaca

gouthampacha:

Switches from the deprecated `os-access_list` POST action to the
GET-based `share-access-rules` API (microversion 2.45+) for listing
access rules. Falls back to the old API for pre-2.45 deployments.

Some OpenStack deployments rate-limit POST/PUT as write requests,
causing unnecessary throttling on what is semantically a read operation.

fixes #2277

The server's max supported microversion (already fetched during client
init validation) is now stored in the Client struct. `GetAccessRights`
conditionally uses `shareaccessrules.List` (GET) when the server
supports >= 2.45, otherwise falls back to `shares.ListAccessRights`
(POST). The `Interface` return type is unchanged — all callers still
receive `[]shares.AccessRight`.

**Release note**:
```release-note
[manila-csi-plugin] Use GET-based share access rules API (microversion 2.45+) instead of deprecated POST action, avoiding write rate-limiting on read operations. Falls back automatically for older OpenStack deployments.
```

### DIFF
--- a/pkg/csi/manila/manilaclient/builder.go
+++ b/pkg/csi/manila/manilaclient/builder.go
@@ -72,11 +72,12 @@ func New(ctx context.Context, o *client.AuthOpts, userAgent string, extraUserAge
 	// Check client's and server's versions for compatibility
 
 	client.Microversion = minimumManilaVersion
-	if err = validateManilaClient(ctx, client); err != nil {
+	serverMaxMicroversion, err := validateManilaClient(ctx, client)
+	if err != nil {
 		return nil, fmt.Errorf("manila v2 client validation failed: %v", err)
 	}
 
-	return &Client{c: client}, nil
+	return &Client{c: client, serverMaxMicroversion: serverMaxMicroversion}, nil
 }
 
 func splitManilaMicroversion(microversion string) (major, minor int) {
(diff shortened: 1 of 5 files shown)


gouthampacha:
Closing this in favor of #3147 which landed the same core change (switching to the GET-based shareaccessrules API).

This PR had fallback logic for pre-2.45 servers, but microversion 2.45 shipped in Queens (2018). Any Manila deployment that old is unlikely to be running a current CSI driver, so the fallback isn't worth the extra complexity.

@yahor-kurachkin @kayrus thanks for picking this up.

PR #3136, opened 2026-06-20
[manila-csi-plugin] Wait for access rule to become active after granting

### LINKED ISSUE #3000: [manila-csi-plugin] wait for the "active" access state
/kind bug

**What happened**:

Manila's grant access API is asynchronous, and it doesn't guarantee that the access rules are applied. There are cases, when the rules state may still be `queued_to_apply`.

Manila CSI driver must wait until the access rules state is `active` before trying to mount the share. If the timeout is reached or the access rule is in error state, CSI driver must delete the access rule and return an error, so the next reconcile loop can create it from scratch.

**How to reproduce it**:

n/a

see also #2277 #2725

supported access rule states: https://github.com/openstack/manila/blob/40107f961671fd55710f70eef30ea526298b5e22/manila/common/constants.py#L62-L71

**Environment**:
- OpenStack version:
- Others:

gouthampacha:

Manila's `GrantAccess` API is asynchronous — the access rule starts in `queued_to_apply` state and transitions through `applying` before reaching `active`. Both the NFS and CephFS share adapters were proceeding to mount immediately after granting access, which could fail if the backend hadn't applied the rule yet.

This PR adds a shared `waitForAccessRuleActive` helper that polls until the access rule reaches `active` state. On `error` state or timeout, the helper revokes the failed rule so the next reconcile retries cleanly. Both adapters now call this helper after `GrantAccess`.

For CephFS, this replaces the previous wait loop that only checked for `AccessKey` population without verifying the rule state. Waiting for `active` state implicitly guarantees the key is populated since Manila assigns the cephx key as part of applying the rule.

fixes #3000

Tested on a DevStack with both LVM (NFS) and native CephFS backends. Both protocols show the `queued_to_apply` → `active` transition, confirming the race window this fixes.

**Release note**:
```release-note
[manila-csi-plugin] Wait for access rule to become active after granting access, fixing mount failures when backends are slow to apply rules.
```

### DIFF
--- a/pkg/csi/manila/shareadapters/accessrights.go
+++ b/pkg/csi/manila/shareadapters/accessrights.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package shareadapters
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/gophercloud/gophercloud/v2/openstack/sharedfilesystems/v2/shares"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/cloud-provider-openstack/pkg/csi/manila/manilaclient"
+	"k8s.io/klog/v2"
+)
+
+const (
+	accessRuleStateActive      = "active"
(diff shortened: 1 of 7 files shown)


gouthampacha:
@kayrus @yahor-kurachkin could you review this when you get a chance?

yahor-kurachkin (approved):
LGTM, nice work!

pkg/csi/manila/shareadapters/accessrights.go
```diff
@@ -0,0 +1,94 @@
...
+	var (
+		backoff = wait.Backoff{
+			Duration: time.Second * waitForAccessRuleTimeout,
+			Factor:   1.2,
+			Steps:    waitForAccessRuleRetries,
+		}
+		result *shares.AccessRight
+	)
+
+	err := wait.ExponentialBackoff(backoff, func() (bool, error) {
```
kayrus: can you use the [ExponentialBackoffWithContext](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#ExponentialBackoffWithContext) here?

gouthampacha: Done.

PR #3138, opened 2026-06-20
[manila-csi-plugin] Add staticVolume parameter to CephFS volume context

### LINKED ISSUE #3137: [manila-csi-plugin] CephFS adapter missing staticVolume parameter for ceph-csi compatibility
**What happened**:

The Manila CSI CephFS share adapter sends `provisionVolume: "false"` in the volume context to indicate a statically provisioned volume to the ceph-csi node plugin. However, ceph-csi introduced a volume ID format validator in [e2a1f6e](https://github.com/ceph/ceph-csi/commit/e2a1f6e8ca47c87d517949c15ee41a48014ebe6a) that rejects volume IDs not matching the ceph-csi internal format (`^[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[a-zA-Z0-9\-\_]+$`).

Manila CSI passes Manila share UUIDs as volume IDs (e.g. `fee6cec9-988a-45fd-b01d-dc1f5e562104`), which don't match this format. The format check is skipped when [`IsStaticVol()`](https://github.com/ceph/ceph-csi/blob/devel/internal/util/validate.go) returns true, but that function checks for `staticVolume: "true"` in the volume attributes — not `provisionVolume: "false"`.

The result is that `NodeStageVolume` fails with: `volumeID has an unexpected format`.

Manila CSI volumes should mount successfully with ceph-csi regardless of the volume ID format, since they are statically provisioned volumes.

**How to reproduce**:

Deploy Manila CSI with CephFS protocol using ceph-csi >= v3.14.0 as the forwarding node plugin. Create a CephFS PVC — the pod will fail to start with the format validation error.

**Fix**:

Add `staticVolume: "true"` to the CephFS volume context alongside the existing `provisionVolume: "false"`. The `staticVolume` parameter is simply ignored by older ceph-csi versions that don't recognize it, making this change backwards compatible.

**Environment**:
- cloud-provider-openstack version: master
- ceph-csi version: v3.14.0+

gouthampacha:

ceph-csi [introduced volume ID format validation](https://github.com/ceph/ceph-csi/commit/e2a1f6e8ca47c87d517949c15ee41a48014ebe6a) that rejects IDs not matching its internal format (`^[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[a-zA-Z0-9\-\_]+$`). The format check is skipped when [`IsStaticVol()`](https://github.com/ceph/ceph-csi/blob/devel/internal/util/validate.go) returns true, which requires `staticVolume: "true"` in the volume attributes. Manila CSI only sets `provisionVolume: "false"`, which ceph-csi does not check for this purpose. This causes `NodeStageVolume` to fail with `volumeID has an unexpected format` on ceph-csi v3.14.0+.

This PR adds `staticVolume: "true"` to the CephFS volume context alongside the existing `provisionVolume: "false"`. The parameter is ignored by older ceph-csi versions that do not recognize it, making this change backwards compatible.

fixes #3137

One-line change. Tested on DevStack with native CephFS backend using ceph-csi v3.12 (confirms backwards compatibility — parameter is ignored) and verified the fix against ceph-csi v3.17 source code.

**Release note**:
```release-note
[manila-csi-plugin] Add staticVolume parameter to CephFS volume context for compatibility with ceph-csi v3.14.0+.
```

### DIFF
--- a/pkg/csi/manila/shareadapters/cephfs.go
+++ b/pkg/csi/manila/shareadapters/cephfs.go
@@ -119,6 +119,7 @@ func (Cephfs) BuildVolumeContext(args *VolumeContextArgs) (volumeContext map[str
 		"monitors":        monitors,
 		"rootPath":        rootPath,
 		"mounter":         args.Options.CephfsMounter,
+		"staticVolume":    "true",
 		"provisionVolume": "false",
 	}
 


gouthampacha:
Closing this — further analysis shows that adding `staticVolume: "true"` without `clusterID` trades one error for another on ceph-csi v3.14+.

The `staticVolume: "true"` parameter fixes the `ValidateVolumeID` format check, but then `NewVolumeOptionsFromStaticVolume` in `internal/cephfs/store/volumeoptions.go` claims the volume and calls `GetClusterInformation()`, which hard-fails on missing `clusterID`. That error is not `ErrNonStaticVolume`, so the fallback to `NewVolumeOptionsFromMonitorList` (which works with Manila CSI's existing volume context) is never reached.

The fix needs to happen upstream in ceph-csi — see #3137 for the full analysis.

PR #3140, opened 2026-06-23
[manila-csi-plugin] Restore --nodeid and --nodeaz as optional overrides

### LINKED ISSUE #3139: manila-csi-plugin: Restore --nodeid/--nodeaz flags as alternative to metadata service
Since #2734, the Manila CSI node plugin requires the OpenStack metadata service (config drive or 169.254.169.254) to discover node identity. The `--nodeid` and `--nodeaz` flags were deprecated and are ignored.

#2734 aligned Manila with how Cinder CSI handles node identity. For Cinder this makes sense: the node must be a Nova VM for volume attach/detach to work, so the metadata service is always available. Manila doesn't have that constraint. Shares are mounted via NFS or CEPHFS on the host. The node doesn't need to be a Nova instance.

This breaks deployments where the CSI node plugin runs on hosts that aren't OpenStack VMs, such as bare metal k8s clusters or edge nodes. `NodeGetInfo` fails with an Internal error and the node plugin can't register.

Before #2734, operators could set `--nodeid` to the Kubernetes node name (via the downward API) and `--nodeaz` to a fixed value or leave it empty. That worked fine and should continue to work.

Metadata auto-detection is a good default, but it shouldn't be the only option. If `--nodeid` is explicitly set, the driver should use it instead of querying metadata.

gouthampacha:

Un-deprecates the `--nodeid` and `--nodeaz` flags so they can be used as
alternatives to the metadata service for node identity.

#2734 aligned Manila CSI with how Cinder CSI handles node identity by making
the metadata service mandatory. For Cinder this makes sense: the node must be
a Nova VM for volume attach/detach. Manila doesn't have that constraint. Shares
mount via NFS or CEPHFS on the host, and the node doesn't need to be a Nova
instance.

When `--nodeid` is set, the flag value is used directly. When unset, the driver
falls back to the metadata service (current behavior). Same for `--nodeaz`.
OpenStack deployments are unaffected.

Fixes #3139

Tested end-to-end on a DevStack environment with Manila LVM:
- With `--nodeid=test-node-from-flag`: CSINode registers with the flag value,
  PVC binds, share mounts, pod reads/writes data
- Without `--nodeid` (default): CSINode registers with the metadata-provided
  UUID, PVC binds, share mounts, pod reads/writes data

**Release note**:
```release-note
[manila-csi-plugin] The `--nodeid` and `--nodeaz` flags are no longer
deprecated. When set, they override the metadata service for node identity,
allowing Manila CSI to run on nodes that are not OpenStack VMs.
```

### DIFF
--- a/cmd/manila-csi-plugin/main.go
+++ b/cmd/manila-csi-plugin/main.go
@@ -89,6 +89,8 @@ func main() {
 				ManilaClientBuilder: manilaClientBuilder,
 				CSIClientBuilder:    csiClientBuilder,
 				ClusterID:           clusterID,
+				NodeID:              nodeID,
+				NodeAZ:              nodeAZ,
 				PVCLister:           csi.GetPVCLister(),
 			}
 
@@ -105,15 +107,16 @@ func main() {
 			}
 
 			if provideNodeService {
-				err = metadata.CheckMetadataSearchOrder(searchOrder)
-				if err != nil {
-					klog.Fatalf("Invalid search-order: %v", err)
+				var md metadata.IMetadata
+				if nodeID == "" || (withTopology && nodeAZ == "") {
+					err = metadata.CheckMetadataSearchOrder(searchOrder)
+					if err != nil {
+						klog.Fatalf("Invalid search-order: %v", err)
+					}
+					md = metadata.GetMetadataProvider(searchOrder)
 				}
 
-				// Initialize metadata
-				metadata := metadata.GetMetadataProvider(searchOrder)
-
-				err = d.SetupNodeService(metadata)
+				err = d.SetupNodeService(nodeID, nodeAZ, md)
 				if err != nil {
 					klog.Fatalf("Driver node service initialization failed: %v", err)
 				}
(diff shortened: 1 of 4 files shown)


gouthampacha:
@kayrus @yahor-kurachkin could you review this when you get a chance?

pkg/csi/manila/options/shareoptions.go
```diff
@@ -43,8 +43,8 @@ type ControllerVolumeContext struct {
 type NodeVolumeContext struct {
 	ShareID        string `name:"shareID" value:"optionalIf:shareName=." precludes:"shareName"`
 	ShareName      string `name:"shareName" value:"optionalIf:shareID=." precludes:"shareID"`
-	ShareAccessID  string `name:"shareAccessID" value:"optionalIf:shareAccessIDs=." precludes:"shareAccessIDs"` // Keep this for backwards compatibility
-	ShareAccessIDs string `name:"shareAccessIDs" value:"optionalIf:shareAccessID=." precludes:"shareAccessID"`
+	ShareAccessID  string `name:"shareAccessID" value:"optional"`  // Deprecated: use shareAccessIDs
```
kayrus, line 46: could you please make this change in a separate PR?

cmd/manila-csi-plugin/main.go
```diff
@@ -104,10 +106,12 @@ func main() {
...
 			if provideNodeService {
-				// Initialize metadata
-				metadata := metadata.GetMetadataProvider("")
+				var md metadata.IMetadata
+				if nodeID == "" || (withTopology && nodeAZ == "") {
+					md = metadata.GetMetadataProvider("")
+				}
 
-				err = d.SetupNodeService(metadata)
+				err = d.SetupNodeService(nodeID, nodeAZ, md)
```
kayrus, line 119: this change needs to be aligned with #3141 
@gouthampacha can you check the #3141 as well?

pkg/csi/manila/metadata_test.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func (f *fakeMetadataProvider) GetAvailabilityZone() (string, error) {
+	if f.err != nil {
+		return "", f.err
+	}
+	return f.availabilityZone, nil
+}
+
+func TestStaticMetadata(t *testing.T) {
```
yahor-kurachkin, line 44: Your approach with table-driven tests in `shareoptions_test.go` and `adapters_test.go` is nice — would you consider following the same pattern in `metadata_test.go`?

yahor-kurachkin (commented):
Just a minor comment about the testing approach. Otherwise LGTM.

pkg/csi/manila/options/shareoptions.go
gouthampacha: ugh, this is my bad, it was a part of a different PR: #3132

cmd/manila-csi-plugin/main.go
gouthampacha: I've left a comment on #3141 about the `/dev` mount scope. If #3141 merges first, I'll rebase this PR on top. The conflict is limited to the metadata init block in main.go and should be straightforward to resolve either way.

pkg/csi/manila/metadata_test.go
gouthampacha: Done, switched to table-driven tests.

cmd/manila-csi-plugin/main.go
gouthampacha: done rebasing, this builds on top of the merged changes..

gouthampacha:
@kayrus @stephenfin wdyt of this?

stephenfin (commented):
Fair enough

/lgtm
/approve

PR #3141, opened 2026-07-01
Fix config drive mount and add flag to control the search order in which the driver retrieves instance metadata

### LINKED ISSUE #3118: Manila CSI registration failing due to non existing medata service on baremetal nodes
/kind bug

**What happened**:
Manila CSI deprecated the` --node-id`  option, and now the value is automatically retrieved from the metadata service.

In our setup, VMs rely on the metadata service, while baremetal nodes rely on config drive. As a result, the CSI fails to register on baremetal nodes:
`I0521 08:58:47.584747       1 main.go:111] "Received NotifyRegistrationStatus call" status="error:\"RegisterPlugin error -- plugin registration failed with err: rpc error: code = Internal desc = [NodeGetInfo] unable to retrieve instance id of node error fetching http://169.254.169.254/openstack/latest/meta_data.json: Get \\\"http://169.254.169.254/openstack/latest/meta_data.json\\\": dial tcp 169.254.169.254:80: i/o timeout\""`

To either have support for auto discovery from both metadata service and config drive, or keep the `--node-id`, which we can populate on the k8s side.

**How to reproduce it**:
Deploy Manila CSI on a node configured with config drive. The registrar container with fail `unable to retrieve instance id of node error fetching http://169.254.169.254/openstack/latest/meta_data.json`.

This also seems to be the case for Cinder CSI.

**Environment**:
- openstack-cloud-controller-manager: v1.35.0
- csi-node-driver-registrar:v2.15.0

dgaponcic:
Currently the manila csi only uses metadata service to get instance metadata, which fails when using config drive.

* This PR adds a new flag `--search-order` (similar to cinder) to allow setting up the order in which the driver should try to retrieve the metadata. By default: configDrive,metadataService
* And this PR adds the missing bits to get the config drive to work.

fixes https://github.com/kubernetes/cloud-provider-openstack/issues/3118

1. Start the node plugin without `--search-order`. It will use the default value of configDrive,metadataService. On a machine using metadata service, this is the output:
```
}
I0701 14:15:22.146174       1 driver.go:327] [ID:2] GRPC call: /csi.v1.Node/NodeGetInfo
...
I0701 14:15:22.146893       1 metadata.go:170] Attempting to mount configdrive on /tmp
I0701 14:15:22.148041       1 mount_linux.go:326] 'umount /tmp/kubelet-detect-safe-umount3371542402' failed with: exit status 1, output: umount: can't unmount /tmp/kubelet-detect-safe-umount3371542402: Invalid argument
...
E0701 14:15:22.148613       1 mount_linux.go:271] Mount failed: exit status 1


I0701 14:15:22.149110       1 metadata.go:204] Attempting to fetch metadata from http://169.254.169.254/openstack/latest/meta_data.json, ignoring proxy settings
I0701 14:15:22.288691       1 driver.go:333] [ID:2] GRPC response: {"node_id":"<node id>"}
```
It tried to mount the config drive, fails, and then tries the metadata service, which works. 

2. Start the node plugin with --search-order=metadataService,configDrive. On an instance using config drive, this is the output:
```
I0701 14:47:56.211444       1 driver.go:327] [ID:2] GRPC call: /csi.v1.Node/NodeGetInfo
I0701 14:47:56.211462       1 driver.go:328] [ID:2] GRPC request: {}
I0701 14:47:56.211486       1 metadata.go:204] Attempting to fetch metadata from http://169.254.169.254/openstack/latest/meta_data.json, ignoring proxy settings

I0701 14:48:26.212841       1 metadata.go:170] Attempting to mount configdrive /dev/disk/by-label/config-2 on /tmp
...
I0701 14:48:26.216486       1 metadata.go:182] Configdrive mounted on /tmp
I0701 14:48:26.216921       1 mount_linux.go:401] Unmounting /tmp
I0701 14:48:26.218503       1 driver.go:333] [ID:2] GRPC response: {"node_id":"<node-id>"}

```
It tried to use the matadata sevice, and when it didn't work, tried the config drive, which worked.

**Release note**:

```release-note
[manila-csi-plugin] Add `--search-order` flag to control the order in which instance metadata is retrieved (configDrive, metadataService)
```

### DIFF
--- a/charts/manila-csi-plugin/values.yaml
+++ b/charts/manila-csi-plugin/values.yaml
@@ -142,4 +142,3 @@ controllerplugin:
 # See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md
 # for description of individual verbosity levels.
 logVerbosityLevel: 2
-
--- a/charts/manila-csi-plugin/values.yaml
+++ b/charts/manila-csi-plugin/values.yaml
@@ -143,4 +143,3 @@ controllerplugin:
 # for description of individual verbosity levels.
 logVerbosityLevel: 2
 
-searchOrder:
--- a/charts/manila-csi-plugin/values.yaml
+++ b/charts/manila-csi-plugin/values.yaml
@@ -48,6 +48,9 @@ csimanila:
     pullPolicy: IfNotPresent
     tag:  # defaults to .Chart.AppVersion
 
+  # Set the search order in which the driver retrieves metadata relating to the instance (s) in which it runs
+  searchOrder: ""
+
 # DeamonSet deployment
 nodeplugin:
   # Component name
@@ -139,3 +142,5 @@ controllerplugin:
 # See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md
 # for description of individual verbosity levels.
 logVerbosityLevel: 2
+
+searchOrder:
--- a/cmd/manila-csi-plugin/main.go
+++ b/cmd/manila-csi-plugin/main.go
@@ -52,6 +52,7 @@ var (
 	userAgentData            []string
 	provideControllerService bool
 	provideNodeService       bool
+	searchOrder              string
 )
 
 func validateShareProtocolSelector(v string) error {
@@ -104,8 +105,13 @@ func main() {
 			}
 
 			if provideNodeService {
+				err = metadata.CheckMetadataSearchOrder(searchOrder)
+				if err != nil {
+					klog.Fatalf("Invalid search-order: %v", err)
+				}
+
 				// Initialize metadata
-				metadata := metadata.GetMetadataProvider("")
+				metadata := metadata.GetMetadataProvider(searchOrder)
 
 				err = d.SetupNodeService(metadata)
 				if err != nil {
@@ -158,6 +164,7 @@ func main() {
 
 	cmd.PersistentFlags().BoolVar(&provideControllerService, "provide-controller-service", true, "If set to true then the CSI driver does provide the controller service (default: true)")
 	cmd.PersistentFlags().BoolVar(&provideNodeService, "provide-node-service", true, "If set to true then the CSI driver does provide the node service (default: true)")
+	cmd.PersistentFlags().StringVar(&searchOrder, "search-order", "configDrive,metadataService", "The search order in which the driver retrieves metadata relating to the instance (s) in which it runs")
 
 	code := cli.Run(cmd)
 	os.Exit(code)
--- a/docs/manila-csi-plugin/using-manila-csi-plugin.md
+++ b/docs/manila-csi-plugin/using-manila-csi-plugin.md
@@ -44,6 +44,7 @@ Option | Default value | Description
 `--provide-controller-service` | `true` | If set to true then the CSI driver does provide the controller service.
 `--provide-node-service` | `true` | If set to true then the CSI driver does provide the node service.
 `--pvc-annotations` | `false` | If set to true then the CSI driver will use PVC annotations as an additional information when creating shares. See [Supported PVC annotations](#supported-pvc-annotations) for more info.
+`--search-order` | `configDrive,metadataService` | The search order in which the driver retrieves metadata relating to the instance (s) in which it runs.
 
 ### Controller Service volume parameters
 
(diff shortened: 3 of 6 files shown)


gouthampacha:
/ok-to-test

@dgaponcic - have you signed the CLA?

manifests/manila-csi-plugin/csi-nodeplugin.yaml
```diff
@@ -80,6 +80,9 @@ spec:
               mountPath: /var/lib/kubelet/plugins/manila.csi.openstack.org
             - name: fwd-plugin-dir
               mountPath: /var/lib/kubelet/plugins/FWD-NODEPLUGIN
+            - name: configdrive-dev
```
kayrus, line 83: can you also adjust a corresponding helm chart template?

cmd/manila-csi-plugin/main.go
```diff
@@ -158,6 +164,7 @@ func main() {
 
 	cmd.PersistentFlags().BoolVar(&provideControllerService, "provide-controller-service", true, "If set to true then the CSI driver does provide the controller service (default: true)")
 	cmd.PersistentFlags().BoolVar(&provideNodeService, "provide-node-service", true, "If set to true then the CSI driver does provide the node service (default: true)")
+	cmd.PersistentFlags().StringVar(&searchOrder, "search-order", "configDrive,metadataService", "The search order in which the driver retrieves metadata relating to the instance (s) in which it runs")
```
kayrus, line 167: please adjust docs accordingly

kayrus:
thanks for the PR. Please set proper `Release note` in your initial message.

charts/manila-csi-plugin/values.yaml
```diff
@@ -139,3 +142,5 @@ controllerplugin:
 # See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md
 # for description of individual verbosity levels.
 logVerbosityLevel: 2
+
+searchOrder:
```
yahor-kurachkin: I think this is a duplication. The `searchOrder` already under `csimanila:` so this top-level line has no effect. Is there a reason it's needed here?

cmd/manila-csi-plugin/main.go
kayrus, line 167: I'm curious about the config format. In cinder it is part of the config file, here it is a CLI argument. Does it make sense to implement a common config pattern?
@gouthampacha @yahor-kurachkin @dgaponcic what do you think?

docs/manila-csi-plugin/using-manila-csi-plugin.md
```diff
@@ -44,6 +44,7 @@ Option | Default value | Description
 `--provide-controller-service` | `true` | If set to true then the CSI driver does provide the controller service.
 `--provide-node-service` | `true` | If set to true then the CSI driver does provide the node service.
 `--pvc-annotations` | `false` | If set to true then the CSI driver will use PVC annotations as an additional information when creating shares. See [Supported PVC annotations](#supported-pvc-annotations) for more info.
+`--search-order` | `configDrive,metadataService` | The search order in which the driver retrieves metadata relating to the instance (s) in which it runs.
```
kayrus: this option is valid only when `provide-node-service` is true, please mention this in docs.

cmd/manila-csi-plugin/main.go
yahor-kurachkin: @kayrus Currently, `manila-csi-plugin` doesn't have a config file. I think that introducing one just for `search-order` is out of scope for this PR. I'd rather add a config file later if we end up having other settings besides the `search-order` that need to be configurable.

docs/manila-csi-plugin/using-manila-csi-plugin.md
kayrus:
```suggestion
`--search-order` | `configDrive,metadataService` | The search order in which the driver retrieves metadata relating to the instance in which it runs. Only effective when `--provide-node-service=true`.
```

gouthampacha:
The `/dev` mount is too broad for Manila CSI. Config drive is at `/dev/disk/by-label/config-2`, so mounting `/dev/disk` should be sufficient. Manila doesn't need access to block devices unlike Cinder.

kayrus:
@dgaponcic could you rebase your changes on recent master HEAD?

kayrus:
> The `/dev` mount is too broad for Manila CSI. Config drive is at `/dev/disk/by-label/config-2`, so mounting `/dev/disk` should be sufficient. Manila doesn't need access to block devices unlike Cinder.

If there is a config-drive, it is accessible as a `/dev/sr0` block device:

```sh
$ readlink -f /dev/disk/by-label/config-2 
/dev/sr0
```

kayrus:
@dgaponcic could you bump the chart version and set `/dev` mount point for a pod due to:

> it is accessible as a `/dev/sr0` block device:
> 
> ```shell
> $ readlink -f /dev/disk/by-label/config-2 
> /dev/sr0
> ```

kayrus:
@dgaponcic it doesn't look that you reverted the `mount /dev/disk instead of /dev` commit. See my arguments above.

dgaponcic:
@kayrus Yeah, you are correct. I tested the change, and it was working. But I didn't realize it's was working not because of the /dev/disk mount, but because the container runs with `priviledged: true`, which grants access to all host devices. I think it makes sense to use /dev, as this is what is has access to anyway.

dgaponcic:
Thank you all for your help!

PR #3143, opened 2026-07-04
[cinder-csi-plugin] Fix panic when backup lookup fails in CreateSnapshot

immanuwell:

This fixes a nil pointer panic in the cinder CSI backup snapshot path.
If `CreateSnapshot` runs with `snapshotType=backup`, `ListBackups` finds one match, and `GetBackupByID` then errors, the code dereferences `backup.ID` and blows up. Kinda nasty.

The fix logs the already known backup ID from `ListBackups` and returns `codes.Internal` cleanly.

fixes #

Special notes for reviewers:

Repro before this patch:
```bash
go test ./pkg/csi/cinder -run TestCreateSnapshotBackupGetBackupByIDError -count=1
```

Before the fix this panics in `CreateSnapshot`.
With this patch it returns a normal gRPC internal error, so no crash, nice and boring.

Release note:
```release-note
[cinder-csi-plugin] Fix panic when backup lookup fails in CreateSnapshot
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -565,9 +565,10 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		if len(backups) == 1 {
 			// since backup.VolumeID is not part of ListBackups response
 			// we need fetch single backup to get the full object.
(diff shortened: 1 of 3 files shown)


immanuwell:
@kayrus @stephenfin please take a look

kayrus:
@immanuwell can you provide a proper release note? in the initial PR message?

PR #3148, opened 2026-07-16
fix: return backup lookup errors when deleting snapshots

immanuwell:
Fixes a path where a Cinder backup lookup failure is ignored during `DeleteSnapshot`.

A non-404 error now reaches the snapshotter, so it retries instead of treating a later snapshot 404 as success.

Repro:
1. Use a VolumeSnapshotClass with `type: backup`
2. Make Cinder return 500 or 503 to `GET /backups/<id>`
3. Delete the VolumeSnapshot. Before this, it can report success and leave the backup behind.

Tests: `make unit`, `go test ./tests/sanity/cinder`

**Release note**:
```release-note
[cinder-csi-plugin] Return Cinder backup lookup failures when deleting a backup-backed VolumeSnapshot so the snapshotter retries deletion.
```

### DIFF
--- a/pkg/csi/cinder/controllerserver.go
+++ b/pkg/csi/cinder/controllerserver.go
@@ -769,12 +769,17 @@ func (cs *controllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
 
 	// If volumeSnapshot object was linked to a cinder backup, delete the backup.
 	back, err := cloud.GetBackupByID(ctx, id)
-	if err == nil && back != nil {
+	if err != nil && !cpoerrors.IsNotFound(err) {
+		klog.Errorf("Failed to get backup %s: %v", id, err)
+		return nil, status.Errorf(codes.Internal, "GetBackupByID failed with error %v", err)
+	}
+	if back != nil {
 		err = cloud.DeleteBackup(ctx, id)
-		if err != nil {
+		if err != nil && !cpoerrors.IsNotFound(err) {
 			klog.Errorf("Failed to Delete backup: %v", err)
 			return nil, status.Error(codes.Internal, fmt.Sprintf("DeleteBackup failed with error %v", err))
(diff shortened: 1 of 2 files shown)


kayrus:
/ok-to-test

@immanuwell please set a proper release note in the initial PR message

pkg/csi/cinder/controllerserver.go
kayrus, line 783: I'd suggest to use this logic with the early return. WDYT?
```suggestion
	// If volumeSnapshot object was linked to a cinder backup, delete the backup.
	back, err := cloud.GetBackupByID(ctx, id)
	if err != nil && !cpoerrors.IsNotFound(err) {
		klog.Errorf("Failed to get backup %s: %v", id, err)
		return nil, status.Error(codes.Internal, fmt.Sprintf("GetBackupByID failed with error %v", err))
	}
	if back != nil {
		err = cloud.DeleteBackup(ctx, id)
		if err != nil && !cpoerrors.IsNotFound(err) {
			klog.Errorf("Failed to Delete backup: %v", err)
			return nil, status.Error(codes.Internal, fmt.Sprintf("DeleteBackup failed with error %v", err))
		}
		return &csi.DeleteSnapshotResponse{}, nil
	}
```

immanuwell: got it, done in latest push here

immanuwell:
@kayrus addressed your suggestions, PTAL

pkg/csi/cinder/controllerserver.go
kayrus: @gouthampacha can you double-check whether an early return is feasible here? since this is an ID, there could not be backups or snapshots with the same ID.

gouthampacha: yeah, that's a safe assumption to make

PR #3058, opened 2026-01-23
feat: add support for load balancer, listener, and pool tags annotations

A10ss:
**What this PR does / why we need it**:
This PR adds support for the add tags to load balancer, listener, and pool via the service annotations in cloud-provider-openstack (OCCM).

**Which issue this PR fixes(if applicable)**:
fixes #3032 #2327 

**Special notes for reviewers**:
You can add annotations to the service like:
```
loadbalancer.openstack.org/listener-tags: '["tags-test"]'
loadbalancer.openstack.org/load-balancer-tags: '["tags-test"]'
loadbalancer.openstack.org/pool-tags: '["tags-test"]'
```
Then check if the tags have been added to the service's load blancer(the service's type should be load balancer)

**Release note**:

```release-note
[openstack-cloud-controller-manager] Support adding tags to load balancers, listeners, and pools via Service annotations.
```

PR #3149, opened 2026-07-24
feat: add support for load balancer, listener, and pool tags annotations

### LINKED ISSUE #3032: Hope OCCM could support add tags to LB, listeners, pools
/kind feature

**What happened**:
We have some requirements that need to add some special tags on LB, listeners, pools. But for now, the LB, listeners, pools are created by OCCM automatically and will add a default tag for LB, listeners, pools, we have to add the tags to LB, listeners, pools manually, could we add a feature that could support add tags to LB, listeners, pools?

Could we add some annotations on service, such as:
```yaml
annotations:
    loadbalancer.openstack.org/load-balancer-tags="["tags-1", "tags-2"]"
    loadbalancer.openstack.org/listeners-tags="["tags-1", "tags-2"]"
```

alexanderstephan:
Superceds https://github.com/kubernetes/cloud-provider-openstack/pull/3058.

This PR adds support for the add tags to load balancer, listener, and pool via the service annotations in cloud-provider-openstack (OCCM).

fixes #3032 #2327 

You can add annotations to the service like:
```
loadbalancer.openstack.org/listener-tags: "tags-test"
loadbalancer.openstack.org/load-balancer-tags: "tags-test"
loadbalancer.openstack.org/pool-tags: "tags-test"
```
Then check if the tags have been added to the service's load blancer(the service's type should be load balancer)

**Release note**:

```release-note
[openstack-cloud-controller-manager] Support adding tags to load balancers, listeners, and pools via Service annotations.
```

### DIFF
--- a/pkg/openstack/loadbalancer.go
+++ b/pkg/openstack/loadbalancer.go
@@ -96,6 +96,13 @@ const (
 	defaultProxyHostnameSuffix      = "nip.io"
 	ServiceAnnotationLoadBalancerID = "loadbalancer.openstack.org/load-balancer-id"
 
+	// ServiceAnnotationLoadBalancerTags The lb tags annotation is used to set tags on the loadbalancer resource itself(support json list).
+	ServiceAnnotationLoadBalancerTags = "loadbalancer.openstack.org/load-balancer-tags"
+	// ServiceAnnotationListenerTags The listener tags annotation is used to set tags on the loadbalancer listener resources(support json list).
+	ServiceAnnotationListenerTags = "loadbalancer.openstack.org/listener-tags"
+	// ServiceAnnotationPoolTags The pool tags annotation is used to set tags on the loadbalancer pool resources(support json list).
+	ServiceAnnotationPoolTags = "loadbalancer.openstack.org/pool-tags"
+
 	// Octavia resources name formats
 	servicePrefix  = "kube_service_"
 	lbFormat       = "%s%s_%s_%s"
@@ -146,6 +153,9 @@ type serviceConfig struct {
 	healthMonitorMaxRetries     int
 	healthMonitorMaxRetriesDown int
 	preferredIPFamily           corev1.IPFamily // preferred (the first) IP family indicated in service's `spec.ipFamilies`
+	lbTags                      string
+	listenerTags                string
+	poolTags                    string
 }
 
 type listenerKey struct {
@@ -197,6 +207,20 @@ func getLoadbalancerByName(ctx context.Context, client *gophercloud.ServiceClien
 	return &validLBs[0], nil
 }
 
+// mergeTags merges existedTags and desiredTags, returns true if all desiredTags are in existedTags.
+func mergeTags(existedTags []string, desiredTags []string) (bool, []string) {
+	if len(existedTags) == 0 || existedTags == nil {
+		return false, desiredTags
+	}
+	desiredTagsSet := sets.NewString(desiredTags...)
+	tagSet := sets.NewString(existedTags...)
+	if tagSet.HasAll(desiredTags...) {
+		return true, nil
+	} else {
+		return false, tagSet.Union(desiredTagsSet).List()
+	}
+}
+
 func popListener(existingListeners []listeners.Listener, id string) []listeners.Listener {
 	newListeners := []listeners.Listener{}
 	for _, existingListener := range existingListeners {
@@ -235,7 +259,13 @@ func (lbaas *LbaasV2) createOctaviaLoadBalancer(ctx context.Context, name, clust
 	}
 
 	if svcConf.supportLBTags {
-		createOpts.Tags = []string{svcConf.lbName}
+		var desiredTags []string
+		if len(svcConf.lbTags) == 0 {
+			klog.V(4).Infof("No load balancer tags found from service annotation key: %s", ServiceAnnotationLoadBalancerTags)
+		} else if err := json.Unmarshal([]byte(svcConf.lbTags), &desiredTags); err != nil {
+			klog.Warningf("unmarshal service annotation load balancer tags from key: %s, error: %s", ServiceAnnotationLoadBalancerTags, err)
+		}
+		createOpts.Tags = append([]string{svcConf.lbName}, desiredTags...)
 	}
 
 	if svcConf.flavorID != "" {
@@ -923,16 +953,43 @@ func (lbaas *LbaasV2) ensureOctaviaPool(ctx context.Context, lbID string, name s
 		}
 	}
 
+	var desiredTags []string
+	if len(svcConf.poolTags) == 0 {
+		klog.V(4).Infof("No pools tags found from service annotation key: %s", ServiceAnnotationPoolTags)
+	} else if err := json.Unmarshal([]byte(svcConf.poolTags), &desiredTags); err != nil {
+		klog.Warningf("unmarshal service annotation pools tags from key: %s, error: %s", ServiceAnnotationPoolTags, err)
+	}
+
 	if pool == nil {
 		createOpt := lbaas.buildPoolCreateOpt(listener.Protocol, service, svcConf, name)
 		createOpt.ListenerID = listener.ID
-
+		if svcConf.supportLBTags {
+			createOpt.Tags = desiredTags
+		}
 		klog.InfoS("Creating pool", "listenerID", listener.ID, "protocol", createOpt.Protocol)
+		klog.V(4).Infof("Pool create options: %+v", createOpt)
 		pool, err = openstackutil.CreatePool(ctx, lbaas.lb, createOpt, lbID)
 		if err != nil {
 			return nil, err
 		}
 		klog.V(2).Infof("Pool %s created for listener %s", pool.ID, listener.ID)
+	} else {
+		if svcConf.supportLBTags {
+			// Update tags if needed
+			if len(desiredTags) > 0 {
+				klog.V(4).Infof("Desired pools tags: %+v from service annotation key: %s", desiredTags, ServiceAnnotationPoolTags)
+				if ok, tags := mergeTags(pool.Tags, desiredTags); !ok {
+					klog.V(4).Infof("Will update pools' tags, current pools tags: %+v, desired tags: %+v", pool.Tags, tags)
+					updateOpts := v2pools.UpdateOpts{
+						Tags: &tags,
+					}
+					klog.InfoS("Updating pool tags", "poolID", pool.ID, "listenerID", listener.ID, "lbID", lbID, "tags", desiredTags)
+					if err := openstackutil.UpdatePool(ctx, lbaas.lb, lbID, pool.ID, updateOpts); err != nil {
+						klog.Warningf("Error updating LB pool tags: %v", err)
+					}
+				}
+			}
+		}
 	}
 
 	if lbaas.opts.ProviderRequiresSerialAPICalls {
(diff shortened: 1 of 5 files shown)


pkg/openstack/loadbalancer.go
```diff
@@ -235,7 +259,8 @@ func (lbaas *LbaasV2) createOctaviaLoadBalancer(ctx context.Context, name, clust
 	}
 
 	if svcConf.supportLBTags {
-		createOpts.Tags = []string{svcConf.lbName}
+		desiredTags := cpoutil.SplitTrim(svcConf.lbTags, ',')
```
kayrus: Rename `desiredTags` to specific resource types (`lbTags`, `listenerTags`, `poolTags`) for better code searchability.

pkg/openstack/loadbalancer.go
```diff
@@ -923,16 +948,34 @@ func (lbaas *LbaasV2) ensureOctaviaPool(ctx context.Context, lbID string, name s
...
+			createOpt.Tags = desiredTags
+		}
 		klog.InfoS("Creating pool", "listenerID", listener.ID, "protocol", createOpt.Protocol)
+		klog.V(4).Infof("Pool create options: %+v", createOpt)
 		pool, err = openstackutil.CreatePool(ctx, lbaas.lb, createOpt, lbID)
 		if err != nil {
 			return nil, err
 		}
 		klog.V(2).Infof("Pool %s created for listener %s", pool.ID, listener.ID)
+	} else if svcConf.supportLBTags && len(desiredTags) > 0 {
```
kayrus: Combine with line 913. Consider refactoring to gather all changes in a single `pools.UpdateOpts`, then make one API call instead of multiple updates. cc @yahor-kurachkin

pkg/openstack/loadbalancer.go
```diff
@@ -1177,7 +1221,13 @@ func (lbaas *LbaasV2) buildListenerCreateOpt(ctx context.Context, port corev1.Se
 	}
 
 	if svcConf.supportLBTags {
-		listenerCreateOpt.Tags = []string{svcConf.lbName}
+		// The LB name is always tagged, plus any desired tags from the Service annotation.
+		desiredTags := cpoutil.SplitTrim(svcConf.listenerTags, ',')
+		tags := []string{svcConf.lbName}
+		if _, merged := mergeTags(tags, desiredTags); merged != nil {
```
kayrus: Use `mergeTags` consistently. Line 1148 checks the `ok` return value, but here it's ignored. Also, use `len(merged) > 0` instead of `merged != nil` to make it clear this is a slice check, not a nil-pointer check.

pkg/openstack/loadbalancer.go
```diff
@@ -1756,6 +1811,20 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 	// Make sure LB ID will be saved at this point.
 	lbaas.updateServiceAnnotation(service, ServiceAnnotationLoadBalancerID, loadbalancer.ID)
 
+	if svcConf.supportLBTags {
```
kayrus: Another `svcConf.supportLBTags` check exists at line 1899. Consider combining these tag updates into a single API call.

kayrus (commented):
Thanks for the PR. See my comments.

pkg/openstack/loadbalancer.go
```diff
@@ -902,37 +932,53 @@ func (lbaas *LbaasV2) ensureOctaviaPool(ctx context.Context, lbID string, name s
...
-	// If LBMethod changes, update the Pool with the new value
+	// Resolve the desired LB method, falling back to OCCM's default.
 	var poolLbMethod string
 	if svcConf.poolLbMethod != "" {
 		poolLbMethod = svcConf.poolLbMethod
 	} else {
 		// if LBMethod is not defined, fallback on default OCCM's default method
 		poolLbMethod = lbaas.opts.LBMethod
 	}
-	if pool != nil && pool.LBMethod != poolLbMethod {
```
kayrus, line 913: Please move the pool tag update logic into this existing block for consistency and easier review.

pkg/openstack/loadbalancer.go
```diff
@@ -1826,18 +1891,6 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
 	// save address into the annotation
 	lbaas.updateServiceAnnotation(service, ServiceAnnotationLoadBalancerAddress, addr)
 
-	// add LB name to load balancer tags.
```
kayrus, line 1829: Same here. Please put the LB update tag logic into the old block.

pkg/openstack/loadbalancer.go
```diff
@@ -910,24 +936,39 @@ func (lbaas *LbaasV2) ensureOctaviaPool(ctx context.Context, lbID string, name s
...
-	if pool != nil && pool.LBMethod != poolLbMethod {
-		klog.InfoS("Updating LoadBalancer LBMethod", "poolID", pool.ID, "listenerID", listener.ID, "lbID", lbID)
-		err = openstackutil.UpdatePool(ctx, lbaas.lb, lbID, pool.ID, v2pools.UpdateOpts{LBMethod: v2pools.LBMethod(poolLbMethod)})
-		if err != nil {
-			err = PreserveGopherError(err)
-			msg := fmt.Sprintf("Error updating LB method for LoadBalancer: %v", err)
-			klog.Errorf(msg, "poolID", pool.ID, "listenerID", listener.ID, "lbID", lbID)
-			lbaas.eventRecorder.Event(service, corev1.EventTypeWarning, eventLBLbMethodUnknown, msg)
-		} else {
-			pool.LBMethod = poolLbMethod
```
kayrus, line 922: Nit: Thinking through this: the `pool.LBMethod = poolLbMethod` assignment is now gone. The `pool` pointer is passed to `openstackutil.SeriallyReconcilePoolMembers`, and it wasn't initially clear whether an updated `LBMethod` mattered there. After reviewing that function, it's confirmed the assignment was redundant, the function doesn't rely on the `LBMethod` field.

kayrus (commented):
The PR looks much better now. I left one nit comment. @yahor-kurachkin could you take a look at this PR when you have a chance?

pkg/openstack/loadbalancer.go
alexanderstephan: So, I can mark this as resolved?

kayrus: yes

yahor-kurachkin (commented):
The PR looks good. Since I also working on `loadbalancer.go` at the moment, there will likely merge conflicts. I don't mind to resolving them on my side

alexanderstephan:
@yahor-kurachkin Okay, thanks! Then, I think we can soon proceed to merge this. @kayrus wanted to verify the logic one more time, then we should be good to go.

kayrus:
I have one more concern. If a user sets a custom tag that happens to match another loadbalancer's name, this could cause conflicts. Should we add validation to prevent users from adding the special tag generated by the `GetLoadBalancerName` method?

alexanderstephan:
Yes, I agree. I added a tag sanitizer helper for checking whether a user provided tag is equal to the lb name. If it finds it, it drops it and emits a warning. I think this is better than erroring since it won't block the whole reconcile. The warning can be observed in the logs. I changed this in my latest commit.

kayrus:
@alexanderstephan that won't work. my concern is about users setting a custom tag that matches a different tenant's loadbalancer name. Your validation compares against the current loadbalancer's name, but that name will end up in the tags anyway. The issue is cross-tenant tag collisions, not self-referential ones.

alexanderstephan:
One more issue regarding this: this only reserves this service's own LB-name tag
A tag matching a different LB's name is just an opaque string, so it can't be blocked here. Fully preventing that would mean namespacing the ownership tag with a dedicated prefix. Happy to do that separately if you think it's worth it. @kayrus, what do you think?

Edit: Nevermind, I didn't see your message yet when typing this.

pkg/openstack/loadbalancer.go
```diff
@@ -197,6 +207,66 @@ func getLoadbalancerByName(ctx context.Context, client *gophercloud.ServiceClien
...
+	seen := sets.NewString()
+	result := make([]string, 0, len(tags))
+	for _, t := range tags {
+		if !seen.Has(t) {
+			seen.Insert(t)
+			result = append(result, t)
+		}
+	}
+	return result
+}
```
kayrus: how about moving this to `pkg/util/util.go` and making it generic?
```go
func Unique[T comparable](slice []T) []T {
	keys := make(map[T]struct{})
	result := make([]T, 0, len(slice))
	for _, item := range slice {
		if _, exists := keys[item]; !exists {
			keys[item] = struct{}{}
			result = append(result, item)
		}
	}
	return result
}
```

pkg/openstack/loadbalancer.go
```diff
@@ -197,6 +207,66 @@ func getLoadbalancerByName(ctx context.Context, client *gophercloud.ServiceClien
...
+// comma-separated tags from the given Service annotation value. Any user tag
+// using the reserved servicePrefix is stripped, and duplicate tags are removed
+// (including ones that duplicate the LB name) so no tag can appear twice.
+func withLBNameTag(lbName, annotation string) []string {
+	userTags := stripReservedTags(cpoutil.SplitTrim(annotation, ','))
+	return dedupTags(append([]string{lbName}, userTags...))
+}
+
+// mergeTags merges existedTags and newTags, returns true if all newTags are already in existedTags.
+func mergeTags(existedTags []string, newTags []string) (bool, []string) {
```
kayrus: this function could also be moved to `pkg/util/util.go` with generic support:

```go
func Merge[T comparable](dest, src []T) ([]T, bool) {
	existing := make(map[T]struct{})
	for _, item := range dest {
		existing[item] = struct{}{}
	}

	merged := false
	result := make([]T, len(dest), len(dest)+len(src))
	copy(result, dest)

	for _, item := range src {
		if _, exists := existing[item]; !exists {
			existing[item] = struct{}{}
			result = append(result, item)
			merged = true
		}
	}

	return result, merged
}
```

pkg/openstack/loadbalancer.go
```diff
@@ -910,24 +980,39 @@ func (lbaas *LbaasV2) ensureOctaviaPool(ctx context.Context, lbID string, name s
...
+	poolTags := dedupTags(stripReservedTags(cpoutil.SplitTrim(svcConf.poolTags, ',')))
+
+	if pool != nil {
+		updateOpts := v2pools.UpdateOpts{}
+		if pool.LBMethod != poolLbMethod {
+			updateOpts.LBMethod = v2pools.LBMethod(poolLbMethod)
+		}
+		if svcConf.supportLBTags && len(poolTags) > 0 {
+			klog.V(4).Infof("Desired pool tags: %+v from service annotation key: %s", poolTags, ServiceAnnotationPoolTags)
+			if ok, tags := mergeTags(pool.Tags, poolTags); !ok {
```
kayrus: if you use a new `utils.Merge` now, you need to:

```suggestion
			if ok, tags := utils.Merge(pool.Tags, poolTags); ok {
```

pkg/openstack/loadbalancer.go
```diff
@@ -1099,11 +1184,11 @@ func (lbaas *LbaasV2) ensureOctaviaListener(ctx context.Context, lbID string, na
...
 
 		if svcConf.supportLBTags {
-			if !slices.Contains(listener.Tags, svcConf.lbName) {
-				var newTags []string
-				copy(newTags, listener.Tags)
-				newTags = append(newTags, svcConf.lbName)
-				updateOpts.Tags = &newTags
+			// Ensure the LB name tag plus any desired tags from the Service annotation.
+			listenerTags := withLBNameTag(svcConf.lbName, svcConf.listenerTags)
+			if ok, tags := mergeTags(listener.Tags, listenerTags); !ok {
```
kayrus: if you use a new `utils.Merge` now, you need to:

```suggestion
			if ok, tags := utils.Merge(listener.Tags, listenerTags); ok {
```

pkg/openstack/loadbalancer.go
```diff
@@ -1826,15 +1917,16 @@ func (lbaas *LbaasV2) ensureOctaviaLoadBalancer(ctx context.Context, clusterName
...
 	if svcConf.supportLBTags {
-		lbTags := loadbalancer.Tags
-		if !slices.Contains(lbTags, lbName) {
-			lbTags = append(lbTags, lbName)
-			klog.InfoS("Updating load balancer tags", "lbID", loadbalancer.ID, "tags", lbTags)
-			if err := openstackutil.UpdateLoadBalancerTags(ctx, lbaas.lb, loadbalancer.ID, lbTags); err != nil {
-				return nil, err
+		lbTags := withLBNameTag(lbName, svcConf.lbTags)
+		klog.V(4).Infof("Desired load balancer tags: %v (LB name plus annotation %s)", lbTags, ServiceAnnotationLoadBalancerTags)
+		if ok, tags := mergeTags(loadbalancer.Tags, lbTags); !ok {
```
kayrus: if you use a new `utils.Merge` now, you need to:

```suggestion
		if ok, tags := utils.Merge(loadbalancer.Tags, lbTags); ok {
```

kayrus (commented):
see my new comments

PR #3153, opened 2026-07-29
occm: convert bash based lbaas e2e tests into go tests

kayrus:

old bash based tests are slow and not modular. new go-based tests are modular, fast and provide more logs for debugging.

The original bash tests had a bug where OVN provider failures were silently ignored. When creating an HTTP listener with OVN (which only supports TCP), the `openstack loadbalancer listener create` command would fail with HTTP 501, but the bash script didn't check the exit code and continued executing, causing the tests to pass despite the failure.

```bash
+ tests/e2e/cloudprovider/test-lb-service.sh:test_shared_user_lb:735 :   openstack loadbalancer listener create --protocol HTTP --protocol-port 80 8ff613e8-339f-43a8-9e81-23b11f7c7430
Provider 'ovn' does not support a requested option: OVN provider does not support HTTP protocol (HTTP 501) (Request-ID: req-98dceeba-f6a4-4aca-bbec-a1d4912eb683)
```
Full logs: https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/cloud-provider-openstack/3154/openstack-cloud-controller-manager-ovn-e2e-test-release-136/2082483820640079872/build-log.txt

**Release note**:

```release-note
NONE
```


tests/e2e/cloudprovider/loadbalancer_test.go
```diff
@@ -0,0 +1,636 @@
...
+		err = createNamespace(ginkgo.GinkgoT().Context(), tstCtx)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+		err = createDeployment(ginkgo.GinkgoT().Context(), tstCtx)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	})
+
+	ginkgo.AfterEach(func() {
+		if autoCleanup {
+			cleanupResources(ginkgo.GinkgoT().Context(), tstCtx)
```
valeryia-hurynovich, line 37: If `setupTestContext` or `createDeployment` fails in `BeforeEach`, `tstCtx` is nil here and this will panic. You can add a guard like: `if autoCleanup && tstCtx != nil {`

tests/e2e/cloudprovider/loadbalancer_test.go
```diff
@@ -0,0 +1,636 @@
...
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	gomega.Expect(len(initialNodePorts)).To(gomega.Equal(2), "Should have 2 NodePorts initially")
+	framework.Logf("Initial NodePorts: %v", initialNodePorts)
+
+	initialMemberPorts, err := getMemberPorts(ctx, tstCtx, lbID)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	framework.Logf("Initial member ports: %v", initialMemberPorts)
+
+	// Validate member ports match NodePorts
+	gomega.Expect(len(initialMemberPorts)).To(gomega.Equal(len(initialNodePorts)),
```
valeryia-hurynovich, line 238: The original bash test explicitly failed if the NodePort did not change after update (`if [[ ${member_port} == ${member_ports} ]]`). This version only logs. If the point of this test is to verify Kubernetes assigns a new NodePort and the LB member is updated to match it, so please restore the assertion

tests/e2e/cloudprovider/loadbalancer.go
```diff
@@ -0,0 +1,599 @@
...
+  name: %s
+`, namespace)
+
+	ns, err := parseYAMLNamespace(namespaceYAML)
+	if err != nil {
+		return err
+	}
+
+	_, err = tstCtx.k8sClient.CoreV1().Namespaces().Create(ctx, ns, metav1.CreateOptions{})
+	if err != nil && !strings.Contains(err.Error(), "already exists") {
```
valeryia-hurynovich, line 252: Use `k8serrors.IsAlreadyExists(err)` / `k8serrors.IsNotFound(err)` from k8s.io/apimachinery/pkg/api/errors instead of `strings.Contains(err.Error(), "already exists")`. String matching on error messages is fragile across API versions and locales

tests/e2e/cloudprovider/loadbalancer_test.go
```diff
@@ -0,0 +1,636 @@
...
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/kubernetes/test/e2e/framework"
+)
+
+var _ = ginkgo.Describe("[cloud-provider-openstack] LoadBalancer Service", func() {
+	var tstCtx *testContext
+
+	ginkgo.BeforeEach(func() {
+		var err error
+		tstCtx, err = setupTestContext(ginkgo.GinkgoT().Context())
```
valeryia-hurynovich, line 25: `BeforeEach` re-authenticates to OpenStack and creates all clients before every single test. Move the client setup to `BeforeSuite`/`AfterSuite` and only reset `createdServices`/`createdFIPs`/`createdLBs` slices in `BeforeEach`. This will significantly reduce test setup time

tests/e2e/cloudprovider/loadbalancer.go
```diff
@@ -0,0 +1,599 @@
...
+			return false, err
+		}
+
+		if lb.ProvisioningStatus == "ACTIVE" {
+			consecutiveActive++
+			if consecutiveActive >= 2 {
+				framework.Logf("Load balancer %s is ACTIVE", lbID)
+				return true, nil
+			}
+		} else {
```
valeryia-hurynovich, line 411: `http.DefaultClient `has no timeout. A connection that is established but never returns data will hang the test indefinitely. It could be better to use` &http.Client{Timeout: 10 * time.Second}` for connectivity checks

tests/e2e/cloudprovider/loadbalancer.go
```diff
@@ -0,0 +1,599 @@
...
+// initAPILogger initializes the API debug logger to write to a file
+func initAPILogger() error {
+	dir := filepath.Join(os.TempDir(), "lb-e2e-logs")
+	err := os.MkdirAll(dir, 0755)
+	if err != nil {
+		return fmt.Errorf("failed to create log directory: %w", err)
+	}
+
+	fileName := time.Now().Format("20060102-150405") + "-api-debug.log"
+	logFile, err := os.Create(filepath.Join(dir, fileName))
```
valeryia-hurynovich, line 97: `logFile` is opened here but never stored or closed. If the test process is killed, buffered log data may be lost. Store the `*os.File` in `testContext` and close it in cleanup, or use `defer` in `initAPILogger` with a registry in `BeforeSuite`

valeryia-hurynovich (changes requested):

tests/e2e/cloudprovider/loadbalancer_test.go
```diff
@@ -0,0 +1,636 @@
...
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+	ipAddr, err := waitForServiceAddress(ctx, tstCtx, serviceName)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+	err = waitForAddressAccessible(ctx, ipAddr)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+	framework.Logf("Sending request to service %s", serviceName)
+	resp, err := http.Get(fmt.Sprintf("http://%s", ipAddr))
```
yahor-kurachkin, line 101: Wouldn't it be better to add a context and timeout here instead of using a plain Get request? For example, by using http.NewRequestWithContext?

same for line:152

tests/playbooks/roles/install-cpo-occm/tasks/run-lb-e2e-tests.yaml
```diff
@@ -0,0 +1,160 @@
...
+        -ginkgo.show-node-events \
+        -ginkgo.focus="LoadBalancer Service" \
+        -ginkgo.timeout=1h \
+        -test.timeout=0 2>&1 | tee /tmp/lb-e2e-tests.log
+
+      # Capture the test exit code from the pipeline
+      exit ${PIPESTATUS[0]}
+  timeout: 3900
+  ignore_errors: true  # Don't stop playbook - always collect logs first
+  failed_when: false   # Don't mark as failed yet - check manually later
```
yahor-kurachkin, line 44: What is the point of using both of these annotations? Would it be enough to keep only failed_when?

winiciusallan:
@kayrus are you still planning to work on this? while reading the source code, I thought that would be nice to have go tests instead of bash on OCCM, and I ended up seeing this PR. if not or don't have to do so, I'd be happy to take this further :).

PR #3155, opened 2026-07-31
[cinder-csi-plugin] Configure topology with --with-topology

### LINKED ISSUE #2993: cinder-csi-plugin: use of .Values.csi.provisioner.topology and --with-topology
/kind feature

Follow up from discussion here: https://github.com/kubernetes/cloud-provider-openstack/pull/2985#issuecomment-3291851278

It sounds like the helm value `.Values.csi.provisioner.topology` should be changed to set `--with-topology` for the node and controller plugins, instead of setting the Topology feature gate.

@stephenfin can you confirm if that is correct?

Alternatively maybe a completely new value should be used instead of reusing `.Values.csi.provisioner.topology` , but then the issue and ambiguity remains with having different settings to disable topology in different ways.

Personally I would lean towards the simplest and cleanest change, even if it is breaking behaviour.
But I don't know all the intricacies involved.

234u34k:

The Cinder chart currently applies `csi.provisioner.topology` only to the external-provisioner feature gate, so the value does not control the Cinder driver topology capabilities.

This uses the existing value for `--with-topology` in both Cinder driver containers and removes the explicit external-provisioner feature-gate argument. The existing `"true"` default is unchanged. The pinned [external-provisioner v5.3.0](https://github.com/kubernetes-csi/external-provisioner/blob/v5.3.0/pkg/features/features.go#L60-L66) enables its GA Topology gate by default, while `"false"` now makes the driver stop advertising and returning topology information.

The affected documentation is updated and an explicit disabled-values render is added to chart CI. This follows #2998, which established the required driver behavior.

Fixes #2993

Validated with Helm v3.10.0:

- `helm lint` passes with defaults and every `ci/*-values.yaml` file.
- The default render passes `--with-topology=true` to both driver containers.
- The disabled render passes `--with-topology=false` to both driver containers.
- Neither render sets the external-provisioner `Topology` feature gate.
- Existing driver `extraArgs` retain their override precedence.
- `git diff --check origin/master...HEAD` passes.

**Release note**:

```release-note
[cinder-csi-plugin] Fixed the Helm chart so `csi.provisioner.topology: "false"` disables topology awareness in both the controller and node plugins.
```

### DIFF
--- a/charts/cinder-csi-plugin/Chart.yaml
+++ b/charts/cinder-csi-plugin/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v1.36.0
 description: Cinder CSI Chart for OpenStack
 name: openstack-cinder-csi
-version: 2.36.2
+version: 2.36.3
 home: https://github.com/kubernetes/cloud-provider-openstack
 icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
 maintainers:
--- a/charts/cinder-csi-plugin/ci/topology-disabled-values.yaml
+++ b/charts/cinder-csi-plugin/ci/topology-disabled-values.yaml
@@ -1,0 +1,3 @@
+csi:
+  provisioner:
+    topology: "false"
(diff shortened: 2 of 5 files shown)


gouthampacha (approved):
/lgtm

thanks @234u34k.

stephenfin (approved):

PR #3159, opened 2026-08-09
[k8s-keystone-auth] Preserve config after failed reload

immanuwell:

A malformed live policy or sync ConfigMap logs a parse error, but still replaces working config with empty or partial state
As a result, auth can start denying requests or syncing can silently stop

This keeps the last good config when parsing or validation fails. Valid reloads still work, and failures use the existing retry queue.

N/A

Special notes for reviewers:

Repro:

1. Start with a valid policy or sync ConfigMap.
2. Set `policies` to `{`, or set `syncConfig` to malformed YAML.
3. Current code logs the parse failure, then logs that config was updated and replaces the active state.

Tests:

`go test -race ./pkg/identity/keystone`

`go test ./pkg/...`

`make check`

Release note:

```release-note
NONE
```

### DIFF
--- a/pkg/identity/keystone/keystone.go
+++ b/pkg/identity/keystone/keystone.go
@@ -171,16 +171,16 @@ func (k *Auth) processNextItem() bool {
 	return true
 }
 
-func (k *Auth) updatePolicies(cm *apiv1.ConfigMap, key string) {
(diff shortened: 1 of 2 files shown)


stephenfin (commented):
/approve
/lgtm

Issue #1935, opened 2022-07-14
`not found` checks for LoadBalancers do not cover all possible cases


ialidzhikov:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

**What happened**:
Recently with https://github.com/kubernetes/cloud-provider-openstack/pull/1913 we addressed https://github.com/kubernetes/cloud-provider-openstack/issues/1912. It seems that this is not enough to address all possible cases for LoadBalancer deletion that does not exists.

The issue I see is that in the `GetLoadBalancer` func (and in other places) we have 2 different ways to fetch the LoadBalancer - by ID, if its `loadbalancer.openstack.org/load-balancer-id` is set; otherwise by name:

https://github.com/kubernetes/cloud-provider-openstack/blob/38938aeb216dc58adebb2367f79b84988ca80b72/pkg/openstack/loadbalancer.go#L674-L684

The problem I see is that both of the ways different `not found` errors:
- `openstackutil.GetLoadbalancerByID` func returns a error that can be checked with the `cpoerrors.IsNotFound(err)` func
- `getLoadbalancerByName` returns `cpoerrors.ErrNotFound` that is not covered by `cpoerrors.IsNotFound(err)`

https://github.com/kubernetes/cloud-provider-openstack/blob/38938aeb216dc58adebb2367f79b84988ca80b72/pkg/util/errors/errors.go#L42-L58

The `cpoerrors.IsNotFound(err)` func currently checks only for gophercloud errors (returned by the OpenStack API) and does not check for `cpoerrors.ErrNotFound`.

**What you expected to happen**:
Non-existing LoadBalancers to be properly ignored in all cases.

**How to reproduce it**:
1. Create a LoadBalancer that fails to be created. For example with LB class set to a non-existing class (loadbalancer.openstack.org/class=foo).

2. Make sure that as expected the LoadBalancer fails to be created with:

```
  Warning  SyncLoadBalancerFailed  4s (x3 over 19s)  service-controller  Error syncing load balancer: failed to ensure load balancer: invalid loadbalancer class "foo"
```

3. Delete the Service

4. Make sure that the deletion fails with:

```
  Warning  SyncLoadBalancerFailed  3s                 service-controller      Error syncing load balancer: failed to check if load balancer exists before cleanup: failed to find object
```

**Anything else we need to know?**:

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: v1.24.2
- OpenStack version:
[...]

PR #3083, opened 2026-03-13
[occm]: fix openstack getServerByName return error code

tyiying:
-
Additional notes:

**What this PR does / why we need it**:
after upgraded occm version from v1.29 to v1.32, we found occm cannot delete node from k8s node list anymore after we remove the node from openstack

**Which issue this PR fixes(if applicable)**:
fixes #2999 - we have fixed this issue in our product using nokia repo, but would like to fix it in the upstream. 

**Special notes for reviewers**:
After delete the node from openstack, we expect occm delete the node from k8s node list. But we still see the node in the node list:
[cloud-admin@hotel-ncs03-control-01 ~]$ sudo kubectl get node
Enter login password:
NAME STATUS ROLES AGE VERSION
hotel-ncs03-control-01 Ready 44h v1.32.8
hotel-ncs03-control-02 Ready 44h v1.32.8
hotel-ncs03-control-03 Ready 44h v1.32.8
hotel-ncs03-edge-01 Ready 44h v1.32.8
hotel-ncs03-edge-02 Ready 44h v1.32.8
hotel-ncs03-storage-01 NotReady 44h v1.32.8             <= this should be gone
hotel-ncs03-storage-02 Ready 44h v1.32.8
hotel-ncs03-storage-03 Ready 44h v1.32.8
hotel-ncs03-worker-01 Ready 44h v1.32.8
hotel-ncs03-worker-02 Ready 44h v1.32.8
[cloud-admin@hotel-ncs03-control-01 ~]$

Here are the logs we see:
E0922 11:38:01.217813 1 node_lifecycle_controller.go:156] error checking if node hotel-ncs03-storage-01 exists: failed to find object

I did a code comparison between 1.29 and 1.32, I can see the return code from getInstance() function was causing the different behavior:
The return code is expected to be cloudprovider.InstanceNotFound for both ccm 1.29 and ccm 1.32
// InstanceExists indicates whether a given node exists according to the cloud provider
func (i *InstancesV2) InstanceExists(ctx context.Context, node *v1.Node) (bool, error) {
_, err := i.getInstance(ctx, node)
if err == cloudprovider.InstanceNotFound {
klog.V(6).Infof("instance not found for node: %s", node.Name)
return false, nil
}

**Release note**:
```release-note
occm: properly handle removed nodes
```

PR #3160, opened 2026-08-13
occm: Fix node with `providerID=""` deletion when underlying Server does not exist

### LINKED ISSUE #3114: Node is not deleted from the cluster, after being deleted from openstack
/kind bug

**What happened**:

When deleting a baremetal node from openstack, occm doesn't detect the deletion, and the node remains in the cluster as NotReady.

The logs say: `error checking if node <node-name> exists: failed to find object`

I expect to see a log line that the node was deleted, and then for the node to not be in the cluster anymore.

It seems this is the same issue reported in: https://github.com/kubernetes/cloud-provider-openstack/issues/2999

I tried the PR proposed in the ticket (https://github.com/nokia/cloud-provider-openstack/pull/1/changes), and it fixed the issue for us.

**Environment**:
- openstack-cloud-controller-manager version: v1.35.0

r4mek:
Currently `getServerByName()` returns `errors.ErrNotFound`, whereas the caller only handles `cloudprovider.InstanceNotFound`. This PR now returns a wrapped error containing `cloudprovider.InstanceNotFound` so that its caller `InstanceExists()` can suppress `cloudprovider.InstanceNotFound` error.
This is crucial as the `cloud-controller-manager` expects `InstancesV2.InstanceExists()` to suppress `cloudprovider.InstanceNotFound` errors and return `exist=false, err=nil`.
Further details are discussion in the linked issue.

There was a similar PR raised earlier which targeted loadbalancers error handling: #1935 

fixes #3114 

I have decided to wrap the error with `cloudprovider.InstanceNotFound` and `errors.ErrNotFound` rather than just returning `cloudprovider.InstanceNotFound` so that downstream systems depending on `errors.ErrNotFound` do not fail.

**Release note**:

```release-note
[openstack-cloud-controller-manager] An issue preventing Node with providerID="" to be deleted when the corresponding OpenStack Server does no longer exists is now fixed.
```

### DIFF
--- a/pkg/openstack/instances.go
+++ b/pkg/openstack/instances.go
@@ -208,7 +208,7 @@ func getServerByName(ctx context.Context, client *gophercloud.ServiceClient, nam
 	}
 
 	if len(serverList) == 0 {
-		return nil, errors.ErrNotFound
+		return nil, fmt.Errorf("%w: %w", cloudprovider.InstanceNotFound, errors.ErrNotFound)
 	}
 
 	return &serverList[0], nil


r4mek:
closing this PR in favor of #3083 
/close

PR #3164, opened 2026-08-25
[k8s-keystone-auth] Don't log tokens

stephenfin:

The Keystone authentication webhook handler writes the raw bearer token submitted in every TokenReview to the log stream when verbosity is 4 or higher. It is pretty common to set this level (-v=4) while e.g. debugging. Additionally, it is common for logs to be sent to a central store, with different access controls than the main cluster. A user with access to this store can use the captured token for replay attacks for any user that authenticates against the cluster until the token's TTL.

Stop logging the token and avoid all of this. The combo of user, any error and the usual timestamp etc. should be more than sufficient.

(none)

**Release note**:

```release-note
NONE
```


mandre (commented):
/lgtm

PR #3165, opened 2026-08-26
[occm] Reduce ClusterRole scope

stephenfin:

A collection of 3 changes, all intended to reduce the cluster-wide access granted by the `system:cloud-controller-manager` ClusterRole.

- `[occm] Limit verbs for node access`: OCCM does not need to create or delete nodes
- `[occm] Prevent access to PVs`: OCCM does not need access to PVs
- `[occm] Add new Role for secrets`: OCCM does not need cluster-wide secret access. It only needs access to the configuration secret (`kube-system / cloud-config` if using the standard manifests).

(none)

**Release note**:

```release-note
[openstack-cloud-controller-manager] Reduce scope of `system:cloud-controller-manager` cluster role
```


winiciusallan:
/lgtm

looks like this needs a bump to a new chart version

charts/openstack-cloud-controller-manager/templates/clusterrole.yaml
```diff
@@ -29,7 +29,7 @@ rules:
   resources:
   - nodes
   verbs:
-  - '*'
+  - 'get,list,watch,patch,update'
```
mandre: Is that even valid? I think it should be a list:

```suggestion
  - get
  - list
  - watch
  - patch
  - update
```

manifests/controller-manager/cloud-controller-manager-roles.yaml
```diff
@@ -26,7 +26,7 @@ items:
     resources:
     - nodes
     verbs:
-    - '*'
+    - 'get,list,watch,patch,update'
```
mandre: ditto

mandre (changes requested):

charts/openstack-cloud-controller-manager/templates/clusterrole.yaml
stephenfin: It absolutely isn't. Bad `yq` invocation on my end :facepalm:

manifests/controller-manager/cloud-controller-manager-roles.yaml
stephenfin: done

mandre (commented):
/lgtm
/approve

PR #2628, opened 2024-07-19
[occm] Add Openstack server hostId as k8s node label

chess-knight:
**What this PR does / why we need it**:
Adds OpenStack server hostId to the node labels via the AdditionalLabels feature. It also introduces label name `topology.openstack.org/host-id` which can be discussed further.

**Which issue this PR fixes(if applicable)**:
fixes #2579

**Special notes for reviewers**:

**Release note**:

```release-note
NONE
```

PR #3166, opened 2026-08-26
[occm] Add OpenStack server host ID as node label

### LINKED ISSUE #2579: [occm] Add Openstack server hostId as k8s node label

/kind feature

**What happened**:
This issue follows a discussion in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1605, as a request to automatically label nodes with underlying hostId information, so e.g. workload can be scheduled on different physical hosts. It can be used as a topology differentiator when all other topology labels are the same.

Issue https://github.com/kubernetes/cloud-provider/issues/67 is closed and potentially resolved by https://github.com/kubernetes/kubernetes/pull/123223 so now based on comment https://github.com/kubernetes/cloud-provider/issues/67#issuecomment-2003176563, AdditionalLabels with hostId information can be added to InstanceMetadata:
https://github.com/kubernetes/cloud-provider-openstack/blob/dab0f067249eea1051e3bb57ed05d2cc3e326aa8/pkg/openstack/instancesv2.go#L136-L142
This information should be available in the server struct AFAIK, so there should not be too much work I think now.

One potential issue that I see is live migration, e.g. see https://github.com/kubernetes/cloud-provider-openstack/issues/1801, where occm will have to update the label because hostId will change.

**Environment**:
- OpenStack version:
- Others:

stblatzheim:

Adds the OpenStack server `hostId` as the
`topology.openstack.org/host-id` Kubernetes node label when Nova
returns a non-empty value.

The label provides a project-scoped, opaque representation of the
underlying compute host. It can be used with topology spread
constraints and pod anti-affinity to distribute workloads across
different OpenStack compute hosts.

The label is applied during cloud node initialization. It is currently
not reconciled after a live migration, which is documented as a known
limitation.

This PR supersedes #2628 and incorporates the latest review feedback.
Credit for the original implementation goes to @chess-knight.

Fixes #2579

@kayrus  as requested a new PR with the requested changes - original PR was #2628 

**Release note**:

```release-note
[openstack-cloud-controller-manager] Nodes are now labeled with `topology.openstack.org/host-id` when Nova returns a non-empty server host ID, enabling topology-aware workload distribution across OpenStack compute hosts.
```


winiciusallan:
while this looks like a desired feature, I saw a lot of discussions in the linked issue about the capacity of reconciling node labels to avoid drift in case, for example, the instance has been migrated.

doesn't the node controller handle label reconciling?

stblatzheim:
@winiciusallan I think reconciling node labels is out of scope here.
As you said there was a lot of discussion regarding this and the main point was that it is not in the intend of kubernetes to have such live migrations. The normal way would be to drain the node and spawn a new node.

Also the affinitys for scheduling pods are done at pod creation time => Even a changed node label afterwards would just be taken into account for new pods.

Maybe we can raise a new issue for updating node label on reconcile afterwards, atm we don't even have node labels to work on ...

kayrus:
> As you said there was a lot of discussion regarding this and the main point was that it is not in the intend of kubernetes to have such live migrations. The normal way would be to drain the node and spawn a new node.

there could be system migration to drain the hypervisor for maintenance on the openstack level.

stblatzheim:
@kayrus 

You are right that an OpenStack operator may live-migrate instances while
draining a hypervisor for maintenance, independently of Kubernetes.

I checked the current cloud-node-controller implementation used here
(`k8s.io/cloud-provider v0.36.0`). `AdditionalLabels` are applied only during
the initial synchronization of a node carrying the external-cloud-provider
taint. The periodic node status update calls `InstanceMetadata()` again, but
currently applies only `NodeAddresses`.

Therefore, this PR provides the OpenStack-specific desired state and labels
newly initialized nodes, while backfilling existing nodes and updating the
label after a migration require a generic change in
`kubernetes/kubernetes/staging/src/k8s.io/cloud-provider`.

There is no strict runtime dependency between the two changes: the generic
controller can reconcile the labels returned in `AdditionalLabels`, including
creating a missing label. Before this provider starts returning the host ID,
that reconciliation is simply a no-op.

There have already been two upstream attempts addressing this:
kubernetes/kubernetes#130888 and kubernetes/kubernetes#131096. I propose to
create a focused follow-up based on #130888 which periodically adds or updates
provider-supplied `AdditionalLabels`.

This also matches the previous conclusion in #2579 that exposing HostID is an
immediate provider-side improvement, while continuous reconciliation is a
generic and somewhat independent cloud-node-controller concern.

Existing pods will still not be automatically rescheduled after a migration,
but updating the label makes the changed placement observable and ensures that
new pods are scheduled against the current host topology.

Would you be comfortable merging this provider-side prerequisite while the
generic reconciliation is tracked in a linked upstream PR, or would you prefer
the upstream PR to be opened first?

kayrus:
@stblatzheim I don't mind merging this right now. Need to discuss this with other maintainers
/assign @zetaab @stephenfin @mandre

zetaab:
at least I have not seen permission in our system that allows normal user to see in which host id the node is running?

stblatzheim:
@zetaab `hostId` is different from the admin-only `OS-EXT-SRV-ATTR:host` and `OS-EXT-SRV-ATTR:hypervisor_hostname` fields.

Nova includes `hostId` in the regular server details response available to the server owner. It is an opaque value derived from the physical host and the `project_id`, so:

* it does not expose the compute hostname;
* it cannot be correlated across different projects;
* it can only be used to determine whether two instances within the same project are running on the same or different physical hosts.

The distinction is documented in the [[Nova Compute API](https://docs.openstack.org/api-ref/compute/)](https://docs.openstack.org/api-ref/compute/): `hostId` is part of the normal response, while `OS-EXT-SRV-ATTR:host` is explicitly marked as available to administrative users only.

The implementation does not call an additional or administrative API. It reads `srv.HostID` from the same server response that `InstanceMetadata()` already retrieves.

This can also be verified with non-admin project credentials:

```bash
openstack server show <server-id> -f value -c hostId
```

pkg/openstack/instances.go
```diff
@@ -285,6 +287,16 @@ func srvInstanceType(ctx context.Context, client *gophercloud.ServiceClient, srv
...
 
+func getAdditionalLabels(srv *servers.Server) map[string]string {
+	if srv.HostID == "" {
+		return nil
+	}
+
+	return map[string]string{
+		labelHostID: srv.HostID,
+	}
+}
```
kayrus, line 299: let's do this just in case, see #2634
```suggestion
func getAdditionalLabels(srv *servers.Server) map[string]string {
	hostID := util.SanitizeLabel(srv.HostID)
	if hostID == "" {
		return nil
	}

	return map[string]string{
		labelHostID: hostID,
	}
}
```

kayrus:
here is a reference how hostid is generated: https://github.com/openstack/nova/blob/2589095f39f6adf994d386744a4320d461c431cc/nova/utils.py#L1075-L1079

pkg/openstack/instances.go
stblatzheim: I personally dont think that there can be none unicode characters. The hostID value is a hash and non user-generated like in the linked PR.

 metadata:
  labels:
    topology.openstack.org/host-id: 29d3c8c896a45aa4c34e52247875d7fefc3d94bbcc9f622b5d204362

mandre (commented):
Thanks, I think it's quite useful to surface the HostID in the node object.

I've recently looked into the HostID implementation in detail when we did port binding in ORC, and can confirm is not discoverable by the user other than scheduling an instance and inspecting its HostID value.

Agreed that keeping the host ID updated after a migration for example is a separate concern (and I see you have [a PR for this](https://github.com/kubernetes/kubernetes/pull/141616/)). Given you've also documented the limitation, I think it's acceptable.

Before we merge, would you be able to squash the commits, and credit the original author of https://github.com/kubernetes/cloud-provider-openstack/pull/2628?

mandre:
> @stblatzheim I don't mind merging this right now. Need to discuss this with other maintainers /assign @zetaab @stephenfin @mandre

@kayrus I'm not yet a maintainer, but we can remediate that :wink: 
I've recently freed up time and plan to dedicate some of it to CPO.

winiciusallan (approved):
thanks for the follow-up PR on kubernetes repo @stblatzheim! lgtm

kayrus:
@stblatzheim can you file a corresponding feature request regarding the hostid update to https://github.com/kubernetes/cloud-provider ?

stblatzheim:
Thanks. The `kubernetes/cloud-provider` repository has issues disabled and asks contributors to file issues against the main `kubernetes/kubernetes` repository.

I filed the corresponding feature request here:

https://github.com/kubernetes/kubernetes/issues/141631

The proposed implementation is available here:

https://github.com/kubernetes/kubernetes/pull/141616

PR #3169, opened 2026-08-27
Rework `check` Makefile target

stephenfin:

Rework the check target to run both golangci-lint and helm's chart-testing (ct) lint command. This lets us reproduce locally what is running in CI.

Additionally, remove an irrelevant header from the file - the Golang CTI has not been relevant in OpenStack, let along anywhere else, for a very long time - as well as a load of PTI-related or dead targets.

(none)

**Release note**:

```release-note
NONE
```


Makefile
```diff
@@ -80,8 +75,39 @@ $(BUILD_CMDS): $(SOURCES)
...
+ifdef CI
+	go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION) run \
+		-v --max-same-issues=50 --timeout=20m
+else
+	mkdir -p ~/.cache/golangci-lint/$(GOLANGCI_LINT_VERSION)
+	$(CONTAINER_ENGINE) run -t --rm \
+		--volume $(shell pwd):/data \
+		--volume ~/.cache/golangci-lint/$(GOLANGCI_LINT_VERSION):/root/.cache \
+		--workdir /data \
+		--env GOFLAGS="-tags=acceptance" \
```
mandre, line 80: There are no files in the repo that use the `//go:build acceptance` build tag. Do we need this? Also, this differs from the command that runs in CI.

Makefile
```diff
@@ -80,8 +75,39 @@ $(BUILD_CMDS): $(SOURCES)
...
+#
+#   directory prefix . does not contain main module or its selected dependencies
+#
+#   failed to initialize build cache at /root/.cache/golangci-lint: mkdir /root/.cache/golangci-lint: permission denied
+#
+# you probably have to fix the SELinux security context for root directory plus your cache
+#
+#   chcon -Rt svirt_sandbox_file_t .
+#   chcon -Rt svirt_sandbox_file_t ~/.cache/golangci-lint
+check:
```
mandre, line 70: Is it intentional to drop the `work` prerequisite?

Makefile
stephenfin: Nope. This was copy-pasta from gophercloud

Makefile
stephenfin: Yes, it's unnecessary. In fact, I can remove this from everywhere. Let me do that.

stephenfin:
This PR has grown a bit now. Turns out we had a lot of targets that were dead or purely there for PTI purposes. The commits view is probably helpful in reviewing these.

stephenfin:
/close

I'm going to break this up. It's gotten too large.

stephenfin:
Split into https://github.com/kubernetes/cloud-provider-openstack/pull/3208 and https://github.com/kubernetes/cloud-provider-openstack/pull/3209

PR #3170, opened 2026-08-27
[magnum-auto-healer] Add certificate validation

stephenfin:

The controller always sets InsecureSkipVerify. The two modern Magnum drivers in use today both use CAPI and thus kubeadm under the hood. kubeadm injects nodes internal IPs as SANs into the certs it generates, meaning certificate validation should pass when using the same root cert. Thus, we add two new configuration knobs: one to point to a CA file to use (defaulting to the well-known location k8s mounts certs to) and another to explicitly disable verification as an escape hatch.

(none)

**Release note**:

```release-note
[magnum-auto-healer] Certificate validation is now enabled by default when using HTTPS
```

### DIFF
--- a/pkg/autohealing/healthcheck/plugin_endpoint.go
+++ b/pkg/autohealing/healthcheck/plugin_endpoint.go
@@ -157,7 +157,7 @@ func (check *EndpointCheck) Check(ctx context.Context, node NodeInfo, controller
 			caPool, err = cert.NewPool(check.CAFile)
 			if err != nil {
 				log.Errorf("Node %s, failed to load CA file %s, error: %v", nodeName, check.CAFile, err)
-				return false
+				return check.checkDuration(ctx, node, controller, false)
 			}
 			tlsConfig.RootCAs = caPool
 		}
--- a/pkg/autohealing/healthcheck/plugin_endpoint.go
+++ b/pkg/autohealing/healthcheck/plugin_endpoint.go
@@ -19,6 +19,7 @@ package healthcheck
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"net/http"
 	"os"
@@ -27,12 +28,14 @@ import (
 	"time"
 
 	"github.com/mitchellh/mapstructure"
+	"k8s.io/client-go/util/cert"
 	log "k8s.io/klog/v2"
 )
 
 const (
 	EndpointType = "Endpoint"
 	TokenPath    = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+	CAPath       = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
 	TimeLayout   = "2006-01-02 15:04:05"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/autohealing/healthcheck/plugin_endpoint.go
```diff
@@ -140,9 +150,18 @@ func (check *EndpointCheck) Check(ctx context.Context, node NodeInfo, controller
...
-		tr := &http.Transport{
-			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		tlsConfig := &tls.Config{InsecureSkipVerify: check.TLSInsecure}
+		if check.CAFile != "" {
+			var caPool *x509.CertPool
+			var err error
+			caPool, err = cert.NewPool(check.CAFile)
+			if err != nil {
+				log.Errorf("Node %s, failed to load CA file %s: %v", nodeName, check.CAFile, err)
+				return true
```
winiciusallan: shouldn't it return `false` since a fail has happened when loading the CA?

pkg/autohealing/healthcheck/plugin_endpoint.go
```diff
@@ -140,9 +150,18 @@ func (check *EndpointCheck) Check(ctx context.Context, node NodeInfo, controller
 	protocol := strings.ToLower(check.Protocol)
 	switch protocol {
 	case "https":
-		tr := &http.Transport{
-			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		tlsConfig := &tls.Config{InsecureSkipVerify: check.TLSInsecure}
+		if check.CAFile != "" {
```
winiciusallan: do we need to load `CAFile` if `TLSInsecure` is set?

stephenfin: Good point. I've updated this to ignore it.

pkg/autohealing/healthcheck/plugin_endpoint.go
stephenfin: I'm conflicted on this. On one hand, this feels like an error that implies an unhealthy node. On the other, we currently return `true` if we require a token but can't read it.

I've changed it to `false` since it feels more correct to me.

stephenfin:
I've just realised I already did this :joy: 

https://github.com/kubernetes/cloud-provider-openstack/pull/3110

I prefer this version though.

pkg/autohealing/healthcheck/plugin_endpoint.go
mandre: How about we call `checkDuration()` so that we respect the `unhealthy-duration` grace period?

pkg/autohealing/healthcheck/plugin_endpoint.go
```diff
@@ -140,9 +150,18 @@ func (check *EndpointCheck) Check(ctx context.Context, node NodeInfo, controller
...
-		tr := &http.Transport{
-			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		tlsConfig := &tls.Config{InsecureSkipVerify: check.TLSInsecure}
+		if check.CAFile != "" && !check.TLSInsecure {
+			var caPool *x509.CertPool
+			var err error
+			caPool, err = cert.NewPool(check.CAFile)
+			if err != nil {
+				log.Errorf("Node %s, failed to load CA file %s, error: %v", nodeName, check.CAFile, err)
+				return false
```
mandre: As an alternative to returning `false` right away, meaning we trigger a node repair, how about we instead call `checkDuration()` so that we respect the `unhealthy-duration` grace period?
```suggestion
				return check.checkDuration(ctx, node, controller, false)
```

stephenfin: Good idea

mandre (commented):
/lgtm
/approve

PR #3171, opened 2026-08-28
[occm] test: add loadbalancer servicemonitor regression

jihyun-huh:

Hi maintainers, 
this PR adds renderering level regression coverage for the OCCM Helm chart ServiceMonitor behavior fixed by #3039

currently, the existing chart CI runs `ct lint` with the default values, where `serviceMonitor.enabled` is disabled. and it therefore does not exercise or semantically validate the ServiceMonitor, metrics Service, and DaemonSet wiring.

The new tests verify this things:
- add metric resources and public metrics access remain disabled by default
- enabling `serviceMonitor` creates matching Service and ServiceMonitor resources
- the Service selects the OCCM DaemonSet pods
- update document that deprecated `bearerTokenFile` authentication is not now used
- update OCCM exposes and authorizes `/metrics` only when monitoring is enabled

references:
- https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/tests/controller-servicemonitor_test.yaml
- https://github.com/kubernetes-sigs/descheduler/blob/master/.github/workflows/helm.yaml
- https://github.com/kubernetes-sigs/descheduler/blob/master/charts/descheduler/tests/servicemonitor_test.yaml

Follow-up to #3039, adds regression coverage for #3144

`helm-unittest` is pinned to the immutable `v1.0.3` commit, because this is the latest release compatible with the Helm `v3.10.0` currently used by the chart workflow.

The suite passes against current `master` branch. 
against the parent of #3039, it fails on the previous selector mismatch, deprecated bearer token configuration, and missing metrics authorization flag.

also, this is a renderering level test and does not require a Kubernetes, Prometheus, or OpenStack environment.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/helm/openstack-cloud-controller-manager/servicemonitor_test.yaml
+++ b/tests/helm/openstack-cloud-controller-manager/servicemonitor_test.yaml
@@ -1,0 +1,105 @@
+suite: OCCM ServiceMonitor
+templates:
+  - templates/clusterrolebinding-sm.yaml
+  - templates/daemonset.yaml
+  - templates/service-sm.yaml
+  - templates/servicemonitor.yaml
+
+release:
+  name: test-release
+  namespace: test-namespace
+
+chart:
+  version: 1.2.3
+  appVersion: v4.5.6
+
+tests:
+  - it: keeps the metrics endpoint private by default
+    asserts:
+      - hasDocuments:
+          count: 0
+        template: templates/servicemonitor.yaml
+      - hasDocuments:
+          count: 0
+        template: templates/service-sm.yaml
+      - hasDocuments:
+          count: 0
+        template: templates/clusterrolebinding-sm.yaml
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: --bind-address=127.0.0.1
+        template: templates/daemonset.yaml
+      - notContains:
+          path: spec.template.spec.containers[0].args
+          content: --authorization-always-allow-paths=/metrics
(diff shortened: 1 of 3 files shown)


jihyun-huh:
Hi @gouthampacha @kayrus, could one of you please add `/ok-to-test` when you have a chance? This PR only adds Helm regression coverage for the ServiceMonitor fix merged in #3039. Thanks!

tests/helm/openstack-cloud-controller-manager/servicemonitor_test.yaml
```diff
@@ -0,0 +1,105 @@
...
+      - notContains:
+          path: spec.template.spec.containers[0].args
+          content: --authorization-always-allow-paths=/metrics
+        template: templates/daemonset.yaml
+
+  - it: wires the ServiceMonitor to the OCCM metrics endpoint
+    set:
+      serviceMonitor.enabled: true
+    asserts:
+      - isKind:
```
mandre, line 44: Should we also check if the ClusterRoleBinding is being created?

I think the following assertion should do:
```yaml
  - isKind:
      of: ClusterRoleBinding
    template: templates/clusterrolebinding-sm.yaml
```

jihyun-huh: sure, that makes sense!
and thanks for your review, i just updated on it

mandre (commented):
/lgtm
/approve

PR #3181, opened 2026-09-04
[cinder-csi-plugin] Reduce caps of manifests, charts

stephenfin:

Restrict capabilities on the node and controller pods, along with their respective sidecar containers. Also set `dnsPolicy: ClusterFirstWithHostNet` for the nodeplugin, since that uses `hostNetwork: true` and set `priorityClassName` for both nodeplugin and controllerplugin to ensure pods don't get evicted under memory pressure, respectively.

Commits:

- `cinder-csi-plugin: Add note on why hostNetwork is required`
- `cinder-csi-plugin: Drop redundant capabilities add`
- `cinder-csi-plugin: Add securityContext for node plugin sidecars`
- `cinder-csi-plugin: Add securityContext for all controller plugin containers`
- `cinder-csi-plugin: Set dnsPolicy`
- `cinder-csi-plugin: Set priorityClassName`
- `cinder-csi-plugin: Bump cinder-csi-plugin chart to 2.36.3`

(none)

**Release note**:

```release-note
[cinder-csi-plugin] Changes to default manifests, charts

The following changes have been made to provided manfiests and charts:

- `securityContext` is now set by default for all pods.
- `dnsPolicy` is now set to `ClusterFirstWithHostNet` for the nodeplugin
- `priorityClassName` is now set to `system-node-critical` for nodeplugin
  and to `system-cluster-critical` for controllerplugin to ensure node
  drain works as expected.
- The `priorityClassName` chart variable is now ignored and has been replaced
  by `csi.plugin.controllerPlugin.priorityClassName` and
  `csi.plugin.nodePlugin.priorityClassName` variables.
```

### DIFF
--- a/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
+++ b/charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
@@ -249,8 +249,8 @@ spec:
       hostAliases:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-    {{- if .Values.priorityClassName }}
-      priorityClassName: {{ .Values.priorityClassName }}
+    {{- if .Values.csi.plugin.controllerPlugin.priorityClassName }}
+      priorityClassName: {{ .Values.csi.plugin.controllerPlugin.priorityClassName }}
     {{- end }}
     {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
--- a/charts/cinder-csi-plugin/values.yaml
+++ b/charts/cinder-csi-plugin/values.yaml
@@ -96,12 +96,13 @@ csi:
       # Optional additional labels to add to the nodePlugin Pods.
       podLabels: {}
       podSecurityContext: {}
-      securityContext: {}
-        # capabilities:
-        #   drop:
-        #   - ALL
-        # seccompProfile:
-        #   type: RuntimeDefault
+      securityContext:
+        # Images use distroless/static:latest which runs as root; runAsNonRoot
+        # and allowPrivilegeEscalation cannot be set until upstream switches to
+        # a nonroot base image.
+        readOnlyRootFilesystem: true
+        seccompProfile:
+          type: RuntimeDefault
       affinity: {}
       nodeSelector: {}
       tolerations:
--- a/charts/cinder-csi-plugin/values.yaml
+++ b/charts/cinder-csi-plugin/values.yaml
@@ -137,13 +137,10 @@ csi:
         # runAsGroup: 65532
         # fsGroup: 65532
         # fsGroupChangePolicy: OnRootMismatch
-      securityContext: {}
-        # capabilities:
-        #   drop:
-        #   - ALL
-        # seccompProfile:
-        #   type: RuntimeDefault
-        # readOnlyRootFilesystem: true
+      securityContext:
+        readOnlyRootFilesystem: true
+        seccompProfile:
+          type: RuntimeDefault
       affinity: {}
       nodeSelector: {}
       tolerations: []
--- a/charts/cinder-csi-plugin/values.yaml
+++ b/charts/cinder-csi-plugin/values.yaml
@@ -91,6 +91,7 @@ csi:
         readOnly: true
     nodePlugin:
       dnsPolicy: ClusterFirstWithHostNet
+      priorityClassName: "system-node-critical"
       # Optional additional annotations to add to the nodePlugin Pods.
       podAnnotations: {}
       # Optional additional labels to add to the nodePlugin Pods.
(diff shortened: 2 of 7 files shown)


charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
```diff
@@ -249,9 +249,7 @@ spec:
       hostAliases:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-    {{- if .Values.priorityClassName }}
-      priorityClassName: {{ .Values.priorityClassName }}
-    {{- end }}
+      priorityClassName: {{ .Values.priorityClassName | default "system-cluster-critical" }}
```
mandre: Do we want to retain the ability to _not_ set `pritorityClassName`?

If so, the easiest is probably to set in `values.yaml`:
```yaml
controllerPriorityClassName: "system-cluster-critical"
nodePriorityClassName: "system-node-critical"
```

And have:
```yaml
    {{- if .Values.controllerPriorityClassName }}
      priorityClassName: {{ .Values.controllerPriorityClassName }}
    {{- end }}
```

stephenfin: It's a behaviour change, but I think *not* setting that is a huge foot gun: if the pod gets evicted due to e.g. memory pressure then you lose the ability to unmount volumes and can't drain nodes. The likes of the AWS EBS CSI driver [do what I've done here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/36d7fd880c3c4ff3b29212e2c5df8aca6628eda6/charts/aws-ebs-csi-driver/templates/_node.tpl#L48) for exactly this reason. I called this out in [the commit message](https://github.com/kubernetes/cloud-provider-openstack/pull/3181/changes/9f1fde9728dadbd5002089bcc0423037e9a2af0a) but probably should have been less terse.

mandre: I agree that we should always have `priorityClassName` mapping to a critical class for the reasons you cited, so I wouldn't push back if you want to merge the patch as it is now. Just wondering if there's any use case for not setting `priorityClassName` (CI or dev?).

stephenfin:
I had to change the `securityContext` set for the containers since the upstream images run as root, meaning `runAsNonRoot` and `allowPrivilegeEscalation` cannot be set. I've added a note in the manifests to this effect.

charts/cinder-csi-plugin/templates/controllerplugin-deployment.yaml
stephenfin: I wasn't aware of any so I asked an LLM. It seems there **are** some reason:

> 1. Namespace restrictions via `ResourceQuota`. The `system-node-critical` and `system-cluster-critical` classes are built-in, but clusters can have `ResourceQuota` objects that restrict their use to `kube-system`. If the cinder-csi-plugin is deployed to a custom namespace (e.g. `cinder-csi`), pods will fail to schedule with a quota error. This is common in hardened/multi-tenant clusters.
> 2. No way to opt out in the Helm chart. The change from a conditional `{{- if .Values.priorityClassName }}` to `{{ .Values.priorityClassName | default "system-cluster-critical" }}` means a user cannot opt out: setting `priorityClassName: ""` in their values still applies the default (Helm's `default` treats empty string as falsy). There's no escape hatch.
> 3. Dev / CI environments. In small clusters (kind, minikube, k3s) with limited resources, using `system-node-critical` causes the CSI pods to preempt other pods if resources are tight. In CI you often don't care about eviction and may prefer that test workloads aren't preempted by infrastructure pods.
> 4. Some PodSecurityAdmission or OPA/Kyverno policies in corporate environments reject pods requesting system-critical priority classes outside `kube-system`, for similar compliance reasons.

I think some of those are weak arguments, but the fact that we're even discussing this indicates it's definitely debatable. I'll revert to the previous version and just override the default.

charts/cinder-csi-plugin/values.yaml
mandre: We should update the helm charts as well to have suitable default for priorityClassName for both the controller plugin and the nodeplugin.

stephenfin: Oh, no idea how I missed this. Done.

charts/cinder-csi-plugin/values.yaml
```diff
@@ -225,7 +225,7 @@ clusterID: "kubernetes"
 # Enable PVC annotations support to create PVCs with extra parameters
 pvcAnnotations: false
 
-priorityClassName: ""
+priorityClassName: "system-node-critical"
```
mandre: That won't work, we need to be able to set different values for the controllerplugin and the nodeplugin, we'll need different variables.

stephenfin: Done. I've just ignored the old value. It's effectively useless, as I've noted in the commit message. Will update the release note now.

stephenfin:
/test openstack-cloud-csi-cinder-e2e-test openstack-cloud-csi-manila-e2e-test

Infra issues

mandre (commented):
/approve

PR #3191, opened 2026-09-07
[occm] Reduce caps of manifests, charts

stephenfin:

Restrict capabilities on the OCCM controller pod.
Also set `dnsPolicy: ClusterFirstWithHostNet` since we use `hostNetwork: true` by default and set `priorityClassName` to ensure pods don't get evicted under memory pressure.
Finally, update the maintainers and remove use of a bitnami dependency to remove an unnecessary external dependency.

Commits:

- `openstack-cloud-controller-manager: Add note on why hostNetwork is required`
- `openstack-cloud-controller-manager: Add securityContext for controller containers`
- `openstack-cloud-controller-manager: Set dnsPolicy`
- `openstack-cloud-controller-manager: Set priorityClassName`
- `openstack-cloud-controller-manager: Drop Bitnami common chart dependency`
- `openstack-cloud-controller-manager: Bump openstack-cloud-controller-manager chart to 2.36.5`

(none)

**Release note**:

```release-note
[occm] Changes to default manifests, charts


The following changes have been made to provided manfiests and charts:

- `securityContext` is now set by default.
- `dnsPolicy` is now set to `ClusterFirstWithHostNet`
- `priorityClassName` is now set to `system-node-critical` to ensure node drain
  works as expected.
```

### DIFF
--- a/charts/openstack-cloud-controller-manager/templates/servicemonitor.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/servicemonitor.yaml
@@ -3,7 +3,7 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: {{ include "occm.name" . }}
-  labels: 
+  labels:
     {{- include "occm.labels.standard" . | nindent 4 }}
   {{- if .Values.serviceMonitor.additionalLabels }}
     {{- toYaml .Values.serviceMonitor.additionalLabels | nindent 4 }}
@@ -23,5 +23,5 @@ spec:
   jobLabel: component
   selector:
     matchLabels:
-      {{- include "occm.labels.standard" . | nindent 6 }}
+      {{- include "occm.labels.matchLabels" . | nindent 6 }}
 {{- end }}
--- a/charts/openstack-cloud-controller-manager/templates/servicemonitor.yaml
+++ b/charts/openstack-cloud-controller-manager/templates/servicemonitor.yaml
@@ -4,7 +4,7 @@ kind: ServiceMonitor
 metadata:
   name: {{ include "occm.name" . }}
   labels: 
-    {{- include "common.labels.standard" . | nindent 4 }}
+    {{- include "occm.labels.standard" . | nindent 4 }}
   {{- if .Values.serviceMonitor.additionalLabels }}
     {{- toYaml .Values.serviceMonitor.additionalLabels | nindent 4 }}
   {{- end }}
@@ -23,5 +23,5 @@ spec:
   jobLabel: component
   selector:
     matchLabels:
-      {{- include "common.labels.standard" . | nindent 6 }}
+      {{- include "occm.labels.standard" . | nindent 6 }}
 {{- end }}
--- a/charts/openstack-cloud-controller-manager/values.yaml
+++ b/charts/openstack-cloud-controller-manager/values.yaml
@@ -78,18 +78,17 @@ podLabels: {}
 # For all available options, see https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#podsecuritycontext-v1-core
 podSecurityContext:
   runAsUser: 1001
-  # seccompProfile:
-  #   type: RuntimeDefault
+  seccompProfile:
+    type: RuntimeDefault
 
 # Set security settings for the controller container
 # For all available options, see https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#securitycontext-v1-core
-securityContext: {}
-# securityContext:
-#   capabilities:
-#     drop:
-#     - ALL
-#   readOnlyRootFilesystem: true
-#   allowPrivilegeEscalation: false
+securityContext:
+  readOnlyRootFilesystem: true
+  seccompProfile:
+    type: RuntimeDefault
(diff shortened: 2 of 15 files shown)


manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml
```diff
@@ -87,6 +87,7 @@ spec:
       # since not all CNIs support link-local addresses, or if the OpenStack
       # control plane is on a management network that is not advertised to pods
       hostNetwork: true
+      priorityClassName: system-node-critical
```
mandre, line 93: Is it expected that the daemonset manifest uses `system-node-critical` while the pod one uses `system-cluster-critical` ?

stephenfin: Yeah, this was intentional. This pod manifest appears designed for development purposes (think a k3s all-in-one cluster: I don't see it used in CI or anywhere else and contemplated deleting it, but decided to defer that to a separate PR at least). Assuming this is the case, `system-cluster-critical` makes sense to me since there's only one node and no `DaemonSet` to ensure it's replicated elsewhere. `system-node-critical` could also work though, and I could even drop this entirely since it's just development. Let me know what you'd like

mandre: ok, that makes more sense now, thanks for the explanation 👍

charts/openstack-cloud-controller-manager/values.yaml
```diff
@@ -78,18 +78,17 @@ podLabels: {}
...
-  # seccompProfile:
-  #   type: RuntimeDefault
+  seccompProfile:
+    type: RuntimeDefault
 
 # Set security settings for the controller container
 # For all available options, see https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#securitycontext-v1-core
-securityContext: {}
-# securityContext:
-#   capabilities:
```
mandre, line 85: isn't it possible to add `capabilities.drop: [ALL]` as well everywhere we enforce the security context?

charts/openstack-cloud-controller-manager/values.yaml
```diff
@@ -75,18 +78,17 @@ podLabels: {}
 # For all available options, see https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#podsecuritycontext-v1-core
 podSecurityContext:
   runAsUser: 1001
```
mandre, line 80: Might want to add `runAsNonRoot: true` too, for extra measure.

charts/openstack-cloud-controller-manager/templates/servicemonitor.yaml
```diff
@@ -23,5 +23,5 @@ spec:
   jobLabel: component
   selector:
     matchLabels:
-      {{- include "common.labels.standard" . | nindent 6 }}
+      {{- include "occm.labels.standard" . | nindent 6 }}
```
mandre: We should update the matchLabels because `occm.labels.standard` contains both `helm.sh/chart` and `app.kubernetes.io/version` that are updated on chart bump or app bump. We should use stable labels:

```yaml
selector:
  matchLabels:
    app.kubernetes.io/name: {{ include "occm.name" . }}
    app.kubernetes.io/instance: {{ .Release.Name }}
```

charts/openstack-cloud-controller-manager/values.yaml
stephenfin: Yeah, it is. I didn't add it here since it wasn't present in the manifests. Let me add it now.

charts/openstack-cloud-controller-manager/values.yaml
stephenfin: I could, but it feels redundant: they say the same things. If we were doing this, I'd rather *just* set `runAsNonRoot: true`

charts/openstack-cloud-controller-manager/templates/servicemonitor.yaml
stephenfin: This would be a change in behavior though since the `common.labels.standard` used before this included these. I'm okay with that, but I'd like to confirm that that was your intention?

charts/openstack-cloud-controller-manager/values.yaml
mandre: Opus tells me that it's good practice to set `runAsNonRoot: true` at the pod level as well, since that also covers init containers or sidecars that might be added.

charts/openstack-cloud-controller-manager/templates/servicemonitor.yaml
mandre: Correct, this was a pre-existing issue. Since we were tweaking the matchLabels, I though this might be a good opportunity to fix it. Or we might do it in https://github.com/kubernetes/cloud-provider-openstack/pull/3192.

stephenfin: Ack, just wanted to confirm your intention. I've done this now. Let's tackle the other ones in #3192.

mandre (commented):
/lgtm
/approve

PR #3195, opened 2026-09-08
Change Chart release process

### LINKED ISSUE #3194: Helm Charts should not be released from `master` / `main`
(Cross-posting from [Slack](https://kubernetes.slack.com/archives/C0LSA3T7C/p1788878292076129)).

I think how we're handling Helm Chart versioning currently is broken and I'd like to change it. Currently, we are required to bump the chart `version` for any change to Helm Charts. This is true for changes to `master` branches as well as `release-*` branches. This presents two issues. Firstly, it means we effectively can't do large breaking changes since doing so would require us to "grab" the minor version corresponding to the next CPO release (e.g. currently, we'd have to "grab" 2.37.0, meaning we'd have to use 2.38.0 when we release CPO v1.37.0). It also means we have to do bugfixes on the `master` branch, since we are incrementing the patch release there and any attempt to do this on the release branch would conflicting tags/releases.

I propose that we change our release process so that, following a new minor release (which is really a "major" version, e.g. the forthcoming v1.37.0 release) we immediately set `version` to e.g. `2.37.0-dev`. Users would continue to push bug fixes to `master`, but would then need to backport them if they wanted these fixes released before the next full release. This backport would need to be partially manual, since they would also need to bump the chart version but we might be able to automate this

In order to get us to this situation for the current branches, I will need to backport all the recent changes to the charts to `release-1.36`. I can either do this via multiple PRs in sequence or via one mega PR. If the former, I will need to first temporarily disable the chart lint and release jobs to avoid us re-tagging from this branch before re-enabling and bumping the patch version of the 3 charts once all PRs are merged. If the latter, I'll propose a mega PR will all those recent changes included as well as a stable-only change that bumps the patch version by 1 to ensure we release from here.

stephenfin:

As discussed in #3194, we currently release charts from `master`. This is confusing, since `master` is intended for development for the next minor version, rather than the next patch release (which should happen on `release-*` branches).

Fix this via changes to our release process and tooling. We now expect the `version` field to change twice per minor release: once immediately after a release, when we'll bump the version to the next minor version with a `-dev` suffix, and once at release, when we'll drop this `-dev` suffix.

Make these changes in the charts, then update the docs. We also update the release scripts, removing the existing bash script with a (mostly LLM-written) Python release script. This will attempt to auto-detect what branch we are on and guide users down the correct path. It's written with PEP-723 metadata so it should be very easy to run for anyone with a functioning Python toolchain.

Commits:

- `charts: Use '-dev' suffix for Helm Charts on master`
- `charts: Change maintainer for remaining Helm Charts`
- `charts: Skip chart version check on master`
- `docs: Add section on helm charts update`
- `hack: Rework bump-chart script`

Fixes #3194.

**Release note**:

```release-note
NONE
```


.github/workflows/release.yaml
```diff
@@ -3,8 +3,9 @@ name: Release Charts
 on:
   push:
     branches:
-      - "master"
       - "release-*"
+    tags:
+      - "v*"
```
mandre: is it really necessary to trigger the workflow on `v*` tags? won't that cause the workflow to run twice because we're tagging a commit that was just pushed into the related `release-*` branch?

docs/release-procedure.md
```diff
@@ -142,14 +151,36 @@ dependency or sidecar container.
...
+process for major releases but with the following changes:
+
+1. You must always bump the Helm Chart `appVersion` and `version` fields.
+
+1. It is not necessary to create a new branch or add new jobs.
+
+## Helm Charts
+
+Chart versions on `master` use a `-dev` pre-release suffix (e.g.
+`2.37.0-dev`) and are **not** bumped for individual PRs. Version bumps only
```
mandre, line 165: Such a nice quality of life improvement for anyone contributing changes to charts, since they won't have to update their PRs each time another change merges.

docs/release-procedure.md
```diff
@@ -76,31 +76,16 @@ dependency or sidecar container.
...
     ```
 
-1. Update the package version with the expected version.
+1. Bump the release version.
 
-    Make changes in the `docs`, `manifests`, `tests`, and `examples`
-    directories using the `hack/bump-release.sh` script. For example:
+    Run `hack/bump-release.py`, which detects the current branch automatically
+    and updates the Helm chart versions (`charts/`) and all image references in
+    `docs/`, `manifests/`, and `examples/`:
```
mandre, line 83: Did we drop `tests` on purpose? If so, could you explain why in the commit message?

We have some files in `tests` referencing versions:
```
❯ rg v1.3 tests
tests/playbooks/test-csi-cinder-e2e.yaml
7:    e2e_test_version: v1.34.1

tests/playbooks/roles/install-k3s/defaults/main.yaml
2:k3s_release: v1.34.1+k3s1
```

docs/release-procedure.md
```diff
@@ -61,73 +61,67 @@ dependency or sidecar container.
...
+    `openstack-cinder-csi-X.Y.Z`, and `openstack-manila-csi-X.Y.Z`).
+
+1. Make a `release-X.Y` release branch and push it to the upstream repository
+
+    ```bash
+    $ git checkout -b release-X.Y
+    $ git push origin release-X.Y
+    ```
+
+1. Reset the `version` field of the Helm Charts to `2.X.Y.0-dev`
```
mandre:
```suggestion
1. Reset the `version` field of the Helm Charts to `2.X.0-dev`
```

hack/bump-release.py
```diff
@@ -0,0 +1,242 @@
...
+
+def load_chart(path: Path) -> tuple[YAML, dict]:
+    yaml = YAML()
+    yaml.preserve_quotes = True
+    with path.open() as f:
+        return yaml, yaml.load(f)
+
+
+def read_versions() -> tuple[str, str]:
+    """Return (appVersion, version), validating that all charts agree."""
```
mandre: So we now enforce that all charts carry the same version? That looks like a big change from the current release procedure where each chart can have independent version numbers.
It should be possible to update the script to keep the existing behavior, with independent versions.

mandre (commented):
I really like it, nice work.

.github/workflows/release.yaml
stephenfin: If we don't do this then we won't release on new minor version (i.e. `v1.37.0`). I'm not sure if I can tighten the glob up to e.g. `v1.*.0`?

docs/release-procedure.md
stephenfin: Hmm, I missed those because I grepped for `1.34.0`. These are k8s/k3s versions, not CPO version. We probably need to update one of the earlier steps to indicate these should be bumped while updating the k8s.io deps

hack/bump-release.py
stephenfin: That was not the intention. This is a bug. Let me fix it.

.github/workflows/release.yaml
mandre: Oh, I see. Could you perhaps add a comment?

stephenfin: Let's try

docs/release-procedure.md
```diff
@@ -61,73 +61,69 @@ dependency or sidecar container.
...
-   `openstack-cinder-csi-X.Y.Z`, and `openstack-manila-csi-X.Y.Z`). Make a `vX.Y.Z`
-   release tag and `release-X.Y` release branch and push both to the upstream
-   repository.
+1. Make a `vX.Y.0` release tag and push it to the upstream repository.
 
     ```bash
     $ git checkout master
     $ git pull upstream master
     $ git tag vX.Y.Z
     $ git push upstream vX.Y.Z
```
mandre: nit
```suggestion
    $ git tag vX.Y.0
    $ git push upstream vX.Y.0
```

docs/release-procedure.md
```diff
@@ -61,73 +61,69 @@ dependency or sidecar container.
...
+    `openstack-cinder-csi-X.Y.Z`, and `openstack-manila-csi-X.Y.Z`).
+
+1. Make a `release-X.Y` release branch and push it to the upstream repository
+
+    ```bash
+    $ git checkout -b release-X.Y
+    $ git push origin release-X.Y
+    ```
+
+1. Reset the `version` field of the Helm Charts to `2.X.0-dev`
```
mandre: Should it actually be x+1 at this point?
Also, it would be cool to have a script to do that (or update the new `bump-release.py` to be able to do post-release operations.

hack/bump-release.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+def git_check(*args: str) -> bool:
+    return subprocess.run(["git", *args], capture_output=True).returncode == 0
+
+
+def find_base_branch() -> str | None:
+    """Return 'master' or a 'release-N.N' name, whichever is the closest ancestor of HEAD."""
+    candidates = ["master"] + [
+        line.strip().removeprefix("origin/")
```
mandre: can we have a flag to change the name of the remote? I expect that in many cases `origin` points somewhere else.

docs/release-procedure.md
stephenfin: Done

docs/release-procedure.md
stephenfin: > Should it actually be x+1 at this point?
It should.

>  Also, it would be cool to have a script to do that (or update the new `bump-release.py` to be able to do post-release operations.

I didn't add this to the script since it would require the script handling `git commit` duties, which I didn't want to do. I could do this, but we'd likely want to have the script do more things then (like bump the `k8s.io` dep versions in `go.mod`). Maybe one for a follow-up?

mandre (commented):
/lgtm

hack/bump-release.py
stephenfin: Done

PR #3208, opened 2026-09-09
make: Add lint-charts target

stephenfin:

We will add a prow job to run this separately, allowing us to remove the GitHub Action workflow.

Note that we lose `yamllint` by doing this. That seems acceptable. If we want, we could hack something together with pipx or uvx but it seems...unrobust for CI purposes.

Ref #3194

This is a step towards addressing the issue @mandre reported in https://kubernetes.slack.com/archives/C0LSA3T7C/p1788939409251989.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,6 +57,8 @@ BUILD_CMDS	?= openstack-cloud-controller-manager \
 				barbican-kms-plugin \
 				magnum-auto-healer \
 				client-keystone-auth
+export GOLANGCI_LINT_VERSION ?= v2.13.2
+export HELM_UNITTEST_VERSION ?= v1.1.2
 
 # CTI targets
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -57,6 +57,8 @@ BUILD_CMDS	?= openstack-cloud-controller-manager \
 				barbican-kms-plugin \
 				magnum-auto-healer \
 				client-keystone-auth
+export GOLANGCI_LINT_VERSION ?= v2.13.2
```
mandre, line 60: nit: technically, GOLANGCI_LINT_VERSION doesn't need to be exported since it's never read from the environment.

mandre (commented):
/approve
/lgtm

Not sure if you want to address the nit, feel free to remove the hold.
/hold

Makefile
stephenfin: Indeed. I did this for consistency with the line below though. It shouldn't affect the parent shell afaik so this is effectively a no-op

PR #3209, opened 2026-09-09
Remove dead targets from Makefile

stephenfin:

There are a lot of targets in here that are either there purely to satisfy the now-dead OpenStack Go PTI or are unused. Remove them all.

- `make: Remove irrelevant header`
- `make: Remove build-cross, dist targets`
- `make: Remove bindep target`
- `make: Remove more PTI targets`
- `make: Remove work target`
- `make: Remove DEST and SOURCES variables`
- `make: Remove unused variables`
- `make: Split PHONY definition up`

(none)

This one is worth reviewing one-by-one.

**Release note**:

```release-note
NONE
```


mandre (commented):
OK, there's a lot going on there. Thanks for splitting that into small commits. They all make sense to me.

/lgtm
/approve

mandre (commented):
/lgtm

PR #3215, opened 2026-09-10
[cinder-csi-plugin] Implement direct attach mode for baremetal nodes

mandre:

Add `--attach-mode=direct` to the Cinder CSI driver, enabling volume attachment without Nova via the Cinder Attachment API and an os-brick gRPC sidecar. This allows pods on bare-metal (Ironic-provisioned) Kubernetes workers to consume Cinder block storage (iSCSI, Fibre Channel, NVMe-oF, RBD, etc.).

The default `--attach-mode=nova` is unchanged.

The os-brick sidecar is a Python container that wraps the `os-brick` library and exposes `ConnectVolume`, `DisconnectVolume`, `ExtendVolume`, and `GetConnectorProperties` over a Unix domain socket. It runs as a privileged sidecar alongside the node plugin.

On the controller side, `ControllerPublishVolume`/`UnpublishVolume` read connector properties from a node annotation and call Cinder's Attachment API instead of Nova. On the node side, `NodeStageVolume`/`UnstageVolume` call the sidecar to connect/disconnect volumes, and `NodeGetInfo` stores connector properties on the node for the controller to read.

fixes #

This is a BIG PR, but I've tried my best to break it into digestible commits.
I used Opus 4.6 to assist me with the PR, and tested these changes with a virtualized node in an environment with iSCSI+LVM. I've also tested an older version of this code with RBD.

All direct-mode logic is gated behind `--attach-mode=direct`. The nova code paths are not modified.

The commit series is structured for incremental review:
1. **Commits 1-4**: os-brick sidecar (protobuf, Python server, Go client, Cinder Attachment API)
2. **Commits 5-8**: driver plumbing (`buildKubeConfig()`, attach-mode fields, `ConnectorPropertiesGetter`, node server wiring)
3. **Commits 9-13**: direct-mode CSI operations (controller publish/unpublish, `NodeStageVolume`, `NodeUnstageVolume`, `NodeGetInfo`, `NodeExpandVolume`)
4. **Commit 14**: nodeplugin DaemonSet manifest for direct-mode deployment

To test: deploy the direct-mode DaemonSet (`cinder-csi-nodeplugin-direct.yaml`) on a bare-metal node with `--attach-mode=direct` and a Cinder backend reachable via iSCSI/FC/RBD. Create a PVC and pod; the sidecar log shows `ConnectVolume`/`DisconnectVolume` calls.

**Release note**:

```release-note
[cinder-csi-plugin] Add `--attach-mode=direct` flag for bare-metal volume attachment via Cinder Attachment API and os-brick sidecar, bypassing Nova.
```


kayrus:
this PR would be quite challenging to review :melting_face:

mandre:
> this PR would be quite challenging to review 🫠

Yeah... I've tried my best to split this into logical chunks but it's still a bit overwhelming. The protobuf and the sidecar container changes shouldn't be too difficult to review as they contain very little logic. The bulk of the work is in the CSI driver, but here again, everything should be gated behind the `--attach-mode=direct` flag, so at least we know we aren't breaking anything...

Would it help if I split this into 4 PRs (the 4 logical groups detailed in the PR description)? They could be reviewed and merged separately, although there's dependency between them and would need to be merged in order.

I still need to figure out how to test the new direct attach mode in CI, and how to publish the new container image.

proto/osbrick/v1/connector.proto
winiciusallan, line 1: nit: Better than commenting on every occurrence... It seems that every word after a period has two spaces between them.

proto/osbrick/v1/connector.proto
```diff
@@ -0,0 +1,126 @@
...
+//   - rescan/extend a connected volume after online resize (ExtendVolume)
+service OsBrickConnector {
+  // GetConnectorProperties returns host-side initiator information
+  // (iSCSI IQN, FC WWPNs, hostname, etc.) that Cinder needs to set up
+  // a volume connection via InitializeConnection.
+  rpc GetConnectorProperties(GetConnectorPropertiesRequest) returns (ConnectorProperties);
+
+  // ConnectVolume asks os-brick to attach a volume described by
+  // connection_info (the JSON dict returned by Cinder's
+  // InitializeConnection API).  Returns the local device path.
```
winiciusallan:
```suggestion
  // InitializeConnection API). Returns the local device path.
```

pkg/util/brick/interface.go
```diff
@@ -0,0 +1,60 @@
+/*
+Copyright 2025 The Kubernetes Authors.
```
winiciusallan, line 2:
```suggestion
Copyright 2026 The Kubernetes Authors.
```

cmd/cinder-csi-plugin/main.go
```diff
@@ -129,7 +147,17 @@ func handle() {
 			}
 		}
 
-		d.SetupControllerService(clouds)
+		var connProps cinder.ConnectorPropertiesGetter
+		if attachMode == cinder.AttachModeDirect {
+			kubeClient := csi.GetKubeClient()
```
winiciusallan, line 152: I think we can get the kube client once and reuse it on this same function.

sidecar/requirements.txt
winiciusallan, line 1: wouldn't it be better to pin versions to avoid pulling a incompatible version?

winiciusallan (commented):
great work, huh @mandre? left a few comments from a first look

proto/osbrick/v1/connector.proto
stephenfin, line 1: Note this (and the hack file) were the only files I reviewed in commit 0.

sidecar/os_brick_grpc/main.py
stephenfin, line 1: Thinking out loud: I assume this isn't something that could go into upstream os-brick and be consumed from there?

stephenfin, line 1: Can we call this package `os_brick_grpc` or something else? The `osbrick` package name confused me for a moment: I expected to find a `gen` subpackage at https://opendev.org/openstack/os-brick/src/branch/master/os_brick 😅

sidecar/osbrick/main.py
```diff
@@ -0,0 +1,117 @@
...
+from osbrick.gen import connector_pb2_grpc
+from osbrick.server import OsBrickConnectorServicer
+
+LOG = logging.getLogger(__name__)
+
+DEFAULT_SOCKET_PATH = "/var/run/osbrick/osbrick.sock"
+SERVICE_NAME = "osbrick.v1.OsBrickConnector"
+
+
+def _configure_logging():
```
stephenfin:
```suggestion
def _configure_logging() -> None:
```

ditto for `_configure_oslo` and `main` below

sidecar/osbrick/server.py
```diff
@@ -0,0 +1,315 @@
...
+LOG = logging.getLogger(__name__)
+
+# Root helper for os-brick operations. In a privileged container this
+# is typically 'sudo' or '' (empty string if already root).
+# Configurable via OSBRICK_ROOT_HELPER environment variable.
+ROOT_HELPER = os.environ.get("OSBRICK_ROOT_HELPER", "sudo")
+
+# Whether to report multipath capability in connector properties.
+# Set OSBRICK_MULTIPATH=false on hosts without multipath-tools.
+MULTIPATH = os.environ.get("OSBRICK_MULTIPATH", "true").lower() in (
```
stephenfin: os_brick pulls is `oslo_utils`. Let's use [`oslo_utils.strutils.bool_from_string`](https://opendev.org/openstack/oslo.utils/src/commit/5f10aaacb05406a5ab690e8ba9deefe570e2b7df/oslo_utils/strutils.py#L187) instead?

sidecar/osbrick/server.py
```diff
@@ -0,0 +1,315 @@
...
+    "true", "1", "yes",
+)
+
+
+def _get_my_ip():
```
stephenfin:
```suggestion
def _get_my_ip() -> str:
```

I'll stop commenting on these going forward, but we could(/should?) add hints to these functions where it makes sense to do so

sidecar/Dockerfile.osbrick-sidecar
```diff
@@ -0,0 +1,99 @@
...
+# Generate Python gRPC stubs from the proto file.
+COPY proto/osbrick/v1/connector.proto /proto/connector.proto
+RUN mkdir -p /build/gen && python -m grpc_tools.protoc \
+    --proto_path=/proto \
+    --python_out=/build/gen \
+    --grpc_python_out=/build/gen \
+    /proto/connector.proto \
+    && sed -i 's/^import connector_pb2/from osbrick.gen import connector_pb2/' /build/gen/connector_pb2_grpc.py
+
+# ---
```
stephenfin, line 54: Intentional?

```suggestion
---
```

sidecar/Dockerfile.osbrick-sidecar
```diff
@@ -0,0 +1,99 @@
...
+RUN pip install --no-cache-dir -r requirements.txt
+
+# Generate Python gRPC stubs from the proto file.
+COPY proto/osbrick/v1/connector.proto /proto/connector.proto
+RUN mkdir -p /build/gen && python -m grpc_tools.protoc \
+    --proto_path=/proto \
+    --python_out=/build/gen \
+    --grpc_python_out=/build/gen \
+    /proto/connector.proto \
+    && sed -i 's/^import connector_pb2/from osbrick.gen import connector_pb2/' /build/gen/connector_pb2_grpc.py
```
stephenfin: side note: I expected there to be a knob exposed in the tooling here to avoid this, but https://github.com/grpc/grpc/issues/29459 suggests not

sidecar/Dockerfile.osbrick-sidecar
```diff
@@ -0,0 +1,99 @@
...
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    open-iscsi \
+    multipath-tools \
+    sg3-utils \
+    sysfsutils \
+    sudo \
+    # Ceph/RBD support: python3-rados and python3-rbd must match the
+    # image's Python ABI.  python:3.13-slim is based on Debian Trixie
+    # whose system Python is also 3.13, so the apt C extensions are
+    # ABI-compatible.
```
stephenfin: Given this fact, surely it would be better to focus on the distro rather than the Python version? That is, couldn't we use the `debian:trixie` image rather than `python:3.13-slim`? fwict, Trixie uses Python 3.13 anyway so this shouldn't make a difference, but it would prevent a fellow contributor missing this and attempting to bump to `python:3.15-slim`

sidecar/Dockerfile.osbrick-sidecar
```diff
@@ -0,0 +1,99 @@
...
+COPY --from=builder /usr/local/bin /usr/local/bin
+
+# Copy generated gRPC stubs.
+COPY --from=builder /build/gen/ /app/osbrick/gen/
+
+# Copy application source.
+COPY sidecar/osbrick/ /app/osbrick/
+
+# Ensure the gen directory has an __init__.py.
+RUN touch /app/osbrick/gen/__init__.py
```
stephenfin: Could we do this up during generation?

sidecar/entrypoint.sh
```diff
@@ -0,0 +1,43 @@
...
+# /run/containers/storage/.  This script symlinks the host runtime
+# directories that os-brick tools need into /run before starting the
+# gRPC server.
+
+set -eu
+
+HOST_RUN="${HOST_RUN_DIR:-/host/run}"
+
+# Directories used by os-brick initiator connectors (iSCSI, FC,
+# multipath, LVM, NVMe).
```
stephenfin, line 29: This describes "what" we're doing here. Could we get a comment on "why"?

stephenfin (changes requested):
Still working through this (I'm at [commit 5/11](https://github.com/kubernetes/cloud-provider-openstack/pull/3215/changes/24b916c5487ba19f2737086e96be0761dbf41afd)). Dumping my first round of comments here.

proto/osbrick/v1/connector.proto
mandre: Yes, the files in `pkg/util/brick/gen/osbrickv1` are generated, no need to review them.

sidecar/os_brick_grpc/main.py
mandre: Possibly: the proto definition and the python server are intentionally generic. However, I'd prefer it to mature here where the consumer (the go gRPC client) lives too before even considering it.

sidecar/Dockerfile.osbrick-sidecar
mandre: Yep, this is apparently standard practice as I also learned myself working on this PR.

sidecar/Dockerfile.osbrick-sidecar
mandre: This is not yaml, but a dockerfile, so it was indeed meant to be commented out.

sidecar/Dockerfile.osbrick-sidecar
mandre: That's a good point. I've switched to `python:3-slim-trixie` to ensure ABI compatibility. Changes to the python version should now be transparent for us.

sidecar/entrypoint.sh
mandre: The explanation of "why" is just above (lines 18 to 22).

sidecar/requirements.txt
mandre: You're right. This is the source of truth for what's in the image.

pkg/csi/cinder/openstack/openstack_attachments.go
```diff
@@ -0,0 +1,116 @@
...
+	mc := metrics.NewMetricContext("volume", "reset_status")
+	err := volumes.ResetStatus(ctx, blockstorageClient, volumeID, opts).ExtractErr()
+	if mc.ObserveRequest(err) != nil {
+		klog.Errorf("Failed to reset volume %s status to %s: %v", volumeID, targetStatus, err)
+		return err
+	}
+
+	klog.V(4).Infof("Reset volume %s status to %s", volumeID, targetStatus)
+	return nil
+}
```
stephenfin: Ooh, I don't know about this. This API is a *massive* footgun that has caused more than it's fair share of customer cases over the years. Was this born out of testing or is it something the LLM suggested? I would be inclined *not* to include this unless absolutely necessary: if an attachment fails, it's up to the user to go investigate and fix it (which may include running this command, but also may not).

pkg/csi/csi.go
```diff
@@ -111,29 +112,44 @@ func GetAZFromTopology(topologyKey string, requirement *csi.TopologyRequirement)
...
-	// get the KUBECONFIG from env if specified (useful for local/debug cluster)
-	kubeconfigEnv := os.Getenv("KUBECONFIG")
-
-	if kubeconfigEnv != "" {
+// buildKubeConfig creates a rest.Config from the --master/--kubeconfig
+// flags and the KUBECONFIG environment variable. QPS, burst, and
+// content-type are set from the shared CLI flags.
+func buildKubeConfig() *rest.Config {
+	kc := kubeconfig
+	if env := os.Getenv("KUBECONFIG"); env != "" {
```
stephenfin: We don't have stacked PRs yet but if you pulled this commit out and proposed it separately we could merge this asap.

cmd/cinder-csi-plugin/main.go
```diff
@@ -101,6 +103,9 @@ func main() {
 	cmd.PersistentFlags().BoolVar(&noClient, "node-service-no-os-client", false, "If set to true then the CSI driver node service will not use the OpenStack client (default: false)")
 	cmd.PersistentFlags().MarkDeprecated("node-service-no-os-client", "This flag is deprecated and will be removed in the future. Node service do not use OpenStack credentials anymore.") //nolint:errcheck
 
+	cmd.PersistentFlags().StringVar(&attachMode, "attach-mode", "nova", "Volume attach mode: 'nova' uses Nova attach/detach, 'direct' uses Cinder Attachment API with os-brick sidecar for direct attachment (typically bare-metal nodes)")
+	cmd.PersistentFlags().StringVar(&brickEndpoint, "brick-endpoint", "unix:///var/run/osbrick/osbrick.sock", "Endpoint of the os-brick gRPC sidecar (only used when --attach-mode=direct)")
```
stephenfin: Do we need two flags? Doesn't `--brick-endpoint` imply direct attach mode?

pkg/csi/cinder/connprops.go
```diff
@@ -0,0 +1,146 @@
...
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+)
+
+const (
+	// ConnectorPropertiesAnnotation is the node annotation key where the
+	// node plugin stores its os-brick connector properties (JSON). The
+	// controller reads this annotation to pass connector properties to
+	// Cinder's Attachment API (AttachmentCreate).
+	ConnectorPropertiesAnnotation = "cinder.csi.openstack.org/connector-properties"
```
stephenfin: Connector properties can store usernames and passwords in them (the scaleio connector does just this [historically](https://github.com/openstack/os-brick/blob/44c01994e45c781cb5fd3a9a9f9e8e3e5522d04c/os_brick/initiator/connectors/scaleio.py#L349-L353)). I wonder if we need to store this in a secret instead and reference the secret via the annotation?

pkg/csi/cinder/controllerserver.go
```diff
@@ -330,14 +330,140 @@ func (cs *controllerServer) ControllerPublishVolume(ctx context.Context, req *cs
...
+				// another controller could be concurrently operating
+				// on this volume. The re-fetch + attachment count
+				// check above mitigates but does not eliminate this
+				// race. Force-reset should be used with caution.
+				klog.Errorf("[ControllerPublishVolume] volume %s stuck in transient state %q with no attachments, force-resetting to available ; this may mask a real problem; in multi-replica deployments this could race with another controller", volumeID, refreshed.Status)
+				if resetErr := cloud.ResetVolumeStatus(ctx, volumeID, openstack.VolumeAvailableStatus); resetErr != nil {
+					return nil, status.Errorf(codes.Internal, "[ControllerPublishVolume] volume %s did not reach available status (%v) and reset failed: %v", volumeID, waitErr, resetErr)
+				}
+			}
+		}
```
stephenfin: Noted this elsewhere, but I don't believe this is something we should do. If the volume is in an error state, we should die and leave this to the user to resolve.

pkg/csi/cinder/controllerserver.go
```diff
@@ -330,14 +330,140 @@ func (cs *controllerServer) ControllerPublishVolume(ctx context.Context, req *cs
...
+	vol, err := cloud.GetVolume(ctx, volumeID)
 	if err != nil {
 		if cpoerrors.IsNotFound(err) {
 			return nil, status.Errorf(codes.NotFound, "[ControllerPublishVolume] Volume %s not found", volumeID)
 		}
 		return nil, status.Errorf(codes.Internal, "[ControllerPublishVolume] get volume failed with error %v", err)
 	}
 
+	// Direct mode: use Cinder Attachment API instead of Nova attach.
+	if cs.Driver.IsDirectMode() {
```
stephenfin: Is is possible to move the remainder of this function into two separate helper functions and call/return those? This had gotten rather large now.

pkg/csi/cinder/controllerserver.go
```diff
@@ -393,6 +519,70 @@ func (cs *controllerServer) ControllerUnpublishVolume(ctx context.Context, req *
 	if len(volumeID) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "[ControllerUnpublishVolume] Volume ID must be provided")
 	}
+
+	// Direct mode: use Cinder Attachment API instead of Nova detach.
+	// Look up the attachment ID from the volume's existing attachments;
+	// no connector properties needed.
+	if cs.Driver.IsDirectMode() {
```
stephenfin: Same comment as above RE: splitting out the remainder of this function into helper functions.

pkg/csi/cinder/nodeserver.go
```diff
@@ -195,11 +218,89 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
+					klog.Errorf("NodeStageVolume: cleanup DisconnectVolume failed for volume %s: %v", volumeID, discErr)
+				}
+			}
+		}()
+
+		// Mark the attachment as "in-use" in Cinder (best-effort).
+		attachmentID := req.GetPublishContext()["AttachmentID"]
+		volCloud := req.GetPublishContext()["Cloud"]
+		cloud := ns.Clouds[volCloud]
+		if cloud != nil && attachmentID != "" {
```
stephenfin: Why would we not have these attributes available? An expanded comment here would be helpful

pkg/csi/cinder/nodeserver.go
```diff
@@ -481,10 +487,43 @@ func (ns *nodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
...
 
+// nodeIDNamespace is a UUID v5 namespace used to derive a
+// deterministic node ID from the Kubernetes node name when the node
+// is not a Nova instance and cannot reach the metadata service
+// (e.g. bare-metal nodes, k3s on devstack host).
+//
+// IMPORTANT: This UUID MUST remain stable across releases. Changing
+// it would alter the derived instance UUIDs that Cinder stores in
+// its attachments, breaking detach/cleanup of existing volumes.
+var nodeIDNamespace = uuid.MustParse("458bfdd0-3a23-4a81-a1a0-b6cd82e37c23")
```
stephenfin: Potentially dumb question but k8s node names in this context are fully-qualified and unique, right? We don't ever expect to see the same deterministic IDs pointing at two different nodes, do we?

stephenfin (commented):
More comments. Still haven't gotten to the end of this 😅

PR #3217, opened 2026-09-11
Stop insisting on chart version bump on stable branches also

stephenfin:

In PR #3195 we modified the chart release process so that we no longer required a bump in the chart version on `master`. Unfortunately, we still have an issue on stable `release-*` branches. Anyone that backports a fix using the `/cherry-pick` slash command will see a failing chart lint job on said stable branch. This can only be fixed by incrementing the `version` field, but no one (including the maintainers) has the ability to push to the `k8s-infra-cherrypick-robot` repo fork that the bot uses. This means the only practical way to backport fixes that include chart changes is to do so manually, which is a lot more work for very little gain.

Rather than doing this, drop the version check entirely. This means we will have to manually bump the chart version periodically, but that's not a huge ask and is very similar to what we already do for CPO itself. To make this easier, we make the `hack/bump-version.sh` smarter so that it can start automatically bumping chart versions.

While here, we also improve the release procedure doc further, clarifying the purpose of different tags and the instances where you should bump different fields.

Ref #3194

(none)

**Release note**:

```release-note
NONE
```


docs/release-procedure.md
```diff
@@ -152,22 +159,68 @@ dependency or sidecar container.
...
-1. You must always bump the Helm Chart `appVersion` and `version` fields.
+1. Check out the `release-X.Y` branch and run the `master` version of
+   `hack/bump-release.py`.
 
-1. It is not necessary to create a new branch or add new jobs.
+    ```bash
+    $ git fetch upstream
+    $ git checkout release-X.Y
+    $ git pull --rebase upstream release-X.Y
+    $ git show upstream/master:hack/bump-release.py | uv run -
```
mandre: This command always outputs "No changes to CPO code or charts detected" since we're in sync with the remote (from the previous command).

We should diff against the latest tag, right?

stephenfin: Oh, of course :sweat_smile: Fixing.

mandre (commented):
/lgtm
/approve

PR #3227, opened 2026-09-15
[manila-csi-plugin] Support mutable NFS access rules

### LINKED ISSUE #3226: [manila-csi-plugin] Support mutable NFS access rules
/kind feature

**What happened**:

The `nfs-shareClient` storage class parameter is only evaluated while dynamically provisioning a Manila NFS share. After the PVC is created, Kubernetes users cannot change the share client IP addresses or CIDRs through the Manila CSI driver. Updating the access list requires operating Manila directly, which moves the share configuration outside the Kubernetes workflow.

The Manila CSI plugin should implement the CSI `ControllerModifyVolume` RPC and advertise the `MODIFY_VOLUME` capability for NFS. When a PVC is assigned a new `VolumeAttributesClass` containing `nfs-shareClient`, the driver should reconcile the Manila share's `rw` IP access rules with the requested list.

The operation should be idempotent and should grant new rules before revoking obsolete rules to avoid an access gap. Read-only rules and non-IP access rules should remain unchanged. Mutable parameters should also be honored during initial volume creation, as required by the CSI specification.

Example:

```yaml
apiVersion: storage.k8s.io/v1
kind: VolumeAttributesClass
metadata:
  name: manila-nfs-private
driverName: nfs.manila.csi.openstack.org
parameters:
  nfs-shareClient: 10.0.0.0/24,192.0.2.10
```

**How to reproduce it**:

1. Provision an NFS PVC using a storage class with `nfs-shareClient`.
2. Create a `VolumeAttributesClass` with a different `nfs-shareClient` value.
3. Assign the class to the PVC.
4. Observe that the current driver does not advertise `MODIFY_VOLUME` and returns `Unimplemented` from `ControllerModifyVolume`, so the Manila access rules are not updated.

This requires Kubernetes support for `VolumeAttributesClass` and a compatible external-resizer. The intended reconciliation manages `rw` rules with access type `ip`; other Manila access rules are preserved.

**Environment**:
- manila-csi-plugin version: current master
- OpenStack version: any version supported by the current Manila CSI plugin
- Others: Kubernetes 1.34 or newer

taaahaaa:

Adds support for updating Manila NFS share access rules after a PVC has been provisioned. The Manila CSI plugin now advertises the CSI `MODIFY_VOLUME` controller capability for NFS and implements `ControllerModifyVolume` so changes to `nfs-shareClient` through a `VolumeAttributesClass` reconcile the share's `rw` IP access rules.

The reconciliation is idempotent, grants and activates new rules before removing obsolete rules, waits for revocations to complete, and preserves read-only and non-IP rules. Mutable parameters are also honored during initial volume creation.

The Helm RBAC grants access to `VolumeAttributesClass` resources, and the Manila CSI documentation includes an example.

Fixes #3226

Validated with:

- `go test ./pkg/csi/...`
- `go vet ./pkg/csi/manila/...`
- `helm lint charts/manila-csi-plugin`
- `git diff --check`

**Release note**:

```release-note
[manila-csi-plugin] Support updating NFS share client access rules through VolumeAttributesClass.
```


taaahaaa:
All required checks are passing, and this is ready for review. @gouthampacha @kayrus, could you please take a look when you have time?

PR #3233, opened 2026-09-21
[occm] Add annotation to override enable-ingress-hostname per service

### LINKED ISSUE #3232: [occm] Per-Service override for enable-ingress-hostname
/kind feature

**What happened**:

With `enable-ingress-hostname=true` in the cloud config, OCCM publishes every PROXY protocol load balancer in the Service status as an `<ip>.nip.io` hostname (perhaps specific to our Cloud Provider). Anything that reads that status to publish DNS, such as ExternalDNS via the Gateway API, can then only create a CNAME to a nip.io name.

A way for one Service to get the IP with `ipMode: Proxy` instead, which OCCM already does when the option is off. On a managed cluster the cloud config belongs to the provider, so a tenant cannot turn the option off, and the provider is understandably reluctant to change a default for every customer.

**How to reproduce it**:

1. Set `enable-ingress-hostname=true` under `[LoadBalancer]` in the cloud config.
2. Create a `type: LoadBalancer` Service with `loadbalancer.openstack.org/proxy-protocol: "true"`.
3. `status.loadBalancer.ingress` shows `hostname: <ip>.nip.io` and no `ip`. No annotation on the Service changes that; `loadbalancer.openstack.org/hostname` only substitutes another hostname.

Proposal: a Service annotation `loadbalancer.openstack.org/enable-ingress-hostname` that overrides the config option, defaulting to it, in the same way `enable-health-monitor` overrides `create-monitor`. Setting it to `false` yields the IP with `ipMode: Proxy`, so kube-proxy still leaves the LB IP alone on Kubernetes 1.32 and later. I have a PR ready.

**Environment**:
- openstack-cloud-controller-manager(or other related binary) version: v1.33.0
- OpenStack version: Octavia on Elastx (managed Kubernetes)
- Others: Kubernetes v1.34.4, Istio Gateway API, ExternalDNS v0.21.0

BeyondEvil:

Adds a `loadbalancer.openstack.org/enable-ingress-hostname` Service annotation that overrides the `enable-ingress-hostname` config option for one Service, the same way `enable-health-monitor` overrides `create-monitor`.

With the option on cluster-wide, every PROXY protocol load balancer is published as an `<ip>.nip.io` hostname, so DNS integrations reading the Service or Gateway status can only create CNAMEs. Setting the annotation to `false` publishes the IP with `ipMode: Proxy` instead, which OCCM already does when the option is off. This matters on managed clusters where the tenant cannot change the cloud config.

fixes #3232

The annotation is read in `createLoadBalancerStatus`, next to the existing `hostname` annotation, so the status builder stays testable from a bare Service. The new unit tests cover both directions of the override; existing tests are unchanged.

This PR was written in part with the assistance of generative AI. I have reviewed and understand every change.

**Release note**:

```release-note
[openstack-cloud-controller-manager] Add the `loadbalancer.openstack.org/enable-ingress-hostname` Service annotation to override the `enable-ingress-hostname` config option per Service.
```


docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
```diff
@@ -388,7 +392,7 @@ This requires that not only the proxy server(e.g. NGINX) should support PROXY pr
 
 This guide uses nginx-ingress-controller as an example.
 
-To enable PROXY protocol support, the either the openstack-cloud-controller-manager config option [enable-ingress-hostname](./using-openstack-cloud-controller-manager.md#load-balancer) should set to `true` or an explicit hostname should be set on the load balancer service via [annotation](./expose-applications-using-loadbalancer-type-service.md#service-annotations) `loadbalancer.openstack.org/hostname`.
+To enable PROXY protocol support, the either the openstack-cloud-controller-manager config option [enable-ingress-hostname](./using-openstack-cloud-controller-manager.md#load-balancer) should set to `true` or an explicit hostname should be set on the load balancer service via [annotation](./expose-applications-using-loadbalancer-type-service.md#service-annotations) `loadbalancer.openstack.org/hostname`. On Kubernetes 1.32 and later neither is required: the load balancer IP is published with `ipMode: Proxy`, and a service can opt out of a cluster-wide `enable-ingress-hostname` with the annotation `loadbalancer.openstack.org/enable-ingress-hostname: "false"`.
```
stephenfin: Let's inverse this: we should reference the preferred/modern way to do this first. Something like "Since Kubernetes 1.32, ... Prior to this..."

stephenfin (changes requested):
One small nit in the docs but otherwise LGTM

docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md
BeyondEvil: Done. Order flipped. 👍

stephenfin (commented):
/ok-to-test
/lgtm
/assign @kayrus

Thanks. This makes sense to me. I'd like to kayrus to take a look since all things load balancers are his baby

winiciusallan (approved):

